PR #4, opened 2015-12-26
Introduced Locations

ThislsAUsername:
Locations hold
a Tile, now "Environment"
a Unit
a CO


src/UI/MapView.java
```diff
@@ -3,7 +3,9 @@
 import java.awt.Color;
 import java.awt.Graphics;
 
-import Terrain.Tile;
+import javax.tools.JavaFileManager.Location;
```
ThatIsAUsername: Remove this import.

src/Terrain/GameMap.java
```diff
@@ -23,65 +23,75 @@ public GameMap()
...
 				}
 			}
 		}
-
+		
 		// Throw down an HQ and a factory for each.
-		map[1][8] = Tile.getTile(Tile.Terrains.HQ, Tile.Weathers.CLEAR);
-		map[1][7] = Tile.getTile(Tile.Terrains.FACTORY, Tile.Weathers.CLEAR);
-		map[2][7] = Tile.getTile(Tile.Terrains.FACTORY, Tile.Weathers.CLEAR);
+		map[1][8] = new Location(Environment.getTile(Environment.Terrains.HQ, Environment.Weathers.CLEAR));
```
ThatIsAUsername: For all the locations that we are replacing, we don't want to re-create the location - just call setEnvironment().

PR #6, opened 2015-12-27
Implements the Combat Engine

ThislsAUsername:
Also added terrain defense to Environment

### DIFF
--- a/src/Terrain/Environment.java
+++ b/src/Terrain/Environment.java
@@ -10,6 +10,7 @@ public class Environment {
 
 	public final Terrains terrainType;
 	public final Weathers weatherType;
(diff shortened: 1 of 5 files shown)


src/Terrain/Environment.java
```diff
@@ -20,6 +21,42 @@ private Environment(Terrains tileType, Weathers weather)
...
+			defLevel = 0;
+			break;
+		case WATER:
+			defLevel = 1;
+			break;
+		case REEF:
+			defLevel = 2;
+			break;
+		default:
+			defLevel = -1;
```
ThatIsAUsername: We should at least print an error here rather than just continue happily along with bad data.

PR #9, opened 2015-12-27
Menuation

ThislsAUsername:
This is terrible and hacky, but it works for now....

### DIFF
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -64,6 +64,10 @@ public class MapController {
 			default:
 				System.out.println("Invalid InputMode in MapController! " + inputMode);
 		}
+		myGame.currentMenu = null;
+		if (inputMode == InputMode.ACTIONMENU || inputMode == InputMode.PRODUCTION) {
+			myGame.currentMenu = actionMenu;
+		}
 	}
 	
 	/**
@@ -103,7 +107,8 @@ public class MapController {
 				System.out.println("found a factory");
 				// TODO: Don't hard-code this. Also, is DamageChart the best place for UnitEnum?
 				DamageChart.UnitEnum[] units = {DamageChart.UnitEnum.INFANTRY};
-				actionMenu = new GameMenu(units);
+				String[] labels = {"Infantry: " + myGame.activeCO.getUnitModel(DamageChart.UnitEnum.INFANTRY).moneyCost};
+				actionMenu = new GameMenu(units, labels);
 				inputMode = InputMode.PRODUCTION;
 			}
 			break;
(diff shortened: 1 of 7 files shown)


src/Engine/MapController.java
```diff
@@ -103,7 +107,8 @@ else if(Environment.Terrains.FACTORY == loc.getEnvironment().terrainType
 				System.out.println("found a factory");
 				// TODO: Don't hard-code this. Also, is DamageChart the best place for UnitEnum?
 				DamageChart.UnitEnum[] units = {DamageChart.UnitEnum.INFANTRY};
-				actionMenu = new GameMenu(units);
+				String[] labels = {"Infantry: " + myGame.activeCO.getUnitModel(DamageChart.UnitEnum.INFANTRY).moneyCost};
```
ThatIsAUsername: I don't like hard-coding this here. MapView can get the string from the Enum value itself, and the cost by querying the activeCO in the GameInstance.

src/Engine/MapController.java
```diff
@@ -64,6 +64,10 @@ public void handleInput(InputHandler.InputAction input)
 			default:
 				System.out.println("Invalid InputMode in MapController! " + inputMode);
 		}
+		myGame.currentMenu = null;
+		if (inputMode == InputMode.ACTIONMENU || inputMode == InputMode.PRODUCTION) {
+			myGame.currentMenu = actionMenu;
```
ThatIsAUsername: Rather than set this here, I'd probably set it in GameInstance wherever the menu is created/deleted.

src/UI/GameMenu.java
```diff
@@ -5,11 +5,13 @@
 public class GameMenu {
 	
 	private Enum options[];
+	public String[] labels;
 	int selectedOption = 0;
 	
-	public GameMenu(Enum[] options)
+	public GameMenu(Enum[] options, String[] labels)
```
ThatIsAUsername: The string array parameter should not be necessary, based on current usage.

src/UI/MapView.java
```diff
@@ -40,9 +40,25 @@ protected void paintComponent(Graphics g) {
...
-		// Draw the current cursor location.
-		g.setColor(new Color(253,171,77));
-		g.fillRect(myGame.getCursorX()*tileSizePx, myGame.getCursorY()*tileSizePx, tileSizePx, tileSizePx);
+		if (myGame.currentMenu == null) {
+			// Draw the current cursor location.
+			g.setColor(new Color(253,171,77));
+			g.fillRect(myGame.getCursorX()*tileSizePx, myGame.getCursorY()*tileSizePx, tileSizePx, tileSizePx);
+		} else {
+			g.setColor(Color.black); // outer border
+			g.fillRect(myGame.gameMap.mapWidth*tileSizePx/4, myGame.gameMap.mapHeight*tileSizePx/4, myGame.gameMap.mapWidth*tileSizePx/2, myGame.gameMap.mapHeight*tileSizePx/2);
```
ThatIsAUsername: Using the GameMap width and height to set the size of the menu will not work once we have different sized maps. Also, it would more readable to pull the value into a local variable instead of repeating a lengthy incantation for each usage.

PR #12, opened 2015-12-31
Moves from DamageChart to discrete Weapons

ThislsAUsername:


src/Engine/MapController.java
```diff
@@ -265,11 +266,11 @@ private void handleActionInput(InputHandler.InputAction input)
...
 				case ATTACK:
 					Unit unitTarget = myGame.gameMap.getLocation(myGame.getCursorX(), myGame.getCursorY()).getResident();
-					if(unitTarget != null && DamageChart.chartDamage(unitActor, unitTarget) != 0)
+					if(unitTarget != null && unitActor.getDamage(unitTarget) != 0)
 					{
 						placeUnit(unitActor, unitActor.x, unitActor.y);
-						Utils.findActionableLocations(unitTarget, GameAction.ATTACK, myGame.gameMap);
-						boolean canCounter = myGame.gameMap.getLocation(unitActor.x, unitActor.y).isHighlightSet() && DamageChart.chartDamage(unitTarget, unitActor) != 0;
+//						Utils.findActionableLocations(unitTarget, GameAction.ATTACK, myGame.gameMap);
+						boolean canCounter = unitTarget.getDamage(unitActor) != 0;
```
ThatIsAUsername: Does this take relative positions into account?

src/Engine/Utils.java
```diff
@@ -21,14 +22,20 @@ public static void findActionableLocations(Unit unit, MapController.GameAction a
...
-						map.getLocation(i, j).setHighlight(true);
-					}
-					else
+					Unit target = map.getLocation(i, j).getResident();
+					if (target != null && target.CO != unit.CO)
 					{
-						map.getLocation(i, j).setHighlight(false);
+						for (Weapon gun :unit.weapons) 
+						{
+							if (gun.getDamage(unit.x, unit.y, target) > 0)
```
ThatIsAUsername: Why not just use Unit.getDamage() here instead?

src/Units/Weapons/Weapon.java
```diff
@@ -0,0 +1,52 @@
+package Units.Weapons;
+
+import Units.Unit;
+import Units.UnitModel;
+
+public class Weapon {
+	// format is [attacker][defender]
+	private static double[][] chart = {{55, 45,  7},
```
ThatIsAUsername: 'chart' is not a very informative name.

src/Units/Weapons/Weapon.java
```diff
@@ -0,0 +1,52 @@
...
+	public double getDamage(UnitModel.UnitEnum defender) {
+		if (ammo == 0 || defender == null)
+			return 0;
+		return chart[model.type][defender.ordinal()];
+	}
+	/**
+	 * @return returns its base damage against defender if the unit is in range,
+	 */
+	public double getDamage(int x, int y, Unit defender) {
+		int dist = Math.abs(defender.y-y) + Math.abs(defender.x-x);
```
ThatIsAUsername: Not checking for ammo or null defender in this version of getDamage?

src/Units/Weapons/Weapon.java
```diff
@@ -0,0 +1,52 @@
...
+		if ((dist >= model.minRange) && (dist <= model.maxRange))
+			return getDamage(defender.model.type);
+		return 0;
+	}
+	
+	public void fire()
+	{
+	if (ammo > 0)
+		ammo--;
```
ThatIsAUsername: Maybe add an "else print ("WARNING! Trying to fire an empty gun!");

src/Units/Weapons/WeaponModel.java
```diff
@@ -0,0 +1,24 @@
...
+import Units.Unit;
+import Units.UnitModel;
+
+public abstract class WeaponModel {
+	public int type;
+	public int maxAmmo;
+	public int minRange;
+	public int maxRange;
+	
+	public WeaponModel(int type, int ammo, int minRange, int maxRange) {
```
ThatIsAUsername: Type should probably be an enum, not a plain int.

src/Units/Unit.java
```diff
@@ -49,12 +59,56 @@ public void initTurn()
 		}
 	}
 	
-	// allows the unit to choose its weapon
-	public UnitEnum getWeapon(UnitEnum target) {
-		return model.type;
+	/**
+	 * @return the weapon of choice
+	 */
+	public Weapon getWeapon(Unit target) {
```
ThatIsAUsername: Is this function ever called? If not, we should remove it.

src/Engine/MapController.java
ThislsAUsername: Yes, yes it does

src/Engine/Utils.java
ThislsAUsername: Good point

src/Units/Weapons/Weapon.java
ThislsAUsername: Ammo is checked in the other version of the function, which is called anyway. Fixing null defender, though.

PR #15, opened 2016-01-01
Implemented a couple COModifiers, with example CO to test.

ThatIsAUsername:


src/Units/Weapons/Weapon.java
```diff
@@ -22,7 +22,7 @@ public Weapon(WeaponModel model) {
 	public double getDamage(UnitModel.UnitEnum defender) {
 		if (ammo == 0 || defender == null)
 			return 0;
-		return damageChart[model.getIndex()][defender.ordinal()];
+		return damageChart[model.getIndex()][defender.ordinal()] * (model.dmgPercent/100d);
```
ThislsAUsername: Damage modifier for the unit is in UnitModel, under the name COStr; Adding another damage modifier to the weapon itself is potentially confusing and doesn't add a huge amount of flexibility.

src/CommandingOfficers/Modifiers/CODamageModifier.java
```diff
@@ -0,0 +1,43 @@
...
+
+	@Override
+	public void apply(Commander commander) {
+		for(UnitModel um : commander.unitModels)
+		{
+			if(um.weaponModels != null)
+			{
+				for(WeaponModel wm : um.weaponModels)
+				{
+					wm.modifyDamageRatio(attackModifier);
```
ThislsAUsername: Should simply manipulate the COStr in the UnitModel

src/Terrain/GameMap.java
```diff
@@ -73,7 +74,7 @@ public GameMap(CommandingOfficers.Commander[] COs)
 		commanders[0].units.add(n);
 		map[11][4].setEnvironment(Environment.getTile(Environment.Terrains.FOREST, Environment.Weathers.CLEAR));
 		map[8][4].setEnvironment(Environment.getTile(Environment.Terrains.FOREST, Environment.Weathers.CLEAR));
-		Unit n2 = new Unit(commanders[1], new InfantryModel());
+		Unit n2 = new Unit(commanders[1], commanders[0].unitModels[UnitModel.UnitEnum.INFANTRY.ordinal()]);
```
ThatIsAUsername: This uses the Infantry model from the wrong CO.

src/Engine/CombatParameters.java
```diff
@@ -43,9 +43,9 @@ public void swap() {
 	
 	private void calculateParameters() {
 		baseDamage			= attacker.getDamage(defender);
-		attackFactor		= attacker.model.COStr;
+		attackFactor		= attacker.model.getDamageRatio();
```
ThatIsAUsername: Had I remembered that we had a COStr variable already, I probably would have named the function [get/modify]COStrength(), but really either name will work.

ThislsAUsername: I think I like the difference of nomenclature: the variable being named "COstr" or "COdef" makes it clear that it's a variable determined by their commander, and is thus effectively external to the unit.
Naming the function (get/modify)(Damage/Defense)Ratio makes it clear the exact function of the variable as it interacts with other units.

ThatIsAUsername: Good enough for me.

PR #16, opened 2016-01-02
Switches to using DamageChart.ini for... well, the damage chart.

ThislsAUsername:


ThatIsAUsername:
So, I guess I'm fine with switching to an external .ini file for weapon damage values, but... why?

ThislsAUsername:
My thought is basically that it's
a) slightly easier to manage than hardcoding the table and
b) good practice to learn/remember some I/O stuff before tackling .ini maps

ThatIsAUsername:
Eh, you are either tweaking a number in a .java file or in a .ini file. The second option doesn't require a recompile, but I don't think these values are going to be volatile. Plus, if we redo the combat system at some point, this might go away anyway.

I'm fine either way; you can merge this if you want (just gotta fix yer conflicts first!).

ThislsAUsername:
.....Right, we're doing that. I forgot. Aight, I'm just gonna hardcode it and re-commit/push.

src/Units/Weapons/MechRocket.java
```diff
@@ -3,7 +3,7 @@
 public class MechRocket extends WeaponModel {
 
 	public MechRocket() {
-		super(WeaponType.MECHROCKET, 3);
+		super(WeaponType.MECHZOOKA, 3);
```
ThatIsAUsername: Is it MECHROCKET or MECHZOOKA? Or are we happy having both?

src/Engine/Driver.java
```diff
@@ -12,6 +12,7 @@
 import Terrain.GameMap;
 import UI.InputHandler;
 import UI.MapView;
+import Units.Weapons.Weapon;
```
ThatIsAUsername: Does Driver need this import now that it isn't initializing Weapon?

src/Units/Weapons/Weapon.java
```diff
@@ -1,13 +1,42 @@
 package Units.Weapons;
 
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.StringBufferInputStream;
+import java.util.Scanner;
+import java.util.regex.Pattern;
```
ThatIsAUsername: I don't think we need these File IO imports any more either.

src/Units/Weapons/Weapon.java
```diff
@@ -1,13 +1,42 @@
...
+
 import Units.Unit;
 import Units.UnitModel;
 
 public class Weapon {
+	
+	private static boolean chartSet = false;
```
ThatIsAUsername: is chartSet actually used for anything?

src/Units/Weapons/MechRocket.java
ThislsAUsername: MECHZOOKA. I suppose it could be MECHROCKET, though.

src/Engine/Driver.java
ThislsAUsername: *sigh* I guess I'll go fix that.

src/Units/Weapons/Weapon.java
ThislsAUsername: Nope. Forgot to take it out.

PR #17, opened 2016-01-03
Created GameAction class, now use it in MapController.

ThatIsAUsername:
No changes are made until a full GameAction is constructed.
Moved Map-control functions to GameMap.
Changed possible-action code to take in a location.
Accidentally formatted MapController.

This change basically uses an object to represent a Unit action. I think it cleans up the MapController somewhat, and will make it easier to implement testing and also AI when we get there (which could be really soon, if we want).

### DIFF
--- a/src/Engine/GameAction.java
+++ b/src/Engine/GameAction.java
@@ -1,7 +1,12 @@
 package Engine;
 
+import Terrain.GameMap;
 import Units.Unit;
 
+/**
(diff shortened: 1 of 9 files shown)


ThatIsAUsername:
That was a fast code review for a pretty significant reorganization. Any
questions or anything?

ThislsAUsername:
Nope. We'd already spoken about it, so I knew what I was looking at.

PR #19, opened 2016-01-03
Testing

ThatIsAUsername:
New testing code. Anytime we find/fix a bug, or add new functionality, we should add a test to make sure we immediately find out if we break it later.

It also wouldn't hurt to add more tests than the two sample ones I included.

### DIFF
--- a/src/Engine/GameAction.java
+++ b/src/Engine/GameAction.java
@@ -24,12 +24,22 @@ public class GameAction
   
   public GameAction(Unit unit)
   {
+    this(unit, -1, -1, ActionType.INVALID, -1, -1);
+  }
+
+  public GameAction(Unit unit, int mx, int my, ActionType action)
+  {
+    this(unit, mx, my, action, -1, -1);
+  }
+
+  public GameAction(Unit unit, int mx, int my, ActionType action, int ax, int ay)
+  {
     unitActor = unit;
-    actionType = ActionType.INVALID;
-    moveX = -1;
-    moveY = -1;
-    actX = -1;
-    actY = -1;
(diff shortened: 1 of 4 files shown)


src/Engine/GameAction.java
```diff
@@ -93,6 +118,8 @@ public boolean isReadyToExecute()
    * Performs the built-up action, using the passed-in GameMap.
    * @return true if the action successfully executes, false if a problem occurs.
    */
+  // TODO: Should this validate that the GameAction is valid (e.g. we aren't trying
+  //  to move an Infantry unit 10 spaces)?
```
ThislsAUsername: I think isReadyToExecute should do it, if anything does. Do we really need more tests for valid action? I think the systems we have in place already should probably be enough.

ThatIsAUsername: Well, it's an interesting problem. We don't want to show the user invalid actions, so we kinda need to validate the user actions at each step (and then in theory it won't be possible to construct a GameAction that isn't correct). But then an AI doesn't go through the same process. If you manually build `GameAction(Infantry, 1, 1, WAIT, 10, 5)` and call execute, it will happily move the infantry 13 spaces regardless of the destination tile type. I'm somewhat inclined to add proper validation of some kind here as well, just as a final fail-safe (and for the extra debugging info it would provide when running test cases).

ThislsAUsername: Eh, sure. Should we make an issue and get back to that later, or what?

ThatIsAUsername: Well, this is merged, so an issue it is.

PR #20, opened 2016-01-03
Fixes counters and a little bit of the damage calculation itself

ThislsAUsername:


src/Engine/CombatParameters.java
```diff
@@ -11,19 +9,21 @@
 	public GameMap map;
 	public boolean isCounter, canCounter;
 	
-	public CombatParameters(Unit pAttacker, Unit pDefender, GameMap pMap, boolean isCounter, boolean canCounter) {
+	public CombatParameters(Unit pAttacker, Unit pDefender, GameMap pMap, boolean isCounter) {
```
ThatIsAUsername: Do we ever initialize CombatParameters with isCounter = true? It looks like we probably don't need this parameter, since it is always false.

src/Engine/CombatEngine.java
```diff
@@ -18,19 +17,22 @@ public static void resolveCombat(Unit attacker, Unit defender, GameMap map, bool
...
 		params.attacker.fire(params); // Lets the unit know that it has actually fired a shot.
-		if (canCounter) {
+		if (params.canCounter) {
 			params.swap();
 			for(int i = 0; i < modifiers.size(); i++) {
 				modifiers.get(i).alterCombat(params);
 			}
-			params.defender.HP -= params.calculateDamage();
-			params.attacker.fire(params);
+			if (params.attackerHP > 0) // stops counterattacks from dead units unless a CombatModifier allows it
```
ThatIsAUsername: Is it possible to build a test case to verify that this works as expected?

PR #25, opened 2016-01-05
Ui reorg

ThatIsAUsername:
I think breaking up the MapView into components may make it easier to maintain/update/replace sections of the drawing code. There will be more changes incoming, as I'm fleshing out how to get action animations to work (I may change MapView's name, and add some animation classes), but I'd like to get your feedback on these changes even if we don't merge them yet, to hopefully catch any issues (also, if you review this now, you'll be able to grok the final version more quickly).

### DIFF
--- a/src/UI/Art/FillRectUnitArtist.java
+++ b/src/UI/Art/FillRectUnitArtist.java
@@ -1,0 +1,63 @@
+package UI.Art;
+
+import java.awt.Color;
+import java.awt.Graphics;
+
+import Engine.GameInstance;
+import Terrain.GameMap;
+import Terrain.Location;
+import UI.MapView;
+import Units.Unit;
+
+public class FillRectUnitArtist implements UnitArtist
+{
+	private int tileSizePx = MapView.tileSizePx;
+	
+	GameInstance myGame;
+	GameMap gameMap;
+	
+	public static final Color COLOR_TIRED = new Color(128,128,128,160);
+	
+	public FillRectUnitArtist(GameInstance game)
+	{
+		myGame = game;
+		gameMap = myGame.gameMap;
+	}
+	
+	public void drawUnits(Graphics g)
+	{
+		// TODO: It would be pretty swell if we could just ask the map for a list of units.
+		for(int w = 0; w < gameMap.mapWidth; ++w)
+		{
+			for(int h = 0; h < gameMap.mapHeight; ++h)
+			{
+				if(gameMap.getLocation(w,h) != null)
+				{
+					Location locus = gameMap.getLocation(w,h);
+					if(locus.getResident() != null)
+					{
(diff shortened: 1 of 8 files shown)


ThatIsAUsername:
FYI, I also pushed up the branch battleAnim, which shows a very basic (think Nobunaga's Ambition) fight animation. It doesn't look like much yet, but it's about building the framework at this point. I didn't make a pull request because there will be more changes, but it's there if you feel like testing it out.

src/UI/Art/FillRectUnitArtist.java
```diff
@@ -0,0 +1,63 @@
...
+	
+	public FillRectUnitArtist(GameInstance game)
+	{
+		myGame = game;
+		gameMap = myGame.gameMap;
+	}
+	
+	public void drawUnits(Graphics g)
+	{
+		// TODO: It would be pretty swell if we could just ask the map for a list of units.
```
ThislsAUsername: Could you just.... get the COs and use their unit arrays to draw?

src/UI/MapView.java
```diff
@@ -16,198 +22,40 @@
...
-			break;
-			default:
-				tileColor = Color.BLACK;
-		}
-		
-		g.setColor(tileColor);
-		g.fillRect(x, y, tileSizePx, tileSizePx);
+		mapArtist.drawMap(g);
+		unitArtist.drawUnits(g);
+		mapArtist.drawHighlights(g);
```
ThislsAUsername: In Advance Wars, the units are on top of the highlights

ThislsAUsername:
I think it's relevant to point out that we should give UnitArtist the current GameAction, which probably entails moving that also into GameInstance.
Then the UnitArtist can just check if the unit it's about to draw is equal to the one in the action (checking for null first, ofc), and draw it in the acting place if it is.

src/UI/Art/FillRectUnitArtist.java
ThatIsAUsername: Probs... but then you would also get units in transports, not just those on the map. Still a good idea though, since that should be easy enough to check.

src/UI/MapView.java
ThatIsAUsername: So they are. If we switch that, I also want to make the highlight more opaque.

ThatIsAUsername:
Actually, I think I'm going a slightly different direction... If you look at the battleAnim branch, I've made MapController own MapView, and I'm thinking we will want to move both the game menu and the current action into MapView, so it will have knowledge of everything it needs for drawing.

ThatIsAUsername:
OK, I've updated battleAnim so that gameMenu and currentAction are in MapView. I'm gonna see about making units *move* rather than teleport.

PR #28, opened 2016-01-09
Makes HP range from 0/1-10, depending on how you count. Also makes it private.

ThislsAUsername:


src/Engine/TestDriver.java
```diff
@@ -86,12 +86,12 @@ private static boolean testCounter()
...
     
     // make him git rekt first attack
-    infB.HP = 5;
+    infB.damageHP(9.5);
     
     // make the attack
     new GameAction(mechA, 1, 1, GameAction.ActionType.ATTACK, 1, 2).execute(testMap);
     
-    boolean result = validate(Math.abs(mechA.HP - 100) < .1, "  attacker lost or gained health.");
+    boolean result = validate(mechA.getHP() == 10, "  attacker lost or gained health.");
```
ThatIsAUsername: Not get preciseHP()? A 1-health infantry might not do a whole tick of damage in an attack.

PR #29, opened 2016-01-09
Implements healing

ThislsAUsername:


src/Terrain/Environment.java
```diff
@@ -54,7 +54,7 @@ public int getDefLevel() {
 		case FACTORY:
 			return 4;
 		case HQ:
-			return 5;
+			return 4;
```
ThislsAUsername: Unrelated change, yes, but I didn't wanna bother with making an issue or a whole new branch for literally one character.

src/Units/Unit.java
```diff
@@ -46,27 +47,35 @@ public Unit(Commander co, UnitModel um)
...
-	
-	public void initTurn()
-	{
+
+	public void initTurn(Location locus) {
 		isTurnOver = false;
 		fuel -= model.idleFuelBurn;
-		if (captureTarget != null && captureTarget.getResident() != this)
-		{
+		if (captureTarget != null && captureTarget.getResident() != this) {
```
ThislsAUsername: Oops, guess my Eclipse has that as its default formatting. Oh well.

src/Units/Unit.java
```diff
@@ -46,27 +47,35 @@ public Unit(Commander co, UnitModel um)
...
-	
-  public double getDamage(Unit target)
-  {
-    return getDamage(target, x, y);
-  }
 
 	/**
 	 * @return how much base damage the target would take if this unit tried to attack it
 	 */
-	// TODO - Consider a canHitFrom function that could return true/false without doing damage calculations.
```
ThislsAUsername: It may as well calculate damage anyway, because you need to know if you can do damage to them with the weapon to know if you can hit them.

src/Units/Unit.java
```diff
@@ -46,27 +47,35 @@ public Unit(Commander co, UnitModel um)
...
+	public void initTurn(Location locus) {
 		isTurnOver = false;
 		fuel -= model.idleFuelBurn;
-		if (captureTarget != null && captureTarget.getResident() != this)
-		{
+		if (captureTarget != null && captureTarget.getResident() != this) {
 			captureTarget = null;
 			captureProgress = 0;
 		}
+		if (HP < 9.99) {
```
ThatIsAUsername: Whoops, I commented on the specific commit, rather than the top-level diff, so let's add them again.

Why not use `HP < 10`? Or better, `HP < maxHP`, so we don't have magic numbers.

src/Units/Unit.java
```diff
@@ -46,27 +47,35 @@ public Unit(Commander co, UnitModel um)
...
+			if (model.canRepairOn(locus) && locus.getOwner() == CO) {
+				int cost = model.moneyCost * Math.min(10 - getHP(), 2)/10; // 1/10th the cost to repair fully
```
ThatIsAUsername: I don't like magic numbers. I assume the 10 from (10 - getHP()) is max health? Use the variable instead, so we know why. Also I'm not sure your comment makes sense; healing 2 HP is 1/10 the cost of healing 10HP? It looks like you are doing proportional healing, i.e. cost for 2 HP is ((unit cost / maxHP) * 2), which doesn't match that comment.

src/Units/Unit.java
```diff
@@ -46,27 +47,35 @@ public Unit(Commander co, UnitModel um)
...
+				if (CO.money >= cost) {
+					CO.money -= cost;
+					alterHP(2);
+				} else if (CO.money >= model.moneyCost/10) {
```
ThatIsAUsername: It looks a little odd to calculate the cost of healing, and then throw it out the door and recalculate from the model again if the CO can only heal 1 HP (especially since the formula is much simpler this time). I'd be tempted to either use the formula both places (model.moneyCost/10 for 1 and model.moneyCost/5 for 2), or calculate both up front and use the one we can afford.

src/Units/Unit.java
ThislsAUsername: I misread the source I got the cost from, and continued in my misunderstanding while using the proper numbers. I think I'm officially crazy.

PR #30, opened 2016-01-16
Map animation

ThatIsAUsername:


src/Engine/MapController.java
```diff
@@ -1,28 +1,34 @@
...
+  private Path currentMovePath;
 
-  public MapController(GameInstance game)
+  public MapController(GameInstance game, MapView view)
   {
     myGame = game;
+    myView = view;
+    myView.setController(this);
     inputMode = InputMode.MAP;
+    myGame.setCursorLocation(6,5);
```
ThatIsAUsername: Hah. This wasn't entirely intended. I just did this so I wouldn't have to move to the unit before trying things out.

ThatIsAUsername:
Uh oh, just found a bug. Will update.

src/Engine/Utils.java
```diff
@@ -143,30 +148,171 @@ private static boolean checkSpace(Unit unit, GameInstance myGame, SearchNode cur
...
+		}
+
+		return canReach;
+	}
+
+	/**
+	 * Calculates and returns the shortest path for unit to take from its current location to map(x, y).
+	 * If no valid path is found, null is returned.
+	 */
+	public static Path findShortestPath(Unit unit, int x, int y, GameMap map)
```
ThislsAUsername: Correct me if I'm wrong, but this looks like it duplicates the stuff from the findPossibleDestinations function. I'm thinking it'd be good stuff to make a separate function that returns the 2D array of moves left.
It could easily be private, I imagine.

ThatIsAUsername: It's close. I'm assigning parents to SearchNodes here, which findPossibleDestinations doesn't do. There's no reason it couldn't though, so we could combine them. I may go ahead and do that.

ThatIsAUsername: OK, did that.

ThislsAUsername:
Looks good. Haven't had a chance to test it, and I'm afraid I'll end up working on more things if I actually do so. You can merge if you're sure.

ThatIsAUsername:
OK, I'll merge. If we do find something wrong, we can just make an issue
and fix it later.

PR #46, opened 2016-12-27
Makes CombatModifiers an actual thing that does actual things. Lifecycle untested, unfortunately.

ThislsAUsername:


src/Engine/Combat/CombatDamageModifier.java
```diff
@@ -0,0 +1,22 @@
...
+public class CombatDamageModifier extends CombatModifier
+{
+  public int boost;
+  
+  public CombatDamageModifier(Commander user, int ATKboost)
+  {
+    super(user);
+    boost = ATKboost;
+  }
+
```
ThatIsAUsername: I like to put the @Override annotation just so it's clear this function is from the parent class.

src/Engine/Combat/CombatModifier.java
```diff
@@ -0,0 +1,27 @@
+package Engine.Combat;
+
+import CommandingOfficers.Commander;
+
+public class CombatModifier
```
ThatIsAUsername: Now that this class has actual functionality, I'd like a description here of what it is, how/where it's used, so a new person can look at the class description and orient themselves.

src/Engine/GameAction.java
```diff
@@ -1,6 +1,7 @@
 package Engine;
 
 import CommandingOfficers.Commander;
+import Engine.Combat.CombatEngine;
```
ThatIsAUsername: Why does GameAction now import CombatEngine if it has not otherwise changed?

src/Engine/MapController.java
```diff
@@ -112,6 +116,11 @@ public boolean handleInput(InputHandler.InputAction input)
         System.out.println("Invalid InputMode in MapController! " + inputMode);
     }
 
+    if (exitMap)
+    {
+      // re-initialize our list of CombatModifiers, so we know it will be empty for next game
+      CombatEngine.modifiers = new ArrayList<CombatModifier>();
```
ThatIsAUsername: Why not just `CombatEngine.modifiers.clear()`?

src/Test/TestCombat.java
```diff
@@ -54,6 +58,7 @@ private boolean testUnitDeath()
 
     // Clean up
     testMap.removeUnit(mechA);
+    CombatEngine.modifiers.remove(mod);
```
ThatIsAUsername: I'm not sure I like needing a separate step to remove modifiers from the engine. That seems like an easy step to overlook, and I'd rather make it more self-contained.

src/Engine/GameAction.java
ThislsAUsername: .....Not sure. I don't think I even opened the file?

src/Engine/MapController.java
ThislsAUsername: Because I didn't know it existed.

src/Engine/Combat/CombatDamageModifier.java
ThislsAUsername: Feel free?

src/Engine/Combat/CombatModifier.java
ThislsAUsername: It's a pretty broad functionality; I'm not sure how I'd go about describing it.

src/Test/TestCombat.java
```diff
@@ -43,7 +46,8 @@ private boolean testUnitDeath()
     Unit infB = addUnit(testMap, testCo2, UnitEnum.INFANTRY, 1, 2);
 
     // Make sure the infantry will die with one attack
-    infB.damageHP(7);
+    CombatDamageModifier mod = new CombatDamageModifier(testCo1,500);
```
ThatIsAUsername: That's... quite a power boost. Do we still want a vanilla test case for combat? I might be more comfortable making a separate test for the combat modifier. Or even make a proper unit test for the CombatModifier class.

src/Engine/Combat/CombatEngine.java
```diff
@@ -11,13 +11,9 @@
...
-    // TODO: make sure to clean up unneeded modifiers
-    //		for(int i = 0; i < modifiers.size(); i++) {
-    //			if (modifiers.get(i).done) {
-    //				modifiers.get(i).initTurn();
-    //				modifiers.remove(i);
-    //			}
-    //		}
+    // The list of modifiers is updated and culled every turn in GameInstance.
+    // It is emptied in MapController when exiting the map.
+    // TODO: Figure out a better way to handle their lifecycle?
```
ThatIsAUsername: Would it make more sense to have Commander own their CombatModifiers? This would clean up the lifecycle quite a bit I think. The Commander's initTurn could be used to remove any that no longer matter, and then we don't have to remember to clean out the engine at the end of the game. Plus, I'm just not sure I like having a global, static entity that persists between games own data for a specific game.

src/Test/TestCombat.java
ThislsAUsername: It's not necessary in normal gameplay. I just did that cleanup to make sure there weren't weird sideeffects.

src/Test/TestCombat.java
ThislsAUsername: It should realistically get its own test case, but I'm lazy and I didn't have stuff set up yet.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: See, this is why I shouldn't code while I'm tired. This would actually allow merging of COModifiers and CombatModifiers, maybe. We should look into this at some point.

ThatIsAUsername: I'd rather look into it now, rather than create duplicate code going forward that we have to address at some point. Clean as you go, and all that.

src/Test/TestCombat.java
ThatIsAUsername: Oh, but it is necessary in normal gameplay. You just put `MapController` in charge of it.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: Yes, but doing that now might make merge conflicts, and that'd be no fun.

ThatIsAUsername: Also, you are right CombatDamabeModifier and CODamageModifier do look suspiciously similar. We should consolidate that if possible.

ThislsAUsername: the only real difference is that the combat mod could probably filter by fun things like terrain or something

src/CommandingOfficers/Modifiers/COModifier.java
```diff
@@ -1,9 +1,31 @@
 package CommandingOfficers.Modifiers;
 
 import CommandingOfficers.Commander;
+import Engine.Combat.CombatParameters;
 
-public interface COModifier
+// COModifier provides static bonuses to units,
```
ThatIsAUsername: You should use Javadoc comments for class/function/variable descriptions.
This allows Eclipse to show the comment when you hover over it from somewhere else.

src/CommandingOfficers/Commander.java
```diff
@@ -62,12 +61,15 @@ public void initTurn(GameMap map)
     }
     money += turnIncome;
 
-    // Un-apply any modifiers that were activated last turn.
-    // TODO: If/when we have modifiers that last multiple turns, figure out how to handle them.
-    for( int i = modifiers.size() - 1; i >= 0; --i )
+    // Update any modifiers that were activated last turn, removing expired ones.
+    for( int i = 0; i < modifiers.size(); i++ )
```
ThatIsAUsername: Why did you remove the iteration direction? If we remove elements while iterating forward, our index will get out of sync.

src/CommandingOfficers/Modifiers/COModifier.java
ThislsAUsername: _shrug_

src/CommandingOfficers/Commander.java
ThislsAUsername: Because I didn't think about it.

src/CommandingOfficers/Modifiers/CODamageModifier.java
```diff
@@ -1,37 +1,66 @@
...
 import Units.UnitModel;
 import CommandingOfficers.Commander;
+import Engine.Combat.CombatParameters;
 
-public class CODamageModifier implements COModifier
+// Provides a damage boost either universally, on one terrain
+public class CODamageModifier extends COModifier
 {
   private int attackModifier = 0;
+  private Environment validTerrains = null;
```
ThatIsAUsername: Why is `validTerrains` plural if there's only one?
Also, is it about validity, or applicability/relevancy/some other word that would make more semantic sense?

src/CommandingOfficers/Modifiers/COModifier.java
ThatIsAUsername: _sigh_

PR #52, opened 2016-12-28
Added terrain to combatParameters for future use.

ThatIsAUsername:


ThatIsAUsername:
Terrain info was already available via Units and map

PR #64, opened 2017-12-16
Adds a map conversion script and some maps from AW as samples.

ThislsAUsername:


ThatIsAUsername:
We don't want thumbs.db to be included in the repo.

ThatIsAUsername:
As discussed, let's avoid output file names with spaces.

res/map/Asphalt_Maze.map
```diff
@@ -0,0 +1,11 @@
...
+  FR  FR  FR  GR  GR  RD  RD  RD  RD  RD  RD  RD 1FC 1HQ 1FC
+  FR  FR  FR  GR  FR  RD  FR  RD  FR  RD  FR  RD  GR 1FC  GR
+  FR  GR  GR  RD  RD  RD  RD  CT  RD  RD  RD  RD  RD  RD  GR
+  FR  GR  FR  RD  FR  RD  FR  RD  FR  RD  FR  RD  FR  RD  GR
+  GR  RD  RD  RD  RD  CT  RD  RD  RD  CT  RD  RD  RD  RD  GR
+  GR  RD  FR  RD  FR  RD  FR  RD  FR  RD  FR  RD  FR  GR  FR
+  GR  RD  RD  RD  RD  RD  RD  CT  RD  RD  RD  RD  GR  GR  FR
+  GR 0FC  FR  RD  FR  RD  FR  RD  FR  RD  FR  GR  FR  FR  FR
+ 0FC 0HQ 0FC  RD  RD  RD  RD  RD  RD  RD  GR  GR  FR  FR  FR
+  GR 0FC  GR  GR  GR  GR  GR  GR  GR  GR  FR  FR  FR  FR  FR
```
ThatIsAUsername: This file looks correct, but in-game, the map is missing the last row.
asphaltmaze

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import Engine.XYCoord;
+import Terrain.Environment;
+import Terrain.MapInfo;
+
+public class MapReader extends IMapBuilder
+{
+  private final static String mapName = "Spann Island";
```
ThatIsAUsername: It looks like you copy-pasta'd and forgot to remove some things. From here through line 37 can disappear I think, plus the getMapInfo() method at the end of the file.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+      for( final File fileEntry : folder.listFiles() )
+      {
+        // we aren't checking subdirectories, yet
+        if( !fileEntry.isDirectory() )
+        {
+          try
+          {
+            // System.out.println();
+            String mapName = fileEntry.getName();
+            if( !mapName.contains(".map") )
```
ThatIsAUsername: Can we verify that the filename ends with ".map", instead of just containing it?

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
+package Terrain.Maps;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
```
ThatIsAUsername: `IOException` doesn't appear to be used.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+          try
+          {
+            // System.out.println();
+            String mapName = fileEntry.getName();
+            if( !mapName.contains(".map") )
+              continue;
+            mapName = mapName.replaceAll("_", " ");
+            mapName = mapName.replaceAll(".map", "");
+            System.out.println("Parsing map: " + mapName);
+            Scanner scanner = new Scanner(fileEntry);
```
ThatIsAUsername: I would probably move the `Scanner` creation closer to the loop where it's used. In general, it's good practice to create variables as close to the time of first use as possible, to make it easier to read and reduce the lifecycle scope.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+            String line = scanner.nextLine();
+            ArrayList<ArrayList<Environment.Terrains>> terrainData = new ArrayList<ArrayList<Environment.Terrains>>();
+
+            ArrayList<ArrayList<XYCoord>> properties = new ArrayList<ArrayList<XYCoord>>(4);
```
ThatIsAUsername: I know this is temporary, since we both want support for more than four players, but I'd prefer to get it right the first time, and not have to go back and revise later. That's not always possible of course, and I'll take a working mostly-solution over a perfect figment of my imagination any day.

On a more general note, you can generally expect to handle zero of something, one of something, or 2+ of something. Once you have to handle more than one (certainly more than two), you should be looking for a way to not care exactly how many there are anymore.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+            properties.add(new ArrayList<XYCoord>());
+            properties.add(new ArrayList<XYCoord>());
+            properties.add(new ArrayList<XYCoord>());
+            properties.add(new ArrayList<XYCoord>());
+
+            for( int i = 0; i < line.length() / 4; i++ )
```
ThatIsAUsername: This whole file is somewhat lacking in comments. Sure, I can figure it out, but it's much faster if there are comments to guide. Plus whenever you next have to edit the code, you'll be glad to have them.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+            {
+              for( int i = 0; i < line.length(); i += 4 )
+              {
+                int sideNumber = 99;
+                try
+                {
+                  sideNumber = Integer.parseInt(("" + line.charAt(i) + line.charAt(i + 1)).trim());
+                }
+                catch (NumberFormatException e)
+                {}
```
ThatIsAUsername: Empty `catch` statements are just the worst. What is the value of `sideNumber` if you get here? Is there anything the user needs to know, or any steps you should take to ensure correct operation after catching an exception? Would it be appropriate to abort? Empty `catch`es are a recipe for hidden bugs that take forever to track down.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+              terrainData.add(new ArrayList<Environment.Terrains>());
+            }
+            for( int yCoord = 0; scanner.hasNextLine(); line = scanner.nextLine() )
+            {
+              for( int i = 0; i < line.length(); i += 4 )
+              {
+                int sideNumber = 99;
+                try
+                {
+                  sideNumber = Integer.parseInt(("" + line.charAt(i) + line.charAt(i + 1)).trim());
```
ThatIsAUsername: This kind of string addition is not very performant, since it generally requires copying the entire string contents as each piece is added (since strings are immutable). Not that it's too expensive here, but FYI. Can we just grab the substring we care about in one shot instead of kludging a char-to-string-to-int conversion?

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+                }
+                catch (NumberFormatException e)
+                {}
+                if( sideNumber < 4 ) // TODO: add more sides
+                {
+                  properties.get(sideNumber).add(new XYCoord(i / 4, yCoord));
+                }
+                String terrainCode = "" + line.charAt(i + 2) + line.charAt(i + 3);
```
ThatIsAUsername: I'm not a super fan of carving up the input code piecemeal at different sections. It's probably cleaner to parse the whole token at once in the same place, so if we ever change our encoding there will be less hunting around for where to make updates. Yes, that might involve declaring the `terrainCode` variable a few lines up--which contradicts my earlier advice about declaring variables close to where they are to be used--but these rules are mostly suggestions, and clarity often trumps purity.

The super-OO way to do this would probably be to have an object that takes in the four-byte map code in its constructor, and has accessors for the owner and terrain type, but I'm not going to insist on that unless you think it provides some benefit for our use case here.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+          // skip the file if you can't find it
+          catch (FileNotFoundException e)
+          {
+            System.out.println("WARNING: Could not find map file " + fileEntry.getName());
+          }
+        }
+      }
+    }
+    catch (NullPointerException e)
+    {}
```
ThatIsAUsername: Another empty `catch` block. Gross.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+    return importMaps;
+  }
+
+  private static Environment.Terrains stringToCode(String input)
+  {
+    // helpful regex:
```
ThatIsAUsername: Helpful for what? Do we need/use this here?

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+                if( sideNumber < 4 ) // TODO: add more sides
+                {
+                  properties.get(sideNumber).add(new XYCoord(i / 4, yCoord));
+                }
+                String terrainCode = "" + line.charAt(i + 2) + line.charAt(i + 3);
+                terrainData.get(i / 4).add(stringToCode(terrainCode));
+              }
+              yCoord++;
+            }
+            ArrayList<Environment.Terrains[]> terrainArrayArray = new ArrayList<Environment.Terrains[]>();
```
ThatIsAUsername: Is there a reason to create `terrainArrayArray` as an intermediate before calling `toArray` on it? Can we just go to a 2D array in one step instead of taking two (and copying everything twice)? Same with `propertyArrayArray`.

If we could find a way to avoid the need to do this conversion alltogether, that would be even better, but I'm not sure we can reasonably avoid that.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,171 @@
...
+      case "SE":
+        return SE;
+      case "SH":
+        return SH;
+      default:
+        return GR;
+    }
+  }
+
+  public static MapInfo getMapInfo()
```
ThatIsAUsername: This doesn't look like it belongs here.

res/map/AWBWMapConverter.py
```diff
@@ -0,0 +1,217 @@
...
+		173: '13XX',
+		174: '13HQ',
+		175: '13LB',
+		176: '13SP',
+	}.get(x, '  XX')    # '  XX' is default if x not found
+	
+	
+
+if __name__ == "__main__":
+	if "-h" in sys.argv:
```
ThatIsAUsername: Instead of an `if`-`elif`, why not just say `if '-h' in sys.argv or '--help' in sys.argv:`?

res/map/AWBWMapConverter.py
```diff
@@ -0,0 +1,217 @@
...
+				print("File ",outname," cannot be written to.")
+		except IOError:
+			print("File ",name," does not exist.")
+		finally:
+			infile.close()
+			outfile.close()
+
+
+#dictionary text generation
+def ugh():
```
ThatIsAUsername: This function doesn't appear to be referenced anywhere, and has a pretty nondescriptive name. Is this intermediate code that you forgot to remove?

res/map/AWBWMapConverter.py
```diff
@@ -0,0 +1,217 @@
...
+	
+	
+
+if __name__ == "__main__":
+	if "-h" in sys.argv:
+		print("Takes in Advance Wars by Web map files, and outputs our own map format. \nFilename is the same, extension is '.map'.")
+	elif "--help" in sys.argv:
+		print("Takes in Advance Wars by Web map files, and outputs our own map format. \nFilename is the same, extension is '.map'.")
+	else:
+		main(sys.argv[1:])
```
ThatIsAUsername: It might be helpful to print the help text if the script is run with no parameters (and maybe include an "invoke as 'AWBWMapConverter.py <filename>'", just so it's clear how to run it).

src/Terrain/Maps/MapReader.java
ThislsAUsername: I only did that much because otherwise it was trying to read in my Python script, and that threw lots of errors.
I don't really care to make that change.

src/Terrain/Maps/MapReader.java
ThislsAUsername: I suppose we could check the size() of properties to make sure each team we put in is accounted for, but I'd much rather just support what AWBW supports - up to 14, and not add more unnecessary logic to the loop.
I'd happily make the 14 instantiations a loop, but 4 was too few to bother with at the time.

src/Terrain/Maps/MapReader.java
ThislsAUsername: Sure. Am planning to add comments, but I was in the zone and had further goals to accomplish.

src/Terrain/Maps/MapReader.java
ThislsAUsername: 0
It's already instantiated, and Integer.parseInt() does not affect the value if the input is invalid.
The user doesn't need to be notified.
No more steps needed.
Aborting would be highly stupid.
We're feeding parseInt() a string that may or may not be empty. I suppose I could make a variable, and check if it's an empty string, but the exception works just as well.
The only reason I'd want an exception to actually be an exception is if the file was improperly formatted. In that case, I don't know what we'd want to do but skip the file and maybe log it.

src/Terrain/Maps/MapReader.java
ThislsAUsername: I don't even know what *could* trigger that catch block, unless the operating system is generally borked in some way.
But Java would scream and wail and carry on if I didn't have it.

src/Terrain/Maps/MapReader.java
ThislsAUsername: I used it in Notepad++ to generate the cases with a line-separated list of the codes.
I suppose I can delete it, but it'd be handy to have around if/when we add more terrains.
Also, adding terrains is painful and it'd be nice to fix that.

src/Terrain/Maps/MapReader.java
ThislsAUsername: I looked into it some while I was doing it, and didn't find anything useful.

res/map/AWBWMapConverter.py
ThislsAUsername: Because I didn't know that would work?

res/map/AWBWMapConverter.py
ThislsAUsername: It's code to generate the text of the dictionary. I suppose it's not useful anymore, so I can remove it if it's unwanted.

res/map/AWBWMapConverter.py
ThislsAUsername: That... is a good point.

src/Terrain/Maps/MapReader.java
ThatIsAUsername: If you want to leave it in I guess that's fine, but maybe expand the comment so the next person knows why it's useful too (and even how to use it).

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,174 @@
...
+          catch (FileNotFoundException e)
+          {
+            System.out.println("WARNING: Could not find map file " + fileEntry.getName());
+          }
+        }
+      }
+    }
+    catch (NullPointerException e)
+    {
+      System.out.println("WARNING: res/maps directory does not exist.");
```
ThatIsAUsername: The printed warning references res/maps, but the actual directory is res/map

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,174 @@
...
+            for( int i = 0; i < 4; i++ ) // TODO: add more sides
+            {
+              properties.add(new ArrayList<XYCoord>());
+            }
+
+            Scanner scanner = new Scanner(fileEntry);
+            // We need the first line so we can pre-fill our terrain data array with the proper number of sub-arrays.
+            String line = scanner.nextLine();
+
+            // The game's internal representation of the map is transposed from what we see.
```
ThatIsAUsername: I think this comment could be misleading. If you manually write out the 2D terain array, then yes, it appears to be transposed from what is drawn. However, if you pick an x/y coordinate on the map, it <i>does</i> correspond to the tile at that [x][y] position in the array. The transposition thing is just a function of how it is necessary to write 2D arrays in code, and only really comes into play when e.g., you try to look at one of our in-code maps. The remaining explanation is useful though.

res/map/AWBWMapConverter.py
```diff
@@ -0,0 +1,211 @@
+
+import sys
+import math
+import re
+import csv
+
+
+def convertFile(infile,outfile):
+	reader = csv.reader(infile)
+	# Fun fact: this used to write out to the file once for every loop iteration. That was... not exactly what you'd call efficient. This is... more so.
```
ThatIsAUsername: While that <i>is</i> a fun fact, I'm not sure we need to preserve it for posterity.

res/map/AWBWMapConverter.py
```diff
@@ -0,0 +1,211 @@
...
+		outstring += "\n"
+	outfile.write(outstring)
+	return
+
+def main(names):
+	for name in names:
+		try:
+			infile = open(name)
+			# Split the file extension off, and replace it with .map
+			outname = re.sub(r'\..*$','.map',name)
```
ThatIsAUsername: I'd like to also replace any spaces in the filename with underscores. We already handle reading those on the Java side. I think you can accomplish this by just tacking `.replace(' ', '_')` to the end if this line.

src/Terrain/Maps/MapReader.java
```diff
@@ -0,0 +1,177 @@
...
+            {
+              properties.add(new ArrayList<XYCoord>());
+            }
+
+            Scanner scanner = new Scanner(fileEntry);
+            // We need the first line so we can pre-fill our terrain data array with the proper number of sub-arrays.
+            String line = scanner.nextLine();
+
+            // The game's internal representation of the map is transposed from the representation you'd see if you wrote out the array.
+            // This is so because that allows you to access mapArray[x][y], which is convenient for handling the map.
```
ThatIsAUsername: I'm not sure this is any more clear. I think the default expectation going in is that the third tile from the left and two down (so, location 2, 1), would correspond to mapArray[2][1]. Adding a bunch of verbiage here about transposed matrices is just going to create confusion, and subvert that initial (and valid) expectation. The explanation here isn't super clear either, which doesn't help. If anything, I'd just point out that each subarray is a column, and rows span across the subarrays at a given index, and if we want to talk about how manually writing out a 2D array in code might look different than it does when drawn out, we can do that in some location where we are actually manually writing out a 2D array in code.

ThislsAUsername: I suppose I was more explaining it to myself than anyone else, and that's not terribly relevant since the code works.
I cut out most of it.

PR #66, opened 2017-12-21
Air and sea

ThislsAUsername:


src/CommandingOfficers/Commander.java
```diff
@@ -106,27 +130,29 @@ public UnitModel getUnitModel(UnitEnum unitType)
...
         break;
       }
     }
 
     return um;
   }
 
-  public ArrayList<UnitModel> getShoppingList()
+  public ArrayList<UnitModel> getShoppingList(Terrains buyLocation)
   { // TODO: will eventually need to take in terrainType so it can separate out air/ground/navy
```
ThatIsAUsername: I think this `TODO` can go away now.

src/CommandingOfficers/Modifiers/CODamageModifier.java
```diff
@@ -15,7 +15,7 @@ public CODamageModifier(int percentChange)
   @Override
   public void apply(Commander commander)
   {
-    for( UnitModel um : commander.unitModels )
+    for( UnitModel um : commander.landModels )
```
ThatIsAUsername: Looks like `CODamageModifier`s cannot apply to air and sea units.

src/CommandingOfficers/Modifiers/COMovementModifier.java
```diff
@@ -25,7 +25,7 @@ public COMovementModifier(int range)
   @Override
   public void apply(Commander commander)
   {
-    for( UnitModel um : commander.unitModels )
+    for( UnitModel um : commander.landModels )
```
ThatIsAUsername: Air/Sea left out again.

src/Engine/MapController.java
```diff
@@ -110,10 +114,10 @@ public boolean handleInput(InputHandler.InputAction input)
         exitMap = handleConfirmExitMenuInput(input);
         break;
       case CO_INFO:
-        if( coInfoMenu.handleInput( input ) )
+        if( coInfoMenu.handleInput(input) )
```
ThatIsAUsername: There seem to be a lot of whitespace changes. Do we have conflicting autoformat settings?

src/Engine/MapController.java
```diff
@@ -666,7 +673,7 @@ public void animationEnded( GameEvent event, boolean animEventQueueIsEmpty )
       }
 
       // If fewer than two COs yet survive, the game is over.
-      if(activeNum < 2)
+      if( activeNum < 2 )
```
ThatIsAUsername: Yeah, we really need to reconcile our autoformat settings. I don't want pull requests with 90% whitespace changes.

src/UI/Art/SpriteArtist/TerrainSpriteSet.java
```diff
@@ -262,7 +262,10 @@ private boolean isTerrainObject(Environment.Terrains terrainType)
       case CITY:
       case FACTORY:
       case FOREST:
+      case AIRPORT:
```
ThatIsAUsername: Hmm... I really don't like having to hunt down all the switch cases to add enum values. We should make a note to re-evaluate how we do this so we have fewer updates when we add terrain types in the future. This is fine for now though.

src/Units/LanderModel.java
```diff
@@ -0,0 +1,32 @@
...
+import Engine.GameAction.ActionType;
+import Terrain.Environment.Terrains;
+import Units.MoveTypes.FloatLight;
+import Units.MoveTypes.MoveType;
+
+public class LanderModel extends UnitModel
+{
+
+  private static final MoveType moveType = new FloatLight();
+  // TODO: limit the valid locations to unload
```
ThatIsAUsername: Can you provide a little more explanation? I'm currently not sure what needs to be done. I can infer that maybe we would currently get an "Unload" prompt when there are no valid drop locations, but a little extra text here would go a long way towards making sure this actually gets done. A descriptive issue would go even farther.

src/Units/MoveTypes/Flight.java
```diff
@@ -6,8 +6,8 @@
   public Flight()
   {
     // format is [weather][terrain]
-    int[][] tempCosts = { { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }, { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 },
-        { 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2 }, { 1, 1, 1, 3, 1, 1, 1, 1, 1, 1, 1 } };
+    int[][] tempCosts = { { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }, { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 },
```
ThatIsAUsername: Ugh. I don't like having to do manual data entry in opaque 2D arrays to support new terrain/movement types. Add this to the list of things to revisit.

src/Engine/MapController.java
ThatIsAUsername: The file `DefPcCodeStyle.xml` at the top level of the repo contains the code settings I thought we both were using. You should set Eclipse to use this file for this project so we don't keep stepping on each other.

src/CommandingOfficers/Modifiers/CODamageModifier.java
ThislsAUsername: I think my preference would be to have a unified model array that various applications can filter, but I dunno how we could easily sort them out again.
This was a stopgap so I didn't have to think more about it since I wanted to move on.

src/Engine/MapController.java
ThislsAUsername: I already am using that file. That's why you never see any braces like
if (thing) {
   then do this
}

src/CommandingOfficers/Modifiers/CODamageModifier.java
ThatIsAUsername: Hm. We could consider adding a 'domain' or 'category' attribute to `UnitModel`, and filter on that. We may want to introduce unit attribute flags at some point anyway. Up to you if that's in the scope of this pull request; we can just make an issue and get it later.

src/Units/APCModel.java
```diff
@@ -16,7 +16,7 @@
 
   public APCModel()
   {
-    super("APC", UnitEnum.APC, 5000, 70, 0, 6, moveType, actions, healHabs, null);
+    super("APC", UnitEnum.APC, ChassisEnum.TRUCK, 5000, 70, 0, 6, moveType, actions, healHabs, null);
```
ThatIsAUsername: I'm curious why you have both tank and truck. I'm also curious why APC is a truck, since it is a treaded vehicle. I would sort of see a more natural distinction as being in terms of domain - land, sea, air (maybe low-air and high-air to distinguish copters and jets), subsea. Do we gain anything by distinguishing truck and tank (and how do we determine which is which)?

src/Units/UnitModel.java
```diff
@@ -15,8 +15,16 @@
     INFANTRY, MECH, RECON, TANK, MD_TANK, NEOTANK, APC, ARTILLERY, ROCKETS, ANTI_AIR, MOBILESAM, FIGHTER, BOMBER, B_COPTER, T_COPTER, BATTLESHIP, CRUISER, LANDER, SUB
   };
 
+  // NB: "Truck" is technically unnecessary, but I feel it may be useful for alternate damage systems.
+  // The rubric for what's a "truck" is whether the damage dealt to it by an infantry is >9%, not movetype.
```
ThatIsAUsername: Oh. I see. I don't think I like infantry damage as the distinguishing factor, especially since, as you say, this could be used for other damage systems (in which infantry may not deal >9% damage to certain of these vehicles). I might be feeling "domain" as a more useful discriminator than "chassis".

src/Units/APCModel.java
ThislsAUsername: It's explained in UnitModel. I called anything that takes more than 9% from an infantry a truck. That distinguishes it from tire/tread, since chassis distinctions are likely going to be used in damage, and those vehicles are a lot less armored.

ThislsAUsername:
The merge conflict is two lines in MapLibrary that both have a right to be there. Just add both if you merge it in.

PR #70, opened 2017-12-26
Window resize

ThatIsAUsername:


src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -349,9 +403,11 @@ private void drawUnitsAndMapObjects(Graphics g)
...
-    for( int y = mapViewY-1; y < mapViewY+mapTilesToDrawY+1; ++y )
+    int drawY = (int)mapViewDrawY;
+    int drawX = (int)mapViewDrawX;
+    for( int y = drawY-1; y < drawY+mapTilesToDrawY+1; ++y )
     {
-      for( int x = mapViewX-1; x < mapViewX+mapTilesToDrawX+1; ++x )
+      for( int x = drawX-1; x < drawX+mapTilesToDrawX+1; ++x )
       {
         // Since we are trying to draw a ring of objects around the viewable space to
         // ensure smooth scrolling, make sure we aren't running of the edge of the map.
```
ThislsAUsername: *off

ThislsAUsername:
I will note that TerrainObjects aren't displayed on the lower edge you're scrolling onto if the screen is not exactly on a number of tiles in size.
Also the main menu gets weird if you resize it after quitting out from a resized game 
image

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: Fixed

ThatIsAUsername:
The TerrainObject thing would be annoying, and the offset main menu would as well. I've fixed both, and I also made the map select and CO select screens dynamically rescalable as well, so in theory the game should just keep your chosen window size throughout; it'll only resize if you open a map that is smaller than the window, and then it will just shrink to fit.

ThislsAUsername:
Are you sure you fixed it?
image
Because I'm sure you didn't.

ThatIsAUsername:
Steps to reproduce? This doesn't seem to happen on my machine.

PR #74, opened 2017-12-27
Sprites tank md tank

ThatIsAUsername:


ThislsAUsername:
I don't really want to complain too much about sprites, but I think the mdTank looks a little too similar to the normal tank.

ThatIsAUsername:
Reworking things

PR #75, opened 2017-12-27
Bug fixes

ThislsAUsername:


src/Engine/BattleInstance.java
```diff
@@ -92,7 +95,13 @@ public BattleParams(Unit attacker, Unit defender, int battleRange, boolean moved
       attackerHP = attacker.getHP();
       defenseFactor = defender.model.getDefenseRatio();
       defenderHP = defender.getHP();
-      terrainDefense = battleground.getDefLevel();
+      terrainDefense = 0;
+      // Air units shouldn't get terrain defense
+      if( ChassisEnum.AIR_HIGH != defender.model.chassis && ChassisEnum.AIR_LOW != defender.model.chassis )
+      {
+        // getDefLevel returns the number of terrain stars. Since we're using %Def, we need to multiply by 10. However, we do that when we multiply by HP in calculateDamage.
```
ThatIsAUsername: If we don't need to know about multiplying by 10 at this point in the code, I might just not mention it here. Too much detail at the wrong time (when it's not relevant, in particular) can be just as confusing as not enough detail.

ThislsAUsername: I put it there because I added a multiplication by 10 because I looked at what getDefLevel did.
That comment is there to keep me from doing the same dumb thing again.

ThatIsAUsername: Ah, fair enough then.

src/Engine/Path.java
```diff
@@ -157,6 +160,17 @@ public int getPathLength()
     return waypoints.size();
   }
 
+  public int getFuelCost(UnitModel model, GameMap map)
+  {
+    int cost = 0;
+    for (int i = 1; i < waypoints.size(); i++)
```
ThatIsAUsername: If you are gonna do something sneaky like iterate from 1, you should leave a comment to call it out and explain why, to keep future developers from trying to "fix" it.

src/Test/TestGameEvent.java
```diff
@@ -182,6 +182,7 @@ boolean testMoveEvent()
     Unit apc = addUnit(testMap, testCo1, UnitEnum.APC, 3, 2);
 
     Path path = new Path(1.0); // TODO: Why do we have to provide a speed here?
+    path.addWaypoint(3, 3); // we need two waypoints to not break compatibility with MoveEvent, since it assumes the first waypoint isn't used.
```
ThatIsAUsername: I would rephrase this to say that Paths should be defined to start where the Unit currently is (really, we should specify each intermediate point as well, so that fuel cost can be calculated correctly. Speaking of which, it would be good to have a test case to verify we are calculating fuel cost correctly).

ThislsAUsername: We're not checking fuel consumption, though, just trying to test if the unit actually moves when the event tells it to.
I could attempt to make the move event or fuel consumption check account for malformed paths, but that would likely break compatibility further.
I don't think a malformed path is a problem here, so accounting for the breaking change seemed like the most logical step.

ThatIsAUsername: This test wasn't made for fuel consumption/path formation, so I'm fine with leaving this as it is, but it wouldn't hurt to have another test to verify that we consume fuel correctly. I'm not too worried about the path formation thing at this point; we may build a test for that when we fix the GameAction issue we have.

PR #78, opened 2018-01-10
Base damage strategy

ThislsAUsername:
Implements the strategy pattern for base damage calculations.
It also fixes unit weapon choice.

Some questions:
Would we actually want it to be possible to change damage systems on the fly in the options menu? If so, then the option for that should probably be something other than a number specifying which to use.

How would we go about testing our strategy pattern? I made a function to find the deltas, and a decent conversion to string for comparison, but IDK what would be appropriate for an automated test.

I made Unit simply store which weapon it last checked damage for so it knows which weapon to deduct ammo from, but I feel like there should be a better way to design it that would remove any doubt of weird things happening and causing bugs.


src/Test/TestDamageStrategy.java
```diff
@@ -0,0 +1,138 @@
...
+
+  }
+
+  @Override
+  public boolean runTest()
+  {
+    setupTest();
+
+    boolean testPassed = true;
+    testPassed &= validate(true, "  Everything you know is wrong.");
```
ThatIsAUsername: Interesting test case.

src/Test/TestDamageStrategy.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * The values in the output array denote how much extra damage each interaction has with the new system compared to the standard. Output can be negative.
+   * @param other: which damage strategy to compare
+   * @return an array comprising the difference between the two damage strategy results
+   */
+  public int[][] getDeltas(DamageStrategy other)
+  {
+    int[][] deltas = new int[numWeaponModels][numUnitModels];
```
ThatIsAUsername: So, you might be able to get away with this in Java because it's garbage-collected, but in e.g. C++ I'd be real nervous about creating a "new" array and then returning it as an argument, unless the function name clearly indicated is was allocating memory that it did not intend to free (like in a factory class, for example). Not sure I care too much here (because again, Java), but just something to be aware of.

src/Test/TestDamageStrategy.java
```diff
@@ -0,0 +1,138 @@
...
+    {
+      out += type.toString() + ", ";
+    }
+    out += "\n\n";
+    for( int i = 0; i < numWeaponModels; i++ )
+    {
+      out += WeaponModel.WeaponType.values()[i].toString() + ": ";
+      for( int j = 0; j < numUnitModels; j++ )
+      {
+        out += deltas[i][j] + ", ";
```
ThatIsAUsername: The "print everything in a giant unformatted grid" approach is not the most readable - probably fine for now, but I might want to play with options at some point.
Also, just to make sure I'm reading this correctly - if I `getDelstaString` on `DSDamage`, it looks like `TANKCANNON`, `MD_TANKCANNON`, and `NEOCANNON` have negative values in the first two slots, which seems to imply that in either standard or DS damage strategies, cannons can shoot infantry?

src/Units/Unit.java
```diff
@@ -91,37 +92,30 @@ public double getBaseDamage(UnitModel target, int range, boolean moved)
     // if we have no weapons, we can't hurt things
     if( weapons == null )
       return 0;
+    double maxDamage = 0;
```
ThatIsAUsername: If we are now calculating the weapon that does the most damage (which is fine by me), then we should update the function comment to reflect that. It currently says we choose the first weapon that can do damage.

src/Units/Unit.java
```diff
@@ -91,37 +92,30 @@ public double getBaseDamage(UnitModel target, int range, boolean moved)
...
-      if( weapons[i].getDamage(target) != 0 )
-      {
-        break;
-      }
-    }
-    if( i == weapons.length )
+    if( null == currentWeapon )
       System.out.println("In " + model.name + "'s fire(): no valid weapon found");
-    weapons[i].fire();
+    currentWeapon.fire();
```
ThatIsAUsername: Ehhh, like you pointed out, the disconnect between `getBaseDamage` and `fire` makes me uneasy. Probably worth looking into an alternate approach here.

src/Units/UnitModel.java
```diff
@@ -15,11 +15,10 @@
     INFANTRY, MECH, RECON, TANK, MD_TANK, NEOTANK, APC, ARTILLERY, ROCKETS, ANTI_AIR, MOBILESAM, FIGHTER, BOMBER, B_COPTER, T_COPTER, BATTLESHIP, CRUISER, LANDER, SUB
   };
 
-  // NB: "Truck" is technically unnecessary, but I feel it may be useful for alternate damage systems.
-  // The rubric for what's a "truck" is whether the damage dealt to it by an infantry is >9%, not movetype.
+  // Subs are ships unless they're submerged.
   public enum ChassisEnum
   {
-    TROOP, TRUCK, TANK, SHIP, AIR_LOW, AIR_HIGH
+    TROOP, TANK, AIR_LOW, AIR_HIGH, SHIP, SUBMERGED
```
ThatIsAUsername: Why no more TRUCK? Just not enough to differentiate it from tanks?

src/Units/Weapons/Weapon.java
```diff
@@ -1,35 +1,19 @@
...
-      { 110, 110, 105, 105, 95, 90, 105, 105, 105, 95, 105, 0, 0, 0, 0, 75, 85, 95, 95 },
-      { 0, 0, 55, 55, 25, 20, 60, 65, 65, 25, 65, 0, 0, 0, 0, 25, 55, 25, 25 },
-      { 75, 75, 30, 6, 1, 1, 20, 25, 35, 6, 35, 0, 0, 65, 95, 0, 0, 0, 0 },
-      { 95, 90, 90, 80, 55, 50, 80, 80, 85, 85, 90, 0, 0, 0, 0, 50, 95, 95, 95 },
-      { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 90 },
-      { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 55, 65, 115, 115, 0, 0, 0, 0 },
-      { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 55, 25, 95, 55 } };
   public WeaponModel model;
   public int ammo;
+  public static DamageStrategy[] strategies = {new StandardDamage(), new DSDamage(), new DoRDamage(), new ChassisDamage()};
```
ThatIsAUsername: Would it make more sense to have this live in `CombatEngine`, or at least in the `Engine.Combat` package?

src/Units/Weapons/Damage/ChassisDamage.java
```diff
@@ -0,0 +1,43 @@
...
+import Units.UnitModel;
+import Units.Weapons.WeaponModel;
+
+public class ChassisDamage extends DamageStrategy
+{
+
+  // format is [attacker][defender]
+  // defenders: TROOP, TANK, AIR_LOW, AIR_HIGH, SHIP, SUBMERGED
+  private static int[][] damageChart = {
+      { 55, 25, 30,  0,  0,  0 }, // INFANTRYMGUN
```
ThatIsAUsername: Am I reading correctly that in this damage strategy, infantry can do 25 damage to any tank type? as in, equal damage to recons and md tanks? I'll assume that means this is a work in progress.

src/Test/TestDamageStrategy.java
ThislsAUsername: The AW2 damage table from Wars World News (made by Xenesis, the guy who does all the AW hacking) has damage numbers for tank cannons VS infantry.
We already had tanks shooting infantry with those numbers, I just fixed them to use machine guns.
Damage chart: https://www.warsworldnews.com/wp/aw2/game-aw2/damage-chart/

ThislsAUsername: FWIW, I'd really like to switch to AW3's damage chart for the time being, if we don't come up with a real damage system.

src/Units/UnitModel.java
ThislsAUsername: Basically, yes.

src/Units/Weapons/Weapon.java
ThislsAUsername: It's only used in Weapon, but I could see a case for that.

src/Units/Weapons/Damage/ChassisDamage.java
ThislsAUsername: ChassisDamage is a base for a more complex damage system that adds attributes to units.
Those are the maximum damage values a unit can inflict to each chassis type.

src/Units/Unit.java
ThislsAUsername: My problem is that I'm not sure what to do about it.
fire() doesn't have the proper information to make that determination, and it would be re-doing getDamage()'s work anyway.
I suppose we could add another boolean to getDamage that tells the unit "hey we're firing for real this time"?

src/Units/Weapons/Weapon.java
ThatIsAUsername: Eh, if it's only used here, this is probably where it should stay.

src/Test/TestDamageStrategy.java
ThatIsAUsername: What is better about AW3's chart?

src/Units/Unit.java
ThatIsAUsername: Made a new pull request to try and address this. Lemme know what you think.

src/Test/TestDamageStrategy.java
ThislsAUsername: Cruisers, mainly. They are effective anti-air (instead of being worse AAs) and take significantly less damage from air units (they take the most damage from Bcopters in AW2 of any ship, I believe.)
They can also engage other ships with their anti-sub torpedoes, which is good since that means it can attack the Ship chassis generically, instead of having to special-case somehow.

src/Units/Weapons/Damage/ChassisDamage.java
ThatIsAUsername: I'm... curious to see the more complex system, and how it extends this one. Carry on for now, I guess.

src/Units/Weapons/Damage/StandardDamage.java
```diff
@@ -0,0 +1,43 @@
+package Units.Weapons.Damage;
+
+import Units.UnitModel;
+import Units.Weapons.WeaponModel;
+
+public class StandardDamage extends DamageStrategy
```
ThatIsAUsername: Standard damage is what, AW2? Might be better to just call it that, since the others are named more specifically.

src/Units/Weapons/Damage/DSDamage.java
```diff
@@ -0,0 +1,48 @@
...
+   * @return returns its base damage against that unit type
+   */
+  public double getDamage(WeaponModel attack, UnitModel defender)
+  {
+    return damageChart[attack.getIndex()][defender.type.ordinal()];
+  }
+
+  public String getDescription()
+  {
+    return "AW3";
```
ThatIsAUsername: Maybe this should be "Dual Strike"

src/Units/Weapons/Damage/DoRChassisDamage.java
```diff
@@ -0,0 +1,63 @@
+package Units.Weapons.Damage;
+
+import Units.UnitModel;
+import Units.Weapons.WeaponModel;
+
+public class DoRChassisDamage implements DamageStrategy
```
ThatIsAUsername: This one isn't included in the menu option. Is that intentional?

src/Units/Weapons/Damage/DoRDamage.java
```diff
@@ -0,0 +1,49 @@
...
+   * @return returns its base damage against that unit type
+   */
+  public double getDamage(WeaponModel attack, UnitModel defender)
+  {
+    return damageChart[attack.getIndex()][defender.type.ordinal()];
+  }
+
+  public String getDescription()
+  {
+    return "DoR";
```
ThatIsAUsername: Consider "Days of Ruin". If this is supposed to be user-facing, it should be easy to read and understand.

src/UI/Art/SpriteArtist/SpriteOptions.java
```diff
@@ -17,12 +18,13 @@
...
   private static int drawScale = DRAWSCALE_DEFAULT;
 
-  private static Dimension dimensions = new Dimension(WINDOWWIDTH_DEFAULT*drawScale, WINDOWHEIGHT_DEFAULT*drawScale);
+  private static Dimension dimensions = new Dimension(WINDOWWIDTH_DEFAULT * drawScale, WINDOWHEIGHT_DEFAULT * drawScale);
 
   // Set up configurable options.
   private static GraphicsOption drawScaleOption = new GraphicsOption("Draw Scale", 1, 6, DRAWSCALE_DEFAULT);
   private static GraphicsOption dummyOption = new GraphicsOption("Dummy option");
-  private static GraphicsOption[] allOptions = {drawScaleOption, dummyOption};
+  private static GraphicsOption damageStrategyOption = new GraphicsOption("Damage Strategy", Weapon.stratDescriptions, 0);
```
ThatIsAUsername: Instead of "Strategy", maybe "System", or "Engine". I know you are using the Strategy Pattern, but that won't be clear to people who aren't into OO lingo.

src/Units/Weapons/Damage/BHRDamage.java
```diff
@@ -0,0 +1,48 @@
...
+   * @return returns its base damage against that unit type
+   */
+  public double getDamage(WeaponModel attack, UnitModel defender)
+  {
+    return damageChart[attack.getIndex()][defender.type.ordinal()];
+  }
+
+  public String getDescription()
+  {
+    return "AW2";
```
ThatIsAUsername: I'd go for "Black Hole Rising". There's space on the page, and I don't like throwing around acronyms without defining them, especially for user consumption.

PR #80, opened 2018-01-20
Weapon selection/use is now a bit cleaner, more idiomatic.

ThatIsAUsername:
An attempt to clean up the whole currentWeapon thing in Unit, and make weapon selection and use a bit more idiomatic.

### DIFF
--- a/src/Test/TestDamageStrategy.java
+++ b/src/Test/TestDamageStrategy.java
@@ -24,7 +24,6 @@ import Units.TankModel;
 import Units.UnitModel;
 import Units.UnitModel.UnitEnum;
 import Units.Weapons.WeaponModel;
-import Units.Weapons.Damage.ChassisDamage;
 import Units.Weapons.Damage.DamageStrategy;
 import Units.Weapons.Damage.StandardDamage;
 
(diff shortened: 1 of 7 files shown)


src/Test/TestDamageStrategy.java
```diff
@@ -24,6 +24,7 @@
 import Units.UnitModel;
 import Units.UnitModel.UnitEnum;
 import Units.Weapons.WeaponModel;
+import Units.Weapons.Damage.ChassisDamage;
```
ThislsAUsername: An extra unneeded import, I see :v

src/Units/Unit.java
```diff
@@ -83,39 +82,61 @@ else if( CO.money >= proportionalCost )
...
-  public void fire(final Unit defender)
+  /**
+   * Select the weapon owned by this unit that can inflict the
+   * most damage against the chosen target
+   * @param target
+   * @param range
+   * @param moved
+   * @return
+   */
+  public Weapon chooseWeapon(UnitModel targetType, int range, boolean moved)
```
ThislsAUsername: Should probably unify nomenclature with canAttack - make the boolean "afterMoving"

src/Test/TestDamageStrategy.java
ThatIsAUsername: Aw, snap. I added that to play with it locally, and added it to the commit on accident. On the other hand, perhaps it *should* be included. Maybe `TestDamageStrategy` should compare each type to `StandardDamage` or something, and give a warning if values are too far diverged. Might help with ensuring a new system is at least vaguely sane. What you think?

PR #83, opened 2018-03-27
The Seek key now also cycles through your available production facilities

ThislsAUsername:
We might wanna rename a variable or two still.


src/Terrain/Location.java
```diff
@@ -51,6 +51,16 @@ public boolean isCaptureable()
     return (environs.terrainType == Environment.Terrains.CITY || environs.terrainType == Environment.Terrains.FACTORY || environs.terrainType == Environment.Terrains.AIRPORT || environs.terrainType == Environment.Terrains.SEAPORT || environs.terrainType == Environment.Terrains.HQ || environs.terrainType == Environment.Terrains.LAB);
   }
 
+  /**
+   * @return true if this Location has an industry, false else.
+   */
+  public boolean isIndustry()
+  {
+    return (environs.terrainType == Environment.Terrains.FACTORY ||
```
ThatIsAUsername: Rather than explicitly checking a set of enum values, is there a more generic/change-proof way to check if the CO can build something from here? Maybe pass the CO as a parameter and then just return `!co.getShoppingList(environs.terrainType).isEmpty()` or something.

src/Engine/Utils.java
```diff
@@ -319,4 +320,35 @@ public int compare(SearchNode o1, SearchNode o2)
...
+   * Returns a list of all vacant industries a commander owns
+   */
+  public static ArrayList<XYCoord> findFreeIndustries(Commander co, GameMap map)
+  {
+    ArrayList<XYCoord> industries = new ArrayList<XYCoord>();
+    // we don't want to bother if we're trying to find nobody's properties :v
+    if( null != co )
+    {
+      // Add all vacant, <co>-owned industries to the list
+      for( int i = 0; i < map.mapWidth; i++ )
```
ThatIsAUsername: Ooh. I don't really like iterating over the entire map here, especially since you do it every time the `SEEK` button is hit. If you look at the `resupply` branch, I've changed all or most of the `Utils` functions to not do that.
I don't think we are super worried about performance, but it might be worth giving the commander a list of the property locations it owns; that would let us just iterate over that list here, and would also be useful for abilities that depend on/use owned properties.

src/Engine/GameInstance.java
```diff
@@ -56,6 +56,11 @@ public void setCursorLocation(int x, int y)
     cursorX = x;
     cursorY = y;
   }
+  public void setCursorLocation(XYCoord coord)
```
ThatIsAUsername: Whuh-oh. I also added this function in the `resupply` branch (I just added it before the overload instead of after like you did). Which might mean it won't conflict if we merge, but will cause build errors. Probably doesn't really matter for now.

src/Engine/MapController.java
```diff
@@ -166,7 +168,10 @@ private void handleMapInput(InputHandler.InputAction input)
       case SEEK: // Move the cursor to the next unit that is ready to move.
         boolean found = false;
         int tries = 0;
-        int maxTries = myGame.activeCO.units.size();
+        int numUnits = myGame.activeCO.units.size();
+        ArrayList<XYCoord> freeIndustries = Utils.findFreeIndustries(myGame.activeCO, myGame.gameMap);
```
ThatIsAUsername: As for naming, since you mentioned it. Maybe `UsableProperties` instead of `FreeIndustries`? It's a little awkward, but is at least more descriptive, I think.

src/Engine/Utils.java
```diff
@@ -319,4 +320,35 @@ public int compare(SearchNode o1, SearchNode o2)
...
     }
   }
+
+  /**
+   * Returns a list of all vacant industries a commander owns
+   */
+  public static ArrayList<XYCoord> findFreeIndustries(Commander co, GameMap map)
+  {
+    ArrayList<XYCoord> industries = new ArrayList<XYCoord>();
+    // we don't want to bother if we're trying to find nobody's properties :v
```
ThatIsAUsername: Sentence starts with lower case letter! I'm surprised this is even runnable.

src/Terrain/Location.java
ThislsAUsername: Yup. We can absolutely do that. Deleting this function and using that instead since if the Location has no CO... then it doesn't know whether it's an industry.
It's true that we only care whether a CO can buy from this territory, but that may not always be the case and doing it outside of this class makes that more explicit IMO.

src/Engine/MapController.java
```diff
@@ -195,11 +198,19 @@ private void handleMapInput(InputHandler.InputAction input)
             nextSelectedUnitIndex = 0;
           }
 
-          // If we find a unit that is ready to go, move the cursor to it.
-          Unit nextUnit = myGame.activeCO.units.get(nextSelectedUnitIndex);
-          if( !nextUnit.isTurnOver )
+          if( nextSelectedUnitIndex < numUnits )
```
ThatIsAUsername: `nextSelectedUnitIndex` isn't really correct anymore... But I'm not sure what a better option would be. Maybe `nextSeekLocation`?

src/Engine/MapController.java
```diff
@@ -186,7 +186,10 @@ private void handleMapInput(InputHandler.InputAction input)
       case SEEK: // Move the cursor to the next unit that is ready to move.
```
ThatIsAUsername: This comment should be updated to reflect that we cycle through both active units and usable properties.

src/Terrain/GameMap.java
```diff
@@ -22,32 +22,36 @@ public GameMap(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
     map = new Location[mapWidth][mapHeight];
 
     // Build the map locations based on the MapInfo data.
-    for(int y = 0; y < mapHeight; ++y)
+    for( int y = 0; y < mapHeight; ++y )
     {
-      for(int x = 0; x < mapWidth; ++x)
+      for( int x = 0; x < mapWidth; ++x )
       {
+        // We set HQs to labs here because we don't yet support using labs as HQs
```
ThatIsAUsername: Isn't this backwards? This is turning Labs into HQs.

ThatIsAUsername: And as a broader topic, weren't we talking about making Labs a production facility for experimental units? If we change all Labs to HQs, then we can't have Labs at all.

src/Terrain/Location.java
```diff
@@ -2,13 +2,15 @@
...
 import CommandingOfficers.Commander;
+import Engine.XYCoord;
 
 public class Location
 {
 
   private Environment environs = null;
   private Commander owner = null;
   private Unit resident = null;
+  private final XYCoord coords;
```
ThatIsAUsername: I'm not sure I want to add the coordinates as a member of the `Location` class; that already exists in the GameMap, and it's not good to repeat data when it's possible to avoid it. Could the Commander just store a list of `XYCoord`s instead of a list of `Location`s?
I know Unit has the same problem, but there didn't seem to be a good way to avoid it at the time, and there is some extra bookkeeping as a result.

src/Engine/MapController.java
```diff
@@ -186,7 +186,10 @@ private void handleMapInput(InputHandler.InputAction input)
...
         boolean found = false;
         int tries = 0;
-        int maxTries = myGame.activeCO.units.size();
+        int numUnits = myGame.activeCO.units.size();
+        ArrayList<XYCoord> usableProperties = Utils.findFreeIndustries(myGame.activeCO, myGame.gameMap);
+        int numFreeIndustries = usableProperties.size();
```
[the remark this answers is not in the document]
ThislsAUsername: Nope. Thought I might be changing more about it and then forgot.

src/Terrain/GameMap.java
ThislsAUsername: ...yup, it's backwards. Whoops.
We could just turn labs into HQs in the import script, but I'd like to support AWBW's use for labs.

src/Terrain/Location.java
ThislsAUsername: It is a good idea to avoid duplication of data, but I feel like the best place to add/remove properties from the CO's list is in the setter for Location.
You can't do that if the CO's list is a set of coordinates (well, I suppose you could, but you'd then still be stuck with storing those coordinates in Location)

ThatIsAUsername: As discussed elsewhere, I think I'm fine with this approach. It does create some data repetition, but does so in a controlled way, and we have that pattern elsewhere already.

src/Terrain/GameMap.java
ThislsAUsername: The labs are meaningless for capture victory if there are HQs present, though it does allow you to mix teams who must hold all HQs and teams who only have labs. Another point is that Labs can start neutral and be captured, while HQs typically just become cities when captured. We *could* support alternate game modes where that's not the case, I suppose?
Labs do serve another function in AWBW: you can soft-ban units as "Lab Units" so that only players who own a lab can build them. This facilitates interesting game modes like the FIRST Stronghold map (a 3v3) where only one player on each side can build footsoldiers; the other two players are stuck with their starting footies until they get a lab (which is roughly halfway through the enemy fort)
More flexibility is ultimately nicer, but the way AWBW uses them IMO gives a nice compromise between consistent mechanics and flexibility for the map maker.

PR #89, opened 2018-05-30
Removed coordinate info from Location.

ThatIsAUsername:
GameMap is once again the sole proprietor of this data.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -38,7 +38,7 @@ public class Commander
   public final CommanderInfo coInfo;
   public ArrayList<Unit> units;
   public ArrayList<UnitModel> unitModels;
-  public ArrayList<Location> ownedProperties;
+  public ArrayList<XYCoord> ownedProperties;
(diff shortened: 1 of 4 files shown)


ThislsAUsername:
The reason I wanted to put the array management into Location is because I wanted the CO property-ownership-arrays to always be current.
You can't, to my knowledge, restrict access to Location's setOwner() to GameMap's function.
Even making the map read-only wouldn't help, because you're just manipulating the members of a non-immutable contained element.
I guess you *could* make the map itself private, and access specific Locations via a function, but that's kinda awkward, and wouldn't technically fix it anyway.

Mostly, I'm looking for a way to minimize code duplication.

ThatIsAUsername:
Yeah, you are correct. This change by itself would create inconsistencies. Maybe the better approach is to remove getLocation() from the GameMap interface, and force everyone to go through GameMap to interact with positions that are on the map, but that would be a fairly large change.
We really haven't been worrying about proper data hiding so far; given that, and the degree of change it would take to fix this, and since we already have Unit storing its own location, your PR seems like a sensible enough approach.

Issue #90, opened 2018-06-02
Teams


ThatIsAUsername:
Right now we only support free-for-alls.
Allow selecting teams before starting, support shared vision (including stealths/subs), and prevent friendly fire.

PR #96, opened 2018-06-03
Teams

ThislsAUsername:
Solves #90

### DIFF
--- a/src/Test/TestCombat.java
+++ b/src/Test/TestCombat.java
@@ -67,7 +67,7 @@ public class TestCombat extends TestCase
     return testPassed;
   }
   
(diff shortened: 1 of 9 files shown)


src/Test/TestCombat.java
```diff
@@ -65,6 +66,37 @@ private boolean testUnitDeath()
 
     return testPassed;
   }
+  
+  /** Test that units actually die, and don't counter-attack when they are killed. */
```
ThatIsAUsername: Comment copy-pasta error.

ThatIsAUsername:
I like this, it looks good. Now we just need a way to choose teams...

PR #99, opened 2018-07-01
Classy terrain mod

ThatIsAUsername:
Per our offline discussions, here are my suggested modifications to the terrain reification effort.


src/Engine/GameAction.java
```diff
@@ -181,7 +182,7 @@ public CaptureAction(GameMap map, Unit actor, Path movePath)
         if( capture.willCapture() ) // If this will succeed, check if the CO will lose as a result.
         {
           // Check if capturing this property will cause someone's defeat.
-          if( (captureLocation.getEnvironment().terrainType.sustainsSide()) && (null != captureLocation.getOwner()) )
+          if( (captureLocation.getEnvironment().terrainType == TerrainType.HEADQUARTERS) && (null != captureLocation.getOwner()) )
```
ThislsAUsername: I still feel we should support the AWBW style of Lab use, at least somewhat.
Maybe that should just be the map converter turning Labs into HQs until we figure out better what we want, though?
(though having unit bans modified by Lab ownership is pretty cool too)

src/Terrain/Environment.java
```diff
@@ -1,49 +1,29 @@
...
 
-  public final BaseTerrain terrainType;
+  public final TerrainType terrainType;
   public final Weathers weatherType;
 
-  // ArrayList can dynamically expand to accomodate arbitrary terrains at runtime
-  // Thankfully, there's no plans to expand our list of weathers
-  private static ArrayList<Environment[]> tileInstances= new ArrayList<Environment[]>();
+  // Maintain a list of all tile types. Each type will be added the first time it is used.
+  private static Map<TerrainType, Environment[]> tileInstances= new HashMap<TerrainType, Environment[]>();
```
ThislsAUsername: Yeah, a Map is better than an Arraylist for this. Guess I'm just not used to having them around yet.

src/Terrain/TerrainType.java
```diff
@@ -0,0 +1,123 @@
...
+    mDefenseLevel = defense;
+    mMainColor = mainColor;
+  }
+  
+  ////////////////////////////////////////////////////////////////////////////////////////////////////////
+  //// Protected constants to initialize/parse the attribute flags, and the methods to query them.
+  private static final int LAND = 1;
+  private static final int WATER = LAND << 1;
+  private static final int CAPTURABLE = WATER << 1; // Whether a Commander can take ownership of this property.
+  private static final int PROFITABLE = CAPTURABLE << 1; // Whether this terrain type grants income to its owner each turn.
```
ThislsAUsername: I think it may be worth supporting a scaling factor on property income.
For instance, having "oil rigs" be more valuable than normal properties might actually entice people to use sea units.
Likewise, having more-valuable properties on land would allow a higher level of funding with less time spent capturing, and less map space devoted to the quite-defensive urban terrain.

src/Terrain/TerrainType.java
```diff
@@ -0,0 +1,123 @@
+package Terrain;
+
+import java.awt.Color;
+
+public class TerrainType
```
ThislsAUsername: FWIW, this is a lot more readable than I thought it would be, so props for that.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -166,90 +151,90 @@ public static TerrainSpriteSet getTerrainSpriteSet(Location loc)
...
     System.out.println("INFO: Loading terrain sprites for " + terrainType);
 
     TerrainSpriteSet ss = null;
     int w = baseSpriteSize;
     int h = baseSpriteSize;
 
-    // This is for sure kinda ugly, but not much more than the equivalent switch/case
-    // This *could* be moved into BaseTerrain, but I feel like perhaps that would be too close integration of model and view (though it'd make it nice and simple in the short term?)
-    if( terrainType == Bridge.getInstance() )
+    // Load the appropriate images and define the necessary relationships for each tile type.
```
ThislsAUsername: Perhaps having integer indices wouldn't be a bad idea...
Switch/case is IMO a bit more readable than if/else, and you seem determined to keep UI-specific knowledge (sprites and such especially) out of TerrainType.
Thoughts?

src/UI/Art/SpriteArtist/SpriteMiniMapArtist.java
```diff
@@ -55,7 +41,7 @@ private static BufferedImage generateMapImage(MapInfo mapInfo)
     {
       for(int x = 0; x < image.getWidth(); ++x)
       {
-        g.setColor( terrainColors.get(mapInfo.terrain[x][y].index) );
+        g.setColor( mapInfo.terrain[x][y].getMainColor() );
```
ThislsAUsername: Perhaps I should have looked more closely at this before my initial change... oh, well. Good catch.

src/UI/Art/SpriteArtist/TerrainSpriteSet.java
```diff
@@ -313,12 +336,37 @@ private boolean checkTileType(GameMap map, int x, int y, boolean assumeTrue)
...
+      terrainBases.put(TerrainType.SEA, TerrainType.SEA);
+
+      if( TerrainType.TerrainTypeList.length != terrainBases.size())
+      {
+        System.out.println("WARNING! TerrainSpriteSet terrains do not align with Environment!");
+      }
+    }
 
-    return Environment.getTerrainTypes().get(baseIndex);
+    return terrainBases.get(terrain);
```
ThislsAUsername: This is definitely more compact than a switch/case, but it's roughly as hard to maintain.
With all the testinfo and such printed in the console now, one warning probably won't be noticed much.
Then again, it'll probably crash on returning null, so maybe that's not a problem...

src/Units/APCModel.java
```diff
@@ -8,15 +8,18 @@
...
+  private static final int MAX_FUEL = 70;
+  private static final int IDLE_FUEL_BURN = 0;
+  private static final int MOVE_POWER = 6;
   private static final MoveType moveType = new Tread();
   private static final ActionType[] actions = { ActionType.RESUPPLY, ActionType.UNLOAD, ActionType.WAIT };
 
   public APCModel()
   {
-    super("APC", UnitEnum.APC, ChassisEnum.TANK, 5000, 70, 0, 6, moveType, actions, null);
+    super("APC", UnitEnum.APC, ChassisEnum.TANK, UNIT_COST, MAX_FUEL, IDLE_FUEL_BURN, MOVE_POWER, moveType, actions, null);
```
ThislsAUsername: Good change for readability.

src/Units/BomberModel.java
```diff
@@ -1,17 +1,24 @@
...
 import Engine.GameAction.ActionType;
+import Units.MoveTypes.Flight;
+import Units.MoveTypes.MoveType;
 import Units.Weapons.BomberBombs;
 import Units.Weapons.WeaponModel;
 
-public class BomberModel extends AirModel
+public class BomberModel extends UnitModel
 {
+  private static final int UNIT_COST = 2200;
```
ThislsAUsername: I suppose I ought to get around to changing these to the DoR numbers?
Maybe when/if we add fog vision radius?

src/Units/MechModel.java
```diff
@@ -7,15 +7,19 @@
 import Units.Weapons.MechZooka;
 import Units.Weapons.WeaponModel;
 
-public class MechModel extends LandModel
+public class MechModel extends UnitModel
 {
+  private static final int UNIT_COST = 16000;
+  private static final int MAX_FUEL = 50;
+  private static final int IDLE_FUEL_BURN = 0;
+  private static final int MOVE_POWER = 5;
```
ThislsAUsername: "Looks like a Medium Tank to me, sir"

src/Engine/GameAction.java
ThatIsAUsername: Not sure how unit bans work, but my thoughts are: If we want to support alternate game modes (multiple HQs, capture all/just one to win, etc), we can do that. However, I think those rules are a property of the game mode, not of the TerrainType. We can also make new terrain types if we want, so there's no need to overload labs or whatever to support it.

src/Terrain/TerrainType.java
ThatIsAUsername: How would you capture oil rigs? Maybe just ensure there is always shoal nearby? It's an interesting idea, anyway.
We can always have income scaling be a game attribute, but I assume you mean allowing different properties to provide different amounts of income. I'm not necessarily opposed (maybe add mines along with oil rigs?), but I think that should be a future change, if/when we decide we want to add that complexity.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Hah, this after you were so dead-set on removing the switch. At this point, I don't think I want to go back to enums, and adding integer indices just so we can do a switch feels backwards to me. I agree it isn't quite as nice and compact as the switch statement was, but it's not too horrible.

src/UI/Art/SpriteArtist/TerrainSpriteSet.java
ThatIsAUsername: Yeah, I'd actually like to do some work cleaning up logging (or making the logging conditional, so we can see logging for specific subsystems we care about or whatever) at some point. As you say, this is also likely to be near the end of the console output before a stack trace, so other noise may not matter. I guess we could just be proactive and throw the exception ourselves here if we really wanted.

src/Units/MechModel.java
ThatIsAUsername: Gah, caught me. Guilty as charged.

src/Engine/GameAction.java
ThislsAUsername: Unit bans can be either a ban or a "lab unit"
Lab units can only be built by players who own a lab currently.
So yes, it actually is kind of tied to the terrain type.
A decent number of AWBW maps mix players with only labs and players with only HQs so that only the lab players only have access to certain unit types, or maybe they'll have neutral labs that *can* be captured in order to gain access to the unit types in question.
Another point that may be worth mentioning is Labs don't give income, so a player can be given a number of them at the start (for e.g. a multi-capture required victory) without changing their day to day income.

src/Terrain/TerrainType.java
ThislsAUsername: Yes, I meant for certain properties to be more valuable (or even less valuable) than others.
My thought was oil rigs would give 3x income, but only while actively occupied by a sea unit.
You wouldn't capture them, and you'd need at least 2 vessels to guarantee income every turn, and likely an APC... but the sea units would also hopefully be useful on their own, so it's not just a sunk cost thing. (This was in my earlier sea-combat pitch, which I will likely revisit once I'm satisfied with the damage system)
Mines... If you mean the places you dig for metal and stuff, I guess they could grant extra income but not heal?
If you mean the kind that explode when you walk over them, uhhh, that'd be interesting...

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: I'll live, I guess.

src/UI/Art/SpriteArtist/TerrainSpriteSet.java
ThislsAUsername: Throwing the exception here sounds much like the approach you took with the testing; you don't get to play the game if it's currently broken; also the fail fast philosophy.

ThatIsAUsername: Fair enough. I think I'll leave it as an unchecked exception though, since it'll fail anytime you try to start a game, and whether it will pass is determined at compile time; it shouldn't be possible to "release" the game with this broken.

src/Terrain/TerrainType.java
ThatIsAUsername: Why would you need two ships to guarantee income every turn? And just... any ship would do?
I'd potentially be open to introducing either type of mine at some point, but I think we should nail down the base game mechanics before we start adding new features.

ThislsAUsername: You need two ships because ships passively run out of fuel, and APCs can't swim.
I was thinking any ship would do; you're likely to use the less expensive ones, but IMO there's no reason to bar the more expensive ones. It makes intuitive sense that a footsoldier can do more with urban areas than a tank... but ships are just ships.
Fair enough.

src/Engine/GameAction.java
ThislsAUsername: Labs are actually used as HQs in the carts (see: the lab missions), and the technology aspect is attractive also.
I'm mildly opposed to making Labs unable to take at least either role, due to map compatibility, but I guess it's not that big of a deal since find/replace is pretty easy.
That said, I'm not at all opposed to having other buildings with just one of the aspects.

PR #102, opened 2018-07-05
Added support for rivers, along with a map that uses them.

ThatIsAUsername:


ThatIsAUsername:
Fixes #101

src/Terrain/TerrainType.java
```diff
@@ -96,6 +96,11 @@ private TerrainType(int attributeFlags, int defense, Color mainColor)
   private static final Color REEF_COLOR = new Color(218, 152, 112);
   public static final TerrainType REEF = new TerrainType( REEF_FLAGS, REEF_DEFENSE, REEF_COLOR );
 
+  private static final int RIVER_FLAGS = LAND; // It's not really a WATER type, since boats can't go here.
```
ThislsAUsername: It might be amusing to change that, but fair enough.

ThislsAUsername:
I like the new factory, it's much less city-y

ThatIsAUsername:
Hah, I was wondering if you would notice the new factory. Pretty sure you missed it when I swapped out the HQ a little while back...

ThislsAUsername:
It shows up in the diffs. How would I miss it?
I just prefer to avoid criticizing art, since there's so much preference tied up in it. (though I guess code style is much the same)

PR #104, opened 2018-07-14
Cmdr abilities

ThatIsAUsername:
Creates UI elements to display each Commander's current power level and whether any abilities are available.
Introduces the CommanderAbility class, and gives Patch and Strong some actual abilities.
Introduces an event listener to subscribe to in-game events.
Does not provide Commanders with a way to gain ability power.


src/CommandingOfficers/Commander.java
```diff
@@ -191,15 +202,63 @@ public UnitModel getUnitModel(UnitEnum unitType)
...
+
+  /** Lets the commander know that he's using an ability,
+   *  and accounts for the cost of using it. */
+  public void activateAbility(CommanderAbility ability)
+  {
+    modifyAbilityPower(-ability.getCost());
+    myActiveAbilityName = ability.myName;
+  }
+
+  protected void modifyAbilityPower(double amount)
```
ThislsAUsername: I feel like maybe the amount of charge the commander can hold in total should be an independent stat, rather than "whatever the most expensive one is"
This should work as well, though.

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -1,40 +1,163 @@
...
+    @Override
+    protected void perform(GameInstance game)
+    {
+      myCommander.addCOModifier(damageMod);
+
+      // Make bazookas buildable from all production buildings.
+      UnitModel mechModel = myCommander.getUnitModel(UnitModel.UnitEnum.MECH);
+      UnitProductionModifier upm = new UnitProductionModifier(TerrainType.AIRPORT, mechModel);
+      upm.addProductionPair(TerrainType.SEAPORT, mechModel);
+      upm.addProductionPair(TerrainType.HEADQUARTERS, mechModel);
```
ThislsAUsername: No infantry from HQ?

src/CommandingOfficers/Modifiers/UnitRemodelModifier.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Map;
+
+import CommandingOfficers.Commander;
+import Units.Unit;
+import Units.UnitModel;
+import Units.Weapons.Weapon;
+
+/** Modifier to temporarily turn one unit into another kind of unit.
+ *  This only applies to active units (newly-built units will not be changed).
+ *  There are definitely no plans to use this to make werewolves. */
```
ThislsAUsername: Oh, good. I'm glad we got that out of the way.

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -1,40 +1,163 @@
...
+      moveMod.addApplicableUnitType(UnitModel.UnitEnum.MECH);
+      myCommander.addCOModifier(moveMod);
+
+      // Grant transports extra capacity
+      UnitRemodelModifier urm = new UnitRemodelModifier();
+      for( UnitModel.UnitEnum type : highCapacityUnitModels.keySet() )
+      {
+        urm.addUnitRemodel(myCommander.getUnitModel(type), highCapacityUnitModels.get(type));
+      }
+      myCommander.addCOModifier(urm);
```
ThislsAUsername: So uh, it's worth noting that swapping in the high capacity unit models actually nerfs their movement range.
The high capacity models do not have even the D2D move boost, and don't benefit from the movement boosts either.

I suppose the simplest solution is to just make separate copies of the remodeled units.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Yeah, I considered that, but it seems a bit silly to be able to stockpile more power than you can use. I could certainly contrive a use case if I really wanted to... but eh. This is sane and simple.

src/CommandingOfficers/CommanderStrong.java
ThatIsAUsername: He gets infantry from HQ by default. Which is maybe a bit much, so maybe we should change that, and put it here instead. Thoughts?

src/CommandingOfficers/CommanderStrong.java
ThatIsAUsername: hrm. Good catch. I should just make it possible to clone UnitModels.

ThatIsAUsername: Done.

src/Engine/GameEvents/GameEventListener.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /** Sign this listener up to receive events. If a listener registers multiple times, it will still
+   *  receive each notification only once. */
+  public static void registerEventListener(GameEventListener listener)
+  {
+    eventListeners.add(listener);
+  }
+
+  /** Unregister this listener. Call this when a listener is no longer needed, so the JVM knows
+   *  it can clean up the object. */
```
ThislsAUsername: Could also use weak references, but ensuring clean code is also good.

ThatIsAUsername: This is a good thought. We'd still need the `unregisterEventListener` though, since some listeners (e.g. as in Patch's ability) may be unregistered and re-registered instead of being destroyed and re-created.
I suppose we could still use weak references as a backup mechanism, to handle the case where someone forgets to unregister a listener before nulling it. I'm not sure about the behavior of garbage collecting in this case though; could a forgotten listener stick around and affect gameplay until the garbage collector deals with it? If so, it would make it harder to track down this kind of bug, because the problem would just disappear at some indeterminate time.
Of course we could always call `System.gc()` before calling the listeners, but that's exactly the kind of thing we aren't supposed to have to deal with in Java.

ThatIsAUsername: Some brief testing indicates that once set to null, a listener won't be called, even if it hasn't been garbage-collected yet. I think I'm fine with making this change, and seeing how things go if you think its worthwhile. Thoughts?

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -1,40 +1,160 @@
...
+      upm.addProductionPair(TerrainType.SEAPORT, mechModel);
+      upm.addProductionPair(TerrainType.HEADQUARTERS, mechModel);
+      myCommander.addCOModifier(upm);
+
+      // Grant foot-soldiers and transports additional movement power.
+      COMovementModifier moveMod = new COMovementModifier();
+      moveMod.addApplicableUnitType(UnitModel.UnitEnum.APC);
+      moveMod.addApplicableUnitType(UnitModel.UnitEnum.INFANTRY);
+      moveMod.addApplicableUnitType(UnitModel.UnitEnum.MECH);
+      myCommander.addCOModifier(moveMod);
```
ThislsAUsername: Yo bro, this doesn't actually work still.
You told me to tell you.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -0,0 +1,248 @@
...
+    {
+      sb.append(" ");
+    }
+    if( cmdr.money < 10 ) // Fewer than 2 digits. You poor.
+    {
+      sb.append(" ");
+    }
+    sb.append(Integer.toString(cmdr.money));
+
+    return sb.toString();
```
ThislsAUsername: I feel like there's a string format function to produce a fixed width result, and I don't like excess if statements.

src/CommandingOfficers/CommanderStrong.java
ThatIsAUsername: Good catch. The `COMovementModifier` was being applied to Strong's normal `UnitModel`s, not the high-capacity ones, so APCs weren't actually getting the benefit (but infantry/mech were).

`COMovementModifier` now accepts `UnitModel`s to operate on instead of unit types, which is more explicit.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThatIsAUsername: Fixed.

src/CommandingOfficers/CommanderStrong.java
ThatIsAUsername: Yeah, I think an extra infantry-production building might be a bit much in the early game. I'm going to push that to his abilities only.

PR #111, opened 2018-11-30
Power charging

ThislsAUsername:
The whitespace changes are from formatting the files in question per our code style definition.


ThatIsAUsername:
I have a little hesitation to having Commander inherit from GameEventListener, but I can see why it's desirable. We can see how it pans out.

ThislsAUsername:
> 
> 
> I have a little hesitation to having Commander inherit from GameEventListener, but I can see why it's desirable. We can see how it pans out.

This was why I mentioned the possibility of making GameEventListener a set of interfaces.
On one hand, if we do eventually want to do that, we're building technical debt.
On the other hand, we don't have a solid use case for it yet.

PR #113, opened 2018-12-22
AI Framework

ThatIsAUsername:
Adds the ability for an AI to play the game, with a do-nothing AI that only builds Infantry (but doesn't move or do anything else).

Next steps will be to make an AI that can attempt to play for real, and a way to set a player to AI control.


src/AI/InfantrySpamAI.java
```diff
@@ -0,0 +1,62 @@
...
+
+import CommandingOfficers.Commander;
+import Engine.GameAction;
+import Terrain.GameMap;
+import Terrain.Location;
+import Terrain.TerrainType;
+import Units.UnitModel;
+
+/**
+ *  Just build tons of Infantry and try to rush the opponent.
```
ThislsAUsername: It really appears as if "rushing" part doesn't happen, but maybe that's just me.

ThislsAUsername:
Looks pretty good overall, though it seems to freak out and freeze if all the players are AI.

ThislsAUsername:
I'm a silly goose, so I tried implementing the "rush" part... getNextAction() becomes an infinite loop.
The reason it halts in the build-units-only case is that it starts creating malformed gameActions.

It *also* halts if I manually set each unit's turn to be over when making its action... and I'm not sure why that should make a difference.

I feel like a better way to handle this might be a state machine: move/attack phase, build phase. The CO'd have to notify it when to reset to the start again, but that's good IMO due to separation of concerns. Having separate functions for initialization and execution seems like a good plan.

That said, I threw up my hacked-together code on a new branch "aiBugginess" if you wanna take a gander.

src/AI/InfantrySpamAI.java
ThatIsAUsername: Not yet. I've just about got it working on my end, but I was thinking I might make it a separate pull request to keep any one from being too huge. Maybe I'll roll it into this one, I dunno.

ThatIsAUsername:
> Looks pretty good overall, though it seems to freak out and freeze if all the players are AI.

Yup. It currently doesn't pause for animation while polling for AI actions, so it just disappears into itself. I've got a tentative fix, but I'm considering ways to make it less janky.

ThatIsAUsername:
> I'm a silly goose, so I tried implementing the "rush" part... getNextAction() becomes an infinite loop.
> The reason it halts in the build-units-only case is that it starts creating malformed gameActions.
> 
> It _also_ halts if I manually set each unit's turn to be over when making its action... and I'm not sure why that should make a difference.
> 
> I feel like a better way to handle this might be a state machine: move/attack phase, build phase. The CO'd have to notify it when to reset to the start again, but that's good IMO due to separation of concerns. Having separate functions for initialization and execution seems like a good plan.
> 
> That said, I threw up my hacked-together code on a new branch "aiBugginess" if you wanna take a gander.

You probably got the infinite loop because you aren't checking `unit.isTurnOver' before creating actions for the guy. I ran into this too; I added `isTurnOver` checks in all the GameActions (locally), so that'll just fail now.

A state machine is a good idea; the current local version of the AI does basically what you suggest, though it doesn't use a proper state machine to do it.

ThatIsAUsername:
I resolved _most of_ the above - it's not ready for primetime yet, but if you just want to watch the dumb AI's duke it out, look at the `perpetualMotion` branch (right now, Patch is always AI).

ThislsAUsername:
> You probably got the infinite loop because you aren't checking `unit.isTurnOver` before creating actions for the guy. I ran into this too; I added `isTurnOver` checks in all the GameActions (locally), so that'll just fail now.

I totally was, though. The specific check is `if( unit != null && unit.CO == myCo && unit.isTurnOver == false )`

Also, at some point we probably want to *not* start everyone with 10k funds, lol.

ThatIsAUsername:
> > You probably got the infinite loop because you aren't checking `unit.isTurnOver` before creating actions for the guy. I ran into this too; I added `isTurnOver` checks in all the GameActions (locally), so that'll just fail now.
> 
> I totally was, though. The specific check is `if( unit != null && unit.CO == myCo && unit.isTurnOver == false )`
> 
> Also, at some point we probably want to _not_ start everyone with 10k funds, lol.

Ah. Not sure then. If it's still an issue when I get the AI updates pushed up, then we'll have a problem.

And yes. Soon. Maybe right after we get some kind of proper AI in.

ThislsAUsername:
I feel like maybe we should leave this unmerged until we have a proper design for interleaving animation and action?
The hack with throwing a no-op down the pipe every frame is kind of questionable, heh.

ThatIsAUsername:
> I feel like maybe we should leave this unmerged until we have a proper design for interleaving animation and action?
> The hack with throwing a no-op down the pipe every frame is kind of questionable, heh.

Hah, I'm open to suggestions. I think we should keep it unmerged until we merge in the finished `InfantrySpamAI`, anyway, since that's going to involve some architectural adjustments either way.

But for this issue specifically: The issue is that if we just prevent a return from `MapController` until the AI finishes executing, the UI freezes and the actions won't be applied to the game (Events are not "applied" until after they are animated).
So we need a way to start an event, wait an arbitrary (but precise) amount of time so it can animate, then give `MapController` control of the thread again so it can fetch the next AI action.
I guess we could give `IController` a secondary function `pokeForStatus` or whatever to avoid the no-op calls to `handleInput`, but I don't like that much either. Thoughts?

ThislsAUsername:
If we don't want to re-architect anything at all, just slap this bad boy in at the end of animationEnded():
 ```
        if( myGame.activeCO.isAI() )
        {
          handleInput(InputHandler.InputAction.BACK);
        }```
It seems to work for everything without dumping 60 keypresses down the pipe per second.
You can also do it indirectly through Driver by throwing an event at it through the gameWindow, but that code is basically equivalent to the above with extra useless indirection. It also probably requires the gameWindow to be public.

Another solution might be to give IController an "isAutonomous" field that Driver checks in actionPerformed() and passes whatever in, and MapController unsets it whenever it hasn't just ordered an AI action.

ThislsAUsername:
> Hmm. I don't think it really makes sense to add an `isAutonomous` field to `IController`, since `MapController` is not the only class that implements it.
MapController is also the only one of them that currently wants to act autonomously from the user.
IMO there's no reason to require all items implementing the feature use all parts of the interface. You even agree with me, based on your implementation of GameActionListener.
 
> Your code snippet seems to work, but using `InputAction.BACK` causes it to cancel some of the animations. This can be fixed by using `InputAction.NO_ACTION` instead, which allows all animations to play out correctly. However, this still has the issue of extending the call stack by a multiple of the number of actions the AI performs until the next human player allows it all to unroll. With only AI teams and a large map, I expect you won't finish a game due to stack overflows.
I'm not 100% sure, but I think it actually *doesn't* extend the stack like that.
It does cause the stack to get a bit longer, but handleInput() seems to terminate before animationEnded() is called again.
I went through animationEnded(), through handleInput(), and hit Driver's actionPerformed() before I got back to animationEnded()... which is where your hook comes from.
In addition, my initial testing with Europe led to blue conquering everybody but salmon and then both sides just getting kinda mutually distracted by various properties that they couldn't really get to. If the call stack was gonna blow up, I think it would have by then. I'm letting it run overnight though, just to be sure.
If I'm wrong, do tell.

> While it does seem like a bit of a hack, I'm still leaning towards the "I just finished a frame of animation, check if anything else needs to happen" approach I have in place now (though maybe repackaging it somewhat could be in order?).
Whatever else happens, we need to get rid of that log message saying invalid input received, if we're gonna be passing in no-ops.
IMO we really need some sort of check to determine whether the input should go in. These input handler functions aren't trivial and with your implementation, my laptop fan goes full blast all the time on Spann Island even with no AI. (With mine, it's stays low on Europe with 4 AIs)

ThatIsAUsername:
> IMO there's no reason to require all items implementing the feature use all parts of the interface. You even agree with me, based on your implementation of GameActionListener.
Eh, yes and no. `GameEventListener` methods are all at least related - any subclass can use whichever ones they care about, but the purpose of the unused methods is clear. If a function doesn't make sense for 90% of classes implementing an interface, it probably doesn't belong in that interface (and/or there's a larger architectural snarl to work out).

> I'm not 100% sure, but I think it actually _doesn't_ extend the stack like that.
You are correct. I didn't look closely enough the first time around. On finishing an animation, the call to `handleInput` just fetches the next action, queues its animation, and then unrolls the stack completely. This works as desired, though there might be a way to package it more nicely/intuitively.

> In addition, my initial testing with Europe led to blue conquering everybody but salmon and then both sides just getting kinda mutually distracted by various properties that they couldn't really get to.
Yeah, that's the last major thing on my list to declare basic AI functionality.

> Whatever else happens, we need to get rid of that log message saying invalid input received, if we're gonna be passing in no-ops.
Done. Agreed, that was annoying. I think I'll suppress that message for `NO_ACTION`s whether or not we switch the AI-trigger mechanism.

> IMO we really need some sort of check to determine whether the input should go in. These input handler functions aren't trivial and with your implementation, my laptop fan goes full blast all the time on Spann Island even with no AI. (With mine, it's stays low on Europe with 4 AIs)
When the input isn't used, it's just extends the stack by a few function calls. It surprises me that it would make that much of a difference in CPU utilization, and it definitely shouldn't affect memory. I might try to quantify this, or at least ballpark it.

ThatIsAUsername:
Some initial testing on my Linux desktop (Europe, AIx4, run for ~5 min) shows that the heartbeat method does run a bit higher (15%-ish CPU, 3%-ish Memory) than the animation-end tail-hook approach (13%-ish CPU, 2.4%-ish memory), but it's not crazy more - odd that your fan feels threatened by an extra couple-percent duty cycle at under 20% use.
Interestingly, making the game full-screen added quite a bit of CPU usage (~6% and ~4%, respectively) that didn't go away after downsizing the window again.
Of course that's just a one-off test, but it does seem to make some difference.

ThislsAUsername:
> > IMO there's no reason to require all items implementing the feature use all parts of the interface. You even agree with me, based on your implementation of GameActionListener.
> 
> Eh, yes and no. `GameEventListener` methods are all at least related - any subclass can use whichever ones they care about, but the purpose of the unused methods is clear. If a function doesn't make sense for 90% of classes implementing an interface, it probably doesn't belong in that interface (and/or there's a larger architectural snarl to work out).

It would apply any time a controller wants to do an action with multiple animation steps that doesn't involve human interaction. Effectively, the interface goes from an "input controller" to a "game controller". Usually, the game only cares about input because it's menus and turn based stuff, but that's not a strict requirement.
It makes sense, they just don't have a use case for it... same as for EventHandler.

> On finishing an animation, the call to `handleInput` just fetches the next action, queues its animation, and then unrolls the stack completely. This works as desired, though there might be a way to package it more nicely/intuitively.

I feel like there is, but I kinda smell a significant rewrite.

> > mutually distracted by various properties that they couldn't really get to.
> 
> Yeah, that's the last major thing on my list to declare basic AI functionality.

My implementation on aiBugginess shouldn't have that problem, though it obviously has other shortfalls.
I feel like we should check if the endpoint is reachable at all, so we don't get blobs of infantry trying to capture the same place while one of them is on it.
Another thing I'd like is a multi-pass system where if the unit doesn't do anything it can be re-evaluated instead of just ending turn.
This is stuff I'm happy to do, but the AI's currently your baby since you've got the PR, lol.
...I may or may not just keep poking at the AI for a while once we get this stuff merged, heh.

> Some initial testing on my Linux desktop (Europe, AIx4, run for ~5 min) shows that the heartbeat method does run a bit higher (15%-ish CPU, 3%-ish Memory) than the animation-end tail-hook approach (13%-ish CPU, 2.4%-ish memory), but it's not crazy more - odd that your fan feels threatened by an extra couple-percent duty cycle at under 20% use.

My fan is set to pre-emptive cooling, so its "threat" radar is much more sensitive than average.

> Interestingly, making the game full-screen added quite a bit of CPU usage (~6% and ~4%, respectively) that didn't go away after downsizing the window again.

Test results from the overnight thing: I left it original size for the duration, and Europe was clocking in at ~155MB of memory. After Blue coated 1/3rd of Europe with infantry for ~9 hours, that number had climbed to ~175MB. Maximizing the window brought it up to ~205MB, which did not drop down again.
It was at 2-5% CPU normally, going up to 17-20% CPU maximized. This caused my fan to start going full blast after a few seconds. The CPU use did revert on re-smallification.

ThislsAUsername:
Okay, I'm looking at perpetualMotion again, and I'm not reproducing the "fan going crazy" thing with the heartbeat with Task Manager up.
CPU use *is* up slightly, but it's staying under 5%
I won't say it's impossible there was some other application pushing the CPU usage up just enough to get the fan going.

Issue #57, opened 2017-12-12
It'd be nice to have an expected-damage display


ThislsAUsername:
Should account for attacker and defender HP, as well as any modifiers.

Should be? relatively straightforward with the BattleSummary class generated by CombatEngine.

PR #117, opened 2018-12-25
Added popup damage preview.

ThislsAUsername:
This is obviously kind of terrible; a bunch of low-level UI code just dumped in SpriteMapView.
The percent sign doesn't even show up.
It also is potentially questionable to make getMapUnitColors() public.

Mostly, I'm making the PR for feedback on where the code *should* go.


ThislsAUsername:
Fixes #57

ThatIsAUsername:
> The percent sign doesn't even show up.

Hah! That's because I didn't include non-alphanumeric characters in the pixel fonts I made. We might have to add some of those, as well as the mapping needed to use them.

ThatIsAUsername:
I made a PR to add the missing symbol characters. As far as direction; this may not be pretty, but it works well enough, and none of the UI code I've written so far is particularly pretty either.

Some things you could do to clean it up a little: Instead of doing the nitty-gritty drawing stuff right there, you could add `SpriteLibrary.buildFrame()` or `getTextFrame()` or whatever you want to call it, that would return a frame of the correct size, with optional text included. This would also be useful in other places (e.g. ResupplyAnimation), and obviate the need to make `getMapUnitColors()` public.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -302,6 +305,45 @@ public void render(Graphics g)
...
+            int y = signTopLeft.yCoord;
+            int w = signWidth;
+            int h = signHeight;
+            Color[] colors = SpriteLibrary.getMapUnitColors(currentActor.CO.myColor).paletteColors;
+            mapGraphics.setColor(colors[2]); // frame
+            mapGraphics.fillRect(x, y, w, h);
+            mapGraphics.setColor(colors[5]); // highlight
+            mapGraphics.fillRect(x + 1, y + 1, w - 1, h - 1);
+            mapGraphics.setColor(colors[4]); // BG
+            mapGraphics.fillRect(x + 1, y + 1, w - 2, h - 2);
```
ThatIsAUsername: The border of the sign is always 1 pixel, regardless of drawScale. This is very apparent if you bump it up to 4x instead of the usual 2x. Yes, it's a problem in `ResupplyAnimation` as well, which is another reason to try and consolidate our pop-up drawings to one place, whether it's a new class or just in `SpriteLibrary`.

PR #119, opened 2018-12-25
Add COs

ThislsAUsername:
Adds two of my pet ideas that I've been tossing around for a while.
Bear & Bull's COP would interact weirdly with Cinder if Cinder opened a build menu without building anything, but I'm not sure at this point how to fix that.

I had to make some changes to other stuff to support their abilities...
Unit cost itself being unmodifiable makes sense, though I can see potential use cases. Having a multiplier with free access seems truer to how AW COs typically work, though.
The change to getShoppingList()'s arguments make sense, and actually reduce code size by a small amount.
Change to AlterHP doesn't hurt anything, and is useful.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -201,9 +201,9 @@ public class Commander extends GameEventListener
   }
 
   /** Get the list of units this commander can build from the given property type. */
-  public ArrayList<UnitModel> getShoppingList(TerrainType buyLocation)
+  public ArrayList<UnitModel> getShoppingList(Location buyLocation)
   {
-    return (unitProductionByTerrain.get(buyLocation) != null) ? unitProductionByTerrain.get(buyLocation)
+    return (unitProductionByTerrain.get(buyLocation.getEnvironment().terrainType) != null) ? unitProductionByTerrain.get(buyLocation.getEnvironment().terrainType)
         : new ArrayList<UnitModel>();
   }
 
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -45,7 +45,7 @@ public class Commander extends GameEventListener
   public ArrayList<Location> ownedProperties;
   public ArrayList<COModifier> modifiers;
   public Color myColor;
-  public static final int DEFAULTSTARTINGMONEY = 10000;
+  public static final int DEFAULTSTARTINGMONEY = 0;
   public static final int CHARGERATIO_FUNDS = 9000; // quantity of funds damage to equal 1 unit of power charge
   public static final int CHARGERATIO_HP = 90; // quantity of HP damage dealt to equal 1 unit of power charge
   public int money = 0;
@@ -249,7 +249,7 @@ public class Commander extends GameEventListener
   public void activateAbility(CommanderAbility ability)
   {
     modifyAbilityPower(-ability.getCost());
-    myActiveAbilityName = ability.myName;
+    myActiveAbilityName = ability.toString();
   }
 
   protected void modifyAbilityPower(double amount)
@@ -299,9 +299,9 @@ public class Commander extends GameEventListener
     {
       double power = 0;
       // Add up the funds value of the damage done to both participants.
-      power += myHPLoss / minion.model.maxHP * minion.model.moneyCost;
+      power += myHPLoss / minion.model.maxHP * minion.model.getCost();
       // The damage we deal is worth half as much as the damage we take, to help powers be a comeback mechanic.
-      power += myHPDealt / enemy.model.maxHP * enemy.model.moneyCost / 2;
+      power += myHPDealt / enemy.model.maxHP * enemy.model.getCost() / 2;
       power /= CHARGERATIO_FUNDS; // Turn funds into units of power
       power += myHPDealt / CHARGERATIO_HP; // Add power based on HP damage dealt; rewards aggressiveness.
 
--- a/src/CommandingOfficers/CommanderBear_Bull.java
+++ b/src/CommandingOfficers/CommanderBear_Bull.java
@@ -28,18 +28,19 @@ public class CommanderBear_Bull extends Commander
     incomeBase = incomePerCity;
     // we start in Bear mode, so swap to it at the start
     isBull = true;
-    swapD2Ds();
+    swapD2Ds(true);
 
     addCommanderAbility(new UpDownTurnAbility(this));
     addCommanderAbility(new BustBoomAbility(this));
   }
 
-  public void swapD2Ds()
+  public void swapD2Ds(boolean setIncome)
   {
     if( isBull )
     {
       isBull = false;
-      incomePerCity = (int) (incomeBase * BEAR_MOD);
+      if( setIncome )
+        incomePerCity = (int) (incomeBase * BEAR_MOD);
       for( UnitModel um : unitModels )
       {
         um.COcost = BEAR_MOD;
@@ -48,7 +49,8 @@ public class CommanderBear_Bull extends Commander
     else
     {
       isBull = true;
-      incomePerCity = (int) (incomeBase * BULL_MOD);
+      if( setIncome )
+        incomePerCity = (int) (incomeBase * BULL_MOD);
       for( UnitModel um : unitModels )
       {
         um.COcost = BULL_MOD;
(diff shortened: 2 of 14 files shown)


ThislsAUsername:
Also, I just now realize I forgot to have Cinder's SCOP kill units that get too low (to prevent abuse like 1HP infantry killing a whole artillery), but I can't fix it at the moment.

src/CommandingOfficers/CommanderBear_Bull.java
```diff
@@ -0,0 +1,180 @@
...
+      // Damage is dealt after swapping D2Ds so it's actually useful to Bear
+      for( Location loc : cmdr.ownedProperties )
+      {
+        if( loc.getOwner() == myCommander )
+        {
+          Unit victim = loc.getResident();
+          if( null != victim )
+          {
+            double delta = victim.alterHP(DOWNUPTURN_LIQUIDATION);
+            cmdr.money += (-1 * delta * victim.model.getCost()) / 10;
```
ThatIsAUsername: I might move the `-1` up a line, and/or drop a comment to clarify that `alterHP` is expected to return a negative value.

src/CommandingOfficers/CommanderBear_Bull.java
```diff
@@ -0,0 +1,180 @@
...
+          }
+        }
+      }
+    }
+
+    @Override
+    public void revert(Commander commander)
+    {
+      CommanderBear_Bull cmdr = (CommanderBear_Bull) commander;
```
ThatIsAUsername: I might just use `coCast` instead of re-casting the input. I'd also add the comment to say this function implements the `COModifier` interface.

src/CommandingOfficers/CommanderBear_Bull.java
```diff
@@ -0,0 +1,180 @@
...
+  {
+    private static final String BUST_NAME = "Bust";
+    private static final String BOOM_NAME = "Boom";
+    private static final int BOOMBUST_COST = 6;
+    private static final double BOOMBUST_BUFF = 0.2;
+    CommanderBear_Bull COcast;
+
+    BustBoomAbility(Commander commander)
+    {
+      // as we start in Bear form, UpTurn is the correct starting name
```
ThatIsAUsername: UpTurn?

src/CommandingOfficers/CommanderCinder.java
```diff
@@ -0,0 +1,181 @@
...
+import Engine.XYCoord;
+import Engine.Combat.BattleSummary;
+import Engine.GameEvents.GameEventListener;
+import Terrain.GameMap;
+import Terrain.Location;
+import Units.Unit;
+import Units.UnitModel;
+
+/*
+ * Cinder is based on getting an edge in the action economy, at the cost of unit health.
```
ThatIsAUsername: I think both of your new Commanders are definitely on the... evil side of the spectrum.

src/CommandingOfficers/Commander.java
```diff
@@ -201,9 +201,9 @@ public UnitModel getUnitModel(UnitEnum unitType)
   }
 
   /** Get the list of units this commander can build from the given property type. */
-  public ArrayList<UnitModel> getShoppingList(TerrainType buyLocation)
+  public ArrayList<UnitModel> getShoppingList(Location buyLocation)
```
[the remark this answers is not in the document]
ThatIsAUsername: Ah. Because Cinder is doing odd things. Hrm.

src/CommandingOfficers/CommanderCinder.java
```diff
@@ -0,0 +1,181 @@
...
+  private static class WitchFireAbility extends CommanderAbility implements COModifier
+  {
+    private static final String WITCHFIRE_NAME = "Witchfire";
+    private static final int WITCHFIRE_COST = 9;
+    private static final int WITCHFIRE_WOUND = -1;
+    private WitchFireListener listener;
+
+    WitchFireAbility(Commander commander)
+    {
+      // as we start in Bear form, UpTurn is the correct starting name
```
ThatIsAUsername: Hm, yes. Bear form.

src/CommandingOfficers/CommanderCinder.java
```diff
@@ -0,0 +1,181 @@
...
+
+    SearAbility(Commander commander)
+    {
+      super(commander, SEAR_NAME, SEAR_COST);
+    }
+
+    @Override
+    protected void perform(GameMap gameMap)
+    {
+      for( Unit unit : myCommander.units )
```
ThatIsAUsername: No accommodation for units that haven't moved yet? Harsh (but in character).

src/CommandingOfficers/CommanderCinder.java
```diff
@@ -0,0 +1,181 @@
...
+      // Determine if we were part of this fight. If so, refresh at our own expense
+      if( battleInfo.attacker.CO == myCommander )
+      {
+        if( battleInfo.attacker.getPreciseHP() > 0 )
+        {
+          battleInfo.attacker.alterHP(refreshCost);
+          battleInfo.attacker.isTurnOver = false;
+        }
+      }
+      else if( battleInfo.defender.CO == myCommander )
```
ThatIsAUsername: Maybe I'm missing something, but I don't think it makes sense to refresh units that are being attacked. If it isn't Cinder's turn, refreshing them won't help anything anyway.

src/CommandingOfficers/CommanderCinder.java
ThislsAUsername: I changed that in my local copy, but forgot to push it up.
You sounded like you think the D2D is bad, so I'm inclined to push the change up.

ThislsAUsername: Then again....
It's super in character, but perhaps it shouldn't be utterly useless.
Maybe Cinder should give a buff of some kind to the units that haven't moved yet, in exchange for hurting them?

src/CommandingOfficers/CommanderCinder.java
ThislsAUsername: Haha, touche.

src/CommandingOfficers/CommanderBear_Bull.java
ThislsAUsername: Hokay.

src/CommandingOfficers/CommanderBear_Bull.java
ThislsAUsername: My reasoning here was that, if for whatever reason, you applied the thing to another CO... you'd crash instead of getting really weird behavior.

ThislsAUsername:
For the record, I don't want to merge this without handling the weirdness of 1HP units and Cinder's SCOP.
I guess it could be a combat modifier Cinder applies when SCOP is active, instead of a post-action.
Another option would be to just not refresh units that would die with another application.

Thoughts?

src/CommandingOfficers/CommanderCinder.java
ThatIsAUsername: I dunno, Bear&Bull skin their troops alive to turn a profit...

src/CommandingOfficers/CommanderCinder.java
ThatIsAUsername: The D2D feels stronger when you start with a decent bankroll - I think I'd have to play with it more to get a feel for how strong it is.
And sure. I dunno. I would avoid making it too complex, but I'm not sure where that line is.

src/CommandingOfficers/CommanderCinder.java
ThislsAUsername: I wasn't thinking of it as "skinning", but perhaps you have a point.

src/CommandingOfficers/CommanderCinder.java
ThislsAUsername: As is only proper, I picked the worst-implying thing I could.
It's also the least work, and situationally useful.

src/CommandingOfficers/CommanderCinder.java
```diff
@@ -0,0 +1,195 @@
...
+    public WitchFireListener(Commander myCo, int HPCost)
+    {
+      myCommander = myCo;
+      refreshCost = HPCost;
+    }
+
+    @Override
+    public void receiveBattleEvent(BattleSummary battleInfo)
+    {
+      // Determine if we were part of this fight. If so, refresh at our own expense
```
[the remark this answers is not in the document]
ThislsAUsername: Hah, okay.

ThatIsAUsername:
Is there anything else we were waiting for to merge this? I don't think so.

Also, random thought. I bet we could break Bed&Breakfast into two COs (one inside the other, or a faux CO that switches between the other two), and change the name/portrait/Ability list based on which one is in charge. How cool would it be for the Bear and the Bull to actually switch places when they swap out?

ThislsAUsername:
> Is there anything else we were waiting for to merge this? I don't think so.

I don't think so either.
> Bed&Breakfast 

uhhh, Bed&Breakfast?

> Also, random thought. I bet we could break Bed&Breakfast into two COs (one inside the other, or a faux CO that switches between the other two), and change the name/portrait/Ability list based on which one is in charge. How cool would it be for the Bear and the Bull to actually switch places when they swap out?

I bet we could, but that might make things ugly if we ever do a certain mechanic from the third entry of a certain series.
I'm not strictly opposed to this happening, but I *am* strictly opposed to spending effort doing it, haha.
You can if you want to.

PR #122, opened 2018-12-26
Path fix

ThatIsAUsername:
I think I fixed the pathing issue. the -99 starting values in the `movesLeftGrid` were causing trouble. I converted `movesLeftGrid` into `costGrid` so that higher values are less desirable, which made it work with the `theoretical` flag you added, and at arbitrary distances.
Lemme know what you think; it was late when I did this, so there might be a better way.

### DIFF
--- a/src/AI/InfantrySpamAI.java
+++ b/src/AI/InfantrySpamAI.java
@@ -78,7 +78,7 @@ public class InfantrySpamAI implements AIController
 
   private void log(String message)
   {
-    System.out.println(message);
+    //System.out.println(message);
     logger.append(message).append('\n');
   }
 
@@ -126,6 +126,7 @@ public class InfantrySpamAI implements AIController
 
       // If no attack/capture actions are available now, just move towards a non-allied building.
       Utils.sortLocationsByDistance( new XYCoord(unit.x, unit.y), unownedProperties);
+      //Utils.sortLocationsByTravelTime(unit, unownedProperties, gameMap);
       if( !unownedProperties.isEmpty() ) // Sanity check - it shouldn't be, unless this function is called after we win.
       {
         log(String.format("  Seeking a property to send %s after", unit.toStringWithLocation()));
--- a/src/AI/InfantrySpamAI.java
+++ b/src/AI/InfantrySpamAI.java
@@ -126,7 +126,6 @@ public class InfantrySpamAI implements AIController
 
       // If no attack/capture actions are available now, just move towards a non-allied building.
       Utils.sortLocationsByDistance( new XYCoord(unit.x, unit.y), unownedProperties);
-      //Utils.sortLocationsByTravelTime(unit, unownedProperties, gameMap);
       if( !unownedProperties.isEmpty() ) // Sanity check - it shouldn't be, unless this function is called after we win.
       {
         log(String.format("  Seeking a property to send %s after", unit.toStringWithLocation()));
--- a/src/Engine/Utils.java
+++ b/src/Engine/Utils.java
@@ -67,7 +67,7 @@ public class Utils
     for( XYCoord loc : locations )
     {
       // Add any location that is empty and supports movement of the cargo unit.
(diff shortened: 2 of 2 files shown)


src/Engine/Utils.java
```diff
@@ -129,32 +130,30 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
 
   /**
    * Determines whether the Location (x, y), can be added to the search queue.
+   * @param theoretical If set, this function will ignore move-power limitations and enemy-unit presence.
    */
-  private static boolean checkSpace(Unit unit, GameMap myMap, SearchNode currentNode, int x, int y, int[][] movesLeftGrid, boolean ignoreUnits)
+  private static boolean checkSpace(Unit unit, GameMap myMap, SearchNode currentNode, XYCoord coord,
+    int[][] costGrid, boolean ignoreUnits)
```
ThislsAUsername: I don't think we need the costGrid here anymore

src/Engine/Utils.java
```diff
@@ -231,39 +230,34 @@ public static Path findShortestPath(Unit unit, int x, int y, GameMap map, boolea
...
-    //System.out.println("Finding new path for " + unit.model.type + " from " + unit.x + ", " + unit.y + " to " + x + ", " + y);
-    // Set all locations to false/remaining move = 0
-    int[][] movesLeftGrid = new int[map.mapWidth][map.mapHeight];
+    int[][] costGrid = new int[map.mapWidth][map.mapHeight];
     for( int i = 0; i < map.mapWidth; i++ )
     {
       for( int j = 0; j < map.mapHeight; j++ )
       {
-        movesLeftGrid[i][j] = (theoretical) ? -99 : 0;
+        costGrid[i][j] = -1;
```
ThislsAUsername: Why not just set it to 99? Then you won't need special handling down the line.

src/Engine/Utils.java
```diff
@@ -381,9 +396,9 @@ public int compare(SearchNode o1, SearchNode o2)
       int firstDist = Math.abs(o1.x - xDest) + Math.abs(o1.y - yDest);
       int secondDist = Math.abs(o2.x - xDest) + Math.abs(o2.y - yDest);
 
-      int firstPow = movesLeftGrid[o1.x][o1.y] - ((hasDestination) ? firstDist : 0);
-      int secondPow = movesLeftGrid[o2.x][o2.y] - ((hasDestination) ? secondDist : 0);
-      return secondPow - firstPow;
+      int firstPow = costGrid[o1.x][o1.y] - ((hasDestination) ? firstDist : 0);
+      int secondPow = costGrid[o2.x][o2.y] - ((hasDestination) ? secondDist : 0);
```
ThislsAUsername: I believe the distances should now be *added* to the cost, and the variable names clarified.

src/AI/InfantrySpamAI.java
```diff
@@ -126,6 +126,7 @@ public GameAction getNextAction(GameMap gameMap)
 
       // If no attack/capture actions are available now, just move towards a non-allied building.
       Utils.sortLocationsByDistance( new XYCoord(unit.x, unit.y), unownedProperties);
+      //Utils.sortLocationsByTravelTime(unit, unownedProperties, gameMap);
```
ThislsAUsername: Adding commented code?

src/Engine/Utils.java
ThatIsAUsername: Yes

src/AI/InfantrySpamAI.java
ThatIsAUsername: Eh, I thought about doing that instead, since it's more correct, and it's possible to construct a scenario where `sortLocationsByDistance` will end up ping-ponging a unit back and forth each turn without getting anywhere.
Sorting by actual travel time is pretty expensive though; you get a noticeable slowdown in Europe, so I switched back. I should remove the comment though.

src/Engine/Utils.java
ThislsAUsername: If it's not, we have bigger problems, probably

PR #125, opened 2018-12-27
Added a placeholder image for COs that don't have art yet.

ThatIsAUsername:


ThislsAUsername:
Having that is great, except that you still can't tell which CO you're picking in the CO select screen.

ThatIsAUsername:
> Having that is great, except that you still can't tell which CO you're picking in the CO select screen.

True. It's still less ugly though. We could add the name on the CO select screen - maybe on a nice bespoke frame from SpriteLibrary.

PR #126, opened 2018-12-28
Refactors text frame drawing into a global util

ThislsAUsername:
Trying this out to reduce code duplication.
Now with 100% more working for ResupplyAnimation.

### DIFF
--- a/src/UI/Art/Animation/ResupplyAnimation.java
+++ b/src/UI/Art/Animation/ResupplyAnimation.java
@@ -77,13 +77,15 @@ public class ResupplyAnimation implements GameAnimation
       int width = (int) (signWidth * percent);
       int height = (int) (signHeight * percent);
 
-      SpriteUIUtils.drawTextFrame(g, MENUBGCOLOR, MENUFRAMECOLOR, "", mapLocation.xCoord, mapLocation.yCoord,width,2* drawScale+height-signHeight);
+      SpriteUIUtils.drawTextFrame(g, MENUBGCOLOR, MENUFRAMECOLOR, "", mapLocation.xCoord, mapLocation.yCoord, width / 2,
+          2 * drawScale + height - signHeight);
 
     }
     else if( animTime < signUpEnd )
     {
       // The sign is legible.
-      SpriteUIUtils.drawTextFrame(g, MENUBGCOLOR, MENUFRAMECOLOR, SUPPLYTEXT, mapLocation.xCoord, mapLocation.yCoord, 2* drawScale,2* drawScale);
+      SpriteUIUtils.drawTextFrame(g, MENUBGCOLOR, MENUFRAMECOLOR, SUPPLYTEXT, mapLocation.xCoord, mapLocation.yCoord,
+          2 * drawScale, 2 * drawScale);
     }
     else if( animTime < signGone )
     {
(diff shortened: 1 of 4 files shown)


src/UI/Art/Animation/ResupplyAnimation.java
```diff
@@ -88,29 +76,26 @@ public boolean animate(Graphics g)
       double percent = (double) animTime / signUpBegin;
       int width = (int) (signWidth * percent);
       int height = (int) (signHeight * percent);
-      XYCoord signTopLeft = new XYCoord(tileCenterX - width / 2, tileCenterY - height / 2);
 
-      drawSign(g, signTopLeft.xCoord, signTopLeft.yCoord, width, height);
+      SpriteUIUtils.drawTextFrame(g, MENUBGCOLOR, MENUFRAMECOLOR, "", mapLocation.xCoord, mapLocation.yCoord, width / 2,
+          2 * drawScale + height - signHeight);
```
ThatIsAUsername: This. This is stupid clever. Wouldn't it be easier to just have `SpriteUIUtils.drawTextFrame` detect that there is no text, and not add that back in?

ThatIsAUsername: Also, I feel like `drawTextFrame` shouldn't need to know anything about the map. I want to just give it the image coordinates, or possibly have it return a `BufferedImage` with the sign for me to draw where I want. This would also allow it to be used in non-map-centric contexts.

src/UI/Art/SpriteArtist/SpriteMenuArtist.java
```diff
@@ -17,14 +18,14 @@
...
+  public static final Color MENUHIGHLIGHTCOLOR = new Color(246, 234, 210);
 
-  private int menuTextWidth;
-  private int menuTextHeight;
-  private int menuHBuffer; // Amount of visible menu to left and right of options;
-  private int menuVBuffer; // Amount of visible menu above and below menu options;
+  public static int menuTextWidth;
+  public static int menuTextHeight;
+  public static int menuHBuffer; // Amount of visible menu to left and right of options;
+  public static int menuVBuffer; // Amount of visible menu above and below menu options;
```
ThatIsAUsername: Same objection to making all the values here `static`, especially since some of them can be accessed/inferred in a static way already from elsewhere.

For colors in particular, I wonder if it would be worth making a static color repository for that stuff (possibly to include team colors). Probably more than is reasonable to bite off for this change.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -25,7 +33,111 @@ public static double calculateSlideAmount(double currentNum, int targetNum)
...
+      }
+    }
+  }
+
+  /**
+   * Populate 'out' with the string versions of the options available through 'menu'.
+   * @param menu
+   * @param out
+   */
+  public static void getMenuStrings(InGameMenu<? extends Object> menu, ArrayList<String> out)
```
ThatIsAUsername: This function used to be a private member function of `SpriteMenuArtist`, which is still the only class that uses it. I'd be inclined to keep this over there unless there's a good reason to make it public/static.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -25,7 +33,111 @@ public static double calculateSlideAmount(double currentNum, int targetNum)
...
+    {
+      XYCoord visualOrigin = SpriteMapView.getVisualOrigin();
+      // Find the dimensions of the menu we are drawing.
+      int menuWidth = getMenuTextWidthPx(items) + hBuffer * 2;
+      int menuHeight = getMenuTextHeightPx(items) + vBuffer * 2;
+
+      // Center the menu over the current action target location, accounting for the position of the map view.
+      int viewTileSize = getTileSize(); // Grab this value for convenience.
+      int drawX = (int) (mapX * viewTileSize - (menuWidth / 2 - viewTileSize / 2));// - origin.xCoord;
+      int drawY = (int) (mapY * viewTileSize - (menuHeight / 2 - viewTileSize / 2));// - origin.yCoord;
```
ThatIsAUsername: Hanging comments. Not original with you, of course.

src/UI/Art/Animation/ResupplyAnimation.java
ThislsAUsername: Heh. UI code draws out my hacky instincts.
Looks like your PR fixed all this jazz up, though.

PR #127, opened 2018-12-28
Spender AI

ThislsAUsername:


src/AI/SpenderAI.java
```diff
@@ -0,0 +1,325 @@
...
+        {
+          Unit unit = travelQueue.poll();
+
+          // Find the possible destinations.
+          ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap);
+
+          Utils.sortLocationsByDistance(new XYCoord(unit.x, unit.y), unownedProperties);
+          if( !unownedProperties.isEmpty() ) // Sanity check - it shouldn't be, unless this function is called after we win.
+          {
+            log(String.format("  Seeking a property to send %s after", unit.toStringWithLocation()));
```
ThatIsAUsername: I feel like this should be limited to units with the ability to capture. Maybe send non-capturing units to a) the nearest enemy, or b) the enemy HQ.
Just for grins, I made local hacks to only consider enemy HQs for units that can't capture, and it shuts down ISAI that much faster since it goes straight for the jugular.

ThislsAUsername: That would objectively improve the AI, yes, but having artillery sit on enemy cities with a halo of other units orbiting them has a certain charm, heh

ThislsAUsername: For the record, my intent with this AI was just to get it to use its resources more fully.
Being *smart* with them is a subject for another pull request.

ThislsAUsername: I like the idea of having a number of AIs to choose from with variable weaknesses and... okay, just variable weaknesses is fine.
You could also note that it almost makes a point of rushing Artillery to the front line, but I'm not sure that's worth fixing here either.

The problem with fixing this particular problem is I don't know what I would want it to do instead.
You could have it go for the HQ, but then it wouldn't move off of *that* either.
You could have it go straight for the enemies, but that might be undesirable in other ways (splitting forces, running headlong into danger).

ThatIsAUsername: I probably wouldn't worry about Artillery placement, since that can be a tricky beast, and though the current behavior is suboptimal, it isn't flat wrong per se... just amateurish.
I think going straight for the HQ is fine, because if they reach it, everyone defending it is probably dead anyway. The deadlock condition for HQs is not as troublesome as for other buildings, and even if it is, I'd rather deadlock on one building than all of them.
Going straight for enemies is probably also fine - no it's not optimal, but it's at least understandable in a way that preventing yourself from capturing a city isn't.

ThatIsAUsername: To be clear, I'm not asking you to make an optimal AI before merging - I just want one that doesn't grotesquely shoot itself in the foot (unless that's like, a feature, man).

ThislsAUsername: It's a feature *now*.
But okay, I'll fix it.

src/AI/SpenderAI.java
```diff
@@ -177,19 +178,34 @@ else if( unitQueue.isEmpty() )
...
           if( !unownedProperties.isEmpty() ) // Sanity check - it shouldn't be, unless this function is called after we win.
           {
             log(String.format("  Seeking a property to send %s after", unit.toStringWithLocation()));
             int index = 0;
             XYCoord goal = null;
             Path path = null;
             boolean validTarget = false;
+            ArrayList<XYCoord> validTargets = new ArrayList<>();
 
+            if( unit.model.chassis == ChassisEnum.TROOP ) // Technically a hack, but we don't have vehicles that can capture... yet.
```
ThatIsAUsername: A better/more futureproof solution would be to see if `Unit.model.possibleActions` contains`ActionType.CAPTURE`.

ThislsAUsername: It would be, if possibleActions was not a basic array.
I tried that first and realized I would have to write a loop.

ThatIsAUsername: Are you saying there's some reason that relying on a basic array may cause problems? Or do you just not want to write a few extra lines of code?

ThislsAUsername: I'm saying I'm lazy.

ThatIsAUsername: Ah. Well then, as a lazy person, you should recognize that fixing it now will be less effort than running into a problem with it later, and having to re-figure out what the issue is before doing what you should have done now. It's really less effort to do it right the first time. Plus, then future you won't hate past you. Always try to do future you a solid.

src/AI/SpenderAI.java
```diff
@@ -0,0 +1,350 @@
...
+    do
+    {
+      // If we have more actions ready, don't bother calculating stuff.
+      if( !actions.isEmpty() )
+      {
+        return actions.poll();
+      }
+      else if( unitQueue.isEmpty() )
+      {
+        stateChange = false; // There's been no gamestate change since we last iterated through all the units, since we're about to do just that
```
ThatIsAUsername: Am I wrong in thinking that `stateChange` will pretty much always have the opposite value of `actions.isEmpty()`? It looks like you pretty much always check and set them together as well.

ThislsAUsername: You are wrong.
stateChange requires the unitQueue be emptied and refilled, and the unitQueue is retained between calls of getNextAction().
When stateChange is set to false and *stays* false after re-evaluating unitQueue, we start waiting units and building things.

ThatIsAUsername: Hmm, I tried it for kicks, and it definitely needs to be there. Good enough.

PR #128, opened 2018-12-28
Witchfire will now execute units that end their turn with one HP.

ThatIsAUsername:
If you want Witchfire to execute one-HP units, here's one way to make that happen. It does sort of open a can of worms in terms of what's possible, but we've also been operating in a "more flexibility is better" paradigm, so I can't say I'm opposed to making this change.
I thought about implementing the "Just publish an extra event right after the battle ends" approach, but I'm not a fan of that one, since that would bypass the animator.

This does of course break with tradition in *ahem* similar games, in terms of killing units outside of combat, but I'm not sure I'll shed tears over that either.

I don't have a strong opinion either way; let me know what you prefer, and of course, if you have a better option, do tell.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -12,6 +12,7 @@ import Engine.XYCoord;
 import Engine.Combat.BattleInstance;
 import Engine.Combat.BattleSummary;
 import Engine.GameEvents.GameEventListener;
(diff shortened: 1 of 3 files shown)


ThatIsAUsername:
Good idea; I had totally intended to do that, and then just failed to add that check.

PR #130, opened 2018-12-31
SpenderAI now ignores HQs that nobody owns because Europe.

ThislsAUsername:


src/AI/SpenderAI.java
```diff
@@ -204,7 +204,8 @@ else if( unitQueue.isEmpty() )
             {
               for( XYCoord coord : unownedProperties )
               {
-                if( gameMap.getEnvironment(coord).terrainType == TerrainType.HEADQUARTERS ) // should we have an attribute for this?
+                Location loc = gameMap.getLocation(coord);
+                if( loc.getEnvironment().terrainType == TerrainType.HEADQUARTERS && loc.getOwner() != null ) // should we have an attribute for this?
```
ThatIsAUsername: No preference. Syntactic sugar can be nice, but it's a simple enough conditional to just type.

PR #132, opened 2018-12-31
Unit cleanup

ThislsAUsername:
Also adds fog stats

### DIFF
--- a/src/AI/SpenderAI.java
+++ b/src/AI/SpenderAI.java
@@ -292,7 +292,7 @@ public class SpenderAI implements AIController
           for( UnitModel unit : units )
           {
             // I only want combat units, since I don't understand transports
-            if( unit.weaponModels != null && unit.weaponModels.size() > 0 && unit.moneyCost <= budget )
+            if( unit.weaponModels.isEmpty() && unit.moneyCost <= budget )
             {
               budget -= unit.moneyCost;
               purchases.put(locShopList.getKey(), unit);
@@ -314,8 +314,7 @@ public class SpenderAI implements AIController
             for( UnitModel unit : units )
             {
               // I want expensive units, but they have to have guns
-              if( budget > unit.moneyCost && unit.moneyCost > currentPurchase.moneyCost && unit.weaponModels != null
-                  && unit.weaponModels.size() > 0 )
+              if( budget > unit.moneyCost && unit.moneyCost > currentPurchase.moneyCost && unit.weaponModels.isEmpty() )
                 currentPurchase = unit;
             }
             // once we've found the most expensive thing we can buy here, record that
--- a/src/AI/SpenderAI.java
+++ b/src/AI/SpenderAI.java
@@ -187,16 +187,7 @@ public class SpenderAI implements AIController
             boolean validTarget = false;
             ArrayList<XYCoord> validTargets = new ArrayList<>();
 
-            boolean canCapture = false;
-            for( ActionType i : unit.model.possibleActions )
-            {
-              if( i == ActionType.CAPTURE )
-              {
-                canCapture = true;
-                break;
-              }
-            }
-            if( canCapture )
+            if( unit.model.possibleActions.contains(ActionType.CAPTURE) )
             {
               validTargets.addAll(unownedProperties);
             }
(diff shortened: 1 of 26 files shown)


src/AI/SpenderAI.java
```diff
@@ -187,16 +187,7 @@ else if( unitQueue.isEmpty() )
...
-            for( ActionType i : unit.model.possibleActions )
-            {
-              if( i == ActionType.CAPTURE )
-              {
-                canCapture = true;
-                break;
-              }
-            }
-            if( canCapture )
+            if( unit.model.possibleActions.contains(ActionType.CAPTURE) )
```
ThatIsAUsername: Hah. In the work I'm doing right now, I added a `canPerformActionType(ActionType)` function to `UnitModel`.

src/AI/SpenderAI.java
```diff
@@ -301,7 +292,7 @@ else if( unitQueue.isEmpty() )
           for( UnitModel unit : units )
           {
             // I only want combat units, since I don't understand transports
-            if( unit.weaponModels != null && unit.weaponModels.length > 0 && unit.moneyCost <= budget )
+            if( unit.weaponModels != null && unit.weaponModels.size() > 0 && unit.moneyCost <= budget )
```
ThatIsAUsername: This null check should no longer be needed (not that it's hurting anything).

ThatIsAUsername: You can also use `isEmpty()` instead of `size() > 0`.

src/Engine/MapController.java
```diff
@@ -75,6 +77,7 @@ public MapController(GameInstance game, MapView view)
     nextSeekIndex = 0;
     contemplatedAction = new ContemplatedAction();
 
+    for (Commander co : myGame.commanders){co.setAIController(new SpenderAI(co));}
```
ThatIsAUsername: No. Bad ThisisAUsername.

src/Units/UnitModel.java
```diff
@@ -35,38 +35,62 @@
...
+      possibleActions.add(action);
+    }
+    for( WeaponModel action : weapons )
+    {
+      weaponModels.add(action);
+    }
+  }
+
+  public UnitModel(String pName, UnitEnum pType, ChassisEnum pChassis, int cost, int pFuelMax, int pIdleFuelBurn, int pVision, int pMovePower,
+      MoveType pPropulsion, ArrayList<ActionType> actions, ArrayList<WeaponModel> weapons)
```
ThatIsAUsername: Is this constructor actually used anywhere? I guess I could check this branch out and look, but

src/Units/UnitModel.java
```diff
@@ -35,38 +35,62 @@
...
+  {
+    this(pName, pType, pChassis, cost, pFuelMax, pIdleFuelBurn, pVision, pMovePower, pPropulsion);
+    possibleActions = actions;
+    weaponModels = weapons;
+  }
+
+  private UnitModel(String pName, UnitEnum pType, ChassisEnum pChassis, int cost, int pFuelMax, int pIdleFuelBurn, int pVision, int pMovePower,
```
ThatIsAUsername: If we are going to have multiple constructors, I'd rather have one be the "real" one, and the others just `this` it if possible. The "real" one should naturally be the one with all the parameters, and other constructors can supply default/placeholder values for the ones they don't have.
This would have the added benefit of being able to remove the weapon argument when calling in from e.g. APC, instead of supplying the mildly-painful `new WeaponModel[0]`

src/Units/UnitModel.java
```diff
@@ -35,38 +35,62 @@
   public int maxFuel;
   public int idleFuelBurn;
   public int movePower;
+  public int vision;
+  public boolean piercingVision = false;
```
ThatIsAUsername: What is `piercingVision`?

src/Units/UnitModel.java
ThislsAUsername: It's used by the cloning logic

src/Engine/MapController.java
ThislsAUsername: Oops?

src/Units/UnitModel.java
ThislsAUsername: Goes through cover, like forests/reefs.
Admittedly, it could probably use a comment.

src/Units/UnitModel.java
ThislsAUsername: Hah, fair. How do you propose we deal with the desire (need?) to have both basic array and ArrayList arguments?
I don't know how to make compile-time ArrayLists, and .toList() doesn't seem to work for passing into a constructor that takes a basic array.

ThatIsAUsername:
Looks like all of this is in the Fog pull request. Closing as OBE.

PR #133, opened 2019-01-01
Fog

ThislsAUsername:
There is an existing "issue" in that adjacent buildings are revealed by your buildings, but I don't care to fix that for now since we don't really know what we want anyway. Or at least, I don't.
There's also a reasonable argument that we shouldn't show the AI's vision while it's its turn, but atm there isn't a way to set up a game with the AI so it doesn't matter much at this point IMO

I've merged all the changes in unitCleanup into this branch.

I guess I probably should have fixed the merge issues there and not here, but suh est lah veeay


src/Engine/GameAction.java
```diff
@@ -80,7 +80,7 @@ public AttackAction(GameMap gameMap, Unit actor, Path movePath, XYCoord atkLoc)
       if( isValid )
       {
         moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
-        defender = gameMap.getLocation(attackLocation).getResident();
+        defender = gameMap.getLocation(attackLocation).getResident(gameMap);
```
ThatIsAUsername: ugh... this line hurts to read a little bit. I have to pass GameMap into GameMap to see what's in GameMap?

src/Engine/GameEvents/UnloadEvent.java
```diff
@@ -50,8 +50,19 @@ public void performEvent(GameMap gameMap)
...
     }
   }
-
+  
+  @Override
+  public boolean shouldPreempt(MapMaster gameMap )
+  {
+    boolean result = false;
+    Unit obstacle = gameMap.getLocation(dropLoc).getResident(gameMap);
+    if( null != obstacle && transport.CO.isEnemy(obstacle.CO) )
```
ThatIsAUsername: If `obstacle` is friendly, then we don't preempt? Shouldn't it still be impossible to drop a unit onto an ally? Granted, that probably couldn't happen because you would see them to start with, but still.

src/Engine/GameInstance.java
```diff
@@ -18,7 +19,7 @@
 
   HashMap<Integer, XYCoord> playerCursors = null;
 
-  public GameInstance(GameMap map, Commander[] cos)
+  public GameInstance(MapMaster map, Commander[] cos)
```
ThatIsAUsername: All these places where you are passing a `MapMaster` around... is there any reason that couldn't still be `GameMap`?

src/Engine/MapController.java
```diff
@@ -504,10 +504,17 @@ private boolean executeGameAction(GameAction action)
 
       if( events.size() > 0 )
       {
+        GameEventQueue usedEvents = new GameEventQueue();
+        for (GameEvent event : events)
+        {
+          usedEvents.add(event);
+          if (event.shouldPreempt(myGame.gameMap))
```
ThatIsAUsername: Nice. I like this.

src/Engine/Utils.java
```diff
@@ -498,4 +498,64 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
     TravelDistanceComparator tdc = new TravelDistanceComparator(unit, map);
     Collections.sort(mapLocations, tdc);
   }
+
+  /** Returns a list of all locations between minRange and maxRange tiles away from origin, inclusive. */
```
ThatIsAUsername: Copypasta.

src/Terrain/GameMap.java
```diff
@@ -1,256 +1,72 @@
 package Terrain;
 
+import Engine.Path;
 import Engine.XYCoord;
 import Units.Unit;
 
-public class GameMap
+public abstract class GameMap
```
ThatIsAUsername: All right, obtuse question time. Why did we have to split this in three and make it abstract? Couldn't we just keep the functionality here, not have a separate `MapMaster` class, and still have `MapWindow` inherit?

src/Terrain/Location.java
```diff
@@ -50,9 +50,14 @@ public void setOwner(Commander owner)
     }
   }
 
-  public Unit getResident()
+  public Unit getResident(GameMap map)
```
ThatIsAUsername: I don't like needing to pass the map in here. I might look into alternatives.

ThatIsAUsername:
Hah, when playing against AI, you get the enemy vision during their turn. It's nice for observing AI games, but sorta kills the mystery if you are playing against them.

src/Engine/GameEvents/MoveEvent.java
```diff
@@ -33,14 +36,15 @@ public GameAnimation getEventAnimation(MapView mapView)
...
+    listener.receiveMoveEvent(this);
   }
 
   @Override
-  public void performEvent(GameMap gameMap)
+  public void performEvent(MapMaster gameMap)
   {
     if( unitPath.getPathLength() > 0 ) // Make sure we have a destination.
     {
+      unitPath.snipCollision(gameMap, unit); // make sure we only go as far as we can go
```
ThatIsAUsername: For some reason, the animation still animates to the original endpoint, even if the unit bumps into someone along the way. It's not immediately clear why.

ThatIsAUsername:
Overall this looks very good; with just a bit of tweaking, and you seem to have knocked it out pretty quickly, so well done on that. I think I'll look at combining the WeaponModel constructors, and maybe see if I can tweak Location a bit, but in the meantime:
The fog color seems a bit flat. It works, but I don't think mid-grey is the most compelling option.
If we give it a bit of a blue tint (72, 72, 96, 128), it looks a bit more natural, I think:
image

If we make it a bit darker (72, 72, 96, 200) it could grant a decent night-ops feel:
image

Alternatively, we could bump it to the white side of the spectrum (255, 255, 255, 180) and get something closer to actual fog, complete with buildings and mountains poking through the clouds:
image

A red tint (255, 96, 72, 96) ends up looking pretty decent too.
image

Any thoughts? Resounding apathy?

src/Engine/GameEvents/BattleEvent.java
```diff
@@ -71,4 +72,10 @@ public void performEvent(GameMap gameMap)
       defender.CO.units.remove(defender);
     }
   }
+  
+  @Override // there's no known way for this to fail after the GameAction is constructed
+  public boolean shouldPreempt(MapMaster gameMap )
```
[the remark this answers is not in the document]
ThislsAUsername: BattleEvent is always preceded by a MoveEvent.
If the MoveEvent succeeds, the BattleEvent *will* be valid since the BattleEvent can't be constructed with an invalid target.
The referenced failure is for the BattleEvent as a whole, not that function.

As for that refactor... you already do that with resupply, so go for it.

src/Engine/GameInstance.java
ThislsAUsername: GameInstance and GameEvents need to have MapMaster because MapMaster never lies.
My idea was that GameActions and UI use GameMap, but final validation is done with MapMaster so the program doesn't decide to actually do something it's not supposed to.
For instance, a MoveEvent couldn't fail/preempt if it had the same fogged view that produced the path it's using.

src/Terrain/GameMap.java
ThislsAUsername: Strictly speaking, you totally could.
MapWindows would have a map-ful of Locations that never gets touched, but that's not really a problem.
As I noted up above though, I like having the ability to require a truth-telling map.

src/Terrain/Location.java
ThislsAUsername: I did. They're all worse.
Unless you have a better?

ThislsAUsername:
> Hah, when playing against AI, you get the enemy vision during their turn. It's nice for observing AI games, but sorta kills the mystery if you are playing against them.

....and now you see why I said having a single source of truth for fog info is technical debt.
This is a super fixable problem, I just didn't bother since we don't even properly support having AI players yet.

> Overall this looks very good; with just a bit of tweaking, and you seem to have knocked it out pretty quickly, so well done on that.

Thanks. It's almost like I had little else to do and no desire to play video games.

> The fog color seems a bit flat. It works, but I don't think mid-grey is the most compelling option.
> Any opinion?

I actually just copied the color you used to make units tired. The buildings and mountains popping out was intentional, for the record.
Any of the first 3 are good with me (some preference for dark?), the red's a little blech IMO.

> OR or, we could go full Wesnoth and day/night cycle this beast, and just use all of them.

I have... questions.

ThislsAUsername:
I guess we should also have non-HQs appear neutral when in fog, but baby steps?

src/Engine/GameEvents/MoveEvent.java
ThatIsAUsername: Oh, of course. Animation happens before execution, so it doesn't know there's something to bump into until after it looks like it didn't happen.

src/Engine/GameEvents/UnloadEvent.java
ThatIsAUsername: Hmm. If I try to unload three things, and the second one is blocked, the first one should still get out, right? We currently store drop locations as a `Map<Unit, XYCoord>`. We should consider making that an ordered list to preserve drop order.

ThatIsAUsername:
> I guess we should also have non-HQs appear neutral when in fog, but baby steps?

Well, unless we've observed them to be captured, but yes.

> The buildings and mountains popping out was intentional, for the record.

How come? It looks kinda wonky if it's night on the floor, and light out 6 stories up. It at least looks a bit less suspect with the white/daytime fog effect.

As for colors, I think I quite like the white fog option - it's unusual, at any rate. The downside is that it's close the the map-highlight color. A darkish blue tint may be the way to go.

src/Engine/GameEvents/CaptureEvent.java
```diff
@@ -54,14 +54,20 @@ public boolean willCapture()
...
         (unit.CO.isEnemy(location.getOwner())) )
     {
       unit.capture(gameMap.getLocation(unit.x, unit.y));
     }
   }
+  
+  @Override // there's no known way for this to fail after the GameAction is constructed
+  public boolean shouldPreempt(MapMaster gameMap )
+  {
+    return false;
```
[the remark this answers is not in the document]
ThislsAUsername: It can *be* preempted, but it cannot itself preempt.
The move action can fail (which will throw a shouldPreempt), which will cause the CaptureEvent to not happen... but the CaptureEvent only governs what happens once you're already on top of the space to be captured.

ThislsAUsername:
> > I guess we should also have non-HQs appear neutral when in fog, but baby steps?
> 
> Well, unless we've observed them to be captured, but yes.

Uhhhh. Hmm. Does that mean MapWindow would *also* need to store owner information? That would probably also require getOwner() to take in a GameMap... ew.
 
> > The buildings and mountains popping out was intentional, for the record.
> 
> How come? It looks kinda wonky if it's night on the floor, and light out 6 stories up. It at least looks a bit less suspect with the white/daytime fog effect.

Hah, fair enough.
I just didn't want the night in the square north of you to affect the top only of the mountain in your square.

> As for colors, I think I quite like the white fog option - it's unusual, at any rate. The downside is that it's close the the map-highlight color. A darkish blue tint may be the way to go.

I think I might be happier with the white fog if it was less transparent.
I'd be even more happier if the highlight for movement didn't look the same, heh.

src/Engine/GameEvents/MoveEvent.java
ThislsAUsername: Did you want me to resolve this, or did you wanna resolve this, or should it just be an issue?
For now, I kinda like having the animation still fire off the intended location since it lets the game tell you "Yup, you *tried* to go where you wanted to"

src/Engine/GameEvents/CaptureEvent.java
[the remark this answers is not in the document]
ThatIsAUsername: Oh, right. `CaptureEvent` cannot be preempted, but `CaptureAction` can be. I was moving the checks for preemption to the action side, and got a wire crossed in my head.

src/Engine/GameEvents/MoveEvent.java
ThatIsAUsername: I got it. I also fixed a problem where units would not update fog after being built. I guess I can push a branch up just so you can see what I'm doing, but I'm not quite done yet.

ThatIsAUsername:
> I just didn't want the night in the square north of you to affect the top only of the mountain in your square.

Ah, so it wasn't "intentional" exactly - but it's also not a trivial problem to fix. I'm not gonna worry too much for now, but I might like to address that at some point.

I think the darker blue might be a bit too much - it hurts visibility of underlying terrain. We could always add an opaque-fog option, but I want the translucent option to have good visibility. I think I'm currently leaning towards either blue-tinted dark with normal highlights, or white clouds with bluish highlights:
image
image

Though I almost do want a day/night flag, and switch between the two for atmosphere.

ThislsAUsername:
>either blue-tinted dark with normal highlights, or white clouds with bluish highlights

I guess I'm okay with either.

>Though I almost do want a day/night flag, and switch between the two for atmosphere.

Hah, that's fine as long as we don't do what a *certain* series did and say a "day" is what happens when everybody gets one turn.

ThatIsAUsername:
> Hah, that's fine as long as we don't do what a certain series did and say a "day" is what happens when everybody gets one turn.

Nah, I'd be inclined to do it on a per-map basis, not switching back and forth in the middle of a game.

PR #135, opened 2019-01-02
Muriel

ThatIsAUsername:
Sorry I've been a bit behind on pull requests. This is why.
This new AI makes units reactively, informed by the enemy loadout.
Muriel has a few other nice unit-control optimizations as well.
I've also introduced `AIUtils`, which should make it easier to compose AIs from existing pieces.
I migrated Isai to use some of the functionality that now resides in `AIUtils`.

I was going to include a combat-simulator as part of this, but Muriel ended up being better without it.

### DIFF
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -149,16 +149,21 @@ public class AIUtils
     return move;
   }
 
-  public static ArrayList<XYCoord> findResupplyPoints(Unit unit)
+  /**
+   * Return a list of all friendly territories at which a unit can resupply and repair.
+   * @param unit
+   * @return
+   */
+  public static ArrayList<XYCoord> findRepairDepots(Unit unit)
   {
-    ArrayList<XYCoord> coords = new ArrayList<XYCoord>();
+    ArrayList<XYCoord> stations = new ArrayList<XYCoord>();
     for( Location loc : unit.CO.ownedProperties )
     {
       if( unit.model.canRepairOn(loc) )
       {
-        coords.add(loc.getCoordinates());
+        stations.add(loc.getCoordinates());
       }
     }
-    return coords;
+    return stations;
   }
 }
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -1,0 +1,117 @@
+package AI;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+import CommandingOfficers.Commander;
+import Engine.GameAction;
+import Engine.GameActionSet;
+import Engine.Path;
+import Engine.Utils;
+import Engine.XYCoord;
+import Terrain.GameMap;
+import Terrain.Location;
+import Units.Unit;
+
+public class AIUtils
+{
+  /**
+   * Figure out all actions unit can perform this turn.
+   * @param unit The unit under consideration.
+   * @param gameMap The world in which the Unit lives.
+   * @return a Map of XYCoord to ArrayList<GameActionSet>. Each XYCoord will have a GameActionSet for
+   * each type of action the unit can perform from that location.
+   */
+  public static Map<XYCoord, ArrayList<GameActionSet> > getAvailableUnitActions(Unit unit, GameMap gameMap)
+  {
+    Map<XYCoord, ArrayList<GameActionSet> > actions = new HashMap<XYCoord, ArrayList<GameActionSet> >();
+
+    // Find the possible destinations.
+    ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap);
+
+    for( XYCoord coord : destinations )
+    {
+      // Figure out how to get here.
+      Path movePath = Utils.findShortestPath(unit, coord, gameMap);
+
+      // Figure out what I can do here.
+      ArrayList<GameActionSet> actionSets = unit.getPossibleActions(gameMap, movePath);
+
+      // Add it to my collection.
+      actions.put(coord, actionSets);
+    }
+
+    return actions;
+  }
+
+  /**
+   * Find locations of every property that is not owned by myCo or an ally.
+   * @param myCo Properties of this Commander and allies will be ignored.
+   * @param gameMap The map to search for properties.
+   * @return An ArrayList with the XYCoord of every property that fills the bill.
+   */
+  public static ArrayList<XYCoord> findNonAlliedProperties(Commander myCo, GameMap gameMap)
+  {
+    ArrayList<XYCoord> props = new ArrayList<XYCoord>();
+    for( int x = 0; x < gameMap.mapWidth; ++x )
+    {
+      for( int y = 0; y < gameMap.mapHeight; ++y )
+      {
+        Location loc = gameMap.getLocation(x, y);
+        if( loc.isCaptureable() && myCo.isEnemy(loc.getOwner()) )
+        {
+          props.add(new XYCoord(x, y));
+        }
+      }
+    }
+    return props;
+  }
+
+  /**
+   * Find the locations of all non-allied units.
+   * @param myCo Units owned by myCo or allies are skipped.
+   * @param gameMap The map to search for units.
+   * @return An ArrayList with the XYCoord of every unit not allied with myCo.
+   */
+  public static ArrayList<XYCoord> getEnemyUnitLocations(Commander myCo, GameMap gameMap)
+  {
+    ArrayList<XYCoord> unitLocs = new ArrayList<XYCoord>();
+    for( int x = 0; x < gameMap.mapWidth; ++x )
+    {
+      for( int y = 0; y < gameMap.mapHeight; ++y )
+      {
+        Location loc = gameMap.getLocation(x, y);
+        if( loc.getResident() != null && myCo.isEnemy(loc.getResident().CO) )
+        {
+          unitLocs.add(new XYCoord(x, y));
+        }
+      }
+    }
+    return unitLocs;
+  }
+
+  /**
+   * Create and return a GameAction.WaitAction that will move unit towards destination, around
+   * any intervening obstacles. If no possible route exists, return false.
+   * @param unit The unit we want to move.
+   * @param destination Where we eventually want the unit to be.
+   * @param gameMap The map on which the unit is moving.
+   * @return A GameAction to bring the unit closer to the destination, or null if it is unreachable.
+   */
+  public static GameAction moveTowardLocation(Unit unit, XYCoord destination, GameMap gameMap)
+  {
+    GameAction move = null;
+
+    // Find the full path that would get this unit to the destination, regardless of how long. 
+    Path path = Utils.findShortestPath(unit, destination, gameMap, true);
+    if( path.getPathLength() > 0 ) // Check that the destination is reachable at least in theory.
+    {
+      path.snip(unit.model.movePower+1); // Trim the path so we don't try to walk through walls.
+      ArrayList<XYCoord> validMoves = Utils.findPossibleDestinations(unit, gameMap); // Find the valid moves we can make.
+      Utils.sortLocationsByDistance(new XYCoord(path.getEnd().x, path.getEnd().y), validMoves); // Sort moves based on intermediate destination. 
+      move = new GameAction.WaitAction(gameMap, unit, Utils.findShortestPath(unit, validMoves.get(0), gameMap)); // Move to best option.
+    }
+    return move;
+  }
+}
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -66,9 +66,9 @@ public class AIUtils
     Map<XYCoord, ArrayList<GameActionSet> > actionsByLoc = getAvailableUnitActions(unit, gameMap);
 
     // Now re-map them by type, irrespective of location.
-    for( Entry<XYCoord, ArrayList<GameActionSet>> entry : actionsByLoc.entrySet() )
+    for( ArrayList<GameActionSet> actionSets : actionsByLoc.values() )
     {
-      for( GameActionSet actionSet : entry.getValue() )
+      for( GameActionSet actionSet : actionSets )
       {
         ActionType type = actionSet.getSelected().getType();
 
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -19,7 +19,7 @@ import Units.Unit;
 public class AIUtils
 {
   /**
-   * Figure out all actions unit can perform this turn, and organizes them by location.
+   * Finds all actions available to unit, and organizes them by location.
    * @param unit The unit under consideration.
    * @param gameMap The world in which the Unit lives.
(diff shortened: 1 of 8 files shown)


src/AI/AIUtils.java
```diff
@@ -0,0 +1,169 @@
...
+    for( ActionType atype : ActionType.values() )
+    {
+      actionsByType.put(atype, new ArrayList<GameAction>());
+    }
+
+    // First collect the actions by location.
+    Map<XYCoord, ArrayList<GameActionSet> > actionsByLoc = getAvailableUnitActions(unit, gameMap);
+
+    // Now re-map them by type, irrespective of location.
+    for( Entry<XYCoord, ArrayList<GameActionSet>> entry : actionsByLoc.entrySet() )
```
ThislsAUsername: You could just use the valueSet()s since it you aren't using the key at all.
Isn't a big deal, though.

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+    {
+      if( myCo.isEnemy(other) )
+      {
+        enemyCos.add(other);
+        otherUnitModels.put(other, new ArrayList<UnitModel>());
+      }
+    }
+
+    // Figure out what I and everyone else can build.
+    for( UnitModel.UnitEnum modelEnum : UnitModel.UnitEnum.values() )
```
ThislsAUsername: Why do you have this convoluted loop when CO.unitModels is public?

src/Units/UnitModel.java
```diff
@@ -147,9 +147,43 @@ public boolean canRepairOn(Location locus)
     return new ArrayList<GameAction>();
   }
 
+  /**
+   * @return True if this UnitModel has at least one weapon with a minimum range of 1.
+   */
+  public boolean hasDirectFireWeapon()
```
ThislsAUsername: I feel like this is maybe of questionable utility. What you're really looking for, I feel, is units that can fire after moving.
All having range 1 means is the ability to counterattack.

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+        {
+          Unit myUnit = new Unit(myCo, myModel);
+          Unit otherUnit = new Unit( oCo, otherModel );
+
+          double myDamage = 0;
+          Weapon myWeapon = myUnit.chooseWeapon(otherUnit.model, 1, false);
+          if( null != myWeapon )
+          {
+            BattleInstance.BattleParams params = new BattleInstance.BattleParams(myUnit, myWeapon,
+                otherUnit, Environment.getTile(TerrainType.ROAD, Weathers.CLEAR));
```
ThislsAUsername: It occurs to me that having the environment of the defender maybe isn't all that useful.
I guess a more comprehensive review is needed, but IDK where/when we'd have that discussion.

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+    for( Unit unit : myCo.units )
+    {
+      if( unit.getCaptureProgress() > 0 )
+      {
+        nonAlliedProperties.remove(unit.getCaptureTargetCoords());
+      }
+    }
+
+    // If the CO has enough AP, preload the CommanderAbilityAction.
+    ArrayList<CommanderAbility> abilities = myCo.getReadyAbilities();
```
ThislsAUsername: I feel like a basic abstraction that could help the AI a lot would be a flag or set of flags that says when the ability is useful.
Witchfire or Strong/Patch's abilities are definitely desirable at the start of the turn, but Sear is something you'd want to use only at the end of your turn.

I don't know what other flags we might want, but my heart sinks a little every time an AI pops Sear at the start of its turn, lol.

ThislsAUsername:
Muriel could benefit from Spender's unit queuing logic, by the way.
I'm not opposed to merging without that, but I'd like it to happen at some point.

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+        for( XYCoord coord : stations )
+        {
+          Location station = gameMap.getLocation(coord);
+          // Go to the nearest unoccupied friendly space, but don't gum up the production lines.
+          if( station.getResident() == null && (station.getEnvironment().terrainType != TerrainType.FACTORY) )
+          {
+            // Plot a course towards a repair station, but only apply the action if it moves us.
+            // If a unit is stuck on the front lines and can't get away past reinforcements, just gotta knuckle up.
+            GameAction goHome = AIUtils.moveTowardLocation(unit, coord, gameMap);
+            if( !goHome.getMoveLocation().equals(new XYCoord(unit.x, unit.y)) )
```
ThislsAUsername: This check causes badly-wounded units to leave a repair station they're on and seek a new one.
It's arguably a good thing... but definitely arguable.

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+    // Figure out what unit types we can purchase with our available properties.
+    CommanderProductionInfo CPI = new CommanderProductionInfo(myCo, gameMap);
+
+    if( CPI.availableProperties.isEmpty() )
+    {
+      log("No properties available to build.");
+      return;
+    }
+
+    // Order the enemy units by model.cost (not number), lowest first.
```
ThislsAUsername: old comment?

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+    ArrayList<UnitModel> enemyModels = new ArrayList<UnitModel>();
```
ThislsAUsername: A queue would communicate intent better

src/AI/Muriel.java
```diff
@@ -0,0 +1,750 @@
...
+    {
+      enemyModels.add(ent.getKey());
+    }
+
+    // Try to purchase units that will counter the most-represented enemies.
+    ArrayList<PurchaseOrder> shoppingCart = new ArrayList<PurchaseOrder>();
+    while( !enemyModels.isEmpty() && !CPI.availableUnitModels.isEmpty())
+    {
+      // Find the first (most numerous) enemy UnitModel, and remove it. Even if we can't find an adequate counter,
+      // there is not reason to consider it again on the next iteration.
```
ThislsAUsername: This doesn't work because the UnitQuantityComparator sorts in *a*scending order, not descending order.

ThislsAUsername:
Aight. I'm all read up for now (though it looks like you're cooking up more changes in another branch I saw?)

src/AI/Muriel.java
ThatIsAUsername: Doesn't it need defender terrain to determine defense, which influences damage?

src/AI/Muriel.java
```diff
@@ -524,88 +515,6 @@ private void queueUnitProductionActions(GameMap gameMap)
...
-        log(String.format("Removing my %s", myModel));
-        myQuantity.remove(myModel); // Ran out of units of this type; hope we have some other unit model to fill the gap.
-        myModels.remove(myModel);
-      }
-      else
-      {
-        myQuantity.put(myModel, myLeft); // Didn't need all of my guys of this type to counter their guys; bring on whatever's next.
-      }
-    }
-  }
```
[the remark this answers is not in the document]
ThatIsAUsername: I was using it to identify the "biggest threats" in the enemy army, but it ended up not being useful. If the enemy had 20 infantry, it would decide that one Md Tank was enough to deal with the threat, and not recommend building anything else. Chasing down 20 infantry with a single Md Tank is not fun.
The current approach tries to build something each turn to counter the most numerous enemy unit type, which fixes the "one unit is enough" philosophy, but I'm still not completely satisfied. It mostly works out because the most cost-efficient units against infantry (Md Tanks, followed by Neos and then AAs or Tanks I think) tend to be generally good against other things too. I may update this to work a different way at some point, time permitting.
But to your other point, yes, this could totally go into AIUtils, and would definitely be useful for "can I take that hill" type scenarios. However, the initial implementation was kinda hardwired into Muriel, and I didn't care to extract it at the time.

src/AI/Muriel.java
ThislsAUsername: ...so it does. That could be easily passed in though, and is directly modifiable.

src/AI/Muriel.java
ThatIsAUsername: I was initially planning to sort out only the direct-fire ones, but then reversed that decision and didn't rewrite the loops. Silly me.

src/Units/UnitModel.java
ThatIsAUsername: Why is that what I really want? Muriel calculates cost-efficiency by using damage ratio and factoring in cost. Damage ratio is either zero or infinity for indirect-fire interactions, so I'm just ignoring those for now so they don't influence my unit-purchase calculations.

src/AI/AIUtils.java
```diff
@@ -0,0 +1,169 @@
...
+import Engine.Utils;
+import Engine.XYCoord;
+import Terrain.GameMap;
+import Terrain.Location;
+import Units.Unit;
+
+public class AIUtils
+{
+  /**
+   * Figure out all actions unit can perform this turn, and organizes them by location.
```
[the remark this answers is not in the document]
ThatIsAUsername: Oof. Me too.

src/AI/AIUtils.java
ThatIsAUsername: True.

src/AI/Muriel.java
ThatIsAUsername: You're right, we should make the `BattleParams` member variables private. I think I like keeping the terrain type in there, so the object is created with everything it needs to calculate the result of the battle.

src/Units/UnitModel.java
ThislsAUsername: I feel like that's a Muriel-specific calculation, then.
Whether a unit can shoot at 1 range is much less important than whether it can fire after moving... and it's a 100% correlation at this point.

src/AI/Muriel.java
ThislsAUsername: Nono, the whole point of BattleParams is to *make* things public.
The COs can then reach their dirty claws in there and change things to suit them, like certain COs could give an attack buff based on attack range or terrain.
I guess I should get around to pulling that stuff out of my AWBW branch and making PRs for here, heh.

ThislsAUsername: I feel like a lot of this design intent stuff should be documented somewhere, but I dunno where.

ThatIsAUsername:
> Muriel could benefit from Spender's unit queuing logic, by the way.
> I'm not opposed to merging without that, but I'd like it to happen at some point.

Are you referring to the `travelQueue` and `waitQueue` you have in there? What benefit does that offer?

src/AI/Muriel.java
ThatIsAUsername: In practice, ascending order seems more robust. Really, I want to rework how it does the production stuff, but that'll have to be a future patch.

src/AI/Muriel.java
ThatIsAUsername: Eh, I dunno. I mean, it's ordered, because I'm sorting it, but plenty of non-queue things are sorted. I'd just as soon leave it as is.

src/AI/Muriel.java
ThatIsAUsername: Hm, yeah. OK. Comments are for documentation. Each class can have a javadoc block describing its use. Some do already. Unless you have a better idea, I'd stick with the simple no-frills approach.

src/Units/UnitModel.java
ThatIsAUsername: Well, we can certainly add utility functions for `canFireAfterMoving` and `hasRangedWeapon` and `canCapture` and whatever silly else we think of. I chose `hasDirectFireWeapon` because that is precisely what I cared about. The fact that it's correlated with another trait doesn't really matter.

src/AI/Muriel.java
ThislsAUsername: Hah, fair enough. I could complain that a lot of this stuff wasn't really set in stone yet, but that's not a very good excuse.

src/AI/Muriel.java
ThislsAUsername: Fair.
I said that because you process each as you remove it, but it's not critical.

src/AI/Muriel.java
ThislsAUsername: How is ascending order more robust in practice? (I suppose I can see not wanting to try to counter infantry each turn, heh.)
At the least, your documentation should be updated.

ThislsAUsername:
> 
> 
> > Muriel could benefit from Spender's unit queuing logic, by the way.
> > I'm not opposed to merging without that, but I'd like it to happen at some point.
> 
> Are you referring to the `travelQueue` and `waitQueue` you have in there? What benefit does that offer?

I have realized since that the `waitQueue` is actually useless, since you don't need to specifically tell units to wait in place, they just kinda do that on their own.

That said, the travelQueue logic makes it so that you order all positive actions that you can order before you order any travel actions. This means if a unit suicides on the front or retreats, you will never have just waited or moved a unit behind it that could have taken its place.

src/Units/UnitModel.java
```diff
@@ -61,9 +61,9 @@ public UnitModel(String pName, UnitEnum pType, ChassisEnum pChassis, int cost, i
     healableHabs = new HashSet<TerrainType>();
     for (TerrainType terrain : TerrainType.TerrainTypeList)
     {
-      if( ((chassis == ChassisEnum.AIR_HIGH) || (chassis == ChassisEnum.AIR_LOW) && terrain.healsAir()) ||
-          ((chassis == ChassisEnum.TANK) || (chassis == ChassisEnum.TROOP) && terrain.healsLand()) ||
-          ((chassis == ChassisEnum.SHIP) || (chassis == ChassisEnum.SUBMERGED) && terrain.healsSea()) )
+      if( (((chassis == ChassisEnum.AIR_HIGH) || (chassis == ChassisEnum.AIR_LOW)) && terrain.healsAir()) ||
+          (((chassis == ChassisEnum.TANK) || (chassis == ChassisEnum.TROOP)) && terrain.healsLand()) ||
+          (((chassis == ChassisEnum.SHIP) || (chassis == ChassisEnum.SUBMERGED)) && terrain.healsSea()) )
```
ThislsAUsername: Whoops. Good catch.

src/AI/Muriel.java
ThatIsAUsername: Yes. Done.

ThatIsAUsername:
> That said, the travelQueue logic makes it so that you order all positive actions that you can order before you order any travel actions. This means if a unit suicides on the front or retreats, you will never have just waited or moved a unit behind it that could have taken its place.

Hmm, clever. I may not bother getting it in on the ground floor here, but I'll definitely add it to the list of nice optimizations I can make "one day", when I have more time again.

src/AI/Muriel.java
ThatIsAUsername: Yup, not intentional. Good catch.

ThislsAUsername:
> > travelQueue logic 
> Hmm, clever.

Yeah... I really should do that documentation thing, shouldn't I?

PR #138, opened 2019-01-04
Fog mod

ThatIsAUsername:
I'm still looking at tweaking some things, but I figured I should put this up sooner rather than later.
I moved `GameAction` validation to the `getEvents` call, which reduces the chance of `GameActions` being wrong by the time they are executed. This will help to catch when AIs do bad things (like give multiple orders to one unit), and it also lets us check sequence-breaking conditions at event-creation time, so the animation matches what really happens, and `MapController` doesn't have to worry about checking for preemption.
The current Unload preemption tries to stop anything from getting out after the first one, but drop order seems to be determined by hash rather than the order provided by the user. I can see going either way in terms of fixing this.
I also removed `GameMap` from `Location.getResident()`, because life is better this way, and I hid ownership of enemy buildings in fog. Still need to mess with this some more though.

Comments appreciated, but there are likely more changes incoming, if I can find the time.


src/Engine/GameAction.java
```diff
@@ -60,27 +60,41 @@ public AttackAction(GameMap gameMap, Unit actor, Path path, int targetX, int tar
...
+        moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
+        if((null != atkLoc) && (null != gameMap) && gameMap.isLocationValid(atkLoc))
+        {
+          defender = gameMap.getLocation(atkLoc).getResident();
+        }
+      }
+    }
+
+    @Override
+    public GameEventQueue getEvents(GameMap gameMap)
```
ThislsAUsername: If this arbitrates fog interactions, then it should require a MapMaster.

src/Engine/GameEvents/CreateUnitEvent.java
```diff
@@ -53,16 +53,11 @@ public void performEvent(MapMaster gameMap)
       myCommander.money -= myNewUnit.model.getCost();
       myCommander.units.add(myNewUnit);
       gameMap.addNewUnit(myNewUnit, myBuildCoords.xCoord, myBuildCoords.yCoord);
+      myCommander.myView.revealFog(myNewUnit);
```
ThislsAUsername: Yeah, I noticed that bug, but I couldn't think of a way to fix it in the brief time I spent considering it before laziness and apathy took over.

src/Engine/Utils.java
```diff
@@ -558,4 +569,25 @@ public static boolean pathCollides(GameMap map, Unit unit, Path path)
...
+  /**
+   * Evaluates the proposed move, creates a MoveEvent describing it, and adds that event to eventQueu
+   * If the move passes over an obstacle, the resulting MoveEvent will have its path shortened accordingly.
+   * @param gameMap The world in which the action is to take place.
+   * @param unit The unit who is to move.
+   * @param movePath The complete sequence of steps the unit proposes to take.
+   * @param eventQueue Will be given the new MoveEvent.
+   * @return true if the move is created as specified, false if the path was shortened.
+   */
+  public static boolean enqueueMoveEvent(GameMap gameMap, Unit unit, Path movePath, GameEventQueue eventQueue)
```
ThislsAUsername: Once again, if we're arbitrating fog interactions here, we need to force a MapMaster so we know what's what.

src/Terrain/GameMap.java
```diff
@@ -64,6 +64,12 @@ public GameMap(int w, int h)
   public void resetFog()
   {}
 
+  /**
+   * Reveals fog along the given unit, if applicable
```
ThislsAUsername: Along the given unit?

src/Terrain/MapWindow.java
```diff
@@ -53,13 +64,22 @@ public Environment getEnvironment(int w, int h)
...
-  public Location getLocation(int w, int h)
+  public Location getLocation(int x, int y)
   {
-    return master.getLocation(w, h);
+    XYCoord coord = new XYCoord(x, y);
+    Location masterLoc = master.getLocation(coord);
+    Location returnLoc = masterLoc;
+    if( isLocationFogged(coord) )
+    {
+      returnLoc = new Location(returnLoc.getEnvironment(), coord);
```
ThislsAUsername: The cloning solution did occur to me, but I didn't feel like outlawing Locations from being used as the keys in maps.
I guess XYCoord works just as well in most contexts, though.
Especially now that every CO and thus every unit now technically can get you full game information, heh.

ThislsAUsername:
Seems pretty okay, overall.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -192,18 +192,31 @@ public void animate( GameEventQueue newEvents )
...
+    // Don't let humans peek in on what the AI is doing.
+    boolean drawFogEverywhere = myGame.isFogEnabled() && (numHumans > 1) && (myGame.activeCO.isAI());
+
     // Keep pulling events off the queue until we get one we can draw.
     while( null == currentAnimation && !eventsToAnimate.isEmpty() )
     {
       GameEvent event = eventsToAnimate.peek();
-      if( null != event )
+      currentAnimation = event.getEventAnimation( this );
+      boolean isEventHidden = !(null == event.getStartPoint()) && gameMap.isLocationFogged(event.getStartPoint()) && gameMap.isLocationFogged(event.getEndPoint());
```
ThislsAUsername: This would crash if CO abilities had animations.
Perhaps null checking should be done in the XYCoord-based getters?

ThatIsAUsername: Would it? `getStartPoint()` would return null, and when it compares positively with null here, it would be inverted to false, which means the whole chain of 'and's cannot succeed, and the subsequent calls to `getStartPoint()` would not trigger; `isEventHidden` would just be set to false. Or did I mess up the logic somhow?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -258,14 +271,28 @@ public void render(Graphics g)
...
+      //   2+ humans - player vs player - draw what the current player sees, IFF the player is human.
+      GameMap gameMap = myGame.activeCO.myView; // Start by assuming zero humans.
+      int numHumans = countHumanPlayers(myGame);
+      if( 1 == numHumans )
+      {
+        // Since there is only one human, always use the human's vision to determine what is drawn.
+        gameMap = getHumanPlayerMap(myGame);
+      }
+      // Don't let humans peek in on what the AI is doing.
+      boolean drawFogEverywhere = myGame.isFogEnabled() && (numHumans > 1) && (myGame.activeCO.isAI());
```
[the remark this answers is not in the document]
ThatIsAUsername: Ehh, I'm not sure I like fake COs, and it's been an explicit design decision to keep the game-logic code unaware of the UI/drawing code - it should be up to the UI side to pull whatever data it needs from the logic side.
On the other hand, I also don't like having that logic in two places; the normal fix is to just push it into a function, but I needed both `drawFogEverywhere` and `gameMap` to come out of that function, so that wasn't ideal. I guess just take a stab at it if you have an idea.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: You're right. I'm wrong.
I guess whenever we actually have animations for CO abilities, we're gonna have locations for them to happen on as well... but I feel like maybe powers should always be animated? I dunno, probably kick that thought down the road.

ThatIsAUsername: I agree that Commander Abilities should always be animated. The current behavior is to perform the animation if the event is not tied to a location.
That being said, a Commander Ability may trigger other events (unit spawns, etc) that _will_ have associated locations - so I think we have full flexibility to make things work however we want.
The "Commander Ability Activated!" splash or whatever should at least be shown.

src/Engine/GameAction.java
```diff
@@ -480,77 +486,90 @@ public String toString()
...
+          {
+            XYCoord dropXY = myDropoffs.get(unit);
+            if( gameMap.isLocationEmpty(actor, dropXY) )
+            {
+              unloadEvents.add(new UnloadEvent(actor, unit, myDropoffs.get(unit)));
+            }
+            else
+            {
+              // Hah, last guy got stuck in the chute. Everyone else is trapped behind them still. Bummer.
+              break;
```
ThatIsAUsername: Did we ever decide definitively how to handle the case where an attempt to drop in fog is interrupted? This still has the "subsequent drops fail too" approach (which is still not properly ordered).

ThislsAUsername: Uhhh.
I *think* it's the actual order, though that's harder to test.
Either way, executing all valid drops is 100% kosher and easy IMO.

ThislsAUsername: Done in FogMod2Mod

PR #140, opened 2019-01-11
Fog mod2

ThatIsAUsername:
This is just FogMod, but with master merged in.


ThislsAUsername:
I feel like the proper design is just to show fog for the "last human to play" regardless of how many humans there are.
"last human to play" would be set to the first in turn order on game init, so there is actually fog on turn 1 if player 1 is AI.

ThatIsAUsername:
> I feel like the proper design is just to show fog for the "last human to play" regardless of how many humans there are.
> "last human to play" would be set to the first in turn order on game init, so there is actually fog on turn 1 if player 1 is AI.

Not sure I agree. If multiple people are playing (and like, sharing a screen and properly looking away between turns), then that gives an advantage to the human player who goes just before an AI player. For consistency, it seems that if one human player can't see what the AI does, no human player should be able to.

ThatIsAUsername:
> I think only human players on different teams should count as separate, for this purpose.
> I'm torn on whether/how players should be allowed to see the actions of an AI ally if they have human opponents, though.

I think that might be a bit more complex than I want to bite off. They can share vision during their turns; no reason to try and reason about who's looking at the screen during AI turns, or between them.

ThislsAUsername:
That's fair, I guess.
It helps a lot that the AI turn is basically instant since nothing is animated.

ThislsAUsername:
There's still the problem that fog is not drawn during turn 1 if player 1 is AI, though.

ThatIsAUsername:
> It helps a lot that the AI turn is basically instant since nothing is animated.

Well, things that you shouldn't be able to see are not animated, anyway.

> There's still the problem that fog is not drawn during turn 1 if player 1 is AI, though.

What? I don't think I'm seeing whatever you are seeing. Turn 1 if player 1 is AI looks like this for me: 
image

ThislsAUsername:
> Well, things that you shouldn't be able to see are not animated, anyway.

That's maybe a point worth thinking about, too. I don't think the call stack is rolled up when there's nothing to animate for the AI, so that could maybe cause stack size problems?
 
> What? I don't think I'm seeing whatever you are seeing. Turn 1 if player 1 is AI looks like this for me:

Make player 1 Cinder.

ThatIsAUsername:
Ah, touché. I see now.

Also yes, the call stack thing is a real problem. but probably beyond the scope of this change.

ThatIsAUsername:
Fog thing is fixed. How opposed are you to putting off fixing the call stack thing? It shouldn't become too much of an issue except in AI-only matches, and then it still won't because those are in observer mode, and thus are animated.

ThislsAUsername:
Cinder is a good bug magnet.

I'm not really opposed to delaying the call stack fix, but an issue probably should be made.

PR #142, opened 2019-01-12
FogMod2 mod

ThislsAUsername:
I have a habit of blindly hitting ctrl+shift+f and saving, it seems.
Perhaps we should not have quite so persnickety a code style?


src/Engine/MapController.java
```diff
@@ -393,7 +394,8 @@ public void onStateChange()
           myGame.gameMap.getLocation(targ).setHighlight(true);
         }
         // Create an option selector to keep track of where we are.
-        myGame.setCursorLocation(myGameInputHandler.getCoordinateOptions().get(myGameInputOptionSelector.getSelectionNormalized()));
+        myGame
+            .setCursorLocation(myGameInputHandler.getCoordinateOptions().get(myGameInputOptionSelector.getSelectionNormalized()));
```
ThislsAUsername: This is a real wart.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -213,10 +214,14 @@ private void loadNextEventAnimation()
       GameEvent event = eventsToAnimate.peek();
       currentAnimation = event.getEventAnimation( this );
       boolean isEventHidden = !(null == event.getStartPoint()) && gameMap.isLocationFogged(event.getStartPoint()) && gameMap.isLocationFogged(event.getEndPoint());
-      if( null == currentAnimation || isEventHidden || drawFogEverywhere )
+      if( null != currentAnimation && (isEventHidden || drawFogEverywhere) )
```
ThatIsAUsername: Why not keep the condition the same, and just always at least animate nothing?

src/Engine/GameInstance.java
```diff
@@ -150,4 +154,39 @@ public void endGame()
...
+      }
+    }
+    return humans;
+  }
+
+  /**
+   * Returns the map owned by the first human Commander found.
+   * Intended to be used when there is only one human player.
+   */
+  public GameMap getHumanPlayerMap()
```
ThatIsAUsername: I'm not sure this really belongs in `GameInstance`. I can see moving `countHumanPlayers` here--maybe--but this is pretty specific to the map/fog-drawing logic - I don't see it ever being needed for actual game stuff.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Didn't wanna break anything accidentally by changing the function behavior too much.

src/Engine/MapController.java
ThatIsAUsername: Yes

src/Engine/GameInstance.java
ThislsAUsername: That's a fair argument. Could go in MapController, I guess?

src/Engine/MapController.java
```diff
@@ -595,8 +597,12 @@ public void animationEnded(GameEvent event, boolean animEventQueueIsEmpty)
               endAITurn = true;
             }
           }
-          else { endAITurn = true; } // The AI can return a null action to signal the end of its turn.
-          if( endAITurn) startNextTurn();
+          else
```
ThatIsAUsername: Sigh. Sometimes I like consolidated code. Oh well.

src/Engine/MapController.java
ThislsAUsername: I'm all for changing the code style... but I don't think my habit of compulsively formatting is gonna go away.

src/Engine/MapController.java
```diff
@@ -622,6 +628,31 @@ private void startNextTurn()
 
     // Kick off the animation cycle, which will animate/init each unit.
     changeInputMode(InputMode.ANIMATION);
+
+    // Here are the fog-drawing rules. If there are:
+    //   zero humans - spectating - draw everything the current player sees.
+    //   one human - player vs ai - draw everything the human player could see.
+    //   2+ humans - player vs player - draw what the current player sees, IFF the player is human.
+    if( !myGame.activeCO.isAI() )
```
ThatIsAUsername: You changed the logic here from what it was before - with no humans in play, `myView.gameMap` isn't set at all.

I'm also just generally uncomfortable with having `MapController` in charge of the `GameMap` used by `SpriteMapView` - with the exception of dumb struct-like classes, I like classes to be in control of their own members. I'd also prefer the UI code to be the one looking at the game-logic code, and the game code to have as little idea as possible that the UI exists.

src/Engine/GameInstance.java
ThatIsAUsername: MapController shouldn't be concerned with drawing logic. I thing this really belongs where it's actually used, back in `SpriteMapView`.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: Yeah, but at the same time, we could easily still have a chain of un-animated events that cause the same overflow error. Since the most common actions currently have animations, that's not super likely, but I'd rather just moot the possibility.
I'd really prefer to not even need a non-animation, but that might be a dream for a future day.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -203,16 +203,11 @@ private void loadNextEventAnimation()
       currentAnimation = event.getEventAnimation(this);
       boolean isEventHidden = !(null == event.getStartPoint()) && gameMap.isLocationFogged(event.getStartPoint())
           && gameMap.isLocationFogged(event.getEndPoint());
-      if( null != currentAnimation && (isEventHidden) )
+      if( null == currentAnimation || isEventHidden )
       {
         // If we want to animate something, but it shouldn't be shown, animate nothing instead.
```
ThatIsAUsername: Comment out of date.

src/Units/Weapons/TankMGun.java
```diff
@@ -5,7 +5,7 @@
 
   public TankMGun()
   {
-    super(WeaponType.TANKMGUN, 9);
+    super(WeaponType.TANKMGUN);
```
ThatIsAUsername: This... is not really in scope for fog. This is minor so I won't fight it, but in general, we should try to keep PRs focused.

src/Engine/GameInstance.java
```diff
@@ -4,13 +4,15 @@
...
 import Engine.GameEvents.GameEventListener;
+import Terrain.GameMap;
 import Terrain.Location;
 import Terrain.MapMaster;
 import Terrain.MapWindow;
 
 public class GameInstance
 {
   public Terrain.MapMaster gameMap;
+  public GameMap foggedMap;
```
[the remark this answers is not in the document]
ThislsAUsername: That's a good idea... but who would make it and when?

PR #145, opened 2019-01-12
Prices return

ThislsAUsername:
It's funny how easy it is to make changes when you've already made them


src/Engine/GameInput/SelectUnitProduction.java
```diff
@@ -10,15 +10,18 @@
...
+  private ArrayList<String> myStrings;
   private ArrayList<UnitModel> myUnitModels = null;
   private XYCoord myProductionLocation = null;
 
+  @SuppressWarnings("unchecked")
   public SelectUnitProduction(StateData data, ArrayList<UnitModel> buildables, XYCoord buildLocation)
   {
     super(data);
     myUnitModels = buildables;
+    myStrings = (ArrayList<String>) data.menuOptions;
```
ThatIsAUsername: You could remove the `SuppressWarnings` and the cast entirely bu making `myStrings` an `ArrayList<? extends Object>`. Whether that represents an actual improvement is left as an exercise for the reader.

ThislsAUsername: I mean, if I really wanted to go hog wild with type safety, I could just iterate and call toString() on every element.

PR #146, opened 2019-01-12
Cinder no longer crashes on trying to build from a freshly-captured property using Sear

ThislsAUsername:
Cinder really is a bug magnet.


ThatIsAUsername:
Blunt, but effective.

PR #147, opened 2019-01-12
Ability AI

ThislsAUsername:
ISAI with Cinder is almost scary now.
3v1 would be quite difficult if we had luck damage, heh.

### DIFF
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -178,7 +178,7 @@ public class AIUtils
     ArrayList<CommanderAbility> abilities = co.getReadyAbilities();
     for( CommanderAbility ab : abilities )
     {
-      if( flags == (ab.AIflags & flags) )
+      if( flags == (ab.AIFlags & flags) )
       {
         retVal = ab;
         if (null != q)
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -3,9 +3,10 @@ package AI;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Map.Entry;
+import java.util.Queue;
 
 import CommandingOfficers.Commander;
+import CommandingOfficers.CommanderAbility;
 import Engine.GameAction;
 import Engine.GameAction.ActionType;
 import Engine.GameActionSet;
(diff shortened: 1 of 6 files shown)


src/AI/AIUtils.java
```diff
@@ -167,4 +168,24 @@ public static GameAction moveTowardLocation(Unit unit, XYCoord destination, Game
     }
     return stations;
   }
+  
+  /**
+   * Find a usable ability that has all specified flags, and add it to the provided queue.
+   */
+  public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Commander co, int flags)
```
ThatIsAUsername: ugh... I'm not sure I can countenance this.

src/AI/AIUtils.java
```diff
@@ -167,4 +168,24 @@ public static GameAction moveTowardLocation(Unit unit, XYCoord destination, Game
...
+  {
+    CommanderAbility retVal = null;
+    ArrayList<CommanderAbility> abilities = co.getReadyAbilities();
+    for( CommanderAbility ab : abilities )
+    {
+      if( flags == (ab.AIflags & flags) )
```
ThatIsAUsername: Might it make more sense to use OR instead of AND? If an ability is useful in more than one place, it will never be triggered here unless you pass in e.g. `CommanderAbility.PHASE_TURN_START | CommanderAbility.PHASE_TURN_END`, which seems silly on its face.
On the other hand, if these are meant to be mutually exclusive, then there's no reason to use flags.

src/AI/AIUtils.java
ThislsAUsername: Cromulent is a totally cromulent word.

src/AI/AIUtils.java
ThislsAUsername: I'm pretty sure you're wrong.
The AND cuts out anything that's *not* in the flags you passed in. Then, if the ability satisfies all criteria, it gets used.
For instance, say I have a power that heals instantly, and another that heals after combat.
Ability 1 would be flagged as a healing ability and start/end.
Ability 2 would be flagged as a healing ability and start.
If I want a healing ability regardless of form, I can grab either.
If I'm at the start/end of my turn and want to filter for that, I can do that too.

src/CommandingOfficers/CommanderAbility.java
```diff
@@ -1,13 +1,16 @@
...
 
 public abstract class CommanderAbility
 {
+  public static final int PHASE_TURN_START = 1;
+  public static final int PHASE_PRE_BUY = PHASE_TURN_START << 1;
+  public static final int PHASE_TURN_END = PHASE_PRE_BUY << 1;
   public final Commander myCommander;
   protected String myName;
   private int myPowerCost;
+  public int AIflags = PHASE_TURN_START;
```
ThatIsAUsername: I'd take `AIFlags`, or even `AiFlags`, but I'm no fan of `AIflags`.

src/AI/AIUtils.java
ThatIsAUsername: Hmm, yeah, I think you are correct. Carry on.

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: Haha, okay.

PR #148, opened 2019-01-12
Fixes #143 - Units can no longer one-two punch HQs from another property

### LINKED ISSUE #143: Anyone can capture HQs instantly
If you begin capturing a property near an HQ, then move to capture the HQ next turn, a full HP infantry will capture the HQ instantly.
This is due to the capture event using the unit's current capture progress to determine whether to add a CO defeat event.

Proposal: move capture progress to Location, and zero it whenever the Resident changes via setResident().

ThatIsAUsername:
I kinda prefer to capture progress in Unit, at least for now - it's the less invasive change, and it may play better with COs who have a different capture rate.

I'm also leery of putting important info into Locations, since they are somewhat ephemeral now that `MapWindow` can invent them.

### DIFF
--- a/src/Engine/GameEvents/CaptureEvent.java
+++ b/src/Engine/GameEvents/CaptureEvent.java
@@ -1,5 +1,6 @@
 package Engine.GameEvents;
 
+import Engine.XYCoord;
 import Terrain.GameMap;
 import Terrain.Location;
 import UI.MapView;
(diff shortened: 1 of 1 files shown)


ThislsAUsername:
Keeping important info in Locations is basically required.
Really, we just need a copy constructor.
Actually, we could be clever too by making the Location's hash be based entirely on its coordinates. Then you could hash by Location without the clones messing you up.

ThatIsAUsername:
> Keeping important info in Locations is basically required.
> Really, we just need a copy constructor.
> Actually, we could be clever too by making the Location's hash be based entirely on its coordinates. Then you could hash by Location without the clones messing you up.

I started down this path and then realized it doesn't work: if I get a Location at (3, 4) from MapMaster, showing that P1 owns a property and has a unit on it, and a Location at (3, 4) from P2's MapWindow showing the property to be unowned and empty, those are really not the same. Locations enable access to important information, but I don't think they can be an authority. For that, you really need some coordinates and a GameMap; all the Location does is package that GameMap's info on that space into a convenient package.

PR #151, opened 2019-01-14
Combat mods

ThislsAUsername:
Venge:
+50/0 against units that have attacked you since your last turn. (This does not boost counterattacks against a "fresh" enemy, and does not count enemy counterattacks.)
xxxXXX
COP: [Iron Will] +0/20 stats, counterattack as if you had not taken any damage.
SCOP: [Retribution] Counterattack first. +40/-20 stats

This is maybe not a very good CO, but I wanted to get all the concepts on the table and not muck with fog.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -140,7 +140,7 @@ public class Commander extends GameEventListener
    * Allows a Commander to inject modifications that must be evaluated at combat time
    * @param striking - lets the Commander know if it's dealing or taking damage 
    */
-  public void applyCombatModifiers(BattleParams params, boolean striking)
+  public void applyCombatModifiers(BattleParams params, boolean amITheAttacker)
   {}
 
   public void addCOModifier(COModifier mod)
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -12,6 +12,7 @@ import CommandingOfficers.Modifiers.COModifier;
 import Engine.GameAction;
 import Engine.XYCoord;
 import Engine.Combat.BattleInstance;
+import Engine.Combat.BattleInstance.BattleParams;
 import Engine.Combat.BattleSummary;
 import Engine.GameEvents.GameEventListener;
 import Engine.GameEvents.GameEventQueue;
@@ -125,12 +126,17 @@ public class Commander extends GameEventListener
   }
 
   /**
-   * Allows a Commander to inject modifications before evaluating a battle.
+   * These functions Allow a Commander to inject modifications before evaluating a battle.
    * Simple damage buffs, etc. can be accomplished via COModifiers, but effects
    * that depend on circumstances that must be evaluated at combat time (e.g. a
    * terrain-based firepower bonus) can be handled here.
+   * applyCombatModifiers() will serve for most combat changes, like the above example.
+   * changeCombatContext() allows the CO to tweak the BattleInstance itself,
+   * to allow for drastic changes to the combat like counterattacking first or at 2+ range.
    */
-  public void applyCombatModifiers(BattleInstance params)
+  public void changeCombatContext(BattleInstance params)
+  {}
+  public void applyCombatModifiers(BattleParams params)
   {}
 
   public void addCOModifier(COModifier mod)
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -136,7 +136,11 @@ public class Commander extends GameEventListener
    */
   public void changeCombatContext(BattleInstance params)
   {}
-  public void applyCombatModifiers(BattleParams params)
+  /**
+   * Allows a Commander to inject modifications that must be evaluated at combat time
+   * @param striking - lets the Commander know if it's dealing or taking damage 
+   */
+  public void applyCombatModifiers(BattleParams params, boolean striking)
   {}
 
   public void addCOModifier(COModifier mod)
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -134,7 +134,7 @@ public class Commander extends GameEventListener
    * changeCombatContext() allows the CO to tweak the BattleInstance itself,
    * to allow for drastic changes to the combat like counterattacking first or at 2+ range.
    */
-  public void changeCombatContext(BattleInstance params)
+  public void changeCombatContext(BattleInstance instance)
   {}
   /**
    * Allows a Commander to inject modifications that must be evaluated at combat time
--- a/src/CommandingOfficers/CommanderVenge.java
+++ b/src/CommandingOfficers/CommanderVenge.java
@@ -84,9 +84,9 @@ public class CommanderVenge extends Commander
   }
 
   @Override
-  public void applyCombatModifiers(BattleParams params, boolean striking)
+  public void applyCombatModifiers(BattleParams params, boolean amITheAttacker)
   {
-    if( striking ) // only boost while I'm attacking
+    if( amITheAttacker )
     {
       if( counterAtFullPower && params.isCounter )
       {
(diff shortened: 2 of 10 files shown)


src/CommandingOfficers/Commander.java
```diff
@@ -125,12 +126,21 @@ protected void addCommanderAbility(CommanderAbility ca)
   }
 
   /**
-   * Allows a Commander to inject modifications before evaluating a battle.
+   * These functions Allow a Commander to inject modifications before evaluating a battle.
```
ThatIsAUsername: FYI, Javadoc comments only apply to the immediately-following function. The comments describing `changeCombatContext` should be moved down to there.

src/CommandingOfficers/CommanderVenge.java
```diff
@@ -0,0 +1,175 @@
...
+  /** How much power I get when beating them up */
+  public final int VENGEANCE_BOOST = 50;
+  public boolean counterAtFullPower = false;
+  public boolean counterFirst = false;
+
+  public CommanderVenge()
+  {
+    super(coInfo);
+
+    // Retraux has +10 defense to mitigate enemy firststrikes, synergizing with the Vengeance boost
```
ThatIsAUsername: Retraux?

src/CommandingOfficers/Commander.java
```diff
@@ -125,12 +126,21 @@ protected void addCommanderAbility(CommanderAbility ca)
...
    * Simple damage buffs, etc. can be accomplished via COModifiers, but effects
    * that depend on circumstances that must be evaluated at combat time (e.g. a
    * terrain-based firepower bonus) can be handled here.
+   * applyCombatModifiers() will serve for most combat changes, like the above example.
+   * changeCombatContext() allows the CO to tweak the BattleInstance itself,
+   * to allow for drastic changes to the combat like counterattacking first or at 2+ range.
    */
-  public void applyCombatModifiers(BattleInstance params)
+  public void changeCombatContext(BattleInstance params)
```
ThatIsAUsername: I wouldn't call the variable `params`, since there's another function with similar usage that takes in a `BattleParams` object.

src/Engine/Combat/BattleInstance.java
```diff
@@ -89,24 +100,33 @@ public BattleSummary calculateBattleResults()
...
+      this.attacker = attacker;
+      this.defender = defender;
       baseDamage = attackerWeapon.getDamage(defender.model);
       attackFactor = attacker.model.getDamageRatio();
       attackerHP = attacker.getHP();
       defenseFactor = defender.model.getDefenseRatio();
       defenderHP = defender.getHP();
       terrainDefense = 0;
+      this.isCounter = isCounter;
+      combatRef = ref;
```
ThatIsAUsername: Is there any need to store `combatRef`, or is this just an "in case it's useful one day" kind of thing?

src/Engine/Combat/CombatEngine.java
```diff
@@ -1,18 +1,32 @@
...
     // Set up our combat scenario.
-    BattleInstance params = new BattleInstance(attacker, defender, map, attackerX, attackerY);
+    BattleInstance params = new BattleInstance(attacker, defender, map, attackerX, attackerY, false);
+
+    return params.calculateBattleResults();
+  }
+
+  /**
+   * Assuming Commanders get weird, this allows for you to check the results of combat without perfect map info.
+   * This also provides un-capped damage estimates, so perfect HP info isn't revealed by the map.
```
ThatIsAUsername: Is there a reason we don't just do un-capped damage estimates by default? If I'm trying to attack an Infantry[4], and my readout says I can do 32% damage (because that's precisely how much HP it has left), I can't tell if that will do the job. I'd rather it just say I could do 45% damage so I know where I stand. I'm not sure I want a separate "simulate battle" flag to complicate things, especially since the fact you are building an attack action against something means you can see it, and therefore know its strength (the exception here is Commanders who can hide unit HP, but those would just always be shown as full-strength).

ThatIsAUsername: I guess I can see a justification for a `simulate` version to ignore luck (not that we support that), where a call to `calculate` will return different results each time. Hmm, perhaps.

src/Engine/Combat/BattleInstance.java
```diff
@@ -6,27 +6,32 @@
...
 
   /**
    * Set up the CombatParams object. Note that we assume the defender is where he thinks he is, but
    * the attacker may move before attacking, so we take that coordinate explicitly.
-   * @param pAttacker
-   * @param pDefender
-   * @param map
-   * @param attackerX
-   * @param attackerY
+   * @param isSim
```
ThatIsAUsername: Hah, you took out the other `@params`, and added this one.

src/Engine/Combat/BattleInstance.java
ThatIsAUsername: Could always add `range` to params as final. Or just figure it out based on unit positions. There shouldn't be a need to change that even in `BattleInstance`, I'd think. The range is what it is.
The Commander shouldn't need to get to the map via `combatRef` either--and probably shouldn't be allowed to, since can't this be a `MapMaster`? He's got his own `MapWindow` locally if he wants to look around.

src/CommandingOfficers/Commander.java
```diff
@@ -125,12 +126,21 @@ protected void addCommanderAbility(CommanderAbility ca)
...
+  {}
+  /**
+   * Allows a Commander to inject modifications that must be evaluated at combat time
+   * @param striking - lets the Commander know if it's dealing or taking damage 
+   */
+  public void applyCombatModifiers(BattleParams params, boolean striking)
```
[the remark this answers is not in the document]
ThatIsAUsername: Ehh, I mean, it feels like duplicating information, which is a [cardinal sin in programming](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself); I wouldn't do it that way (Like what if `striking` and the CO check disagree? They never should, but now there's a possibility because there are two places you can check). I won't make you change it if you really want to keep it, since it's in a place that shouldn't change much (ideally), but at least change the variable name to something more informative, like `amITheAttacker` or whatever.

src/Engine/Combat/BattleInstance.java
ThislsAUsername: Can't figure it out based on unit positions, necessarily, since the attacker is at the start of his move path.
Could use the combatRef's attackerX/Y, though.
It can be a `MapMaster`, yes. The idea is that you might have a CO that needs map info to make proper damage calculations, and that should be based on full map info.

src/CommandingOfficers/Commander.java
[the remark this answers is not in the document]
ThislsAUsername: ...the way it's implemented, it's actually impossible for the pieces of information to disagree (provided COs don't intentionally introduce bugs).
It passes in true to the CO of the unit that's attacking, which means it's basically just doing the exact same check in reverse.
If the system becomes more complicated somehow, then removing it might make sense... but I like how it is at this point.
How about "onOffense"?

ThatIsAUsername:
I generally prefer to be more explicit - "amITheAttacker" makes it clear
that params.attacker is my unit. "onOffense" takes an extra mental
connection or two to tie those together. It's clear enough to figure out,
but I like to cater to the average rock when possible. That's my take
anyway, call it what you like.

ThislsAUsername:
"amITheAttacker" sounds good.
I partially pushed back on this because I've written a lot of these functions, heh.
Will merge when I get this pushed in.

src/Engine/Combat/BattleInstance.java
ThatIsAUsername: So in short, yes - it is an "in case it's useful one day" thing. I'm not sure I like it though, from a design/data encapsulation perspective. I also shudder a bit whenever a private variable changes to public (like everything in BattleInstance, for example). This is just more of the flexibility vs correctness discussion, and how much time we want to spend on getting it "right". I like to lean towards clean implementations when possible, but I also like to see results in my lifetime.

PR #153, opened 2019-01-16
Muriel no longer knows about enemy units that she can't see.

ThatIsAUsername:


src/AI/AIUtils.java
```diff
@@ -130,6 +130,27 @@
...
+  {
+    Map<Commander, ArrayList<Unit> > unitMap = new HashMap<Commander, ArrayList<Unit> >();
+
+    for( int x = 0; x < gameMap.mapWidth; ++x )
+      for( int y = 0; y < gameMap.mapHeight; ++y )
+      {
+        Unit resident = gameMap.getLocation(x, y).getResident();
+        if( null != resident )
+        {
+          if( !unitMap.containsKey(resident.CO) ) unitMap.put(resident.CO, new ArrayList<Unit>());
```
ThislsAUsername: Why not just do this initially, rather than making it a conditional in the main loop?

ThatIsAUsername: Because I only want to insert entries for COs that exist, and I don't know which COs exist until I find a unit they own... unless I missed a trick somewhere.

ThislsAUsername: You know which COs exist when the game starts. That's 100% fair knowledge.
What models they have is also 100% fair knowledge.
An empty list of units just means a) they haven't built anything yet or b) you can't see them yet.

ThatIsAUsername: All true. However, a) I don't need to get an empty list back for a commander with nothing, and b) this function is in AIUtils, not Commander, so it doesn't have memory, so it doesn't know what commanders exist.
I mean, I guess you could run through the map and round up the Commanders based on HQ location... I'm just not sure why you would want to.

PR #154, opened 2019-01-21
Polished up combat modifier logic

ThislsAUsername:
BattleInstance no longer mutates when calculateDamage() is called.
BattleSummary is now guaranteed to have the units you expect in the
place you expect, assuming no bugs are introduced intentionally.

### DIFF
--- a/src/Engine/Combat/BattleInstance.java
+++ b/src/Engine/Combat/BattleInstance.java
@@ -1,5 +1,10 @@
 package Engine.Combat;
 
+import java.util.AbstractMap;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
 import Terrain.Environment;
 import Terrain.GameMap;
 import Units.Unit;
@@ -8,21 +13,19 @@ import Units.Weapons.Weapon;
 
 /**
  * BattleInstance provides COs with two levels of control of what goes on in combat:
- * changeCombatContext(), which takes in the BattleInstance itself and allows the COs to fiddle with just about anything.
+ * changeCombatContext(), which takes in a CombatContext and allows the COs to fiddle with just about anything.
  * applyCombatModifiers(), which is for directly modifying a unit's firepower in a given attack.
- * All variables are public so that the COs can actually modify them.
  */
 public class BattleInstance
 {
-  public Unit attacker, defender;
-  public Weapon attackerWeapon = null, defenderWeapon = null;
-  public int attackerX, attackerY, defenderX, defenderY;
-  public final GameMap gameMap; // for reference, not weirdness
-  public boolean canCounter = false;
-  public boolean attackerMoved;
+  private final Unit attacker, defender;
+  private final Weapon attackerWeapon, defenderWeapon;
+  private final int attackerX, attackerY, defenderX, defenderY;
+  private final GameMap gameMap;
+  private final boolean attackerMoved;
   /** Determines whether to cap damage at the HP of the victim in question */
-  public final boolean isSim;
-  int battleRange;
+  private final boolean isSim;
+  private final int battleRange;
 
   /**
    * Set up the CombatParams object. Note that we assume the defender is where he thinks he is, but
@@ -45,10 +48,10 @@ public class BattleInstance
     if( 1 == battleRange )
     {
       defenderWeapon = defender.chooseWeapon(attacker.model, battleRange, false);
-      if( null != defenderWeapon )
-      {
-        canCounter = true;
-      }
+    }
+    else
+    {
+      defenderWeapon = null;
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/Engine/Combat/BattleInstance.java
```diff
@@ -59,42 +62,85 @@ public BattleInstance(Unit pAttacker, Unit pDefender, GameMap map, int attackerX
    */
   public BattleSummary calculateBattleResults()
   {
+    CombatContext context = new CombatContext(gameMap, attacker, attackerWeapon, defender, defenderWeapon, battleRange, attackerX, attackerY);
+    Map<Unit, Entry<Weapon,Double>> unitDamageMap = new HashMap<Unit, Entry<Weapon,Double>>();
+    unitDamageMap.put(attacker, new AbstractMap.SimpleEntry<Weapon,Double>(attackerWeapon, 0.0));
```
ThatIsAUsername: This whole `AbstractMap.SimpleEntry<>` thing works, but is not idiomatic or fun to read - I might suggest making a custom class to encapsulate the relationship you are representing. You are basically abusing `Map` to make a genericized tuple class - from what I've heard, the designers of Java specifically didn't include a generic Pair/Tuple class because it's semantically meaningless; better to define your own relationships with more meaning.
You also don't need a `Map` here at all - you will always be associating two sets of three things.

ThatIsAUsername: That is, until we add units with splash damage or something... cross that bridge when (IF) we get there.

src/Engine/Combat/BattleInstance.java
```diff
@@ -45,10 +48,10 @@ public BattleInstance(Unit pAttacker, Unit pDefender, GameMap map, int attackerX
...
     {
       defenderWeapon = defender.chooseWeapon(attacker.model, battleRange, false);
-      if( null != defenderWeapon )
-      {
-        canCounter = true;
-      }
+    }
+    else
+    {
+      defenderWeapon = null;
```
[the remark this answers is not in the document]
ThislsAUsername: Because I made `defenderWeapon` final, it wouldn't let me change it if it was initialized to null. By the same token, it *required* that all code paths set the value.
I could just make it not final, but I want it final.

src/Engine/Combat/BattleInstance.java
ThislsAUsername: Hm, that's fair. I can poke at it a bit more, I guess.

ThislsAUsername:
While Venge's preemptive counter ability is active, your BattleInstance would get the attacker and defender swapped. Combat would then proceed as "normal".
However, that meant that the BattleSummary *also* had its attacker/defender swapped. That's the reason I ended up making the UI code so janky for the damage preview; I couldn't assume that the attacker and defender I passed in were the attacker and defender I got out.
This change fixes all that, so CombatEngine becomes a black box that acts like you expect it to again, regardless of the sorcery that Commanders might get up to.

ThislsAUsername:
I meant to mention it as well, but it's also worth noting that previously, it would be entirely possible for two subsequent calculations by a given BattleInstance to return different results.
Again, taking preemptive counters as the easy example:
I make a BattleInstance with Venge A attacking Venge B. Both Venges have their preemptive counter active.
I calculate damage, and both A and B get to modify the combat context in order. B swaps combatants, so now B is attacking A, and does more damage.
I calculate again, and now B is the attacker. B then A modify the combat context in order. A goes second, being the defender, and swaps the combatants *again*. Now A gets to deal more damage.
The outcome of the fight is based on how many times you ask what the outcome is. In practice, this aspect didn't really matter since BattleInstances are throwaway, but it bothered me.

src/Engine/Combat/BattleInstance.java
ThislsAUsername: >You are basically abusing `Map` to make a genericized tuple class - from what I've heard, the designers of Java specifically didn't include a generic Pair/Tuple class because it's semantically meaningless; better to define your own relationships with more meaning.

I prefer using generic tuples than defining a class to be used in exactly one piece of code, but maybe that's just me.
The duplication between BattleInstance and CombatContext is because one of them should be immutable, the other should mutable. I could make CombatContext a child of BattleInstance with getters and setters, but I didn't feel like rewriting a bunch of code :P

>You also don't need a `Map` here at all - you will always be associating two sets of three things.

Hmm. Okay. I took another look at the code and I'm prepared to argue my case. I'm not against changing it, necessarily, but hear me out here.
....I *will* agree that this could be better documented whatever we decide, though.
I don't need arbitrary capacities; that's entirely true. I *do* need an "unhinged" mapping for the values, though, and a collection is the easiest way I know of (least LoC and conditionals) to do that.
I want to update the value associated with the attacker or the defender, but at compile time I can't know whose damage was just calculated. All I can do is match Unit instances and call it a day.

> That is, until we add units with splash damage or something... cross that bridge when (IF) we get there.

I recommend against giving me ideas like this. I might just go implement them, instead of adding more core functionality.

src/Engine/Combat/BattleInstance.java
```diff
@@ -59,42 +62,85 @@ public BattleInstance(Unit pAttacker, Unit pDefender, GameMap map, int attackerX
...
-      BattleParams defendInstance = new BattleParams(defender, defenderWeapon,
-          attacker, gameMap.getEnvironment(attackerX, attackerY), true, this);
+      BattleParams defendInstance = new BattleParams(context, true);
       defendInstance.attackerHP -= defenderHPLoss; // Account for the first attack's damage to the now-attacker.
 
       // Modifications apply "attacker first", and the defender is now the attacker.
-      defender.CO.applyCombatModifiers(attackInstance, true);
-      attacker.CO.applyCombatModifiers(attackInstance, false);
+      context.defender.CO.applyCombatModifiers(attackInstance, true);
+      context.attacker.CO.applyCombatModifiers(attackInstance, false);
```
ThatIsAUsername: Shouldn't these two be `defendInstance`?

ThatIsAUsername: This kind of thing (or maybe, code that uses these hooks) should really have test cases to prove they do what we expect, especially since nobody's been using it until now. And yes, I know we don't have a lot of Commander test cases, but that's partly because they've been fairly simple until now, and not mucking about with game mechanics too much.

src/Engine/Combat/BattleInstance.java
ThatIsAUsername: Define "unhinged" - the attacker/defender don't change after this point, so there shouldn't be a problem with using two objects for `attackerDamageStats` and `defenderDamageStats` or whatever, unless I'm missing something. Why can't you know whose damage was just calculated at compile-time?

I ain't gonna make you change it, since it's only used in this small section of code, and it's easy enough to figure it out. I'm just saying if I ever have a reason to touch this code, I might not be able to resist changing it myself.

And speaking of lines of code - counting just straight lines in files (non-syntax-aware counting) we seem to be solidly over 16k now. So that's cool.

ThislsAUsername: > Define "unhinged" - the attacker/defender don't change after this point, so there shouldn't be a problem with using two objects for `attackerDamageStats` and `defenderDamageStats` or whatever, unless I'm missing something. Why can't you know whose damage was just calculated at compile-time?

"Unhinged" in that I can just put in or pull out a value corresponding to a particular object, and have "someone else" manage which object it is.
You're missing something (unless I am?). Venge *can* swap the attacker and defender just a few lines down, at line 72. I no longer know that the attacker and defender in the `CombatContext` are the same as those in my `BattleInstance` variables.
The `Params` give me the damage for "attacker" and "defender", but I don't know which one is which anymore. So I just plop the battle results in a map and call it a day.

> I ain't gonna make you change it, since it's only used in this small section of code, and it's easy enough to figure it out. I'm just saying if I ever have a reason to touch this code, I might not be able to resist changing it myself.

Make a class to be used in one function? I guess if you're touching it again, it would be more likely to be used outside of that...

> And speaking of lines of code - counting just straight lines in files (non-syntax-aware counting) we seem to be solidly over 16k now. So that's cool.

Hah, swanky. ~~Does that count the AWBW branch?~~

ThatIsAUsername: > Venge can swap the attacker and defender just a few lines down, at line 72

The `unitDamageMap` is not read for a while, it's written again first on line 86, which is after the calls to `changeCombatContext`. The creation of `unitDamageMap` could be moved to there and work just the same. Though I think it is possible with the current architecture, I don't think you intend for `applyCombatModifiers` to be used for swapping combatants.

But I'll concede there isn't much need to change it in this case. In general though, idiomatic code is nice.

src/Engine/Combat/BattleInstance.java
ThislsAUsername: ....you are absolutely correct.

As for testing, I agree with CombatEngine/BattleInstance tests, and probably bringing in Venge (Cinder too?) to mess with stuff.

I guess I should work on that, then.

src/Engine/Combat/BattleInstance.java
ThislsAUsername: > Though I think it is possible with the current architecture, I don't think you intend for `applyCombatModifiers` to be used for swapping combatants.

It occurs to me that I could restrict this by making BattleInstance the combatRef, but ugh...

ThatIsAUsername: Yeah, I feel like this whole deal could still benefit from some restructuring, but it's working for now at least.

src/Test/TestCombatMods.java
```diff
@@ -0,0 +1,131 @@
...
+  {
+    // Add our test subjects
+    Unit infA = addUnit(testMap, cinder, UnitEnum.INFANTRY, 7, 3);
+    Unit infB = addUnit(testMap, venge, UnitEnum.INFANTRY, 7, 5);
+    
+    // Check our damage for each first strike pre-power...
+    BattleSummary normalAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 4);
+    BattleSummary normalBA = CombatEngine.simulateBattleResults(infB, infA, testMap, 7, 4);
+
+    venge.modifyAbilityPower(42); // juice up
```
ThatIsAUsername: Maybe do `modifyAbilityPower( venge.getReadyAbilities().get(1).getCost() )` to ensure this test case survives any future balance changes?

ThatIsAUsername: And also maybe check the name of the ability to ensure are using the one we think we are. That's less likely to change though, I would think.

src/Test/TestCombatMods.java
```diff
@@ -0,0 +1,131 @@
...
+
+    testPassed &= validate(retribBA.attacker == infB, "    infB attacked, but isn't the attacker.");
+    testPassed &= validate(retribBA.defender == infA, "    infA was attacked, but isn't the defender.");
+    testPassed &= validate(retribBA.defenderHPLoss > retribBA.attackerHPLoss, "    Retribution somehow deprived Venge of first strike?");
+
+    // Clean up
+    testMap.removeUnit(infA);
+    testMap.removeUnit(infB);
+
+    return testPassed;
```
ThatIsAUsername: Should we also test that Venge's defenses are lower with Retribution active, and that firepower is actually higher? Maybe those would be better as tests for the two modifier classes specifically, but verifying they are correctly used in the ability here doesn't hurt either.

What about a test for IronWill, since it's mucking with counter-attack damage?

I lean towards thorough testing when I can, but I know overkill is also a thing, so add what you think makes sense.

src/Test/TestCombatMods.java
ThislsAUsername: `getReadyAbilities()` doesn't help if you don't already have any CO energy, it'll just return empty.
...and I don't think any power will ever cost 42+ stars at base, haha.

Checking the name could work, but it's really unnecessary. If anything changes, you're gonna try to run the game to test it.... and the test log will tell you.

src/Test/TestCombatMods.java
ThislsAUsername: I wouldn't be able to test the defense combat-wise without assuming I know all the internals of the damage formula. If that ever changes, it would also be annoying to update this.
I could build my own BattleParams, but that feels like a little much.

I can check the stats and Venge's attack damage, though.

src/Test/TestCombatMods.java
```diff
@@ -87,6 +88,41 @@ private boolean testBasicMod()
...
+    // Check our damage for each first strike pre-power...
+    BattleSummary normalAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 4);
+
+    venge.modifyAbilityPower(42); // juice up
+    venge.getReadyAbilities().get(0).activate(testMap); // activate Iron WIll
+    
+    // ...and after power
+    BattleSummary retribAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 4);
+    
+    // Check that Venge's Retribution works properly without breaking things (other than balance)
```
ThatIsAUsername: Retribution should be Iron Will

src/Test/TestCombatMods.java
```diff
@@ -87,6 +88,41 @@ private boolean testBasicMod()
...
+    boolean testPassed = true;
+    testPassed &= validate(infB.model.getDefenseRatio() > 100, "    Retribution didn't buff defense.");
```
ThatIsAUsername: Retribution

PR #155, opened 2019-01-21
Wally

ThislsAUsername:
Wally is all about building and busting walls.
This makes him a little overly cautious. Oh well.


ThatIsAUsername:
When Wally gets backed into a corner, he has a tendency to just shut down and stop shooting. If the happens to be backed into a corner on a friendly tile, and Muriel is pushing with Infantry, it can turn into a stalemate, since the infantry can't out-damage the per-turn healing, and the artillery aren't shooting at all.

ThislsAUsername:
Yeah, Wally definitely has a list of things he still wants.
Muriel's comparative buying logic is a big one; without it, he's hosed on maps with air.
He also needs to prioritize targets for his N-hit KO algorithm.
Finally, he definitely needs to stop thinking that using his Neo to hit an infantry with a Md Tank in range of it is "safe".

Hmm. I can't say I've seen that stalemate situation, but I know why it would happen.
"Siege phase" won't attack from a factory. An indirect *can* attack in the "unit value" phase, based on dealing more damage than it is worth. Infantry don't fit that second clause, so yeah.
I guess what Wally should do is check for possible attacks after attempting to travel safely. That will fix this and some other issues I've been seeing.

I'm inclined to just make an issue of the above, since what I mostly wanted was a "first draft" of the AI that had been eating at me. It feels a little pointless to continue fine-tuning my AI when it's not even playable normally.

I think my current near-term wants are as follows, FYI:
Some sort of Commander-driven icon system. If I'm Venge, I'd like to see an icon on the enemies I can get +50 attack against this turn. My thought was for the CO to be able to specify a color and a char to display on a given unit, since having COs pull out images would be bad for UI/engine separation. I do have a system thought out for how it should work, I just want to know what you think the CO should return for drawing. Thoughts?
I'd like to port over my pre-deployed units (FTA counters, yo) and fixes for countries 4+ from my mapJazz branch, though that will take a little bit of extricating, I think. (unless you're fine with empty silos and pipes existing?)
I want to fix my Ability power storage and stuff. I'll get on that at some point, possibly soon.
Game setup menus!

ThislsAUsername:
Are you set on getting a fix for this stuff, or can I merge and make an issue for the Wally enhancements?

ThatIsAUsername:
No, I just have been pretty busy lately, and haven't been able to look through it yet.

I did notice that on Europe with a decent stack of units, Wally takes noticeably longer per-unit to make moves than the others; computational complexity will be something to think about as we get more complex with AI logic.

I think I'm fine with merging this. I just haven't had time to review it properly, but I'm not sure when that will be. I don't want to hold this up if you are happy with it - my only objection would be to try and reduce the stalemate potential, but that isn't critical right this moment, as long as we get to it.

src/AI/AIUtils.java
```diff
@@ -323,4 +355,27 @@ public int getNumFacilitiesFor(UnitModel model)
       return num;
     }
   }
+
+  /**
+   * Compares SearchNodes based on the amount of movePower they possess, and optionally
```
ThatIsAUsername: SearchNodes should probably be Units

src/AI/AIUtils.java
```diff
@@ -214,6 +216,36 @@ public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Comman
     }
     return retVal;
   }
+  
+  /**
+   * @return The area threatened by the unit, against the specified target type
```
ThatIsAUsername: This returns not just area, but also potential damage at each location.

src/AI/AIUtils.java
```diff
@@ -214,6 +216,36 @@ public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Comman
...
+   */
+  public static Map<XYCoord, Double> findThreatPower(GameMap gameMap, Unit unit, UnitModel target)
+  {
+    XYCoord origin = new XYCoord(unit.x, unit.y);
+    Map<XYCoord, Double> shootableTiles = new HashMap<XYCoord, Double>();
+    ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap);
+    for( Weapon wep : unit.weapons )
+    {
+      if( wep.ammo > 0 )
```
ThatIsAUsername: Should we also filter for whether `wep` can affect `target`? It doesn't look like there's a good reason to include spaces where the possible damage is zero.

src/Engine/XYCoord.java
```diff
@@ -16,6 +16,15 @@ public boolean equals(int x, int y)
   {
     return x == xCoord && y == yCoord;
   }
+  
+  public int getDistance(XYCoord other)
+  {
+    return getDistance(other.xCoord, other.yCoord);
+  }
+  public int getDistance(int x, int y)
```
ThatIsAUsername: I'm...  surprised we didn't already have this.

src/Units/UnitModel.java
```diff
@@ -197,6 +197,26 @@ public boolean hasDirectFireWeapon()
     return hasDirect;
   }
 
+  /**
+   * @return True if this UnitModel has at least one weapon that cannot fire after moving.
+   */
+  public boolean hasImmobileWeapon()
+  {
+    boolean hasSiege = false;
+    if(weaponModels != null && weaponModels.size() > 0)
```
ThatIsAUsername: Technically we don't need to verify `weaponModels.size() > 0` - the for loop can just trivially iterate over zero items.

src/Units/UnitModel.java
```diff
@@ -197,6 +197,26 @@ public boolean hasDirectFireWeapon()
     return hasDirect;
   }
 
+  /**
+   * @return True if this UnitModel has at least one weapon that cannot fire after moving.
+   */
+  public boolean hasImmobileWeapon()
```
ThatIsAUsername: This name feels pretty unintuitive. I don't anticipate asking the question "Does my unit have a weapon that it can't use after moving?" - a much more likely question is "Does this unit have a siege/ranged weapon?"

Is there a reason not to rename this to `hasRangedWeapon`? I know those aren't exactly the same question (they would have different answers for BShips that can move and shoot), but I'm not sure the current name is the more useful/relevant one.

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+import Terrain.Location;
+import Terrain.TerrainType;
+import Units.Unit;
+import Units.UnitModel;
+import Units.UnitModel.UnitEnum;
+import Units.Weapons.Weapon;
+import Units.Weapons.WeaponModel;
+
+/**
+ *  Wally makes(?) and breaks walls.
```
ThatIsAUsername: (?)?  Are you not sure what Wally does? A little more info in general might be useful.
At some point, I might want to allow for short AI descriptions from the CO select screen just to provide the gist of what to expect.

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+      while (actions.isEmpty() && !unitQueue.isEmpty())
+      {
+        Unit unit = unitQueue.poll();
+        if( (gameMap.getEnvironment(unit.x, unit.y).terrainType == TerrainType.FACTORY) || !unit.model.hasImmobileWeapon() )
+        {
+          tempQueue.offer(unit);
+          continue;
+        }
+
+        // Find the possible destinations.
```
ThatIsAUsername: Is this just very tongue-in-cheek, or was this comment copied from somewhere?

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+              {
+                Location loc = gameMap.getLocation(xyc);
+                Unit resident = loc.getResident();
+                if( null != resident && resident.CO == myCo && !resident.isTurnOver &&
+                    resident.canAttack(target.model, xyc.getDistance(target.x, target.y), false))
+                {
+                  damage += CombatEngine.simulateBattleResults(resident, target, gameMap, xyc.xCoord, xyc.yCoord).defenderHPLoss;
+                  neededAttacks.put(xyc, resident);
+                  if( damage >= target.getPreciseHP() )
+                    break;
```
ThatIsAUsername: Couldn't you also set `foundKill` to true here and avoid going into `findAssaultKills`?

ThatIsAUsername: I guess that's the first condition you check in `findAssaultKills`, so it works out, but still.

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+        Unit unit = assaultQueue.poll();
+        if( !unit.model.hasDirectFireWeapon() || neededAttacks.containsValue(unit) ) // don't try to attack twice with one unit
+          continue;
+
+        int dist = Math.abs(unit.x - target.x) + Math.abs(unit.y - target.y);
+
+        // Figure out how to get here.
+        Path movePath = Utils.findShortestPath(unit, xyc, gameMap);
+
+        if( movePath.getPathLength() > 0 && unit.canAttack(target.model, dist, true) )
```
ThatIsAUsername: Correct me if I'm wrong, but I don't think this condition can ever pass. Consider:
a) We are specifically looking for mobile attacks in this function; we've already looked at target-adjacent units above on lines 227-241. Movement will be required.
b) `dist` is calculated using the current `unit` and `target` positions, which by a) will not be adjacent. Given that mobile weapons are currently all 1-range, `dist` must therefore always be too far.

I suppose you could pass the check if you had a unit placed outside of direct-fire range, that had both a direct-fire and an indirect-fire weapons, and that somehow didn't use the ranged attack during the siege-assignment phase.
At any rate, I ran a sample game with a breakpoint inside this if, and it never got hit.

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+      // try to get unit value by capture or attack
+      while (actions.isEmpty() && !unitQueue.isEmpty())
+      {
+        Unit unit = unitQueue.poll();
+        XYCoord position = new XYCoord(unit.x, unit.y);
+        
+        if (unit.getCaptureProgress() > 0)
+        {
+          actions.offer(new GameAction.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, position, gameMap)));
+          capturingProperties.add(position);
```
ThatIsAUsername: Based on the logic in `initTurn`, `position` should already be a member of `capturingProperties` is there any reason to add it twice? Alternatively, perhaps `capturingProperties` should be a set?

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+    // if we're safe, we're safe
+    if( isSafe(gameMap, threatMap, unit, xyc) )
+      return true;
+    // don't stand on a factory or in danger for no good reason
+    if( gameMap.getEnvironment(xyc).terrainType == TerrainType.FACTORY || !unit.model.hasDirectFireWeapon())
+      return false;
+
+    // TODO: Determine whether the ally actually needs a wall there. Mechs walling for Tanks vs inf is... silly.
+    // if we'd be a nice wall for a worthy ally, we can pretend we're safe there also
+    XYCoord[] adjacentCoords = { new XYCoord(xyc.xCoord + 1, xyc.yCoord), new XYCoord(xyc.xCoord - 1, xyc.yCoord),
```
ThatIsAUsername: This might be a good candidate for `Utils.findLocationsInRange()`.

src/AI/AIUtils.java
ThislsAUsername: Sure. It's up to the caller to determine what "threat" means.

src/AI/AIUtils.java
ThislsAUsername: Fair enough

src/Engine/XYCoord.java
ThislsAUsername: Nope, but we do have `ManhattanDistanceComparator`

src/Units/UnitModel.java
ThislsAUsername: I feel like I ran into problems with it, at some point...

src/Units/UnitModel.java
ThislsAUsername: Yes, there's a reason.
Immobile weapons, by their nature, cannot reposition to attack something that's not immediately in range.
I want to make sure any "siege"/immobile weapons get a chance to attack first since they are liable to be powerful and will free up attacking space for mobile attackers... and because more attacks is more attacks, period.

Immobile weapons are un-complicated to calculate, and it's important that they attack if possible. Ranged weapons (e.g. DoR battleships) can be *more* complicated to calculate, since they have so many valid places to attack from (and Wally's code fully accounts for that, when trying for N-hit KOs).

src/AI/WallyAI.java
ThislsAUsername: I tried to make him make walls, but I'm not so sure he's good at it, haha.
I'm not sure what exactly *should* go here, and to what degree of precision.

I agree it would be cool, yeah. Could be a COInfo page, possibly? A tooltip might be better, though.

src/AI/WallyAI.java
ThislsAUsername: Uhhh. Let's go with yes?

src/AI/WallyAI.java
ThislsAUsername: I didn't think of it.

src/AI/WallyAI.java
ThislsAUsername: Yeah, that's fair.

src/AI/WallyAI.java
ThislsAUsername: Hah, good catch.

src/AI/WallyAI.java
ThislsAUsername: Huh. I could have sworn I've seen it trigger.

Regardless, the calculation of `dist` is wrong. It should be from `xyc` to `target`. Good catch.

src/Engine/XYCoord.java
ThatIsAUsername: Yeah. Useful, but not the same.

src/AI/WallyAI.java
ThatIsAUsername: > I'm not sure what exactly should go here
Whatever was your guiding principle in constructing this AI. ISAI focuses on infantry, Spender tries to spend everything, Muriel builds units to counter whatever the opponent has. Wally's current phrase "makes(?) and breaks walls" is not very clear. As in, he tries to form proper British-style battle lines and prevent his opponent from doing likewise? No gory algorithm details of course; just a high-level description of the way decisions are made, or the value function(s) in play.

Yeah, I might lean towards either a floating box at the bottom of the COSelect screen with a blurb about whatever is highlighted, or a Q-menu a la the CO Info page. This is the wrong PR for that discussion though.

src/AI/WallyAI.java
ThatIsAUsername: "Possible Destinations? Right where I'm currently sitting. Done."

ThislsAUsername: I guess it'd be fair to remove the "s" then, eh?

src/AI/WallyAI.java
```diff
@@ -0,0 +1,794 @@
...
+  {
+    Double threat = threatMap.get(myCo.getUnitModel(unit.model.type)).get(xyc);
+    int threshhold = unit.model.hasDirectFireWeapon() ? DIRECT_THREAT_THRESHHOLD : INDIRECT_THREAT_THRESHHOLD;
+    return (null == threat || threshhold > threat);
+  }
+
+  /**
+   * @return whether it's safe or a good place to wall
+   */
+  private boolean canWallHere(GameMap gameMap, Map<UnitModel, Map<XYCoord, Double>> threatMap, Unit unit, XYCoord xyc)
```
ThatIsAUsername: This function checks if either 1) `unit` is save (no enemies can substantially threaten it), or 2) `unit` can park next to a more-valuable ally. Right? The current comments ("if we'd be a nice wall for a worthy ally") seems to imply taking into account position relative to the enemy, but that doesn't seem to be the case; we are just as likely to "wall" on the home side of an exposed arty as on the threatened side. Or am I misreading?

src/AI/WallyAI.java
ThatIsAUsername: Hah, I guess.

src/AI/WallyAI.java
ThislsAUsername: > This function checks if either 1) unit is safe (no enemies can substantially threaten it), or 2) unit can park next to a more-valuable ally. Right?

Yes.

> The current comments ("if we'd be a nice wall for a worthy ally") seems to imply taking into account position relative to the enemy, but that doesn't seem to be the case; we are just as likely to "wall" on the home side of an exposed arty as on the threatened side. Or am I misreading?

Not quite. This function would have already returned true as the back-side spot *is* actually safe.

I suppose a future version of the function might end up checking if the spot is somewhere an enemy could stand on to attack. Blocking off the enemy such that it can't even approach the spot in the first place would be cool too, but harder to figure out probably.

ThislsAUsername: But also, bear in mind that the AI can't currently account for powers at all, so whether an enemy can get here this turn with their current stats doesn't necessarily determine whether walling here is a good idea.

src/AI/WallyAI.java
```diff
@@ -0,0 +1,788 @@
...
+
+  public WallyAI(Commander co)
+  {
+    myCo = co;
+    for( UnitModel um : myCo.unitModels )
+    {
+      for( WeaponModel wm : um.weaponModels )
+      {
+        if( wm.canFireAfterMoving )
+          mobileAttackRange = Math.max(mobileAttackRange, wm.maxRange);
```
ThatIsAUsername: So I can see why you record this and how you use it below, but storing the maximum theoretical attack range across all units as a class-level variable just feels wrong on a very fundamental level.

I mean, it works, and it's your AI, so don't change it if you don't want to, but this smells something awful.

ThislsAUsername: Haha.
I guess I could just calculate it as part of the function, but Wally already calculates possibly too much all the time (as you yourself have mentioned the performance impact).

Realistically, we shouldn't be handing out buffs to mobile unit range anyway, so leaving it in the constructor is fine.

ThatIsAUsername: Oh, man. I just got a _great_ idea for a new CO...

ThislsAUsername: I say "shouldn't" because it's nigh-impossible to balance reasonably, not because it's infeasible.

ThatIsAUsername: Actually; I thought I was just being sarcastic, but how will this play with your proposed battleship rework?

src/AI/WallyAI.java
ThislsAUsername: He attempts to meatshield his strongest assets, and tries to only commit units when he can get as much value as he has committed.

It doesn't entirely work, but I'm not really interested in poking at it more until gameSetup is merged 'cause I'm tired of hacking in AI.

src/AI/WallyAI.java
ThislsAUsername: We don't have any range-boosting COs, do we?

The B-ship would read as having 5 maximum range, so it would result in Wally calculating more squares than is really needed.

ThatIsAUsername: > We don't have any range-boosting COs, do we?

None that boost mobile range, no. Who knows how long it'll take to fix that though?

ThislsAUsername: > None that boost mobile range, no. Who knows how long it'll take to fix that though?

If you don't count the test branch, we don't have any that boost range at all.
Probably less than an hour.

ThatIsAUsername:
Is this PR OBE now that the new one is open?

ThislsAUsername:
Yessir.

ThatIsAUsername:
Superseded by PR #197

PR #158, opened 2019-01-26
Ability costs are now granular

ThislsAUsername:
I originally thought I wanted this change a lot. I now discover that it's less satisfying than I wanted it to be, heh. It simplifies the charging logic... in exchange for a similar increase in complexity for the UI.

I can take or leave the change to funds; TBH, it might be better to leave it as is since leaving it as "discretes" instead of funds makes it more human-readable.
There's two bits of this I really care about/want:
Granular ability cost scaling
The little UI refactor of pulling the constant into `buildCOPowerBar()` so there isn't quite so much magic

The granularity can be gotten by just using a double... and frankly, I'm not sure why I didn't do that in the first place.
I really don't know why I decided I really needed a datamodel change *now* rather than earlier when I added power charging.


src/CommandingOfficers/Commander.java
```diff
@@ -333,13 +333,14 @@ public void receiveBattleEvent(final BattleSummary summary)
...
+      double power = 0; // value in funds of the charge we're getting
+      
       // Add up the funds value of the damage done to both participants.
       power += myHPLoss / minion.model.maxHP * minion.model.getCost();
       // The damage we deal is worth half as much as the damage we take, to help powers be a comeback mechanic.
       power += myHPDealt / enemy.model.maxHP * enemy.model.getCost() / 2;
-      power /= CHARGERATIO_FUNDS; // Turn funds into units of power
-      power += myHPDealt / CHARGERATIO_HP; // Add power based on HP damage dealt; rewards aggressiveness.
+      // Add power based on HP damage dealt; rewards aggressiveness.
+      power += myHPDealt * CHARGERATIO_HP;
```
ThatIsAUsername: Why change division to multiplication? Was this an error before?

src/CommandingOfficers/CommanderAbility.java
```diff
@@ -9,14 +9,16 @@
   public static final int PHASE_TURN_END = PHASE_BUY << 1;
   public final Commander myCommander;
   protected String myName;
-  private int myPowerCost;
+  public final double numDiscretes; // Ability cost in a chunky form. Can serve as a reference to the "original cost" of the ability.
```
ThatIsAUsername: "a chunky form"? I have no clue what this is supposed to mean.

ThatIsAUsername: Also, if it's supposed to be the original cost of the ability, `numDiscretes` is a terrible name.

ThatIsAUsername: Also, it doesn't appear to ever be referenced?

src/Test/TestCombatMods.java
```diff
@@ -98,7 +95,7 @@ private boolean testRetribution()
     BattleSummary normalAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 4);
     BattleSummary normalBA = CombatEngine.simulateBattleResults(infB, infA, testMap, 7, 4);
 
-    venge.modifyAbilityPower(42); // juice up
+    venge.modifyAbilityPower(Commander.CHARGERATIO_FUNDS*42); // juice up
```
ThatIsAUsername: I'm not a fan of needing to know to invoke `CHARGERATIO_FUNDS` to turn a number into a power level.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -101,6 +101,7 @@ private static void updateAnimIndex()
    */
   public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, double currentPower, boolean leftSide)
   {
+    final double powerDrawScaleW = 4.0/Commander.CHARGERATIO_FUNDS;
```
ThatIsAUsername: Magic 4?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -128,14 +129,14 @@ public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, d
     for( int i = 0; i < abilityPoints.length; ++i )
     {
       int requiredPower = abilityPoints[i];
-      double diff = requiredPower - currentPower;
+      double diff = (requiredPower - currentPower)/Commander.CHARGERATIO_FUNDS;
```
ThatIsAUsername: Adding all of the random `CHARGERATIO_FUNDS` everywhere does not help with intuitiveness.

src/CommandingOfficers/Commander.java
ThislsAUsername: It wasn't. I changed the metric of CO power charge from "discretes" (or stars) to funds.
The exchange rate then changed to 100 funds/HP instead of 90 HP/star

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -128,14 +129,14 @@ public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, d
...
       atLeastOne |= (diff < 0);
       BufferedImage segment = (diff > 1) ? powerBarPieces.getFrame(0)  // empty
           : ((diff > 0.5) ? powerBarPieces.getFrame(1)                 // 1/3 full
               : ((diff > 0) ? powerBarPieces.getFrame(2)               // 2/3 full
                   : ((slowAnimIndex == 0) ? powerBarPieces.getFrame(3) // filled
                       : powerBarPieces.getFrame(4))) );                // blinking
-      int drawLoc = (requiredPower * 2) - imageBufferW - 3; // -3 to center the image around the power level.
+      int drawLoc = ((int) (requiredPower*powerDrawScaleW)) - imageBufferW - 3; // -3 to center the image around the power level.
```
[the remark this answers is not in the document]
ThislsAUsername: That does sound like a better name, yes.
Seems reasonable to change it along with whatever we eventually decide on the whole discretes/storing charge as funds or stars/scaling stuff, rather than separately.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThatIsAUsername: The two magic 2's were not multiplicative with each other. The new power bar is twice as long as the old one.

src/CommandingOfficers/CommanderAbility.java
ThatIsAUsername: Two different game modes where you add free power, but in different increments? Why not just have a "free power [amt]" game mode then? There is no reason to do it two ways.
I don't like `discretes` is a noun here - I get why, but this is not ideal. We also add at minimum one power unit each time the ability is used, but I guess you are trying to change that here too, by keeping the same number of units, but each unit is "bigger". Which means the "discretes->funds" problem you describe is self-inflicted by this pull request.

> This uh, wasn't super-thoroughly thought out. 

Yes.

> But it was bugging me

What, exactly? Just that the power costs are not double-precision? That is easily changed without doing the rest of this stuff.

ThatIsAUsername:
I get the distinct impression that this pull request doesn't know what problem it's trying to solve.

Explain?

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: > a "free power [amt]" game mode

See, that's where things get a little awkward. How do you define [amt]?
My initial concept had been you get a certain number *of stars*. So you can keep using powers over and over again with little slowing, but the contributions made by combat will become less relevant over time as the costs expand.
It's equally valid to make a game mode that gives you a free charge amount of *funds*, but that's a very different game mode, even with similar amounts, because the stars will scale up with your power costs but the funds will not.

>I don't like discretes is a noun here - I get why, but this is not ideal.

Do you have a better term?

> We also add at minimum one power unit each time the ability is used, but I guess you are trying to change that here too, by keeping the same number of units, but each unit is "bigger". Which means the "discretes->funds" problem you describe is self-inflicted by this pull request.

I didn't like the minimum one ever since I realized it was a whole quanta. The PR was merged by that time, so I couldn't really complain at that point....
It is. It's the result of trying to understand power charge to be two different things at the same time. The carts don't have this problem, because they scale power costs together... but I don't think I want to make that change for the main branch since I think it encourages you to make the best use of *all* tools in your kit, not just the "most cost effective" one.

> What, exactly? Just that the power costs are not floating-point numbers? That is easily changed without doing the rest of this stuff.

I want to be able to treat powers as costing a certain number of quanta, but I also want to be able to scale their costs independently.
Having power charge stored in quanta rather than funds makes treating the power as if it costs only a certain number of quanta rather silly, since it's two copies of the same unit.

I didn't do a terribly good job of communicating my intent, though.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: You're correct, but I like that scaling better anyway. (I actually applied that change in the AWBW branch, it helps the shorter meters make a little more sense to look at)

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: I want to get my intent done, which is not exactly the same as what I did, haha.

ThislsAUsername:
> Explain?

The summary is this (copied from the above long comment chain):
I want to be able to treat powers as costing a certain number of quanta, but I also want to be able to scale their costs independently.
Having power charge stored in quanta rather than funds makes treating the power as if it costs only a certain number of quanta rather silly, since it's two copies of the same unit.

ThatIsAUsername:
Why do you want to treat ability power as a multiple of some discrete (variably-defined) ability power unit (I'll call it an APU)?

> Having power charge stored in quanta

Is "charge" here supposed to be "cost"? Why would storing cost in APUs make it silly to treat it as if it is stored in APUs? I think I don't understand the second part there.

I think I'm partial to the system we have, where ability-power units are a constant, and the bar changes size if the ability becomes more expensive. If we aren't changing the visuals, there's no reason to refactor the backend to cling to a notion of an unchanging number of variable APUs.

ThislsAUsername:
> Why do you want to treat ability power as a multiple of some discrete (variably-defined) ability power unit (I'll call it an APU)?

Because I'd like to support that cool game mode idea where you can specify N stars free per turn.
As specified above, that's a different mode to something that gives you N funds-worth of charge per turn, since the stars scale with your power cost.

> > Having power charge stored in quanta
> 
> Is "charge" here supposed to be "cost"? Why would storing cost in APUs make it silly to treat it as if it is stored in APUs? I think I don't understand the second part there.

Okay.
I have a power that costs 3 APU, and I set up my game to feed all players 3 APU per turn.
If my second cast of the power is delayed because I actually have to get 3.3 APU, that's not going to make a lot of sense to our target audience.
If I set up my game to give 24k funds charge every turn, then it would make total sense for the second cast of the 3 APU power to be delayed.

This could be pretty consistently represented by each power requiring X APU, but the cost of those APU scales... if powers didn't scale differently.
I guess I probably just need to sit down and figure out what I want on that, 'cause the only way to do that with the way DefendPeace works now... would be to have *both* APU and funds charge stores, and the CO would have to account for the scaling :P

> I think I'm partial to the system we have, where ability-power units are a constant, and the bar changes size if the ability becomes more expensive. If we aren't changing the visuals, there's no reason to refactor the backend to cling to a notion of an unchanging number of variable APUs.

My reason is that I'd like to have a game mode that clings to the notion, heh.

src/CommandingOfficers/CommanderAbility.java
ThatIsAUsername: > See, that's where things get a little awkward. How do you define [amt]

We do assign the abilities a cost - and they'd build intuition that a cost of 3 generally takes x time to charge, etc. We could also just have [amt] be low, medium, or high, and not say exactly how much they are. Or give a range of options over a reasonable span [0-5] or whatever, and the users would figure it out. While these may not be super intuitive, I don't think a funds-based approach (with a scaling factor of 9000) is going to be any more approachable.

> combat will become less relevant over time as the costs expand

This sounds... kinda antithetical to the whole premise of the game.

> I didn't like the minimum one ever since I realized it was a whole quanta

Fair. I'm inclined to agree - this makes the scaling of low-cost abilities much higher than the scaling of more expensive ones. I like the idea of changing ability cost to a double, which would also encourage treating it as a non-discretized thing, since it could get more expensive by 1.34 or whatever.

> the result of trying to understand power charge to be two different things at the same time

An argument for not doing that, I think. Plus drawing things might get strange if your small "stars" are all of a sudden more expensive than your large ones.

ThatIsAUsername:
> My reason is that I'd like to have a game mode that clings to the notion, heh.

Yeah, but we fundamentally don't work that way, and complicate it further by scaling ability costs independently.

> If my second cast of the power is delayed because I actually have to get 3.3 APU, that's not going to make a lot of sense to our target audience.

I think it would, once they know that abilities get more expensive each time you use them. We don't draw APUs, we draw a charge bar, and it gets longer with every cast. Incidentally, this is a reason I'd probably avoid doing the 4x multiplier on power-bar length - we are gonna be off the screen in no time if we do that.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -99,13 +99,13 @@ private static void updateAnimIndex()
   /**
    * Generate an ability-power bar for the given Commander at 1x size. The requester is responsible for applying any scale factors.
    */
-  public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, double currentPower, boolean leftSide)
+  public static BufferedImage buildCoPowerBar(Commander co, double[] abilityPoints, double currentPower, boolean leftSide)
   {
-    final double powerDrawScaleW = 4.0/Commander.CHARGERATIO_FUNDS;
+    final double powerDrawScaleW = 2.0;
```
ThatIsAUsername: This should also be pixelsPerPowerUnit.

src/CommandingOfficers/Commander.java
```diff
@@ -54,7 +54,7 @@
   public Color myColor;
   public static final int DEFAULTSTARTINGMONEY = 0;
   public static final int CHARGERATIO_FUNDS = 9000; // quantity of funds damage to equal 1 unit of power charge
-  public static final int CHARGERATIO_HP = 90; // quantity of HP damage dealt to equal 1 unit of power charge
+  public static final int CHARGERATIO_HP = 100; // Funds value of 1 HP damage dealt, for the purpose of power charge
```
ThatIsAUsername: I don't think this comment makes sense. Funds value per 1HP varies by unit type.

src/CommandingOfficers/Commander.java
ThatIsAUsername: If we are going back to using "ability power" as our measurement unit for now, should we change this back too?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: I coulda sworn I changed it... oops.

src/CommandingOfficers/Commander.java
ThislsAUsername: Actually, it makes perfect sense. It's putting a funds value to the HP damage you deal on your turn.

ThislsAUsername: "for the purpose of power charge" could be clarified, though

ThatIsAUsername: Wait, are you still converting HP to funds, instead of to power? I don't think the math is working out the same was as it did before - we are now multiplying HP dealt by 100 instead of dividing by 90.

src/CommandingOfficers/Commander.java
ThislsAUsername: Nah, the math works out the same, and it's less complicated.

PR #161, opened 2019-01-27
Co marks

ThislsAUsername:
I decided against actually returning a color for two reasons:
It's not actually useful in any of the use cases even the AWBW branch has.
I didn't feel like coming up with an appropriate domain class (which would have to be UI-agnostic and thus mostly useless or perhaps handle multiple UI implementations), and you would have complained if I'd done my tuple shenanigans again.

The symbol drawing code should probably have a null check on the image instead of checking for the null character (actually, both might make sense since that's a non-zero amount of memory/processing being allocated to get the images set up).

And yes, I'm hopelessly addicted to Maps.


src/CommandingOfficers/CommanderVenge.java
```diff
@@ -61,6 +61,20 @@ public void endTurn()
     aggressors.clear();
   }
 
+  @Override
+  public char getSymbol(Unit unit)
```
ThatIsAUsername: Instead of overriding and re-implementing the function, I'd prefer more of this to be moved into Commander, I think. Have a Map (you love those, right) of Units to the markers, and just have `add/removeUnitMarker` functions subclasses can call. It shouldn't be necessary to override `getSymbol` at all that way.

src/CommandingOfficers/Commander.java
```diff
@@ -235,6 +235,20 @@ public UnitModel getUnitModel(UnitEnum unitType)
 
     return um;
   }
+  
+  /**
+   * Returns a character to be displayed on the unit while the map is seen from your perspective.
+   * Primary usage should be pieces of info that aren't otherwise immediately apparent from the map.
+   */
+  public char getSymbol(Unit unit)
```
ThatIsAUsername: I think a different name would be more meaningful. Maybe `getUnitMarker/ing` or `getCustomSymbol` something. The current name sounds like maybe each Commander has an emblem.

src/CommandingOfficers/CommanderVenge.java
ThatIsAUsername: There should also be at least a comment (and maybe an error) informing/enforcing that only valid characters can be used (Currently [A-Z0-9], no lowercase). Should we accept lowercase and just pretend it was uppercase?

ThislsAUsername: I guess a reasonable priority might be:
Owner's mark
my mark
anybody else's mark

Then we could just pass in the list of COs and let 'em go to town.
Then again, if we're going so far as to store all that jazz manually, maybe we should provision for showing all applied marks with cycling or something.

ThatIsAUsername: I wasn't trying to change this to accommodate displaying anybody's mark; I guess you could, but I'm not sure that's what we actually want.

All what jazz? It would be a map of unit to mark, which you have to have already anyway.
Cycling marks is not a bad idea.

ThislsAUsername: I don't have to have any such map, I only need a CO's domain-specific knowledge and a method to convert that to real data, if needed.
Having a map of unit:mark means that there are now two places to keep synched with what should be marked; the domain, and the marks.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -169,13 +172,26 @@ public void drawUnitIcons(Graphics g, Commander activeCO, Unit u, int drawX, int
...
+      if( '\0' != symbol ) // null char is our sentry value
+      {
+        coSymbols.put(co, Character.toUpperCase(symbol));
+      }
+    }
+
+    // Draw one of them, based on our animation index
+    if( !coSymbols.isEmpty() )
+    {
+      @SuppressWarnings("unchecked") // Can't create an array of generic Entries to pass into toArray(), so we have to cast after the fact
```
ThatIsAUsername: You could avoid the warning by just grabbing the nth entry from the KeySet, and then accessing the value using the key.

Also, I'm not sure a map is really the best option here. You could just drop the commanders who have symbols into an array, and then make a second call to `getCustomSymbol` on the one at your chosen index.

ThislsAUsername: Since it's a set and sets are unordered, I wasn't able to.

....and your second paragraph is why we have code review.

ThatIsAUsername:
The symbol rotation speed is a bit fast. You could probably use some kind of modulus magic to slow it down.

When selecting a unit to move, all symbols (including health) disappear until another unit is selected or the action is performed.

Originally, symbols were going to be shown to a CO if a) I placed the symbol, or b), another CO marked their own unit. This means that CO C wouldn't be able to see that CO A marked CO B but it looks like the current implementation allows that. Is this on purpose?

ThislsAUsername:
> The symbol rotation speed is a bit fast. You could probably use some kind of modulus magic to slow it down.

Thought about that, but didn't come up with anything

> When selecting a unit to move, all symbols (including health) disappear until another unit is selected or the action is performed.

Huh, that's weird. I'm not sure how that's happening, but it isn't doing that on master.

> Originally, symbols were going to be shown to a CO if a) I placed the symbol, or b), another CO marked their own unit. This means that CO C wouldn't be able to see that CO A marked CO B but it looks like the current implementation allows that. Is this on purpose?

That's correct, and it is on purpose 'cause it's useful info.
The other one was 'cause I didn't want to do cycling (due to laziness) and I had a "clever" way of showing what's "most important" :P

ThatIsAUsername:
> That's correct, and it is on purpose 'cause it's useful info.

Useful, perhaps, but I'm not sure C has any business knowing that A marked B. Like, is A advertising whomever he marks?

The new mark-rotation speed is definitely better, though.

ThislsAUsername:
> 
> Useful, perhaps, but I'm not sure C has any business knowing that A marked B. Like, is A advertising whomever he marks?

The idea is to provide tactical awareness that's not available by other means.
If there's no fog, C will be able to watch B attack A (who happens to be Venge), and C should be able to remember what he saw (and C knows A is Venge and what Venge does).
It does get a little more awkward to explain in fog, but I'm pretty sure I think that's a bad rabbit hole to go down.

PR #163, opened 2019-02-03
Game setup

ThislsAUsername:
For the palettes:
Read in the first 6 colors on each row.
Top colors are buildings, bottom are units.
The key color is the last color on the top row.

The palettes included are uggo because it's just all the replaced colors from the faction in question.

I assume you aren't big on the idea of having an expanded palette, since that would mean redoing unit art?


src/AI/AILibrary.java
```diff
@@ -0,0 +1,44 @@
...
+  public static ArrayList<AIMaker> getAIList()
+  {
+    if( null == AIList )
+    {
+      buildCommanderList();
+    }
+    return AIList;
+  }
+
+  private static void buildCommanderList()
```
ThatIsAUsername: `buildCommanderList`?

ThislsAUsername: AIs are in fact commanders. You have *absolutely* no evidence that this code was copied from somewhere else.
At all.
:P

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -288,17 +278,103 @@ public static Sprite getMoveCursorArrowSprite()
...
+      mapUnitColorPalettes.put(Color.CYAN, new ColorPalette(cyanMapUnitColors));
+      mapUnitColorPalettes.put(Color.ORANGE, new ColorPalette(orangeMapUnitColors));
+      mapUnitColorPalettes.put(PURPLE, new ColorPalette(purpleMapUnitColors));
+
+      // Throw some color names in there for the defaults
+      // toString() is not user-friendly
+      colorNames.put(Color.PINK, "salmon");
+      colorNames.put(Color.CYAN, "cyan");
+      colorNames.put(Color.ORANGE, "citrus");
+      colorNames.put(PURPLE, "sparking");
```
ThatIsAUsername: Gotta ask - where did Sparking come from? Salmon and Citrus at least make sense, but that one's out of the blue - er, purple maybe.
I think I want to do some re-branding with these color names, at any rate.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -288,17 +278,103 @@ public static Sprite getMoveCursorArrowSprite()
...
+        for( final File fileEntry : folder.listFiles() )
+        {
+          // If it's a file, we assume it's a palette
+          if( !fileEntry.isDirectory() )
+          {
+            String colorName = fileEntry.getAbsolutePath();
+            if( colorName.contains(".png") )
+            {
+              BufferedImage bi = loadSpriteSheetFile(colorName);
+              // Grab the last color on the first row as our "banner" color
```
ThatIsAUsername: Why is banner color the last one? As the "key" I would naturally expect it to come first.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -288,17 +278,103 @@ public static Sprite getMoveCursorArrowSprite()
...
+              Color key = new Color(bi.getRGB(bi.getWidth() - 1, 0));
+              Color[] cUnits = new Color[defaultMapColors.length];
+              Color[] cStructs = new Color[defaultMapColors.length];
+              for( int i = 0; i < defaultMapColors.length; i++ )
```
ThatIsAUsername: Oh. Probably the "banner" color is last just to make this indexing more natural, huh.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -288,17 +278,103 @@ public static Sprite getMoveCursorArrowSprite()
...
     return moveCursorArrowSprite;
   }
+  
+  /**
+   * Sets up the available palettes for use.
+   * Reads in the first 6 colors on each row.
+   * Bottom colors are buildings, top are units.
+   * The key color is the last color on the top row.
+   */
+  private static void initResources()
```
ThatIsAUsername: `initResources` is too general - it's just handling the colors and factions.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -288,17 +278,103 @@ public static Sprite getMoveCursorArrowSprite()
...
   public static ColorPalette getMapUnitColors(Color colorKey)
   {
+    initResources();
     return mapUnitColorPalettes.get(colorKey);
   }
 
+  public static String getColorName(Color colorKey)
+  {
+    initResources();
+    return colorNames.get(colorKey);
```
ThatIsAUsername: Do you think it's worth raising some kind of flag if for some reason `colorKey` is not in `colorNames`?

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -24,6 +24,8 @@
 {
   // This is the physical size of a single map square in pixels.
   public static final int baseSpriteSize = 16;
+  
+  public static final String DEFAULT_SPRITE_KEY = "DEFAULT";
```
ThatIsAUsername: Maybe `DEFAULT_FACTION_NAME` would be more descriptive of how it's actually used?

src/UI/COSetupInfo.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+  
+  public String getCurrentFaction()
+  {
+    return factions[currentFaction.getSelectionNormalized()];
+  }
+  
+  public int getCurrentTeam()
+  {
+    return currentTeam.getSelectionNormalized() - 1; // -1 means "no team"
```
ThatIsAUsername: This means that the actual team number will not match the team number the user chooses. They don't see it again, so it's probably fine, but still odd. It also appears to conflict with the comment back on line 53.

ThatIsAUsername:
> The palettes included are uggo because it's just all the replaced colors from the faction in question.

Yeah. I want to purge those. Can't stand uggo palettes.

> I assume you aren't big on the idea of having an expanded palette, since that would mean redoing unit art?

Correct. I'm not sure why you would want to expand them either, really. I don't think AW itself has more colors.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: There's an AWBW team named Purple Lightning, so I couldn't just use "purple".
"Violet" wouldn't really fit either.
"Sparking" fits with "Lightning" just fine, and kinda-sorta fits. *shrug*

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: Because it's usually the lightest (or one of the lightest) colors in our pre-defined sets.
....And also 'cause I was working with the super-ugly palettes.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: I added a default to salmon and... default... in the test branch.
We don't really have any sort of proper "hey, user, you did a wrong thing" other than stuff going weird or crashing, yet.
That's a system that probably should exist, but it's not super-useful to me at this time and I don't have any ideas.

src/UI/COSetupInfo.java
ThislsAUsername: A *selection* of 0 is "no team"
A *result* of -1 is "no team"
They come from the same selection value, but are just stated differently.

I agree it's a little odd. In fairness, it should maybe instead be special cased in the CO-generation logic as 0 -> -1, but I dunno.

ThislsAUsername:
> > The palettes included are uggo because it's just all the replaced colors from the faction in question.
> 
> Yeah. I want to purge those. Can't stand uggo palettes.

My intent was to eventually turn them into not-uggo palettes, hah.

> > I assume you aren't big on the idea of having an expanded palette, since that would mean redoing unit art?
> 
> Correct. I'm not sure why you would want to expand them either, really. I don't think AW itself has more colors.

A few of the AWBW teams do, but it's not all that big of a deal.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Since I'm not sure my phrasing is clear - it's just the name that is too general; maybe something like `loadFactionsAndColors` or whatever, but `initResources` is less descriptive than I prefer.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Those both seem like orthogonal considerations?

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Why does "Violet" not fit? Look, some Violets:
https://duckduckgo.com/?q=violet&atb=v87-5_g&iax=images&ia=images

Also, what about Argent? It looks like that's normally more of a silver-grey color, not a deep blue. Plus I feel like that color scheme isn't balanced as well as it could be.
Lavender looks excellent, though.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: The missing piece is perhaps that the ugly palettes are sorted dark-to-light?
Also, our text is black.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: ....Well, okay. "Violet" wouldn't exactly thrill me as part of a country name, but I'm okay with it if you want it lol.

Argent and Lavender were both made by other people; I simply stole with permission (and credit).
Argent's building palette's just a copy-paste with no basis, so I think it's fair to go wild and change that if desired.
Lavender was made by BlueLink; I'm sure he'd appreciate hearing your praise, hah.

ThatIsAUsername: > Argent's building palette's just a copy-paste with no basis

Copy-paste from what? Also, I dunno if changing it to silver would be appreciated... why is it named Argent, anyway?
> "Violet" wouldn't exactly thrill me as part of a country name

Heh. Wait until you hear what I've got in mind for Salmon.

ThislsAUsername: > > Argent's building palette's just a copy-paste with no basis
> 
> Copy-paste from what? Also, I dunno if changing it to silver would be appreciated... why is it named Argent, anyway?

Copy-paste of the unit colors, of course.
Original: https://cdn.discordapp.com/attachments/455901857798291459/542494961484955648/unknown.png
Post with context: https://discordapp.com/channels/313453805150928906/455901857798291459/542494961484955649

> > "Violet" wouldn't exactly thrill me as part of a country name
> 
> Heh. Wait until you hear what I've got in mind for Salmon.

Oh, goody. I can't wait.
(It'd better not be Pink, that's taken already)

ThatIsAUsername: Oh, the building colors are just copies of the unit colors? That explains why the units look fine, but the buildings look wonky. As for the name, Azure (which someone at source suggested) is closer than Argent to what they actually are. I might tweak the building colors if I have time.

Nope, not Pink. It's gonna be _classy_. Plus it dovetails nicely with DefendPeace, so there's that.

ThatIsAUsername: Oooh, they are pretty close to Periwinkle too.

ThislsAUsername: Azure works for me.

I'm... less fine with Periwinkle.

ThatIsAUsername: Robin's-Egg? Uranian? Ruddy? Argentinian? Slate? Cornflower? Steel? So many options.

ThatIsAUsername:
> My intent was to eventually turn them into not-uggo palettes, hah.

I'm not gonna be a fan of merging until that happens or we remove them. Of course, the Discord folks might just take that task off our hands...

> A few of the AWBW teams do, but it's not all that big of a deal.

Constraints beget creativity. Also I don't want to rework everything, but let's stick with the first sentence as the real reason.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: Uranian is a little out there, but palatable.
Ruddy's okay.
Slate is too.
Steel... is a little weird.
The others are kind of terrible IMO.

ThatIsAUsername: Haha, the only one's I'd actually consider are Slate/Steel and maybe Uranian... maybe. Azure would also be fine. Maybe I'll roll a die or something.

ThislsAUsername: I feel like Azure fits best.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: For something like this that should (in theory) be less error-prone/more visible when it fails, maybe this is fine, but I think I'd prefer at least a printed warning and a default like you mentioned. If/when we want something more robust, we can add it.

src/UI/COSetupInfo.java
ThatIsAUsername: Is there any reason not to use 0 as "no team"? That would avoid the extra math and confusion. I'd suggest defining `NO_TEAM` as a constant somewhere, but that would conflict with using an `OptionSelector` index for the value.
I can also see the case for using `-1` as the sentinel value though, since it is clearly out of bounds. I guess I'm not gonna fight too hard either way on this one.

ThatIsAUsername:
When adding a faction folder, it's currently necessary to add sprites for every unit type. We should implement a way to specify only those sprites that differ from the default ones - maybe reference another folder/faction as being the "basis" for the current one. That way, Star/Comet/etc can default to the Wars World sprites, and Strong/Patch can default to ours, and the non-infantry sprites would only have to be in one place.

ThislsAUsername:
> When adding a faction folder, it's currently necessary to add sprites for every unit type. We should implement a way to specify only those sprites that differ from the default ones - maybe reference another folder/faction as being the "basis" for the current one.

Hmm. I like it. Would reduce the amount of duplicated data by a lot.
My first instinct was to just add a softlink/shortcut to the "basis" faction folder, but that seems like it might cause more problems than it solves, hah.
My only other idea is to add an optional "basis faction" name either in parenthesis on the folder name, or in a text file in the folder. Both feel ugly, but I don't have any better ideas.
Also, any of the above could result in recursion, or even a circular loop. Neither is thrilling to me.

> That way, Star/Comet/etc can default to the Wars World sprites, and Strong/Patch can default to ours, and the non-infantry sprites would only have to be in one place.

Strong/Patch are COs, not factions?

ThatIsAUsername:
> Both feel ugly, but I don't have any better ideas.

Yeah, my first idea was a text file (probably always "basis.txt") with only the basis name inside. Arguably inelegant, but it works.

> Also, any of the above could result in recursion, or even a circular loop. Neither is thrilling to me.

So... we don't set it up recursively. If we are really paranoid, we could check for recursion, but I'm not really too worried about it.

> Strong/Patch are COs, not factions?

Nope, silly me. I mean Default (which I want to give a proper name) and any future factions we make.

src/UI/COSetupInfo.java
ThislsAUsername: > Is there any reason not to use 0 as "no team"? That would avoid the extra math and confusion. I'd suggest defining `NO_TEAM` as a constant somewhere, but that would conflict with using an `OptionSelector` index for the value.
> I can also see the case for using `-1` as the sentinel value though, since it is clearly out of bounds. I guess I'm not gonna fight too hard either way on this one.

Indexing from zero is natural (err, whole? heh), and I didn't wanna mess with the existing logic.

I think this is where I pull out your "if it ain't broke, don't fix it" argument.

ThislsAUsername:
> Yeah, my first idea was a text file (probably always "basis.txt") with only the basis name inside. Arguably inelegant, but it works.

People would figure it out pretty quickly, and it does make sense with the file-based approach we've taken.
~~If it's good enough for the kernel, it's good enough for us?~~

> > Also, any of the above could result in recursion, or even a circular loop. Neither is thrilling to me.
> 
> So... we don't set it up recursively. If we are really paranoid, we could check for recursion, but I'm not really too worried about it.

Stop confusing me with your "practical solutions" and "not overengineering". We are true auteurs, are we not?

> I mean Default (which I want to give a proper name) and any future factions we make.

If it's getting a proper name, should we abolish the special-casing and treat all factions equally?

ThatIsAUsername:
> If it's getting a proper name, should we abolish the special-casing and treat all factions equally?

Special casing? I'd give it a proper name, but still leave it as the default for anyone who forgets to add basis.txt.

ThislsAUsername:
> Special casing? I'd give it a proper name, but still leave it as the default for anyone who forgets to add basis.txt.

It has a special-snowflake-path: res/unit, rather than res/unit/faction/<name>

ThatIsAUsername:
> > Special casing? I'd give it a proper name, but still leave it as the default for anyone who forgets to add basis.txt.
> 
> It has a special-snowflake-path: res/unit, rather than res/unit/faction/[name]

Right, yes. I think we would have to move it into its own folder like all the others.

ThislsAUsername:
Lemme know whenever you get around to making a name, then.

I have... limited faith... in the efforts of others with regard to palettes, and I realize now that it shouldn't be all that hard to do, so I might get on that soon-ish.

Yes, that's not my declared next feature, but I like the idea of cleaning up this mess of branches.

ThatIsAUsername:
> Lemme know whenever you get around to making a name, then.

I'd go with Thorn. Nice, defensively-thematic option to go with the game's name

> Yes, that's not my declared next feature, but I like the idea of cleaning up this mess of branches.

What isn't? properly-defaulted unit sprites, more color palettes, or something else?

And yeah, I want to clean this mess up too. We've made good progress in the past few days, at least.

src/UI/UIUtils.java
```diff
@@ -0,0 +1,188 @@
...
+            String basis = DEFAULT_FACTION_NAME;
+            try
+            {
+              BufferedReader br = new BufferedReader(new FileReader(fileEntry.getAbsolutePath() + "/basis.txt"));
+              basis = br.readLine().trim();
+              br.close();
+            }
+            catch (IOException ioex)
+            {
+              // if there's no valid basis, there's no problem
```
ThatIsAUsername: I'd rather check whether the file exists before trying to read, instead of just trying to read and catching an exception if it fails. Exceptions should be for when bad/unexpected/exceptional things happen; it's normal for `basis.txt` to be absent.

src/UI/UIUtils.java
```diff
@@ -126,15 +126,19 @@ private static void initCosmetics()
...
-              // if there's no valid basis, there's no problem
+              try
+              {
+                BufferedReader br = new BufferedReader(new FileReader(basisPath));
+                basis = br.readLine().trim();
+                br.close();
+              }
+              catch (IOException ioex)
+              {
+                System.out.println("WARNING! Exception loading faction basis " + basisPath);
```
ThatIsAUsername: Whenever I print a scary warning like this that doesn't stop execution, I like to print what I'm doing instead that allows me to not halt execution. In this case probably just something like `WARNING! Exception loading faction sprite basis! Defaulting to Thorn sprites`. That way the user knows how much they need to care about the warning or if they should try to fix it.

PR #164, opened 2019-02-09
Merge fta mod

ThatIsAUsername:
Added that missing check and swapped empty silos for a bunker. If either of these changes are undesirable, do tell.


src/Terrain/Maps/MapReader.java
```diff
@@ -221,7 +221,7 @@ private static TerrainType stringToCode(String input)
       case "PI":
         return PI;
       case "SU":
-        return SU;
+        return BK;
```
ThislsAUsername: This was my only real objection to it. Now we don't have a two-letter-shortname:two-letter-string correlation.

ThislsAUsername:
I... I could have sworn I pushed up the unit array size check. I have the commits. I just never pushed them up.

src/Terrain/Maps/MapReader.java
```diff
@@ -220,7 +220,7 @@ private static TerrainType stringToCode(String input)
         return SH;
       case "PI":
         return PI;
-      case "SU":
+      case "BK":
```
ThislsAUsername: Now you have to update the maps that use it...

ThatIsAUsername:
> I... I could have sworn I pushed up the unit array size check. I have the commits. I just never pushed them up.
> EDIT: now pushed up on mergeFTAOtherMod
> EDIT EDIT: derp I was reading it backwards. It's already checked in on mergeFTA

Oh. Snap. You totally did. I just uh, wanted to make double sure. I guess I'm the blind one now. Maybe I shouldn't try to program while tired.

ThislsAUsername:
IDK, guess it looks good. I'll just have to convert all the other maps I have :P 

Bunker sprite makes possibly more sense, possibly less, depending on whether it's a vehicle hiding in there or not...

ThatIsAUsername:
> Bunker sprite makes possibly more sense, possibly less, depending on whether it's a vehicle hiding in there or not...

Hah, true enough. Though to be fair, the empty silo doesn't look like it provides cover for anyone. If you have suggestions for something better, do tell.

ThislsAUsername:
> Hah, true enough. Though to be fair, the empty silo doesn't look like it provides cover for anyone. If you have suggestions for something better, do tell.

IDK if it's better *per se*, but a small fort/ring of sandbags could work.

PR #165, opened 2019-02-10
Weather

ThatIsAUsername:
It pretty well works at this point - the main caveat is that the map will only update if you publish a `MapChangeEvent`, though I think that requirement could be removed if we want.

I might want to tweak things some more, but I figured it was better to get this up now and tweak later if needed.


src/UI/Art/SpriteArtist/SpriteMapArtist.java
```diff
@@ -45,15 +45,15 @@ public SpriteMapArtist(GameInstance game, MapView view)
     // This nice little hack will semi-randomly decide what color fog is used for this game.
     // It switches every 10 seconds, so two games started 10 seconds apart will have different fog settings.
     String strTime = Long.toString(System.currentTimeMillis());
-    if( Integer.parseInt(new String() + (strTime.charAt(strTime.length() - 5))) % 2 == 0 )
+    if( myGame.isFogEnabled() || (Integer.parseInt(new String() + (strTime.charAt(strTime.length() - 5))) % 2 == 0) )
```
ThislsAUsername: Shouldn't this be *not* `myGame.isFogEnabled()`?

ThatIsAUsername: Yes, it should. And because you are looking for context, probably. Better to have some silly comments than to miss something important.

src/UI/Art/SpriteArtist/SpriteMapArtist.java
```diff
@@ -227,7 +239,9 @@ private void buildMapImage()
     @Override
     public void receiveMapChangeEvent(MapChangeEvent event)
     {
-      myArtist.buildMapImage();
+      XYCoord xyc = event.getStartPoint();
+      if( null != xyc ) myArtist.redrawBaseTile(xyc); // Redraw one tile if that's all that changed.
+      else myArtist.buildMapImage(); // Redraw the whole map if we have to.
```
ThislsAUsername: I feel like this is probably a weird edge case. Why would you only change the weather on one tile?

ThatIsAUsername: uh... no comment

ThislsAUsername:
From looking at the code as it stands, it doesn't seem like anybody publishes those terrain change events.
I think I wanna make a PR against this branch with my weather queuing and make sure it matches nicely?

I've got a pretty nice "test case" CO I can steal (with permission), too.
If I don't get permission, I'm sure I can come up with something :v

src/UI/Art/SpriteArtist/SpriteMapArtist.java
ThislsAUsername: Actually, that's a good point. How about giving the MapChangeEvent a set of coordinates that are changing?
....for that matter, did you check that file in?

ThislsAUsername: It might be a decent optimization to use this for buildings as well, and just abolish the TerrainObject idea.
That way, the only sprites that are drawn every frame are the map and units (and UI and stuff obv)

ThatIsAUsername:
> From looking at the code as it stands, it doesn't seem like anybody publishes those terrain change events.
> I think I wanna make a PR against this branch with my weather queuing and make sure it matches nicely?
> 
> I've got a pretty nice "test case" CO I can steal (with permission), too.
> If I don't get permission, I'm sure I can come up with something :v

I've got a test case... could benefit from weather queueing, too. I'll probably make a PR against this one soon. It's almost ready, just have to iron out a coupla things.

ThislsAUsername:
> I've got a test case... could benefit from weather queueing, too. I'll probably make a PR against this one soon. It's almost ready, just have to iron out a coupla things.

Aight. Should I wait for that, then? Actually, I'm going to since I'm moving soon and don't wanna spend too much time on this right now...

Anyhow, I got permission. For your perusal:
Emperor Chill (a Cobalt Ice CO, it's a penguin)
D2D: CI properties and spaces adjacent to CI properties have "snow" conditions.
COP "Chilling Wind": Expands this "snow" area to spaces within 3 of a CI property, and removes 2 terrain stars for enemy units within it.
SCOP "Ice Storm": As Chilling Wind, except spaces within 2 of a CI unit are also affected, and all terrain stars are removed.

ThatIsAUsername:
Well, that's... shockingly similar. I almost went with the penguin idea too, but the  name "Pingu" was already taken.

ThislsAUsername:
Having actually taken a look at how the snow looks...
Can't say I'm a super fan of the snowy grass tile.
It's a little hard to tell whether a road is snowed or not.
~~Also, I feel like it might be cool/good to adjust Ave's snow threshold based on the terrain's defense~~-- I mean what?
(Also also, I think XYCoord might want a utility function that returns the 4 adjacent coordinates, as that logic keeps cropping up in places...)

ThatIsAUsername:
Hmm. What about the snowy grass tile are you not a fan of? Snowy grass just looks like... white grass, or just white. Not a whole lot to work with, especially since it can't be too busy, but I'm open to suggestions.

Snow threshold based on terrain defense is an interesting idea... you could get a kind of flowing effect around mountains or whatever. I might play with this idea, time permitting.

ThislsAUsername:
I guess it's just that the blue is too dark.
Something lower contrast would be nice, like here I just shaded it all towards the whiter color:
grass_snow

PR #167, opened 2019-02-16
Cmdr ave

ThatIsAUsername:
Apparently you discovered this, so I figured I may as well put it out there.

I'm a little concerned Ave might be overpowered right now - a 5-tile radius is significant on smaller maps. There are a few parameters we can tweak though, so we can try some things if needed and see what works.
Her CO ability Glacio knocks down trees, damages enemies, and creates snow near her troops. No secondary ability for now. Not sure what it would be, unless I change her name to Elsa and add snow goblins... Jadis was another excellent name possibility, but I don't wanna copy IP (too much).

The whole snow mechanic is maybe more complicated than it should be - it works well, but I'm considering simplifying it considerably - just have a radius of effect that grows with capture time or something, and make the ability range a fixed larger value, rather than depending on how "deep" the snow is. Thoughts? I guess try it out first and see how it goes.

So any way, stuff left to do:
- Make snow disappear if Ave is defeated. It already recedes over time when she loses property, but after defeat she loses the ability to make updates. Maybe integrating your weather stuff will fix that.
- Consider simplifying weather influence effect. Current system is probably more complex than necessary, and may not be intuitive.
- Consider overriding weather-change events for tiles that are covered in "deep enough" snow. E.g., if weather changes to rain, tiles deep in Ave's domain would just stay snow, but her borders could shrink. There's no mechanism in place to do that now (and maybe letting it just change until her next turn is fine, really).
- Create a portrait. Maybe I'll have time to come up with something by the time this is ready to merge.


ThislsAUsername:
> Apparently you discovered this, so I figured I may as well put it out there.

You were betrayed by a little bird named `git fetch`

> I'm a little concerned Ave might be overpowered right now - a 5-tile radius is significant on smaller maps. There are a few parameters we can tweak though, so we can try some things if needed and see what works.

I would be more worried about that if she wasn't competing with the likes of Strong, haha.

> Her CO ability Glacio knocks down trees, damages enemies, and creates snow near her troops. No secondary ability for now. Not sure what it would be, unless I change her name to Elsa and add snow goblins... Jadis was another excellent name possibility, but I don't wanna copy IP (too much).

A weaker power could easily just be the snow spreading without the damage. I think there's room for that between costs of 0 and 15 stars...

> The whole snow mechanic is maybe more complicated than it should be - it works well, but I'm considering simplifying it considerably - just have a radius of effect that grows with capture time or something

It did seem to work well/look nice, but I can definitely see the desire to make it simpler.
You shouldn't probably have to do complicated math to predict what your opponent's D2D is gonna do next turn. (I'd generalize, but the strike powers are kind of a violation of that)

> and make the ability range a fixed larger value, rather than depending on how "deep" the snow is. Thoughts? I guess try it out first and see how it goes.

I agree on the ability range. Since it's 15 stars, I might even make it just expand your existing snow area (and next to units) by 2 tiles and damage/change anything in the whole area you have snowed on.

As a side note, I'm actually rather a large fan of localized weather/mass damage effects.
People being able to coat the battlefield in snow makes sense when it's a skirmish, much less so when you're fighting over the continent of Europe...
(To that end, my brain has attacked me and I have 3 CO concepts playing off of/contrasting with Ave; one snow, two rain. I'd tackle sandstorm, but it has occurred to me that we don't have a good way of reducing unit range based on where you happen to be or where you happen to be targeting...)

> So any way, stuff left to do:
> 
>     * Make snow disappear if Ave is defeated. It already recedes over time when she loses property, but after defeat she loses the ability to make updates. Maybe integrating your weather stuff will fix that.

Yes, weather queuing should fix this. Should I make a PR against this branch, then?
I've realized I also now (via the ideas mentioned above) have a good use case for wanting to forecast weather a few turns in advance without affecting it this turn, which isn't yet supported.

>     * Consider simplifying weather influence effect. Current system is probably more complex than necessary, and may not be intuitive.

I think it's fairly intuitive, but I also skimmed the code before I playtested it with AI, heh.
Could combine the terrain defense angle and your proposed solution:
Everything "in range" of a building receives snowfall, with reduced effect at range (say, 100-20\*distance?) and a threshold based on terrain defense (I'm thinking 100+50\*stars?).
Once snow stops being accumulated on a tile, it disappears at full speed.

>     * Consider overriding weather-change events for tiles that are covered in "deep enough" snow. E.g., if weather changes to rain, tiles deep in Ave's domain would just stay snow, but her borders could shrink. There's no mechanism in place to do that now (and maybe letting it just change until her next turn is fine, really).

I think it could work to have a "weather strength" value for our weather queue.
Zero does nothing, above that can change the weather from default.
Above that, it's just competition between different COs changing the weather.
What do you think? (I personally think it's cool, but I thought of it about 5 seconds ago so grains of salt and all that)

It might also make sense to have the weather changing erase her snow depth, though would definitely get into hard counter territory if someone had access to global weather changes, especially without the above proposal for "weather strength".

>     * Create a portrait. Maybe I'll have time to come up with something by the time this is ready to merge.

You're the art man, man.

ThislsAUsername:
Oh yeah, it would also make sense to give Glacio the post-turn power flag, since snow doesn't help you and mass damage is always nice to have (and it works better after moving directs anyways)

ThatIsAUsername:
> I would be more worried about that if she wasn't competing with the likes of Strong, haha.

Yeah, I have been leaning towards balancing things overpowered... initial testing with Muriel indicates Ave is more than a match for Strong though. Speaking of which, I kinda want to change him up some.

> I think there's room for that between costs of 0 and 15 stars

True. The 15 figure was completely made up; I just knew it should be "high". Part of why this is still a work in progress.

> the strike powers are kind of a violation of that

Strike powers? Like Venge's revenge thing, or something else?

> don't have a good way of reducing unit range based on where you happen to be

I don't think we handle range modifications at all right now - when we do, I imagine it would fall into place. Probably just alter range based on the attacker's location?

> Should I make a PR against this branch, then?

Sounds good to me.

> a good use case for wanting to forecast weather a few turns in advance without affecting it this turn, which isn't yet supported.

Not sure what those use cases are, but couldn't you set a forecast for three days out by just specifying a forecast that happens to match the existing one for the first two days?

> Everything "in range" of a building receives snowfall,

Interesting idea. I kinda like the "spreading outward" effect, but that could be achieved by scaling the snowfall based on distance too. Hmm. Probably worth exploring.

> I think it could work to have a "weather strength" value for our weather queue.

I think this plays hard with the "is this intuitive to users" boundary. Interesting idea; it would be useful for preventing random weather changes from overriding CO weather changes, but I'd caution against using this for much else - I might even just use a boolean `canOverride` which would be false for non-CO changes.

> It might also make sense to have the weather changing erase her snow depth

I considered that and rejected it for the reason you stated - the random weather setting in particular would hard-counter her effects, which is why I considered doing a "at some depth her tiles just don't change" approach, but I'm not sure about that either.

> Oh yeah, it would also make sense to give Glacio the post-turn power flag

Not pre-turn also? Knocking enemy units down a couple pegs before getting in front of their guns seems like a good idea to me.

ThislsAUsername:
> Yeah, I have been leaning towards balancing things overpowered... initial testing with Muriel indicates Ave is more than a match for Strong though. Speaking of which, I kinda want to change him up some.

Muriel *would* tend to make her look OP, I think. Strong will just try to dive into the snow, and Ave will punish.
What did you have in mind?

> True. The 15 figure was completely made up; I just knew it should be "high". Part of why this is still a work in progress.

I'd also like to note that I think the default behavior for mass damage should be to charge your opponent's meter as if it was combat damage.
Then again, maybe that's not as big a deal with our faster charging and potentially making any mass damage limited and/or expensive.

> > the strike powers are kind of a violation of that
> 
> Strike powers? Like Venge's revenge thing, or something else?

Nah, the powers that damage a cluster of units.
You have to calculate where they'll hit based on whatever metric they use, if you're gonna try to counterplay against the power.

> > don't have a good way of reducing unit range based on where you happen to be
> 
> I don't think we handle range modifications at all right now - when we do, I imagine it would fall into place. Probably just alter range based on the attacker's location?

I have it over on the AWBW branch as a subclass of the UnitRemodelModifier, as it's a weapon attribute that alters behavior substantially.
Altering based on the attacker's location sounds right, but I'm not sure exactly how we'd want to put those hooks in.

> Not sure what those use cases are, but couldn't you set a forecast for three days out by just specifying a forecast that happens to match the existing one for the first two days?

You could, but that's a function I would prefer be handled in one place rather than (possibly) many.
The use cases are for 3 other weather CO concepts my brain attacked me with, sort of based on Ave's mechanics. Taking Ave to be a snow/property-based-CO, I've got ideas for snow/movement, rain/units, and rain/attacks. IDK how worth they would be to add, but I will refine my ideas a bit and send 'em to you to see how you feel about them.

> > I think it could work to have a "weather strength" value for our weather queue.
> 
> I think this plays hard with the "is this intuitive to users" boundary. Interesting idea; it would be useful for preventing random weather changes from overriding CO weather changes, but I'd caution against using this for much else - I might even just use a boolean `canOverride` which would be false for non-CO changes.

Okay, but what if you have Ave vs a hypothetical rain-Ave? If they had two cities next to each other, it'd be Shroedinger's weather.
If you made it granular, it might still shift back and forth, but only where the effects are equal in strength rather than everywhere all the time.

> > It might also make sense to have the weather changing erase her snow depth
> 
> I considered that and rejected it for the reason you stated - the random weather setting in particular would hard-counter her effects, which is why I considered doing a "at some depth her tiles just don't change" approach, but I'm not sure about that either.

It's also worth noting that she has no penalties in one of the 3 weather types. A chance of global snow vs a chance of resetting some of her snow spread just might come out overall positive for her.

> > Oh yeah, it would also make sense to give Glacio the post-turn power flag
> 
> Not pre-turn also? Knocking enemy units down a couple pegs before getting in front of their guns seems like a good idea to me.

All powers have pre-turn by default; I was suggesting to have both.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,534 @@
...
+  {
+    GameEventQueue returnEvents = super.initTurn(gameMap);
+
+    // Initialize our snow tracker if needed.
+    if( null == snowMap )
+    {
+      System.out.println("First time init: adding one snow");
+      // Start by initializing owned properties to SNOW_THRESHOLD
+      snowMap = new int[gameMap.mapWidth][gameMap.mapHeight];
+      addSnow(SNOW_THRESHOLD, gameMap, returnEvents);
```
ThislsAUsername: We add snow twice on turn one? Is this intended?

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,534 @@
...
+      for( XYCoord front : frontier )
+      {
+        sortedLeaves.add(new SnowPail(front, oldSnowMap[front.xCoord][front.yCoord]));
+      }
+      frontier.clear();
+
+      // We want to process the current batch of leaves deepest-first, so unload our queue (which
+      // is shallowest-first) into a stack and use that instead.
+      Stack<XYCoord> leafStack = new Stack<XYCoord>();
+      while(!sortedLeaves.isEmpty()) leafStack.push(sortedLeaves.poll().snowCoord);
```
ThislsAUsername: Now, I could be wrong here, but couldn't you just adjust the sorting function? No other cases of its use are immediately jumping out at me...

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,534 @@
...
+      while(!leafStack.isEmpty())
+      {
+        Iterator<XYCoord> stackIter = leafStack.iterator();
+        StringBuffer stackBuf = new StringBuffer("current stack: ");
+        while( stackIter.hasNext() ) stackBuf.append(stackIter.next()).append(" ");
+        log(stackBuf.toString());
+
+        // Get the leaf; only noted with at least SNOW_THRESHOLD are able to be expanded.
```
ThislsAUsername: *nodes

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,534 @@
...
+  }
+
+  /** Ave's units take less cover from forests. */
+  @Override
+  public void applyCombatModifiers(BattleParams params, boolean amITheAttacker)
+  {
+    if( params.defender.CO == this // We are defending, in a FOREST
+        && (params.combatRef.gameMap.getEnvironment(params.defender.x, params.defender.y).terrainType == TerrainType.FOREST)
+        && (params.defender.model.chassis != UnitModel.ChassisEnum.AIR_HIGH)  // And our unit is actually on the ground.
+        && (params.defender.model.chassis != UnitModel.ChassisEnum.AIR_LOW))
```
ThislsAUsername: I think UnitModel.isSurfaceUnit() sounds like a good idea, to centralize this sort of logic.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,534 @@
...
+
+    @Override
+    public int compareTo(SnowPail other)
+    {
+      // Sort them shallowest first.
+      return snowDepth - other.snowDepth;
+    }
+  }
+
+  private static class SnowSpreadDistanceModifier implements COModifier
```
ThislsAUsername: I don't know how much I like the idea of making this a COMod rather than just having the ability boost it and the CO reset it to default on turn init.
It feels like a large gain in LoC without a commensurate boost in correctness/functionality.
Thoughts?

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: If `null == snowMap`, yes. That means this is the first call to `initTurn`, so I add +100 to all of her properties so they have snow, and then fall through to the rest of the initialization routine where we get the +400 per property. This makes sure it spreads to the immediately-adjacent tiles right away. Same reason I have the `CitySnowifier` listener below - to add 100 on capture so the snow spreads as expected the next turn.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: A `PriorityQueue<SnowPail>` called `workingSet` is used below, and is sorted the natural (least first) way.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Fair. I might do that. It's a turn-on turn-off thing, which `COModifiers` are made to do, so I didn't put too much thought into it.

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: Okay, but I feel like she doesn't really need to spread snow on turn 1.
It happens the next turn at the latest, and where multiple properties sit close, it is turn 1.

ThatIsAUsername:
> What did you have in mind [for Strong]?

I don't think his expanded transports are really worth anything, since they are just during his ability - I'd like to just make that passive.  Beyond that, maybe not much, though I might consider something fun like airdropping 8-HP mechs behind enemy lines.

ThatIsAUsername:
- [ ] More predictable ability dynamics
- [ ] Max snow depth/forecast length
- [ ] Adjust Power flags
- [ ] Draw Portrait
- [ ] Balance

src/CommandingOfficers/CommanderAve.java
```diff
@@ -260,27 +257,42 @@ private void relevelSnow(GameMap gameMap, GameEventQueue outEvents)
...
+          }
+          else
           {
-            log(" adding " + workingSet.peek().snowCoord + " to working set");
-            shallowTiles.add(workingSet.poll().snowCoord);
+            mostShallow = snowMap[leaf.xCoord][leaf.yCoord];
+            if( mostShallow == MAX_SNOW_DEPTH )
+              shallowTiles.remove(leaf); // If leaf is max depth, just push extra to other unprocessed tiles.
+            nextMostShallow = mostShallow + snowToSpread;
           }
```
ThislsAUsername: I don't understand what this else branch is supposed to do. The rest of this section only seems to spread snow if shallowTiles has something in it.

ThatIsAUsername: Correct. But `shallowTiles` can have more than one thing in it. If we get to the point where `leaf` and its neighbors are all at `MAX_SNOW_DEPTH`, then we don't want to add any more snow to `leaf`, so we remove it from `shallowTiles` and adjust `nextMostShallow` so that the remaining snow will be evenly spread across the (as-yet unprocessed) adjacent nodes.
This statement can also be hit when we aren't yet at `MAX_SNOW_DEPTH` - in this case we don't remove `leaf` before the final snow-spreading.

ThatIsAUsername: Oh, and if there are no adjacent unprocessed tiles (i.e. if we are running into a corner), then removing `leaf` will cause `shallowTiles` to be empty, and any remaining snow just vanishes in a puff of logic.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Solid idea.

src/Units/UnitModel.java
```diff
@@ -216,4 +216,24 @@ public boolean hasActionType(ActionType actionType)
     }
     return hasAction;
   }
+
+  public boolean isSurfaceUnit()
+  {
+    return (ChassisEnum.SHIP == chassis) || (ChassisEnum.TANK == chassis) || (ChassisEnum.TROOP == chassis);
+  }
```
ThislsAUsername: Hmm. I was kind of planning to use that for determining whether terrain defense applies... I'd forgotten we have a separate chassis for submerged units.
I'm not sure what to say about that.

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: Okay, I see what you're saying.
It's now clear to me that I just don't understand exactly how this all works, and my brain is too addled at this point to fix that.

ThislsAUsername:
> 
> 
> > What did you have in mind [for Strong]?
> 
> I don't think his expanded transports are really worth anything, since they are just during his ability - I'd like to just make that passive. Beyond that, maybe not much, though I might consider something fun like airdropping 8-HP mechs behind enemy lines.

....and here I thought you were considering a nerf. Well, follow your dreams, I guess.

src/Units/UnitModel.java
ThatIsAUsername: Could just do `if( !unit.model.isAirUnit() )`. Not quite as idiomatic as you probably wanted though, but adding `isAffectedByTerrain()` feels a little obtuse.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -423,72 +481,38 @@ protected void perform(MapMaster gameMap)
...
-                ? TerrainType.GRASS : loc.getEnvironment().terrainType;
-            tileChanges.add(new MapChangeEvent.EnvironmentAssignment(coord, Environment.getTile(newTerrain, Weathers.SNOW), 1));
-          }
-
-          // Damage each enemy nearby.
-          if( null != loc.getResident() && myCommander.isEnemy(loc.getResident().CO) )
-          {
-            loc.getResident().alterHP(-2);
-          }
+          // TODO: Apply Freezing.
```
ThislsAUsername: Should frozen units be able to counterattack?
Should they take extra damage?
If you care at all, this is the diff for when I added stunning for VB https://github.com/ThislsAUsername/DefendPeace/commit/a6ae33b9fd7858694154bc07f1327151373faf5d

src/CommandingOfficers/CommanderAve.java
```diff
@@ -501,6 +525,67 @@ protected void perform(MapMaster gameMap)
...
+      ArrayList<MapChangeEvent.EnvironmentAssignment> tileChanges = new ArrayList<MapChangeEvent.EnvironmentAssignment>();
+
+      // Change terrain to snow around each of Ave's units and buildings, and damage trees and enemies.
+      HashSet<XYCoord> affectedTiles = Utils.findLocationsNearProperties(gameMap, Ave, OBLIDO_RANGE);
+      affectedTiles.addAll(Utils.findLocationsNearUnits(gameMap, Ave, OBLIDO_RANGE));
+
+      // Smash things. Don't add snow though.
+      for( XYCoord coord : affectedTiles )
+      {
+        // Add snow around friendlies.
```
ThislsAUsername: *cough\*

src/Engine/Utils.java
```diff
@@ -591,4 +592,24 @@ public static boolean enqueueMoveEvent(MapMaster gameMap, Unit unit, Path movePa
...
+
+  public static HashSet<XYCoord> findLocationsNearUnits(GameMap gameMap, Commander cmdr, int range)
+  {
+    HashSet<XYCoord> tilesInRange = new HashSet<XYCoord>();
+    for( Unit unit : cmdr.units )
+    {
+      tilesInRange.addAll(Utils.findLocationsInRange(gameMap, new XYCoord(unit.x, unit.y), 0, range));
+    }
+    return tilesInRange;
+  }
```
ThislsAUsername: I like these.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: I was kind of assuming stunned == not able to counterattack, but no extra damage taken. What does VB do on the carts? It seems sensible to just use one mechanism for stunning, whatever the cause.

ThislsAUsername: Units VB has stunned can shoot back on the carts. I know it's silly, but it's definitely a thing.
Also worth noting is that Eagle can make stunned units act out of stun.

The nomenclature you used was "freeze", so I assumed it might have some different mechanics.
Disabling counterattacks is something you'd kind of assume any sort of stun would do.
Boosting damage taken is something freezing status frequently does in games.

Also, both of the above points are use cases for third-party combat modifiers. You don't want to do a UnitRemodel on any other CO's units, so fiddling with defense is right out.
Buffing attacks won't work, since that's a global effect.
Disabling counterattack or boosting damage (unless it becomes a feature of Unit, which seems silly) has to be a combat modifier, and Ave can't do it (atm) if someone else attacks the enemy she froze.

ThatIsAUsername: > both of the above points are use cases for third-party combat modifiers

Disabling counterattacks wouldn't require third-party interference - the combat engine would just have to know what stun means. Same would go for increasing damage, potentially, but I wasn't looking to make that happen.

> Disabling counterattacks is something you'd kind of assume any sort of stun would do.

Yes, but it also feels pretty strong, which is probably why they allowed stunned counterattacking on the carts. Any preference which way we go? I'd like to stick with one "stun" effect unless there's an awfully good reason not to. Should we match VB for tradition's sake, or go our own way and do the more intuitive/OP thing?

ThislsAUsername: > > both of the above points are use cases for third-party combat modifiers
> 
> Disabling counterattacks wouldn't require third-party interference - the combat engine would just have to know what stun means. Same would go for increasing damage, potentially, but I wasn't looking to make that happen.

Perhaps stun should natively negate counterattacks (though, as you say, 'tis quite strong), but I wouldn't want to add "target takes increased damage" as a rider to "stun", especially since it should be of configurable magnitude.
I was also thinking the counterattack negation might be a special thing for Ave, seeing as she supposedly freezes the enemies/coats 'em in ice, not just puts 'em out of sorts.
Perhaps it would satisfy everyone to just have a stat delta per unit? That seems like it could get out of whack pretty easily, if someone forgets or applies twice or what-have-you.

> > Disabling counterattacks is something you'd kind of assume any sort of stun would do.
> 
> Yes, but it also feels pretty strong, which is probably why they allowed stunned counterattacking on the carts. Any preference which way we go? I'd like to stick with one "stun" effect unless there's an awfully good reason not to. Should we match VB for tradition's sake, or go our own way and do the more intuitive/OP thing?

Looks like you've solved this already.

ThatIsAUsername:
I used stun as the stun icon because it works well with Ave's ability (and I think generally connotes "not moving"), but other contenders were stunB (Which might work better with VB) or even stunC.

Any thoughts?

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Yeah, I guess. For icons though, there isn't really one that matches both "freeze from cold" and "stun from lightning". Almost seems worth differentiating just for that, except then we'd come up with more causes of stunning.

ThislsAUsername:
> 
> 
> I used stun as the stun icon because it works well with Ave's ability (and I think generally connotes "not moving"), but other contenders were stunB (Which might work better with VB) or even stunC.
> 
> Any thoughts?

IDK, I feel like maybe the icon should be added *by* Ave.
That way it'd actually be settable by the CO... then again, I don't think there's a character we can easily return for that, which would mean COs would need to know that the UI uses images or something else.

ThatIsAUsername:
> IDK, I feel like maybe the icon should be added by Ave.

Ehh, I think I disagree here. I considered something like that, but besides requiring knowledge of UI code, it would create confusion if you had multiple icons for the same effect... and if you have one icon per effect (which I think I'd advocate), the CO doesn't need to choose the icon.
Plus I don't want every future CO that stuns to have to remember to draw the icon.

ThislsAUsername:
> > IDK, I feel like maybe the icon should be added by Ave.
> if you have one icon per effect (which I think I'd advocate), the CO doesn't need to choose the icon.
> Plus I don't want every future CO that stuns to have to remember to draw the icon.

You say "one icon per effect". I would argue that "got a snow dump and stun" is different from "got hit by 3 damage and stun".
If we're being generic here, maybe the little red sign or an X would be appropriate. Or maybe a turtle, if you can fit it in the space...

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -209,6 +209,17 @@ public void drawUnitIcons(Graphics g, Commander[] COs, Unit u, int animIndex, in
...
     }
 
+    if( u.isStunned )
+    {
+      // Get the icon and characterize the draw space.
+      BufferedImage stunIcon = SpriteLibrary.getStunIcon();
+      int iconW = stunIcon.getWidth() * drawScale;
+      int iconH = stunIcon.getHeight() * drawScale;
+
+      // Draw transport icon.
```
ThislsAUsername: "transport" indeed

ThislsAUsername:
Two more silly ideas to throw at the wall:
Glacio should freeze rivers into roads for a turn.
Ave should listen for MapChangeEvents and if the weather is changing from snow and she has 2x `SNOW_THRESHOLD` there, subtract 1x and publish an event to set it back.

ThatIsAUsername:
> You say "one icon per effect". I would argue that "got a snow dump and stun" is different from "got hit by 3 damage and stun".

I'd argue that "got a snow dump and stun" is two effects, as is "3 damage and stun". We don't need a unit icon for "snow dump" and we already have one for "took damage". Just need a stun icon now.

> Or maybe a turtle

Hah, while I do like turtles, I think that would evoke more of a "slow" effect, not a "rooted" effect. Could do a root maybe? 

> Glacio should freeze rivers into roads for a turn.

That... would look confusing. And is probably overkill. Maybe just grant Ave road movement on frozen rivers?

> Ave should listen for MapChangeEvents...

Thought about doing that. Trouble is you don't have the `MapMaster` in the listener events, so you can't publish events in response to events... which is frankly probably for the best, anyway. I think I'm fine with making Ave reclaim her territory at the beginning of her turn.
Alternatively, we could use your idea of weather precedence to make this work. "Natural" weather events would be non-preemptive, and Commander weather would be preemptive. So Ave won't lose snow coverage to a rainy day, but would give and take territory vs a hypothetical rain-Ave - each would get their turf on their turn.

ThislsAUsername:
> I'd argue that "got a snow dump and stun" is two effects, as is "3 damage and stun". We don't need a unit icon for "snow dump" and we already have one for "took damage". Just need a stun icon now.

You win.

> > Or maybe a turtle
> 
> Hah, while I do like turtles, I think that would evoke more of a "slow" effect, not a "rooted" effect. Could do a root maybe?

Or maybe a little swirl, as in confusion/dizziness.?

> > Glacio should freeze rivers into roads for a turn.
> 
> That... would look confusing. And is probably overkill. Maybe just grant Ave road movement on frozen rivers?

I'm groovy with that, so long as it's only for the Glacio turn.

> Alternatively, we could use your idea of weather precedence to make this work. "Natural" weather events would be non-preemptive, and Commander weather would be preemptive. So Ave won't lose snow coverage to a rainy day, but would give and take territory vs a hypothetical rain-Ave - each would get their turf on their turn.

I like that too. Quite a bit, actually.

ThatIsAUsername:
> Or maybe a little swirl, as in confusion/dizziness.?

Yeah, that could work. Like maybe stun (copy)? Or like a light blue/cyan color. Or just stick with team-color, so it's clear who's unit is stunned even if they are covered in icons.

> I'm groovy with that, so long as it's only for the Glacio turn.

Either she can always treat snowy rivers as roads, or never. I'm fine with never.

> I like that too. Quite a bit, actually.

Cool, let's plan on that. I'm inclined to make that a future change though, since it can be implemented independent of this PR, and I don't want to drag this PR out too much more, and there aren't any interactions where it matters yet.

> Except you totally can, just look at Cinder.

Ah, yes. I forgot Commanders can inject events. At least a little bit frightening, but certainly useful.

ThislsAUsername:
> > Or maybe a little swirl, as in confusion/dizziness.?
> 
> Yeah, that could work. Like maybe stun (copy)? Or like a light blue/cyan color. Or just stick with team-color, so it's clear who's unit is stunned even if they are covered in icons.

Any of those options works for me.

> Either she can always treat snowy rivers as roads, or never. I'm fine with never.

*shrug*
Never works for me, too.

> > I like that too. Quite a bit, actually.
> 
> Cool, let's plan on that. I'm inclined to make that a future change though, since it can be implemented independent of this PR, and I don't want to drag this PR out too much more, and there aren't any interactions where it matters yet.

Hunky dory.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,646 @@
...
+      GameEvent event = new MapChangeEvent(tiles);
+      outEvents.add(event);
+    }
+  }
+
+  /**
+   * Spread snow around.
+   * outEvents will be populated with any MapChangeEvents that result.
+   */
+  private void relevelSnow(GameMap gameMap, GameEventQueue outEvents)
```
ThislsAUsername: This is admittedly a pretty cool algorithm, but I think maybe it's a little overly-complex.
I'm not a huge fan of having functions that won't comfortably fit on a monitor, especially if it's a whole bunch of actual complex calculations. (sidenote: for this reason, I was planning to make a "meta AI" that does flow-of-control stuff, so other AIs can just define whatever phases they want.)

Outside of making the logic less complex, would you consider at least just breaking it up into sub-functions so it's easier to follow?

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,646 @@
...
+
+  private static class SnowPail implements Comparable<SnowPail>
+  {
+    public final XYCoord snowCoord;
+    public final int snowDepth;
+
+    public SnowPail(XYCoord coord, int snowAmt)
+    {
+      snowCoord = coord;
+      snowDepth = snowAmt; // Truncate to two-digit precision.
```
ThislsAUsername: ??

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: > EDIT: I feel like this is especially relevant since her powers no longer use the snow map at all.

Sounds like the real solution is to change her powers to rely on the snow map again, eh?

Yeah, I agree this is more complex than it really needs to be at this point. However, it also works very well without issues, so in the spirit of "don't fix what ain't broken" I just wasn't gonna mess with it unless I found out it was broken.
If you think it's worth finding a simpler alternative despite that we can though.

I'm not a fan of using `log(turnsSinceCapture)` - it levels off too quickly, meaning farther spaces would take basically forever. It also doesn't account for the current behavior of allowing existing snow to overflow into new spaces so they fill faster (though this is probably not all that important).
Another simple approach might be something like:
```
def isSnow(dist, turns):
...   needed = 2*dist-1
...   if( needed <= turns ): return True
...   else: return False
```
This would fill in distance 1 turn, and an additional space every two subsequent turns.

Alternatively, do a Game of Life approach where tiles are filled in once they have neighbors filled in (maybe for a couple of turns), though this works out almost the same as the previous thing.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Hah. Obsolete comment.

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: I guess that's fair.

I think I probably want to poke around and test a bit to be sure, but from a code-inspection standpoint I guess I'm done with this review.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      // Increase Ave's sphere of influence.
+      Ave.MAX_SNOW_SPREAD_RANGE++;
+
+      // Buff units.
+      damageMod.apply(Ave);
```
ThislsAUsername: This line produces an ATK leak.
It might be thematic for Ave to get a firepower bonus in snow (that she can permanently increase), but giving her permanent, stacking global firepower seems a little much.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -0,0 +1,646 @@
...
+
+    CommanderAve Ave;
+    COModifier damageMod = null;
+
+    GlacioAbility(CommanderAve commander)
+    {
+      super(commander, GLACIO_NAME, GLACIO_COST);
+      Ave = commander;
+      damageMod = new CODamageModifier(GLACIO_BUFF);
+      AIFlags = PHASE_TURN_START | PHASE_TURN_END;
```
ThislsAUsername: I think I'd recommend Glacio be turn end only, while Oblido can stay both.
Glacio doesn't produce all that much benefit on your turn, but it's almost Lightning Strike if you use it at the end of your turn.
Oblido can help you get kills during your turn, though, and Oblido having an extra use phase means the AI might choose one or the other, with them at equal costs.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Ahh, good catch. Maybe we should make `apply` private, and make some way to signal if the modifier should be permanent or transient.

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Using Oblido at turn end means you don't get to use your extra firepower though. It does hurt nearby enemies, which might slow their roll next turn, but I'm not sure if that's compelling enough not to just make it turn-begin only.
also Oblido and Glacio are currently not the same cost. I made Glacio cheaper since it doesn't do damage. Think they should be at parity?

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: > Ahh, good catch. Maybe we should make `apply` private, and make some way to signal if the modifier should be permanent or transient.

I don't know that we could make `apply` private without making it impossible for the CO to do it, haha.

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: > Using Oblido at turn end means you don't get to use your extra firepower though.

20 attack is nice, but not exactly game-ending. It's literally Strong's passive, haha.

> It does hurt nearby enemies, which might slow their roll next turn, but I'm not sure if that's compelling enough not to just make it turn-begin only.

2 HP damage, reaching into the enemy formation, is pretty good, especially if you just commit an inf to stare down some tanks (like Muriel seems to enjoy).
Also, powers are supposed to be "ice breakers", so you likely won't have directs committed to get value out of it by the time you use it.

> also Oblido and Glacio are currently not the same cost. I made Glacio cheaper since it doesn't do damage. Think they should be at parity?

As much as I hate mass damage... look at it this way: Olaf and Eagle are considered pretty balanced with each other, and both are mostly bland D2D.
Olaf produces crippling weather and 2 mass damage for 7 stars.
Eagle gets a second turn for non-infantry for *9*.
Neither of Ave's abilities really match up to their global effects (and it's fair to note that our faster charging and Ave's D2D should factor into the costs as well), but we can draw some parallels here.
Oblido is kind of like Olaf's SCOP, but without any of that nasty snow.
Glacio is kind of like Lightning Strike, but it *also* has snow. Not only that, but it works just as well for infantry as it does for anything else - you might very well get to simulate Victory March by freezing your enemy's own units in position to protect your capping infantry.

Especially if we go through with my sinister plans to make mass damage charge COP meters, I think Glacio is perhaps deserving of the greater cost.

ThatIsAUsername: > 20 attack is nice, but not exactly game-ending. It's literally Strong's passive, haha.

So you are saying I should bump it to 30? Also, I think I want to nerf Strong's passive at some point, in exchange for some more unique features.

> 2 HP damage, reaching into the enemy formation, is pretty good, especially if you just commit an inf to stare down some tanks 

Good point. Worth leaving as both, if just to see how it plays.

> blah blah ... I think Glacio is perhaps deserving of the greater cost.

I would hesitate to draw such a direct parallel between Glacio and Lightning Strike - Glacio doesn't affect as much. Also there isn't any precedent for deforestation, which skews any estimates on Oblido as well. Given the reduced scope compared to Olaf, I could see reducing the cost though. Maybe just move them both to 7? Or do you strongly feel that Glacio is worth more/Oblido less?

ThislsAUsername: > > 20 attack is nice, but not exactly game-ending. It's literally Strong's passive, haha.
> 
> So you are saying I should bump it to 30? Also, I think I want to nerf Strong's passive at some point, in exchange for some more unique features.

No especially big opinion there.
Nerfing Strong sounds good. Assuming we change the dynamics around infantry enough, we might even be able to leave the extra infantry production in.

> > blah blah ... I think Glacio is perhaps deserving of the greater cost.
> 
> I would hesitate to draw such a direct parallel between Glacio and Lightning Strike - Glacio doesn't affect as much. Also there isn't any precedent for deforestation, which skews any estimates on Oblido as well. Given the reduced scope compared to Olaf, I could see reducing the cost though. Maybe just move them both to 7? Or do you strongly feel that Glacio is worth more/Oblido less?

Glacio affects as much as Oblido does, so it's as comparable to LS as Oblido is to Olaf SCOP.
The deforestation is something I hadn't really taken into account, though.

I feel like keeping them both at 8 sounds about right, TBH. The caveat I'd add here is that Oblido shouldn't scale cost as quickly as Glacio - possibly not at all. Deforestation is very powerful, but you can only remove a forest once, haha.

ThatIsAUsername: Or maybe just swap them so that they won't be the same cost? I feel like any cost we assign is probably up for debate, so I don't really have strong feelings on this. Somewhere in this ballpark, anyway.

ThislsAUsername: > > Oblido shouldn't scale cost as quickly as Glacio - possibly not at all. Deforestation is very powerful, but you can only remove a forest once, haha.
> 
> Maybe just make Oblido cheaper then? I don't think it should _not_ scale, but starting at 6 instead of 8 might be reasonable then, I dunno.

My reasoning here is that Oblido *should* have a high entrance fee: it makes the map much more favorable for Ave, at the direct cost of her opponents.
At the same time, it's not quite as useful after the first cast, which is why I suggest slower scaling.

I also think both powers should be pretty expensive, as her passive is extremely strong. Wally, or even Spender, has a pretty good shot against Strong/Muriel (which is certainly somewhat due to Muriel's lack of indirects, but hey).

ThatIsAUsername:
> My reasoning here is that Oblido should have a high entrance fee: it makes the map much more favorable for Ave, at the direct cost of her opponents.

Conversely, giving it a lower entrance fee would cause it to be used sooner, which would mean it would cover less ground (since presumably she wouldn't be as spread out by then).

Ahh, never mind, that makes sense. I'll start them both at 8 and lower Oblido's power scaling.

ThislsAUsername:
I'm good with merging, if we don't have conflicts.

ThatIsAUsername:
I'm a little hesitant about messing with power scaling - I don't think I want that to be super common, and Ave has _two_ abilities with custom scaling now. I guess we can see how it shakes out though.

PR #168, opened 2019-02-18
Weather queuing

ThislsAUsername:
This is just the basic queuing logic I added on the AWBW branch.

I still like the idea of adding a weather "strength" value, and I'm not sure my implementation of how the event should work is as good as it could be.

There's a *very* noticeable delay on Ave's turns after the first bit, since Ave has to push events for every snowed tile, to keep them updated.

Perhaps the proper solution is for Ave to `setForecast()` herself, and only push events for the changing tiles... but then she'd have to redraw the map anyway.
Maybe the only weather change event should be the one pushed by GameInstance.

IDK, thoughts?

### DIFF
--- a/src/Terrain/Environment.java
+++ b/src/Terrain/Environment.java
@@ -13,7 +13,7 @@ public class Environment
   
   public enum Weathers
   {
-    CLEAR, RAIN, SNOW, SANDSTORM
+    CLEAR, SNOW, RAIN, SANDSTORM
   };
 
   public final TerrainType terrainType;
--- a/src/Terrain/MapMaster.java
+++ b/src/Terrain/MapMaster.java
@@ -121,7 +121,9 @@ public class MapMaster extends GameMap
   /** Returns the Location at the specified location, or null if that Location does not exist. */
   public Location getLocation(XYCoord location)
   {
-    return getLocation(location.xCoord, location.yCoord);
+    if (null != location)
+      return getLocation(location.xCoord, location.yCoord);
+    return null;
   }
 
   /** Returns the Location at the specified location, or null if that Location does not exist. */
--- a/src/Terrain/MapWindow.java
+++ b/src/Terrain/MapWindow.java
@@ -71,7 +71,7 @@ public class MapWindow extends GameMap
   /** Returns the Location at the specified location, or null if that Location does not exist. */
   public Location getLocation(XYCoord location)
   {
-    return getLocation(location.xCoord, location.yCoord);
+    return master.getLocation(location);
   }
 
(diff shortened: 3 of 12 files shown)


src/Terrain/MapMaster.java
```diff
@@ -121,7 +121,9 @@ public Environment getEnvironment(int w, int h)
   /** Returns the Location at the specified location, or null if that Location does not exist. */
   public Location getLocation(XYCoord location)
   {
-    return getLocation(location.xCoord, location.yCoord);
+    if (null != location)
+      return getLocation(location.xCoord, location.yCoord);
+    return null;
```
ThatIsAUsername: I might just make this a NPE.

src/Terrain/MapWindow.java
```diff
@@ -71,7 +71,7 @@ public Environment getEnvironment(int w, int h)
   /** Returns the Location at the specified location, or null if that Location does not exist. */
   public Location getLocation(XYCoord location)
   {
-    return getLocation(location.xCoord, location.yCoord);
+    return master.getLocation(location);
```
ThatIsAUsername: Why change this to call master directly? Now the behavior is different depending on whether I call `getLocation(XYCoord)` vs `getLocation(int, int)`. I think this enables anybody to see truth data about what's behind fog, too.

src/Terrain/MapMaster.java
ThislsAUsername: It doesn't violate the constraints of the function, since it can already return null for a location outside the map.
"Coordinate null" sounds like it's outside the map to me :V

src/Terrain/MapWindow.java
ThislsAUsername: You're completely right, I'm blind.
I wanted to be lazy/DRY, but I missed the oncoming headlights.

src/Terrain/Environment.java
```diff
@@ -13,7 +13,7 @@
   
   public enum Weathers
   {
-    CLEAR, RAIN, SNOW, SANDSTORM
+    CLEAR, SNOW, RAIN, SANDSTORM
```
ThatIsAUsername: OK, I gotta ask; why?

ThislsAUsername: I was originally going to implement a priority system: clear is the "weakest", snow is superseded by other weather types (since it's almost strictly stronger than rain), and sandstorm is defined by the removal of moisture, which kinda deflates the other weather types.
On more thought, it didn't make a lot of sense, but I'd already made the commit.
And it doesn't affect anything, so I didn't revert.

ThislsAUsername:
I don't have any problem merging as-is.

PR #169, opened 2019-02-22
Weather q mod

ThatIsAUsername:


src/CommandingOfficers/CommanderAve.java
```diff
@@ -432,10 +442,8 @@ protected void perform(MapMaster gameMap)
           if((tileEnvi.weatherType != Weathers.SNOW) || (tileEnvi.terrainType == TerrainType.FOREST))
           {
             // Destroy any forests. Big hail, man.
-            TerrainType newTerrain = (loc.getEnvironment().terrainType == TerrainType.FOREST) ? TerrainType.GRASS : loc.getEnvironment().terrainType;
```
ThislsAUsername: You definitely still need this conditional
image

ThislsAUsername: It might also be good to redraw the terrain around changed squares, so that transitions can be updated.
image

src/Engine/GameEvents/MapChangeEvent.java
```diff
@@ -45,30 +38,56 @@ public GameAnimation getEventAnimation(MapView mapView)
...
   @Override
   public void performEvent(MapMaster gameMap)
   {
-    Location loc = gameMap.getLocation(where);
-    if( null != loc )
+    for( EnvironmentAssignment ea : changes )
     {
-      loc.setEnvironment(myEnvironment);
-      if (turns > 0)
-        loc.setForecast(myEnvironment.weatherType, (gameMap.commanders.length * turns) - 1);
```
ThislsAUsername: This math is needed because the basic unit of duration is a turn rather than a full round.
1 is subtracted in order to ensure the weather reverts on the same CO's turn as it started.

src/UI/Art/SpriteArtist/SpriteMapArtist.java
```diff
@@ -237,11 +239,18 @@ private void redrawBaseTile(XYCoord coord)
...
     @Override
-    public void receiveMapChangeEvent(MapChangeEvent event)
+    public void receiveTerrainChangeEvent(ArrayList<EnvironmentAssignment> terrainChanges)
     {
-      XYCoord xyc = event.getStartPoint();
-      if( null != xyc ) myArtist.redrawBaseTile(xyc); // Redraw one tile if that's all that changed.
-      else myArtist.buildMapImage(); // Redraw the whole map if we have to.
+      for( EnvironmentAssignment ea : terrainChanges )
+      {
+        if( null != ea.where ) myArtist.redrawBaseTile(ea.where); // Redraw one tile if that's all that changed.
```
ThislsAUsername: Possibly also redraw adjacent tiles, and update comment

ThislsAUsername:
I'm not sure exactly where the change is that causes it, but Glacio's unit snow no longer reverts all at once.
It reverts top to bottom, which is weird since I'd expect it to revert as soon as she ends her turn (duration of 0)

src/CommandingOfficers/CommanderAve.java
ThatIsAUsername: Hah! That's what I get for pushing stuff when I was supposed to be asleep. Fantastic.

And good thought on transitions. For weather changes, it isn't necessary, but for other things, yeah.

ThislsAUsername:
Looks like you've addressed the technical issues.
Do you really intend Glacio's generated snow to disappear for her opponents' turns, rather than lasting a full round?
Likewise, it seems reasonable to apply a default duration of 1 round to the global weather event constructor.

ThatIsAUsername:
Ah, yeah. Fixed Glacio and global weather.  The releveling logic wasn't correct after the transition to queuing, so I fixed that too.

Got anything else?

src/CommandingOfficers/CommanderAve.java
```diff
@@ -344,7 +330,23 @@ private void relevelSnow(GameMap gameMap, GameEventQueue outEvents)
...
+      if( snowLoggingEnabled ) log("Snow at " + dis + " melting from " + oldVal + " to " + snowMap[dis.xCoord][dis.yCoord]);
+    }
+
+    // Update weather forecast.
+    ArrayList<MapChangeEvent.EnvironmentAssignment> tiles = new ArrayList<MapChangeEvent.EnvironmentAssignment>();
+    for (XYCoord coord : toSnow)
+    {
+      Environment newEnvi = Environment.getTile(gameMap.getEnvironment(coord).terrainType, Weathers.SNOW);
+      int duration = snowMap[coord.xCoord][coord.yCoord] / SNOW_MELT_RATE;
+      if( coord.xCoord == 8 && coord.yCoord == 0 )
```
ThislsAUsername: *cough*

src/CommandingOfficers/CommanderAve.java
```diff
@@ -344,7 +330,23 @@ private void relevelSnow(GameMap gameMap, GameEventQueue outEvents)
...
-      log("Snow at " + dis + " melting from " + oldVal + " to " + snowMap[dis.xCoord][dis.yCoord]);
+      if( snowLoggingEnabled ) log("Snow at " + dis + " melting from " + oldVal + " to " + snowMap[dis.xCoord][dis.yCoord]);
+    }
+
+    // Update weather forecast.
+    ArrayList<MapChangeEvent.EnvironmentAssignment> tiles = new ArrayList<MapChangeEvent.EnvironmentAssignment>();
+    for (XYCoord coord : toSnow)
+    {
+      Environment newEnvi = Environment.getTile(gameMap.getEnvironment(coord).terrainType, Weathers.SNOW);
+      int duration = snowMap[coord.xCoord][coord.yCoord] / SNOW_MELT_RATE;
```
ThislsAUsername: This should probably still have the +1; it looks a little weird for Ave's snow to disappear and reappear every turn, heh.

ThatIsAUsername: Hmm. There is a bit of a flicker for tiles that aren't 200+ deep. On the other hand, adding a spurious extra 1 to duration causes them to stay under snow longer than they should once they become disjointed.

ThislsAUsername: Hmm.
I don't think it should make a *huge* difference in gameplay either way; an extra turn of snow after Ave loses a property probably isn't a huge deal.
I guess it makes a lot less difference with only a frame or two of flicker. Merge if you like, I guess.

ThatIsAUsername: Eventually I'd like to support zero-animation game advancement. I'll leave the flicker for now because a) it's technically more correct, and b) so it can be auto-fixed once we get zero-animation working.

PR #173, opened 2019-03-11
InGameMenu input now uses a single source of truth

ThislsAUsername:
Without this fix, MapController would pipe LEFT/RIGHT inputs into the OptionSelector, but then the menu itself would throw them out.

### DIFF
--- a/src/UI/InGameMenu.java
+++ b/src/UI/InGameMenu.java
@@ -39,6 +39,8 @@ public class InGameMenu<T>
 
   public InGameMenu(Collection<T> options, OptionSelector selector)
   {
+    if (options.size() != selector.size())
+      throw new IllegalArgumentException("Number of options doesn't match the size of the OptionSelector");
     menuOptions = new ArrayList<T>();
     menuOptions.addAll(options);
     optionSelector = selector;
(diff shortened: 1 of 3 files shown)


src/UI/InGameMenu.java
```diff
@@ -23,20 +24,24 @@
...
+  public InGameMenu(Collection<T> options, OptionSelector selector)
   {
     menuOptions = new ArrayList<T>();
-    for( int i = 0; i < options.length; ++i )
-    {
-      menuOptions.add( options[i] );
-    }
-    optionSelector = new OptionSelector(menuOptions.size());
+    menuOptions.addAll(options);
+    optionSelector = selector;
```
ThatIsAUsername: I feel like we should at least verify `options.size() == selector.numOptions`. That would require adding an accessor to `OptionSelector`, but it's at least a sensible one.

ThislsAUsername: And throw an exception if the check doesn't pass?

ThatIsAUsername: mmm, yeah. I think that's probably best; probably with an unchecked exception type. IllegalArgumentException strikes me as a likely candidate, but do whatever you think makes sense.

PR #174, opened 2019-03-15
Cereal

ThislsAUsername:
This PR adds saving via serialization; it's definitely not feature-complete, though playtesting is proceeding apace with no known issues at this time.
There are a whole bunch of warnings about serialization IDs not being explicitly specified. I have two reasons for not specifying: 1) that's a lot of annoying busy-work 2) that bases the ID on the various features of the class, meaning old saves will automatically become incompatible when things change, which is good since then we don't have to handle weird errors.

The biggest missing bit is a menu to actually pick which save you want to load, and retaining the save file name for when you save again seems like a helpful addition.
It also would probably be good to make the "continue" graphic not horrible to behold.

I did merge in the other existing PRs because I wanted to make sure they all worked together before merging it into the playtest branch.

I went with serialization over a custom save format since this is liable to be less error prone and automagically includes most anything a new CO or such would want to persist. If we went with a custom format, e.g. Venge would have to somehow store references to individual enemy units.
On top of that, we'd have to figure out how to store COModifiers and UnitModels and it'd just be a crazy hassle.
The only(?) things that need special handling in this system are TerrainTypes and strings that you intend to compare by reference rather than content.

I pointed it at gameSetup since that's probably the single largest batch of changes merged into this branch.


ThislsAUsername:
OBE, commits moved to `moldyCereal`

PR #175, opened 2019-03-23
Co info

ThislsAUsername:
Known issues:
Needs more comments
There's no punctuation. That's a biggie.
There's no runtime flexibility on the descriptions, so costs and names of abilities cannot be updated as the situation calls for.

(Yes, I'm aware I made some extra changes in here, but they were cromulent in that *not* changing them now would result in more overall work.)


src/CommandingOfficers/CommanderBear_Bull.java
```diff
@@ -17,9 +17,30 @@
 {
   private static final long serialVersionUID = -3048055535977630509L;
   
-  private static final CommanderInfo coInfo = new CommanderInfo("Bear&Bull", new instantiator());  
-  private static class instantiator implements COMaker
+  private static final CommanderInfo coInfo = new CommanderInfo("Baron Bull", new instantiator());
```
ThatIsAUsername: Aww, no more dual-portrait potential.

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -17,8 +18,31 @@
...
   
   private static final CommanderInfo coInfo = new CommanderInfo("Strong", new instantiator());  
-  private static class instantiator implements COMaker
+  private static class instantiator extends COMaker
   {
+    public instantiator()
+    {
+      infoPages.add(new InfoPage(
+          "Commander Strong is real strong.\n" +
+          "That's it.\n"));
```
ThatIsAUsername: Top notch.

ThatIsAUsername: Though that's a lot of why I want to change him up a bit.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -215,23 +204,12 @@ private void loadNextEventAnimation()
   public void render(Graphics g)
   {
     GameMap gameMap = getDrawableMap(myGame);
+    
+    DiagonalBlindsBG.draw(g);
```
ThatIsAUsername: We don't want to waste time drawing the background behind the map.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -320,16 +298,77 @@ else if( getCurrentGameMenu() == null )
       {
         menuArtist.drawMenu(mapGraphics, mapViewX, mapViewY);
       }
+      
+      // When we draw the map, we want to center it if it's smaller than the view dimensions
```
ThatIsAUsername: Ah, and this is why you are drawing the background behind the map. At least on Linux, this results in white areas below and to the right of the map, covering the background.
I guess you stopped the size-snapping behavior in case the user expands the info menu? Hrm. I think I might prefer to just ensure the window is usable at the smallest size, and scales nicely if it gets expanded.
Failing that... maybe just ditch the animated background behind the map anyway, and use a nice neutral tone so there isn't non-game movement going on?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -108,6 +113,51 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
...
+  /**
+   * Returns an image with the input string printed within the specified width, in normal text.
+   * @param reqWidth: Actual UI size in pixels that you want to fit the text into.
+   */
+  public static BufferedImage paintTextNormalized(String prose, int reqWidth)
+  {
+    // Find the dimensions of the menu we are drawing.
+    int drawScale = SpriteOptions.getDrawScale();
+    int menuTextWidth = SpriteLibrary.getLettersUppercase().getFrame(0).getWidth() * drawScale;
+    int menuTextHeight = SpriteLibrary.getLettersUppercase().getFrame(0).getHeight() * drawScale;
```
ThatIsAUsername: It might be more clear to use `charWidth`/`charHeight`. The current name sounds like it might be for the whole menu text.

ThatIsAUsername:
Good stuff. This has been on the back burner for basically forever, so it'll be nice to make this a real thing.

src/CommandingOfficers/CommanderBear_Bull.java
ThislsAUsername: I mean, I'm quite willing to swap it back.
....I think I might have switched it due to being freaked out about having the name as the save file name?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: Hah, uncooked pasta. Good catch, though.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: I don't have super strong opinions on this,
You'll note my TODO about making the thing scroll freely, so it's viable at any size.

I personally liked the size-snapping, but it wouldn't work for the current iteration of the info menu and it bothered some other folks.
That being said, I think letting the UI hang out outside of the map when applicable is good too, to let you see the whole thing at once.

ThislsAUsername:
I feel like it's also weird that the actual info itself is in the `COMaker`, and come to think of it, I don't think there's any solid reason to *have* `COMaker` since you can just subclass `COInfo`.

Also, the symbol set we have now doesn't work on my end. The interrobang is interpreted as 3 characters, and parenthesis get rolled over to the start of the charset.
Any opposition to swapping the interrobang to an apostrophe?

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: Well, I'm certainly willing to entertain feedback from beta-testers, though I'm curious that anyone would want a window size bigger than the map. Was there any more specific feedback? If it's just that the info pages don't look right at the standard size, that's fixable.

ThatIsAUsername:
>I feel like it's also weird that the actual info itself is in the COMaker, and come to think of it, I don't think there's any solid reason to have COMaker since you can just subclass COInfo.

Yeah, sounds reasonable.

I'm fine with removing interrobang. It worked on Linux, I think, but Windows doesn't like it.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Well, for one thing, the map selection screen quickly gets to have too many to fit on screen at once. (though the real solution there is probably to add proper scrolling) (also I should get around to making it support subfolders)
They've been adding competitive maps for playtests, and there are a lot of those.

Showed it to a guy and he was visibly put out by the game not being able to take up his whole screen.
I think the next step might be to change the drawscale automagically when you resize the thing big enough, but that has its own possible issues.

ThatIsAUsername: > he was visibly put out by the game not being able to take up his whole screen.

Has he... played Advance Wars? I guess it does take up the whole screen on a DS.

> change the drawscale automagically

I'm not really a fan of this idea... DrawScale is already its own option, and I don't want to automatically override that for the user.

ThislsAUsername: > Has he... played Advance Wars? I guess it does take up the whole screen on a DS.

Nope.

> I'm not really a fan of this idea... DrawScale is already its own option, and I don't want to automatically override that for the user.

Yeah, that's fair. It'd be hard to know when to *down*scale it too, and one-way automatic transitions are horrible UX.

src/CommandingOfficers/CommanderBear_Bull.java
ThatIsAUsername: Aww, no more clever pun.

ThislsAUsername: I really can't win here, can I?

ThislsAUsername: I did put it in a separate commit so it can be reverted easily, though

ThatIsAUsername: Nope, no winning on my watch.

I really don't have a stake in the matter; he's your Commander, do whatchu want.

src/Engine/Utils.java
```diff
@@ -501,6 +501,63 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
     TravelDistanceComparator tdc = new TravelDistanceComparator(unit, map);
     Collections.sort(mapLocations, tdc);
   }
+  
+
+  /** Returns a string describing the game statistics of a given CO */
+  public static String getGameStatusData(GameMap map, Commander viewed)
```
ThatIsAUsername: This feels like it should probably belong in `GameInstance`. Unless there's a reason it should go here instead?

src/Engine/Utils.java
```diff
@@ -501,6 +501,63 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
...
+          if (resident.model.chassis != ChassisEnum.TROOP)
+            vehCount++;
+          unitFunds += resident.model.getCost() * resident.getHP() / resident.model.maxHP;
+        }
+      }
+    }
+
+    sb.append("Income:         ").append(income).append("\n");
+    sb.append("Unit count:     ").append(unitCount).append("\n");
+    sb.append("Vehicle count:  ").append(vehCount).append("\n");
```
ThatIsAUsername: Are we gonna split out air and sea units next too?

src/Engine/Utils.java
```diff
@@ -501,6 +501,63 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
...
+    sb.append("Unit funds:     ").append(unitFunds).append("\n");
+
+    // ability stats
+    double[] abilityCosts = viewed.getAbilityCosts();
+    if (abilityCosts.length > 0)
+    {
+      double abilityPower = viewed.getAbilityPower();
+      double untilNextPower = abilityCosts[abilityCosts.length - 1]; // init to our biggest cost, so we can only go down
```
ThatIsAUsername: Are CO abilities guaranteed to be stored in order of cost?

src/Engine/Utils.java
ThislsAUsername: Uh, not particularly? AFAIK, it's not much different from figuring out where a unit can move.
Why do you think it should go in GameInstance?

src/Engine/Utils.java
ThislsAUsername: Nah. People didn't ask for that. These categories were specifically requested.

src/Engine/Utils.java
ThislsAUsername: I was about to say "no, but they are for all current Commanders"... but that state of affairs can change in `master`.
Will fix.

ThislsAUsername:
Should this get re-pointed to `moldyCereal` or `master` now, or at some point later?

src/Engine/Utils.java
ThatIsAUsername: It's asking for a bunch of game-specific data. I guess it's the difference between asking GameInstance to tell about itself vs asking Utils to tell us about GameInstance. Utils stuff is generally functional; figuring out what can be done or how to do it, and this feels slightly out of place since it's just providing statistics. Maybe UIUtils? Though that's more about factions and colors right now. I dunno, I'm not sure how much it matters where it goes, really.

ThatIsAUsername:
> Should this get re-pointed to `moldyCereal` or `master` now, or at some point later?

I would say "why wait", but re-pointing this at `moldyCereal` looks like it might be an involved operation.

ThislsAUsername:
OBE due to moldyInfo

PR #178, opened 2019-04-03
Co info mod

ThislsAUsername:
I decided to implement rather than wait. I put it on a separate branch though, so yay me?

### DIFF
--- a/src/UI/Art/SpriteArtist/SpriteInfoView.java
+++ b/src/UI/Art/SpriteArtist/SpriteInfoView.java
@@ -1,0 +1,137 @@
+package UI.Art.SpriteArtist;
+
+import java.awt.Color;
+import java.awt.Dimension;
+import java.awt.Graphics;
+import java.awt.image.BufferedImage;
+import java.util.HashMap;
+import java.util.Queue;
+
+import CommandingOfficers.COMaker.InfoPage;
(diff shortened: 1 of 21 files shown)


src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,141 @@
...
+        int overlayHeight = 30*drawScale;
+        int heightOffset = 0;
+        for (Commander CO : myControl.getGame().commanders)
+        {
+          BufferedImage overlayPic = SpriteLibrary.createTransparentSprite(100*drawScale, overlayHeight);
+          CommanderOverlayArtist.drawCommanderOverlay(overlayPic.getGraphics(), CO, true);
+          g.drawImage(overlayPic,2*paneOuterBuffer, 2*paneOuterBuffer + heightOffset, null);
+          heightOffset += overlayHeight;
+        }
+        break;
```
ThislsAUsername: I'd like to add income and unit count here, but I don't want to duplicate code and I'm not sure how best to split up the parsing that happens down below. Any ideas?

It's probably also worth caching the results of all these calculations so they aren't done every frame, but I'm not super worried about that at this point.

ThislsAUsername:
OBE due to moldyInfo

PR #180, opened 2019-04-15
Ai fixes

ThatIsAUsername:


src/Engine/GameAction.java
```diff
@@ -94,22 +94,24 @@ public GameEventQueue getEvents(MapMaster gameMap)
       isValid &= (movePath != null) && (movePath.getPathLength() > 0);
       if( isValid )
       {
-        moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
+        XYCoord moveCoord = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
```
ThislsAUsername: New variable shadows private member?

ThislsAUsername:
I think ResupplyAction probably needs the fix as well.
While you're at it, maybe pass in a minimum range of 1 so that the APC doesn't resupply itself, hah.

ThatIsAUsername:
> I think ResupplyAction probably needs the fix as well.
> While you're at it, maybe pass in a minimum range of 1 so that the APC doesn't resupply itself, hah.

Doesn't look like ResupplyAction has the variable-shadowing issue.
Default min range is 1 - the APC already doesn't resupply itself.

ThislsAUsername:
> > I think ResupplyAction probably needs the fix as well.
> > While you're at it, maybe pass in a minimum range of 1 so that the APC doesn't resupply itself, hah.
> 
> Doesn't look like ResupplyAction has the variable-shadowing issue.
> Default min range is 1 - the APC already doesn't resupply itself.

Nah, no shadowing. But you might be able to move the APC onto a Lander if there's something there to supply.
Huh. maybe I'm just seeing things. Or on a branch where that got broken somehow. Ah, well.

ThatIsAUsername:
> But you might be able to move the APC onto a Lander if there's something there to supply.

Ah, right. Fair enough.

ThislsAUsername:
> Huh. maybe I'm just seeing things. Or on a branch where that got broken somehow. Ah, well.

Hah, I *wasn't* just seeing things. Managed to repro it.
IFF an APC can be resupplied and it moves 1 space to resupply on its turn, it will resupply itself.

ThatIsAUsername:
>Hah, I wasn't just seeing things. Managed to repro it.
IFF an APC can be resupplied and it moves 1 space to resupply on its turn, it will resupply itself.

Nice find, that's subtle. Fixed.

PR #181, opened 2019-04-29
Range booster fix

ThislsAUsername:
Because this never should have been a unit remodel

### DIFF
--- a/src/CommandingOfficers/Modifiers/IndirectRangeBoostModifier.java
+++ b/src/CommandingOfficers/Modifiers/IndirectRangeBoostModifier.java
@@ -1,43 +1,63 @@
 package CommandingOfficers.Modifiers;
 
+import java.util.ArrayList;
 import java.util.Map;
 
 import CommandingOfficers.Commander;
 import Units.UnitModel;
 import Units.Weapons.WeaponModel;
 
-public class IndirectRangeBoostModifier extends UnitRemodelModifier implements COModifier
+public class IndirectRangeBoostModifier implements COModifier
 {
+  ArrayList<WeaponModel> modelsToModify;
   private int boost;
 
   public IndirectRangeBoostModifier(int boost)
   {
-    super();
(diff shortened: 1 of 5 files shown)


src/CommandingOfficers/Modifiers/IndirectRangeBoostModifier.java
```diff
@@ -1,43 +1,63 @@
...
   {
-    super();
     this.boost = boost;
+    modelsToModify = new ArrayList<WeaponModel>();
   }
 
-  /** Add a new unit transform assignment. */
-  public Map<UnitModel, UnitModel> init(Commander commander)
+  /** Collects all the ranged weapon models */
+  public void init(Commander commander)
```
ThatIsAUsername: I'm curious why this is a separate function instead of doing it in the constructor?

ThislsAUsername: Because historical reasons.
Probably makes sense to be rid of it, yes.

ThatIsAUsername:
Looks groovy to me (though as an aside, I would have made those last two commits as a single commit, since the COs were still calling `init`, which no longer exists - granular commits are good, but you should try to always have the code in a working state after each individual commit. If you ever have to use git bisect, you will thank yourself).

ThislsAUsername:
> (though as an aside, I would have made those last two commits as a single commit, since the COs were still calling `init`, which no longer exists - granular commits are good, but you should try to always have the code in a working state after each individual commit. If you ever have to use git bisect, you will thank yourself).

That's true and good and all, but this should allow cherry-picking the commit(s) across to `master` without cross-pollinating the COs.

ThatIsAUsername:
Ah, fair enough. I might just address that via `git cherry-pick --no-commit`, then unstage the stuff I don't want, and then commit. Plenty of ways to skin the cat.

PR #182, opened 2019-04-29
Added real com towers

dtch1997:
- Added a com tower terrain type
- Removed universal 1 tower
- COs now get a damage buff upon capturing a tower, damage buff is reset when tower is captured by enemy

TODO: 
- make an actual tower icon


src/CommandingOfficers/Commander.java
```diff
@@ -443,6 +443,32 @@ private void writeObject(ObjectOutputStream stream) throws IOException {
...
+    {
+      if( unit.CO == myCommander && location.getOwner() == myCommander)
+      {
+        // Captured a new com tower. Apply damage boost. 
+        new CODamageModifier(10).apply(this);
+      }
+      else if (unit.CO != myCommander && previousOwner == myCommander) 
+      {
+        // Friendly com tower has been captured. Revert damage boost. 
+        new CODamageModifier(10).revert(this);
```
ThatIsAUsername: I know it shouldn't be possible to lose a tower if you haven't captured one first, but it will be safer to count how many towers this CO actually owns instead of trusting implicit math. For example, if each Commander starts with a Tower, that won't register here and you'll still get nerfed if you lose one.

Another approach might be: Give Commander an `adjustTowerPower` function that counts towers and creates/applies a `CODamageModifier` if needed (it would have to track the current power level explicitly, instead of implicitly as you do here). Then just call `adjustTowerPower` during `initTurn` (It would actually be sufficient to call it once at game start instead of every turn, but there's not an obvious place to do that) and also whenever `receiveCaptureEvent` is called.
This would also make it so we don't need to track the previous owner of a captured property (which shouldn't really be necessary anyway, since the new/previous owner should always be different).

dtch1997:
I refactored it according to what you described. I tested it and the game seems to run correctly (yay) but i couldn't test the actual towers because I didn't have the sprites, so everything showed up as black tiles. I *think* this should work though.

src/Terrain/Maps/IMapBuilder.java
```diff
@@ -25,5 +26,5 @@
   
   protected static final TerrainType LB = TerrainType.HEADQUARTERS; // lab -> HQ
   protected static final TerrainType SR = TerrainType.BUNKER; // full silo, but empty because lol
-  protected static final TerrainType TW = TerrainType.BUNKER; // tower
+   // tower
```
ThislsAUsername: Rather than deleting the line, you should set the TW abbreviation to TerrainType.TOWER

.gitignore
```diff
@@ -1,2 +1,4 @@
 *.class
 Thumbs.db
+/.metadata/
+/bin/
```
ThislsAUsername: @ThatIsAUsername do we even have a policy on changes to .gitignore?
I just hid it so I don't have to mess with it. thoughts?

ThatIsAUsername: It's pretty reasonable to ignore /bin/. I'm semi surprised we didn't have that on the list already.

ThatIsAUsername:
Overall it looks pretty solid. A few comments though:

The last commit says it added a comm tower sprite... but I still don't see one.

Also, you are not using the DefPcCodeStyle settings, which is going to end up causing some real formatting headaches in the long run.

One more change I'd suggest: Add `TerrainType.TOWER` to the `propertyTypes` list in `SpriteMapSelectMenuArtist` so we can see how many Towers there are in a map at the select screen.

ThislsAUsername:
> The last commit says it added a comm tower sprite... but I still don't see one.

I'm just as happy without it, given what it is. Then again, maybe we aren't worried about that?

> Also, you are not using eclipse autoformat settings (in DefPcCodeStyle.xml), which is going to end up causing some real formatting headaches in the long run.

I'm not super worried about the long run, given that this is the test branch.

> One more change I'd suggest: Add `TerrainType.TOWER` to the `propertyTypes` list in `SpriteMapSelectMenuArtist` so we can see how many Towers there are in a map at the select screen.

We should probably get around to displaying more than 99 properties, too, eventually...

PR #183, opened 2019-04-30
Added JOIN command.

ThatIsAUsername:


src/Engine/GameAction.java
```diff
@@ -610,6 +611,93 @@ public String toString()
...
+      isValid &= (null != movePath) && (movePath.getPathLength() > 0);
+      isValid &= (null != gameMap);
+      if( isValid )
+      {
+        pathEnd = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
+        isValid &= gameMap.isLocationValid(pathEnd);
+
+        if( isValid )
+        {
+          // Find the transport unit.
```
ThislsAUsername: "transport"

ThislsAUsername:
It does print a warning about being unable to move when you join, but it's probably not a big deal.
...especially as I'm planning to muck around with everything soon.

ThatIsAUsername:
> It does print a warning about being unable to move when you join, but it's probably not a big deal.
> ...especially as I'm planning to muck around with everything soon.

Yeah, it does the same thing for load actions. Which I inadvertently broke, but it's fixed now.

ThislsAUsername:
> but it's fixed now.
Still printing the warning on my end, and I made sure to pull.

ThatIsAUsername:
Sorry, wasn't clear. Load commands have always printed that warning. No intention to fix that at this time.

I did however, accidentally make it impossible to put a unit on a transport in the initial PR. That's what I fixed.

PR #184, opened 2019-05-03
Added proper HQ/lab semantics. 

dtch1997:
In the presence of HQs, labs do not affect game victory in any way. 
In the absence of HQs, the last lab being captured eliminates player from the game.


src/Engine/GameAction.java
```diff
@@ -296,14 +296,45 @@ public GameEventQueue getEvents(MapMaster map)
...
+              Commander CO = captureLocation.getOwner();
+              for (XYCoord xy : CO.ownedProperties) {
+                Location loc = map.getLocation(xy);
+                if (loc.getEnvironment().terrainType == TerrainType.LAB ) 
+                {
+                  numLabs += 1;
+                }
+                if (loc.getEnvironment().terrainType == TerrainType.HEADQUARTERS ) 
+                {
+                  numHQs += 1;
```
ThislsAUsername: You can just `break` here, actually.

src/Engine/GameAction.java
```diff
@@ -296,14 +296,45 @@ public GameEventQueue getEvents(MapMaster map)
...
+            else if ((propertyType == TerrainType.LAB) && (null != captureLocation.getOwner())) 
+            {
+              int numLabs = 0;
+              int numHQs = 0;
+              Commander CO = captureLocation.getOwner();
+              for (XYCoord xy : CO.ownedProperties) {
+                Location loc = map.getLocation(xy);
+                if (loc.getEnvironment().terrainType == TerrainType.LAB ) 
+                {
+                  numLabs += 1;
```
ThislsAUsername: Probably `break` here if labs > 1, since one `CaptureAction` can only affect one property

src/Engine/GameAction.java
```diff
@@ -294,8 +296,45 @@ public GameEventQueue getEvents(MapMaster map)
...
+              int numLabs = 0;
+              int numHQs = 0;
+              Commander CO = captureLocation.getOwner();
+              for (XYCoord xy : CO.ownedProperties) 
+              {
+                Location loc = map.getLocation(xy);
+                if (loc.getEnvironment().terrainType == TerrainType.LAB ) 
+                {
+                  numLabs += 1;
+                  if (numLabs > 1)
```
ThatIsAUsername: This `if` is redundant. Since `numLabs` starts at zero, and you incremented it on the previous line, this check can never fail.
It doesn't hurt anything, but it doesn't help either.

ThislsAUsername: Nah, it's not redundant. He's checking that it's *greater* than 1, so he'll count up to 2 labs and then break.

ThatIsAUsername: Oh. Right. Silly me.

PR #185, opened 2019-05-03
Strong rework

ThatIsAUsername:


src/Terrain/MapMaster.java
```diff
@@ -33,8 +33,7 @@ public MapMaster(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
     }
 
     // Print a warning if the number of Commanders we have does not match the number the map expects.
-    if( COs.length != mapInfo.COProperties.length ||
-      ( mapInfo.mapUnits.size() > 0 && COs.length != mapInfo.mapUnits.size() ) )
```
ThislsAUsername: If you're removing this, you might also wanna set initOK to false in the other if statement?

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -125,50 +139,56 @@ protected void perform(MapMaster gameMap)
...
-      UnitRemodelModifier urm = new UnitRemodelModifier();
-      for( UnitModel.UnitEnum type : highCapacityUnitModels.keySet() )
+      // Lastly, all troops are refreshed and able to move again.
+      for( Unit unit : myCommander.units )
       {
-        urm.addUnitRemodel(myCommander.getUnitModel(type), highCapacityUnitModels.get(type));
+        if( unit.model.chassis == UnitModel.ChassisEnum.TROOP )
+        {
+          unit.isTurnOver = false;
+        }
```
ThislsAUsername: I guess we should consider adding an event for this so we can animate it, but that's probably best to leave for the future.

PR #186, opened 2019-05-03
Mo sprites

ThatIsAUsername:


ThislsAUsername:
I feel like maybe the Battleship should have moar gun; having it have the same amount of gun as the Cruiser feels a little odd.

Overall looks pretty good, though.

ThatIsAUsername:
> I feel like maybe the Battleship should have moar gun; having it have the same amount of gun as the Cruiser feels a little odd.
> 
> Overall looks pretty good, though.

On the one hand, yeah, but on the other hand, you can't really make a gun smaller than that, and trying to make the BShip's gun larger makes it not really fit.
Think of it as the two ships being drawn at different scales, just like how inf/tank/mdTank map sprites are all close to the same size, despite being actually much different in size.

PR #190, opened 2019-05-06
Moldy cereal

ThislsAUsername:
This PR adds saving via serialization; it's definitely not feature-complete, though playtesting is proceeding apace with no known issues at this time.

There are a whole bunch of warnings about serialization version IDs not being explicitly specified. I have two reasons for not specifying: 1) that's a lot of annoying busy-work [addressable by Eclipse, I think] 2) having no serial version bases the ID on the various features of the class, meaning old saves will automatically become incompatible when things change, which is good since then we don't have to handle weird errors.
That said, I don't think we're gonna be overhauling stuff too much after this next release, so maybe specifying all this stuff could be worth doing so we don't invalidate *all* saves with a new release, going forward. (though with the current scheme, adding new terrains would have to be done at the end of the static array every time)

The biggest missing bit is a menu to actually pick which save you want to load.
It also would probably be good to make the "continue" graphic not horrible to behold. (sidenote, BlueLink made some images that we could probably use, but I think you'd probably rather use your own art?)

I went with serialization over a custom save format since this is liable to be less error prone and automagically includes most anything a new CO or such would want to persist. If we went with a custom format, e.g. Venge would have to somehow store references to individual enemy units.
On top of that, we'd have to figure out how to store COModifiers and UnitModels and it'd just be a crazy hassle.

The only(?) things that need special handling in this system are TerrainTypes and strings that you intend to compare by reference rather than content (which I avoid by only comparing content).

I did also just opt to not serialize the AI at all, but bring it back via index on deserialize. That may be a poor choice if we eventually go for stateful AI.


src/CommandingOfficers/Commander.java
```diff
@@ -384,4 +394,66 @@ public GameAction getNextAIAction(GameMap gameMap)
...
+    {
+      for( AIMaker AI : AILibrary.getAIList() )
+      {
+        if( AI.getName().equalsIgnoreCase(aiController.getAIInfo().getName()) )
+          stream.writeInt(AILibrary.getAIList().indexOf(AI));
+      }
+    }
+
+    // Write out our shopping list as a flattened 2D array of booleans; for each terrain, for all unit types, can I build it?
+    // This is probably inefficient use of space, but just look at all this not-caring I can do!
```
ThatIsAUsername: I'm sure you leave comments just like at your job too, don't you?

ThatIsAUsername:
> There are a whole bunch of warnings about serialization version IDs not being explicitly specified.

Is this still the case? Didn't you run into a Windows/Linux incompatibility issue due to how IDs are auto-generated?

src/CommandingOfficers/Commander.java
ThislsAUsername: Not quite, but mildly close.

It occurs to me that this function will break any time we add new terrains, so I guess the idea of updating the list of terrains without breaking saves is busted unless we find a better way.

src/CommandingOfficers/Commander.java
```diff
@@ -384,4 +394,66 @@ public GameAction getNextAIAction(GameMap gameMap)
...
     return null;
   }
+
+  /**
+   * Private method, same signature as in Serializable interface
+   *
+   * @param stream
+   * @throws IOException
+   */
+  private void writeObject(ObjectOutputStream stream) throws IOException
```
ThatIsAUsername: It appears that this function (and readObject below) are never used. After deleting them, I was still able to save/reload a game.

ThislsAUsername:
> > There are a whole bunch of warnings about serialization version IDs not being explicitly specified.
> 
> Is this still the case? Didn't you run into a Windows/Linux incompatibility issue due to how IDs are auto-generated?

Yes and no, respectively.
I *thought* I ran into an incompatibility between Windows and Linux, but it was *actually* an incompatibility between versions of DefendPeace (entirely expected and desirable), and I was being a dope trying to use the old save with a newer version of the game.

The IDs are generated at compile-time, so as long as we don't change compilers without *otherwise* keeping saves compatible, it shouldn't matter at all since the saves *should* be incompatible anyway.

src/CommandingOfficers/Commander.java
ThislsAUsername: Try saving and reloading a game with an AI in it.

I suspect it might prevent you from buying new troops, as well.

src/Engine/GameInput/GameInputState.java
```diff
@@ -21,8 +21,8 @@
   /** The current GameActionBuilder state data,
    *  shared with all State instances. */
   protected final StateData myStateData;
-  protected final OptionSet myOptions;
-  protected final OptionSelector mySelector;
+  protected OptionSet myOptions;
```
ThatIsAUsername: These kinds of changes make me nervous. Also, I re-added final and it looks like it still works.

ThislsAUsername: That was done under `a quick hack to keep things rolling`.
I was using the same inputstate for both quit and exit, and the only way I could figure out to make that work (without too much effort) was to make it no longer final so I could re-set it in the constructor of the sub-class.
I believe it's no longer needed since I ended up making them into two separate classes.

Full commit message:
```
Quick hack to keep things rolling
I feel like we need a better design for this
I really just wish we could set subclass fields up before super()
```

src/Engine/MapController.java
```diff
@@ -435,6 +445,28 @@ public void onStateChange()
       case LEAVE_MAP:
         // Handled as a special case in handleGameInput().
         break;
+      case SAVE:
+        String filename = "save/" + myGame.saveFile; // "svp" for "SaVe Peace"
```
ThatIsAUsername: This comment might be more cromulent in `GameInstance.getSaveName()`.

src/Engine/MapController.java
```diff
@@ -435,6 +445,28 @@ public void onStateChange()
...
+          out.writeObject(myGame);
+
+          out.close();
+          file.close();
+        }
+        catch (IOException ex)
+        {
+          System.out.println(ex.toString());
+        }
+        myGameInputHandler.reset(); // CO_INFO is a terminal state. Reset the input handler.
```
ThatIsAUsername: `CO_INFO`

src/Engine/GameInput/GameInputState.java
ThatIsAUsername: > I was using the same inputstate for both quit and exit

Haven't those always been separate? `ConfirmExit` either advances to `ExitToMainMenu` or straightforwardly terminates the application.

> I really just wish we could set subclass fields up before super()

Yeah, that _is_ one downside of the current approach. The thought is that `myStateData` should have all information necessary to initialize `myOptions`, but in practice there are many kinds of state info, so it can end up being annoying. I've had a thought or two on how to address that, but so far I haven't had a use case that has forced me to change things.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Hmm, true enough. Better question. Why declare `unitProductionByTerrain` as `transient` and then serialize a representation of its contents manually?

src/Terrain/TerrainType.java
```diff
@@ -161,9 +163,10 @@ private TerrainType(int attributeFlags, int defense, Color mainColor, String nam
   public static final TerrainType SHOAL = new TerrainType( SHOAL_FLAGS, SHOAL_DEFENSE, SHOAL_COLOR, SHOAL_NAME );
 
   // List of all terrain types.
-  public static final TerrainType[] TerrainTypeList = {
-    AIRPORT, BRIDGE, CITY, BUNKER, PILLAR, DUNES, FACTORY, FOREST, GRASS, HEADQUARTERS, LAB, MOUNTAIN, REEF, RIVER, ROAD, SEA, SEAPORT, SHOAL
-  };
+  public static final ArrayList<TerrainType> TerrainTypeList =
```
ThatIsAUsername: This feels a little bit like re-introducing the `TerrainType` enum we killed a while back.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Hm. Is it because the `TerrainType` objects used as keys are static, so Java ignores them? That's... inconvenient.

src/Terrain/TerrainType.java
ThatIsAUsername: Why change this to an `ArrayList`?

src/UI/UIUtils.java
```diff
@@ -157,13 +158,25 @@ else if(fileEntry.canRead() && !fileEntry.getName().endsWith(DEFAULT_FACTION_NAM
   public static ColorPalette getBuildingColors(Color colorKey)
   {
     initCosmetics();
-    return buildingColorPalettes.get(colorKey);
+    ColorPalette palette = buildingColorPalettes.get(colorKey);
+    if (null == palette) // Uh oh, the player's messing with us. Make stuff up so we don't crash.
+    {
+      buildingColorPalettes.put(colorKey, buildingColorPalettes.get(Color.PINK));
```
ThatIsAUsername: At least print a warning here, and also in `getMapUnitColors`.

src/CommandingOfficers/Commander.java
ThislsAUsername: I'm pretty sure they're not ignored, but deserialized data is a *copy*, by definition. That makes all of our identity comparisons moot.

I actually ran into the same problem with strings. My hack to prevent ability energy from accumulating during powers was to check if your active ability string was equal to the empty string. Worked marvelously. Until save/reload.
Turns out it was a copy of the empty string, so "" != "" and nobody could *ever* charge powers.

src/Terrain/TerrainType.java
ThislsAUsername: Easier to work with. I didn't know how before.

src/CommandingOfficers/Commander.java
ThislsAUsername: To answer the static question, it doesn't matter.
The map that I'm storing is *not* static, so it's irrelevant that the value I'm referencing happens to be defined statically.
I can only assume that the reference coming *out* of serialization would be shared by all COs that were deserialized, but that doesn't help because it still won't match the statically defined copies (this holds even if you don't exit the game, as deserialized objects are copies.).

src/Terrain/TerrainType.java
ThatIsAUsername: Huh. I don't think I see any new (kinds of) interactions with `TerrainTypeList` that weren't there before.
I mean, I guess it works either way, but I kinda prefer the simplicity of the standard array when the size is known a priori and it will never change.

ThislsAUsername: `IndexOf()`, for putting the index of the terrain type out into a file without having to write a helper function, haha.

src/UI/UIUtils.java
ThislsAUsername: Fixed in `moldyInfo`

src/Engine/MapController.java
ThislsAUsername: Fixed in `moldyInfo`

src/Engine/MapController.java
ThislsAUsername: Fixed in `moldyInfo`

PR #192, opened 2019-05-11
Moldy info

ThislsAUsername:
All of the comments on the original PR aside from moving `getGameStatusData()`  have been addressed in the `coInfoMod` branch; should I just cherry-pick those commits over here as well?


ThatIsAUsername:
> should I just cherry-pick those commits over here as well?

Yeah, probably. Just get it all together.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -108,6 +117,51 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
...
+  public static BufferedImage paintTextNormalized(String prose, int reqWidth)
+  {
+    // Figure out how big our text is.
+    int drawScale = SpriteOptions.getDrawScale();
+    int characterWidth = SpriteLibrary.getLettersUppercase().getFrame(0).getWidth() * drawScale;
+    int characterHeight = SpriteLibrary.getLettersUppercase().getFrame(0).getHeight() * drawScale;
+
+    ArrayList<String> lines = new ArrayList<String>();
+    // Unload our prose into the lines it already has
+    lines.addAll(Arrays.asList(prose.split("\\R"))); // \R matches all newline formats, yay convenience
```
ThatIsAUsername: Java 7 doesn't support `"\R"`.  How about `"[\n\r]+"`?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -108,6 +117,51 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
     return menuImage;
   }
 
+  /**
+   * Returns an image with the input string printed within the specified width, in normal text.
+   * @param reqWidth: Actual UI size in pixels that you want to fit the text into.
+   */
+  public static BufferedImage paintTextNormalized(String prose, int reqWidth)
```
ThatIsAUsername: What does `Normalized` mean in this context? I feel like there's probably a better name for this function.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -108,6 +117,51 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
...
+    // Unload our prose into the lines it already has
+    lines.addAll(Arrays.asList(prose.split("\\R"))); // \R matches all newline formats, yay convenience
+
+    for( int i = 0; i < lines.size(); ++i ) // basic for, since we care about indices
+    {
+      String line = lines.get(i);
+      if( line.length() * characterWidth <= reqWidth ) // if the line's short enough already, don't split it further
+        continue;
+
+      // TODO: check for word boundaries
```
ThatIsAUsername: I'm not a huge fan of splitting words across lines. This is a good TODO to get done.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -108,6 +117,51 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
...
+      lines.remove(i);
+      lines.add(i, line.substring(reqWidth / characterWidth)); // put in the second half
+      lines.add(i, line.substring(0, reqWidth / characterWidth)); // and then the first half behind it
+    }
+
+    int totalTextHeight = ((lines.isEmpty()) ? 0 : getMenuTextHeightPx(lines, characterHeight));
+    // Build our image.
+    BufferedImage menuImage = null;
+    if( reqWidth == 0 || totalTextHeight == 0 )
```
ThatIsAUsername: if `reqWidth == 0`, you would have hit an infinite loop condition where you are breaking lines apart. If you want to validate inputs, do it at the beginning of the function.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -63,8 +64,29 @@
...
+          "Glacio ("+GlacioAbility.GLACIO_COST+"):\r\n" + 
+          "Boosts the snow-aura around her buildings;\r\n" + 
+          "Hail falls in a two-space radius around her units,\r\n" + 
+          "- changing the weather in those tiles to snow,\r\n" + 
+          "- damaging enemies for up to 2HP, and\r\n" + 
+          "- destroying any forests (reducing them to grass)."));
+      infoPages.add(new InfoPage(
+          "Likes: Sleigh rides, Hot Chocolate\r\n" + 
+          "Dislikes: Jungles, Large Cats"));
+    }
```
ThatIsAUsername: Text is out of date.

src/CommandingOfficers/CommanderPatch.java
```diff
@@ -13,8 +13,24 @@
...
+          "- Patch gets a turn's worth of income from any property he captures\n"));
+      infoPages.add(new InfoPage(
+          "Plunder ("+PLUNDER_COST+"):\n" + 
+          "Gives an attack boost of "+PLUNDER_ATTACK_BUFF+"%\n" + 
+          "Gives "+PLUNDER_INCOME+"x of the value of any funds damage Patch deals.\n"));
+      infoPages.add(new InfoPage(
+          "Plunder ("+PILLAGE_COST+"):\n" + 
+          "Gives an attack boost of "+PILLAGE_ATTACK_BUFF+"%\n" + 
+          "Gives "+PILLAGE_INCOME+"x the value of any funds damage Patch deals.\n"));
+    }
```
ThatIsAUsername: We should probably tweak this guy's numbers. Costs of 5 and 10 seem a bit steep for his abilities.

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -22,8 +26,31 @@
...
+          "Grants footsoldiers and APCs one extra point of movement\n" +
+          "Allows Strong to put an extra unit into any transport\n" +
+          "Allows Strong to build inf and mechs on air/sea ports as well as the HQ\n"));
+      infoPages.add(new InfoPage(
+          "Mobilize ("+MobilizeAbility.MOBILIZE_COST+"):\n" + 
+          "Gives an attack boost of "+MobilizeAbility.MOBILIZE_BUFF+"%\n" +
+          "Grants footsoldiers and APCs two extra points of movement\n" +
+          "Allows Strong to put an extra unit into any transport\n" +
+          "Allows Strong to build inf and mechs on air/sea ports as well as the HQ\n"));
+    }
```
ThatIsAUsername: Outdated.

src/Engine/Utils.java
```diff
@@ -508,6 +508,67 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
...
+  {
+    StringBuilder sb = new StringBuilder();
+
+    // map state
+    int income = 0;
+    int unitCount = 0;
+    int vehCount = 0;
+    int unitFunds = 0;
+    
+    for( int w = 0; w < map.mapWidth; ++w )
```
ThatIsAUsername: Instead of looping over every square on the map, why not just iterate through `Commander.ownedProperties` and `Commander.units`?

src/UI/Art/FillRectArtist/FillRectEngine.java
```diff
@@ -42,4 +44,10 @@ public Dimension getScreenDimensions()
     // TODO: hard-coding is bad.
     return new Dimension(1280, 960);
   }
+
+  @Override
+  public IView createInfoView(InfoController control)
```
ThatIsAUsername: I kinda want to get rid of the FillRect stuff. I think it's served its purpose, and can be safely excised. Not as part of this PR of course (and we do still have one dependency on it).

src/CommandingOfficers/CommanderAve.java
ThislsAUsername: Good point. Haven't gotten around to removing the COMaker interface either.

src/Engine/Utils.java
ThislsAUsername: Fog

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,125 @@
...
+    // add the actual info
+    g.setColor(SpriteUIUtils.MENUFRAMECOLOR); // outer buffer
+    g.fillRect(  paneOuterBuffer,   paneOuterBuffer, paneHSize + 2*paneOuterBuffer, paneVSize + 2*paneOuterBuffer);
+    g.setColor(SpriteUIUtils.MENUBGCOLOR);    // inside of the pane
+    g.fillRect(2*paneOuterBuffer, 2*paneOuterBuffer, paneHSize                    , paneVSize                    );
+    
+    // TODO: consider drawing all pages as one big image, so the user can scroll smoothly through it regardless of screen size
+    switch (page.pageType)
+    {
+      case CO_HEADERS:
```
ThislsAUsername: (copied from the old PR)
I'd like to add income and unit count here, but I don't want to duplicate code and I'm not sure how best to split up the parsing that happens down below. Any ideas?

It's probably also worth caching the results of all these calculations so they aren't done every frame, but I'm not super worried about that at this point.

src/Engine/Utils.java
ThatIsAUsername: Ah, right. We are potentially looking at other CO's stuff here. Carry on.

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public Dimension getPreferredDimensions()
+  {
+    return new Dimension(viewWidth, viewHeight);
+  }
+
+  @Override
+  public void setPreferredDimensions(int width, int height)
+  {
+    // The user wants to use a specific amount of screen. Figure out how many tiles to draw for them.
```
ThatIsAUsername: I don't see much tile figuring.

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,125 @@
...
+    
+    // TODO: consider drawing all pages as one big image, so the user can scroll smoothly through it regardless of screen size
+    switch (page.pageType)
+    {
+      case CO_HEADERS:
+        int overlayHeight = 30*drawScale;
+        int heightOffset = 0;
+        for (Commander CO : myControl.getGame().commanders)
+        {
+          BufferedImage overlayPic = SpriteLibrary.createTransparentSprite(100*drawScale, overlayHeight);
```
ThatIsAUsername: eww, magic numbers

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThatIsAUsername: So then you'd have income and unit count on more than one page?

To answer your question, my first thought is:
You could have `Utils.getGameStatusData` return an object instead of a formatted string. The object could even have methods to extract strings for each bit of info it knows about, or just call toString() to get all of them.
Then you can call `getGameStatusData` before the switch here and reference it wherever needed.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -24,6 +24,8 @@
   // This is the physical size of a single map square in pixels.
   public static final int baseSpriteSize = 16;
 
+  static final String charKey = "%./-~,;:!?'&()";
```
ThatIsAUsername: Is `package private` the intended scope here?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -85,7 +94,7 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
     // Build our image.
     BufferedImage menuImage = null;
     if( menuWidth == 0 || menuHeight == 0 )
-      menuImage = SpriteLibrary.createDefaultBlankSprite(1, 1); // zero-dimensioned images aren't kosher
+      return SpriteLibrary.createDefaultBlankSprite(1, 1); // zero-dimensioned images aren't kosher
```
ThatIsAUsername: Should this throw an exception instead? Maybe it should.

ThatIsAUsername:
Bug report:
1) Start a new game
2) Press Enter to open the MetaAction menu
3) Select "CO Stats" or "CO Info" to enter that menu
4) Press Back to exit the CO_INFO menu
5) Press Enter to open the MetaAction menu
6) Select Quit Game
7) Select Leave Map
8) The application will close instead of falling back to the main menu.

And now that I took the time to write out steps, I think I might know why this happens.
If memory serves, I don't think I actually have a proper stack of Controllers, just a "previous" controller. Since I only had two (Driver and MapController), it worked just fine. I'm guessing that when you move to InfoController and back, the "previous" controller is now nothing, and there's nothing to fall back to after leaving the map.

src/UI/CO_InfoController.java
```diff
@@ -0,0 +1,109 @@
...
+    boolean goBack = false;
+    switch( action )
+    {
+      case DOWN:
+      case UP:
+        // Up/Down changes which CO has focus.
+        coOptionSelector.handleInput( action );
+        break;
+      case LEFT:
+      case RIGHT:
```
ThatIsAUsername: I feel like it might make sense to keep the different pages in sync. Like, If I'm on Strong and cycle over to his Passive page, and then move down to Patch, it should also show his Passive page instead of jerking me back over to the info page.

We'd still have to figure out how to handle COs with more than two abilities though. Maybe have a little navigation pane that shows a "map" of the info screens so you are always oriented.

src/UI/COSetupController.java
```diff
@@ -94,6 +95,20 @@ public boolean handleInput(InputAction action)
             coSelectors[playerSelector.getSelectionNormalized()].setSelectedOption(0);
         }
         break;
+      case SEEK: // Display the CO Info menu so we can window-shop
```
ThatIsAUsername: From the CO Setup screen, I kinda want the info screen to cycle through all available COs, not just those that are already selected... at least, that's what I naturally expected when I tried it. It feels a bit cumbersome to back out to cycle to a different person. Maybe it's just me.

I kinda want to rework this whole screen anyway if I ever find time, so don't worry too much about messing with it right now.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: I guess it could be proper private? Doesn't seem useful to other classes

src/UI/CO_InfoController.java
ThislsAUsername: Hmm, that's fair. Making the UI for that sounds hard, though.
Also, I have a mild ambition to make it one, contiguous image that you can scroll down, at some point, but that hasn't happened and probably won't soon.

src/UI/COSetupController.java
ThislsAUsername: Hmm. That would make sense. Just import the entire list of COs, and start from the currently-selected one.
Having it at all is better than the carts, though, haha.

> I kinda want to rework this whole screen anyway if I ever find time, so don't worry too much about messing with it right now.

I'm glad you think so highly of my designs :P

ThatIsAUsername: Mostly I'd like to be a bit more space efficient, remove the map, and include more info (such as actually showing the infantry sprite being selected). Probably also make a better way to select COs/Colors/Sprites than viewing one at a time when the list is potentially huge.

ThislsAUsername: > remove the map

Hmm. It does take a lot of space, but IMO it'll be useful with arbitrary players in unknown positions: you could change the color to see which player is where on the map (obviously that'd require that be done, though).

ThatIsAUsername: That's a fair point, but then AW doesn't show the map during CO select, and I don't think I've ever felt the need for it there (the exceptions being those final missions where you pick three COs to gang up on the big bad, but those are different).
Maybe I'll send you mockups of what I have in mind. We can keep the map if you think there's value, but I might still want to tweak some things.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: Hmm. It works just fine on my end.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: "Normalized" meaning "fits in the specified width"

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThislsAUsername: Right, so. The actual dimensions of the overlay sprite won't work, since the power bars can and do extend beyond it.
So one way or another, there will be a magic number.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: I discovered it kinda does already(?) if you shrink the window enough that it can't even fit one letter in. The window just goes unresponsive.
I ain't sure exactly where that comes from, but I got it to stop doing it for this next commit.

src/CommandingOfficers/CommanderPatch.java
ThislsAUsername: As part of this PR?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Are you sure you are running 7? My laptop chokes on it.

src/CommandingOfficers/CommanderStrong.java
```diff
@@ -28,25 +28,28 @@ public instantiator()
...
+          "Well-positioned soldiers with proper equipment and training are guaranteed to bring results.\n" +
+          "His focus is therefore on achieving this in his own forces, and denying it of his opponent.\n" +
+          "As a result, his transport units and boots on the ground are among the best, and he brings some extra anti-personnel firepower to keep his opponents in check."));
       infoPages.add(new InfoPage(
           "Passive:\n" + 
-          "- Strong gets an attack bonus of 20% on all units\n" +
-          "- Strong can build infantry from air/sea ports\n" +
-          "- Strong's APCs move 1 space further than average\n"));
+          "- Strong's footsoldiers get an attack bonus of 15%\n" +
+          "- When attacking footsoldiers, all Strong's units get an attack bonus of 15%\n" +
```
ThatIsAUsername: Should be "all _of_ Strong's units".

src/Engine/Driver.java
```diff
@@ -111,29 +106,33 @@ public static void main(String args[])
   @Override // From KeyListener
   public void keyPressed(KeyEvent arg0)
   {
+    // Store off the old controller/view so we can return to them.
+    IView oldView = gameView.view;
```
ThatIsAUsername: The scope of `oldView` could be reduced. Also, I'm not sure the comment here is accurate to how `oldView` is being used.

src/Engine/Driver.java
```diff
@@ -111,29 +106,33 @@ public static void main(String args[])
...
+        ControlState destination = gameStatus.pop(); // grab the old one
+        
         // Reinstate the previous controller/view.
-        // If the top-level controller returns done, it's time to exit.
-        changeGameState( oldController, oldView );
+        changeGameState( destination.controller, destination.view );
 
-        // We've switched back to these, now make sure we don't do it again (forever).
+        // Clean up our old view
         oldView.cleanup();
```
ThatIsAUsername: Perhaps do cleanup before calling `changeGameState` (and maybe before `System.exit(0)` just for good practice)?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: Looks like it to me, boss.
image

src/UI/UIUtils.java
```diff
@@ -175,6 +176,7 @@ public static ColorPalette getMapUnitColors(Color colorKey)
     {
       mapUnitColorPalettes.put(colorKey, mapUnitColorPalettes.get(Color.PINK));
       palette = mapUnitColorPalettes.get(colorKey);
+      System.out.println(String.format("WARNING!: Failed to retrieve unit palette for color %s, defaulting to PINK", colorKey));
```
ThatIsAUsername: Instead of specifying PINK, maybe use `UIUtils.getColorName(Color.PINK)`. Magic strings, and all that.

src/CommandingOfficers/CommanderPatch.java
ThatIsAUsername: Nah. Out of scope.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Huh. That matches what mine says. And yet...
```
Exception in thread "main" java.util.regex.PatternSyntaxException: Illegal/unsupported escape sequence near index 1
\R
 ^
	at java.util.regex.Pattern.error(Unknown Source)
	at java.util.regex.Pattern.escape(Unknown Source)
	at java.util.regex.Pattern.atom(Unknown Source)
	at java.util.regex.Pattern.sequence(Unknown Source)
	at java.util.regex.Pattern.expr(Unknown Source)
	at java.util.regex.Pattern.compile(Unknown Source)
	at java.util.regex.Pattern.<init>(Unknown Source)
	at java.util.regex.Pattern.compile(Unknown Source)
	at java.lang.String.split(Unknown Source)
	at java.lang.String.split(Unknown Source)
```
Weird.

ThislsAUsername: It... could be that I'm using the Zulu stuff rather than the Oracle stuff?
My reason for switching was their change in license I read about recently.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: `paintTextInFrame`? `drawTextToWidth`? `drawTextWithinSpecifiedBoundaryConstraints`?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Huh. Never heard of Zulu stuff. They are just a... repackaged OpenJDK?

Also seems odd that they named themselves after a time zone.

ThislsAUsername: Yes, but with different licensing.
If you want me to change this, I guess I can, but it looks like it's Java 7 compliant to me, hah.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: Hmm. Can I just change it in the `actionRework` branch to avoid conflicts?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: Also, matching `"[\n\r]+"` would negate the ability to add multiple-line breaks.

ThatIsAUsername: `"\r?\n?"` would technically match nothing, wouldn't it? If you want to match against just one of a given character set, just drop the '+': `"[\n\r]"`

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: I suppose.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: But then `\r\n`, the Windows newline, will produce 2 lines.
Based on [this question](https://stackoverflow.com/questions/3445326/regex-in-java-how-to-deal-with-newline), I'm thinking `"\r\n|\n|\r"`

ThatIsAUsername: Works for me.

src/CommandingOfficers/CommanderAve.java
```diff
@@ -68,23 +68,32 @@ public instantiator()
...
       infoPages.add(new InfoPage(
-          "Likes: Sleigh rides, Hot Chocolate\r\n" + 
-          "Dislikes: Jungles, Large Cats"));
+          "Glacio ("+GlacioAbility.GLACIO_COST+"):\n" +
+          "Ave's units gain a "+GlacioAbility.GLACIO_BUFF+"% increase in firepower.\n" +
+          "Increases the snow-aura around her buildings by "+GlacioAbility.GLACIO_SNOW_SPREAD+" spaces for the next turn.\n" +
+          "Snows on every tile in a "+GlacioAbility.GLACIO_SNOW_SPREAD+"-space radius around each of her units.\n" +
+          "Stuns any enemy unit within "+GlacioAbility.GLACIO_FREEZE_RANGE+" spaces of one of Ave's units or buildings."));
+      infoPages.add(new InfoPage(
+          "Oblido ("+OblidoAbility.OBLIDO_RANGE+"):\n" +
```
ThatIsAUsername: `OBLIDO_COST`

src/CommandingOfficers/CommanderInfo.java
```diff
@@ -1,15 +1,42 @@
...
+  public CommanderInfo(String name)
   {
     this.name = name;
-    maker = whichCo;
+    infoPages = new ArrayList<InfoPage>();
+  }
+
+  public abstract Commander create();
+
+  public static class InfoPage implements Serializable
```
ThatIsAUsername: Curious: Why does `InfoPage` have to implement `Serializable`?  I would  assume it could just be re-created on load or whatever, instead of needing to read from disk again since they are all static.

src/UI/InfoController.java
```diff
@@ -0,0 +1,19 @@
+package UI;
+
+import CommandingOfficers.Commander;
+import CommandingOfficers.CommanderInfo;
+import CommandingOfficers.CommanderInfo.InfoPage;
+import Engine.GameInstance;
+import Engine.IController;
+
+public interface InfoController extends IController
```
ThatIsAUsername: "InfoController" seems unnecessarily non-specific.  Maybe CommanderInfoController or something would convey intent better. I know there's a lot of dependent pull requests, so I'm probably fine putting this in the "address later" bin.

src/CommandingOfficers/CommanderInfo.java
ThislsAUsername: It's *also* stored by each instance of `Commander`. I could make it `transient`, but then I'd need to add logic to recall that and that seems like too much work.

src/UI/InfoController.java
ThislsAUsername: I was leaving it generic because there's no reason we couldn't use it for something else later.
It's just a paginated controller with relatively generic inputs.

src/Engine/Utils.java
```diff
@@ -508,6 +508,7 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
     TravelDistanceComparator tdc = new TravelDistanceComparator(unit, map);
     Collections.sort(mapLocations, tdc);
   }
+
```
ThatIsAUsername: Aaaaagh. You can't just go writing viruses into our game like this. People will notice.

ThislsAUsername: Viruses?

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,142 @@
...
+        COPic.getWidth()*drawScale, COPic.getHeight()*drawScale, null);
+
+    if (null != thisCO)
+      CommanderOverlayArtist.drawCommanderOverlay(g, thisCO, false);
+    
+    // add the actual info
+    g.setColor(SpriteUIUtils.MENUFRAMECOLOR); // outer buffer
+    g.fillRect(  paneOuterBuffer,   paneOuterBuffer, paneHSize + 2*paneOuterBuffer, paneVSize + 2*paneOuterBuffer);
+    g.setColor(SpriteUIUtils.MENUBGCOLOR);    // inside of the pane
+    g.fillRect(2*paneOuterBuffer, 2*paneOuterBuffer, paneHSize                    , paneVSize                    );
```
ThatIsAUsername: I'll admit I'm not a huge fan of the giant borders on this thing. I didn't have to build it though, so I guess I can't complain too much.

ThislsAUsername: It felt wrong to have the border be small like the game menus'

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    // TODO: consider drawing all pages as one big image, so the user can scroll smoothly through it regardless of screen size
+    int drawingWidth = paneHSize - paneOuterBuffer;
+    switch (page.pageType)
+    {
+      case CO_HEADERS:
+        int overlayHeight = 42*drawScale;
+        int heightOffset = 0;
+        for (Commander CO : myControl.getGame().commanders)
```
ThatIsAUsername: This page doesn't seem to differ based on which Commander is in focus. The info on this page is also mostly covered on the next page (with the exception of the CO panel, which is already in the top right).

src/UI/Art/SpriteArtist/SpriteInfoView.java
```diff
@@ -0,0 +1,142 @@
...
+        break;
+      case BASIC:
+        BufferedImage infoText = SpriteUIUtils.paintTextNormalized(page.info, drawingWidth);
+        g.drawImage(infoText,3*paneOuterBuffer, 3*paneOuterBuffer, null);
+        break;
+    }
+  }
+
+  
+  // Vestigial method stubs from MapView
```
ThatIsAUsername: Vestigial stubs make me uneasy.

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThislsAUsername: It's *all* of the CO headers at once, for a summary of the overall gamestate (and so you can see your opponents' meters).

src/CommandingOfficers/CommanderInfo.java
ThatIsAUsername: Gotcha. Laziness FTW. As long as our save-file sizes don't get out of hand. Which they wouldn't, from this.

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThislsAUsername: They aren't used, and MapView is where the viewable map finding logic lives.

I guess I could move that to UIUtils?

src/UI/InfoController.java
ThatIsAUsername: Fair enough. I might buy that. Then again, it's got `getSelectedCO` and `getSelectedCOInfo`, which would make using it for non-CO-related things awkward, at best.

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThatIsAUsername: Sure. It just feels weird to cycle the CO in the background and have nothing else change.

I won't make you change it, but I might want to change it one day.

src/UI/InfoController.java
ThislsAUsername: Pretty sure at least `getSelectedCO` is allowed to return `null`. It's no worse than `StateData`.

src/UI/Art/SpriteArtist/SpriteInfoView.java
ThatIsAUsername: ehhh, I'd just leave it for now. MapView seems like a logical place to have map-viewing logic. I'll let the uneasiness fester until it either goes away or causes actionable heartburn.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -119,18 +119,6 @@ public void setPreferredDimensions(int width, int height)
...
-    // Cap the view width/height to the dimensions of the map.
-    if( mapTilesToDrawX > myGame.gameMap.mapWidth )
-    {
-      mapTilesToDrawX = myGame.gameMap.mapWidth;
-      mapViewWidth = mapTilesToDrawX * tileSize;
-    }
-    if( mapTilesToDrawY > myGame.gameMap.mapHeight )
-    {
-      mapTilesToDrawY = myGame.gameMap.mapHeight;
-      mapViewHeight = mapTilesToDrawY * tileSize;
```
ThatIsAUsername: I get that folks want to be able to make it full-screen, but I also kind of want to preserve a way to set the window to the map size. Didn't you make a change somewhere that the window would size to the map when starting the game, and just allow resizing after that?

ThislsAUsername: Totally did. Beats me what PR it's in, though.

ThislsAUsername: ....and *that'd* be why. It's this PR. https://github.com/ThislsAUsername/DefendPeace/pull/192/commits/216db00a2eb0f7df9ce6e9ef219cc685eb793f81

src/UI/InfoController.java
ThatIsAUsername: And now I need to decide if that means "Oh, this is fine then" or "Maybe I also need to worry about `StateData`".
Haha. I think.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: I'd still lean towards an exception of some kind. Fail fast, and all that, but up to you.

ThislsAUsername: Not gonna go for it. I don't wanna make the game crash, and I don't want to bother catching the exception either.
Linux lets you resize the game window to smaller than the toolbar, leaving no space at all for the text.

ThatIsAUsername:
I'm curious: Why'd you close the `otherActions` PR and merge it in manually? You'd have to do the same steps either way, yeah?

ThislsAUsername:
> I'm curious: Why'd you close the `otherActions` PR and merge it in manually? You'd have to do the same steps either way, yeah?

Maybe.
I was gonna use the in-browser merge-deconflictor, but the file wouldn't load.
Also, being able to merge and immediately test (in an ideal world, before pushing, yes I'm an idjit) is beneficial.

ThatIsAUsername:
Wanting to test before pushing is not an idjit thing. It's also totally possible without cancelling the PR or using the in-browser deconflictor.
Just merge `target` into `dev` locally, resolve conflicts, and then push `dev`. Then you are resolved, and you can still see the in-browser PR diff view.

ThislsAUsername:
> Wanting to test before pushing is not an idjit thing. It's also totally possible without cancelling the PR or using the in-browser deconflictor.

I see you misunderstood. The idjit part is that I pushed before I even tried compiling. For some reason.

> Just merge `target` into `dev` locally, resolve conflicts, and then push `dev`. Then you are resolved, and you can still see the in-browser PR diff view.

That's fair, I suppose. I may do that next time (and there will probably be one, sigh)

PR #193, opened 2019-05-17
Reified action types work as advertised

ThislsAUsername:
Things to consider:
Having to create 3 separate objects just to add an action is pretty cumbersome (which may or may not be a bad thing?)
> I think a solution could be to add a "wait+" generic GameAction, which could cut out one of the classes for future instances of adding an action (unless it needs a target...).

Having to have those objects pass around a reference back to the CO is also kind of awkward.
> I think we could set it up so that the event listener interface is generic. Then each instance of it could specify what gets passed into the function.
> That will definitely require runtime management of types: you're gonna have to keep the Class around and use it to cast, which is a little awkward.

IDK how bad of an idea making GameInputState et. al. public is.
> I'm thinking maybe there should be an enum with a generic select type that the InputStates can pull out and use, but there's no support for multi-step processes there.
> Also, there should possibly be generic `GameInputState`s that just let you pick tiles, and outsource the validation of what should happen next to some other object (I'd say the `UnitActionType`, but not all `GameInputStates` are unit actions)

Thoughts?


src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+public class Meridian extends Commander
+{
+  private static final CommanderInfo coInfo = new instantiator();
+  private static class instantiator extends CommanderInfo
+  {
+    public instantiator()
+    {
+      super("Meridian");
+      infoPages.add(new InfoPage(
+          "While other commanders may focus on fine-grain tactics, Meridian has taken a different line: her basic tanks and artillery are built from the same parts, and can be re-configured on the field at no cost.\n" +
```
ThatIsAUsername: "fine-grained". Also, "line" feels like it should be "tack"

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+    CODamageModifier damageMod = null;
+    CODefenseModifier defenseMod = null;
+    Meridian COcast;
+
+    ChangeAndFlow(Meridian commander)
+    {
+      super(commander, STRONGARM_NAME, COST);
+
+      damageMod = new CODamageModifier(BASIC_BUFF);
+      defenseMod = new CODefenseModifier(BASIC_BUFF);
```
ThatIsAUsername: Since so many COs do the +10/10 thing, I'm tempted to make a modifier type that does both, just to make this less cumbersome. It will happen. Eventually. Probably.

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+    public Commander create()
+    {
+      return new Meridian();
+    }
+  }
+
+  /** A list of all the units I've refreshed and need to nerf. */
+  private ArrayList<Unit> justTransformed = new ArrayList<Unit>();
+  private ArrayList<Unit> toBeNerfed = new ArrayList<Unit>();
+  private static final int POST_REFRESH_STAT_ADJUSTMENT = -25;
```
ThatIsAUsername: This feels like it should be a public member of `VehicularCharge`. Also, a 25% firepower nerf (well, effectively 15%) feels kinda bad on a 7-cost ability.

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      // Grant the base firepower/defense bonus.
+      myCommander.addCOModifier(damageMod);
+      myCommander.addCOModifier(defenseMod);
+
+      // Lastly, all troops are refreshed and able to move again.
```
ThatIsAUsername: Hah, troops are the only ground units who are not refreshed.

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+        // Store the destination for later.
+        waitLoc = new XYCoord(path.getEnd().x, path.getEnd().y);
+      }
+      type = pType;
+    }
+
+    @Override
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
+      // WAIT actions consist of
```
ThatIsAUsername: WAIT

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    @Override
+    public void sendToListener(GameEventListener listener)
+    {
+      // listener.receiveUnitDieEvent( this );
+    }
+
+    @Override
+    public void performEvent(MapMaster gameMap)
```
ThatIsAUsername: Huh. I probably would have just used a one-way `UnitRemodelModifier`.

ThatIsAUsername: Which, to be fair, doesn't attempt to adjust ammo, but it could.

src/CommandingOfficers/Meridian.java
```diff
@@ -0,0 +1,375 @@
...
+
+    @Override
+    public UnitActionType getUnitActionType()
+    {
+      return type;
+    }
+  }
+
+  /** Effectively a wait, but the unit ends up as a different unit at the end of it. */
+  public static class UnitTransformEvent implements GameEvent
```
ThatIsAUsername: Should new action types be put in the common action repository just in case they see future use?

src/Engine/GameAction.java
```diff
@@ -165,6 +166,12 @@ public String toString()
       return String.format("[Attack %s with %s after moving to %s]",
           defender.toStringWithLocation(), attacker.toStringWithLocation(), moveCoord );
     }
+
+    @Override
+    public UnitActionType getUnitActionType()
```
ThatIsAUsername: Having both `getType` and `getActionType` feels duplicative.

src/Engine/UnitActionType.java
```diff
@@ -0,0 +1,285 @@
...
+    public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor)
+    {
+      XYCoord moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
+      if( map.isLocationEmpty(actor, moveLocation) )
+      {
+        if( actor.heldUnits.size() > 0 )
+        {
+          ArrayList<GameAction> unloadActions = new ArrayList<GameAction>();
+
+          // TODO: This could get messy real quick for transports with more cargo space. Figure out a
```
ThatIsAUsername: I think this TODO is OBE.

ThatIsAUsername:
> I think a solution could be to add a "wait+" generic GameAction

Why not just extend `WaitAction`?

>Having to have those objects pass around a reference back to the CO is also kind of awkward.

Which objects? Neither `GameEvent`s nor `GameAction`s are required to have a Commander reference.

> I think we could set it up so that the event listener interface is generic.

As in, the `receiveXEvent` call? Wouldn't that effectively require us to have multiple listener objects?

> IDK how bad of an idea making GameInputState et. al. public is.

Yeeeeah, I'm not sure I'm a fan of that either.

> I'm thinking maybe there should be an enum with a generic select type that the InputStates can pull out and use, but there's no support for multi-step processes there.

No? I mean, the `GameInput` state machine would just have to know where to go next, like it currently does.

> Also, there should possibly be generic GameInputStates that just let you pick tiles, and outsource the validation of what should happen next to some other object

I don't think I agree with this - I'd prefer to keep the state machine self-contained; outside actors shouldn't need to know how it works, and definitely shouldn't affect how it works. If we need a new way of doing things (box-select or whatever), then we should modify the state machine to support it.

src/CommandingOfficers/Meridian.java
ThislsAUsername: The tactics are not fine-grained, they are *along* the fine grain. I wouldn't describe Max as using "fine-grained" tactics, but his tactics are better than others' when in a fine-grain analysis i.e. on the level of individual tanks in a unit.

"Line" is a pun, I'm not changing it.

src/CommandingOfficers/Meridian.java
ThislsAUsername: Or we can just keep copy-pasting that code over and over again...

Honestly, it might make sense to make a subclass of `CommanderAbility` to do that for us.

src/CommandingOfficers/Meridian.java
ThislsAUsername: Hmm. I get what you're saying, but it's a substantial chunk of Eagle's Lightning Strike utility and she's got a better D2D.

Could drop it to 6, though, I guess. It's a substantial stat difference and she doesn't refresh air/sea.

src/CommandingOfficers/Meridian.java
ThislsAUsername: It would make sense, but these require a strongly-typed reference back to the original Meridian (since there's no event for tranforming) so that's not viable.

src/Engine/GameAction.java
ThislsAUsername: I agree, but I dunno how to fix it.

ThislsAUsername:
> Why not just extend `WaitAction`?

That... would be a good idea. I think I'll try it and see how it goes.

> Which objects? Neither `GameEvent`s nor `GameAction`s are required to have a Commander reference.

All 3 of Meridian's new unitactiontype, action, and event need a reference back to her since there' no event to make the round trip back to let her know it happened.

> As in, the `receiveXEvent` call? Wouldn't that effectively require us to have multiple listener objects?

What do you mean by "multiple listener objects". It could probably just be an interface `receiveEvent<T>(T event)`, and you can implement as many instantiations of that interface as you want to.
When publishing events, either you'd keep separate lists of listeners or check to make sure they implement `EventListener<T>` before calling the receive method.

> > I'm thinking maybe there should be an enum with a generic select type that the InputStates can pull out and use, but there's no support for multi-step processes there.
> 
> No? I mean, the `GameInput` state machine would just have to know where to go next, like it currently does.

Right, so `GameInput` needs to know the details about what all actions are possible, so it can handle complex actions. Like, technically I didn't need to make them public *for this*, but custom actions in the future might.
An example might be the ability used in the start of this video: https://www.youtube.com/watch?v=wn905lLHAo0

> I don't think I agree with this - I'd prefer to keep the state machine self-contained; outside actors shouldn't need to know how it works, and definitely shouldn't affect how it works. If we need a new way of doing things (box-select or whatever), then we should modify the state machine to support it.

IMO we need some other object to inform the `GameInput` machine what input it should be taking. This might even fix the weirdness that is `StateData` if done correctly.

I'm not saying the other object should know about the `GameInput` object *types*, but it should know that there's free tile/constrained tile/menu select and be able to use them to get multi-step input (like, say, a custom Unload that has more than 1 range and works without needing to be on ground the other unit can traverse)

src/CommandingOfficers/Meridian.java
ThislsAUsername: That mildly feels like using a vise when pliers would be sufficient, but that would be a way to re-use code.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: uhhhh, ok. I guess. "Unit-level" might be more clear than "fine-grain" though.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: How so? You are remodeling the unit, after all. And we use `COModifiers` without intent to revert in lots of places.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: Hm. Meridian could be reworked to not require a strongly-typed reference in the `UnitTransformEvent` (and that does seem like an odd requirement anyway), but I see what you mean, so I ain't gonna make you change it this time.

src/CommandingOfficers/Meridian.java
ThislsAUsername: "Unit-level" feels too game-y IMO.
Could just say "micro tactics" or "micro-level tactics"?

src/CommandingOfficers/Meridian.java
ThislsAUsername: Sure, but this is an event for 1 unit. `COModifiers` are for Commanders, not so much individual units.

src/CommandingOfficers/Meridian.java
ThislsAUsername: I mean, I *guess*, since she can keep a list of what type each unit is, and check to see if it transformed after every move action, but that seems like a really terrible idea.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: I mean. This is a game. But: "granular", "small-scale", "individual", "Lilliputian-esque"
I'm not sure that transforming units counts as the opposite of all that, but eh.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: You wouldn't have to check after every action, just when you used her ability. And really, I don't see why you couldn't do it the same way, just instead of having the event call `user.justTransformed.add(unit);`, you could just implement `receiveUnitTransformEvent(Unit unit, UnitModel newType)` and add it to the collection there (assuming `unit.CO == this`, of course).
That removes the need for a Meridian reference in `TransformEvent`, and `TransformActionType`/`TransformAction` only need a reference to Meridian so they can hand it to `TransformEvent`.

ThislsAUsername: > you could just implement `receiveUnitTransformEvent(Unit unit, UnitModel newType)`

That's of dubious utility. Since you'll be receiving *after* the event happens, you'd want the old type not the new one.
Also, the idea is to allow Commanders to add new action types with minimal effect on other parts of the codebase. A generic event publishing system would work for that, but I'm not convinced I want to add listening for a transform event to the `GameEventListener` overall.

Like, yes, I *could* add it to the `GameEventListener`, but that defeats the point of the exercise.

src/CommandingOfficers/Meridian.java
ThislsAUsername: I like "granular".
Jess's tanks and artillery are a little better at killing things than average.
Meridian's tank can turn into an artillery to be effective vs a Md. Tank, and her arty can become a tank to win against an opposing arty.

The former seems more like a tactical benefit, and the latter a strategic benefit.
Tactical advantages can create strategic advantages, but they're not exactly the same.

ThatIsAUsername:
> When publishing events, either you'd keep separate lists of listeners or check to make sure they implement EventListener<T> before calling the receive method

Yeah, that does feel a bit messy. I'd avoid going down this route unless we have a real good reason.

> technically I didn't need to make them public for this

Then I would vote to not make them public yet.
> An example might be the ability used in the start of this video: https://www.youtube.com/watch?v=wn905lLHAo0

We could implement that basically the same way as Unload, though it would require adding to the state machine.

> IMO we need some other object to inform the GameInput machine what input it should be taking. This might even fix the weirdness that is StateData if done correctly.

You are talking about a state machine with externally-defined states. I can see how that might work, but it also might cause issues if every new commander can generate their own new bugs/features. It might still be worth investigating though.

ThislsAUsername:
> > When publishing events, either you'd keep separate lists of listeners or check to make sure they implement EventListener before calling the receive method
> 
> Yeah, that does feel a bit messy. I'd avoid going down this route unless we have a real good reason.

Hmm. I suppose the complexity might override the benefit in this case?


> > IMO we need some other object to inform the GameInput machine what input it should be taking. This might even fix the weirdness that is StateData if done correctly.
> 
> You are talking about a state machine with externally-defined states. I can see how that might work, but it also might cause issues if every new commander can generate their own new bugs/features. It might still be worth investigating though.

They already can.
Ave creates persistent weather zones.
Cinder can take up to 10 turns per unit in one turn.

IMO the benefit outweighs the risk since we shouldn't ever end up having two different sources of truth for what input state we're in. Either it's a meta-action, controlled by the `StateData`, or it's a unit action (or maybe CO power?) that's controlled only by the specific action type.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: > the idea is to allow Commanders to add new action types with minimal effect on other parts of the codebase

Sure, but "minimal" might still mean "more than you initially hoped".

> but that defeats the point of the exercise.

I dunno about that. If a new kind of thing is happening, it makes sense to have a new kind of event.

ThislsAUsername: Should I make Transform a proper basic action type, then?
It *is* fairly extensible ~~and could be used for your eventual werewolf CO~~

ThatIsAUsername: I wasn't suggesting to make it a standard action type, though I suppose you could. My natural leaning would be to go in the other direction and make all the other actions implicit, like how we didn't have to specify WAIT/LOAD/JOIN in units' available action types before; the game just knows those are available. I'm not sure why we couldn't do the same for ATTACK (has a weapon) and CAPTURE (is infantry), though RESUPPLY would require something different. Then we could do away with the canonical `UnitActionType` arrays. COs could still add new action types to specific models as in this PR.

~~Also, what? I have no idea what you are talking about.~~

ThislsAUsername: Hmm. I think I like LOAD/JOIN being explicit, since we could theoretically add units we don't want doing those things.
I'm not sure I'm a super fan of making things implicit in general, either.

~~Mmmhmm. For sure. Also, that entirely depends on the kind of werewolf it is. My Wesnoth mod made it so when you got a werewolf whose day form was a Mage of Light, it gained the ability to transform to a wolf manually any time it wanted (and gained a darkness aura while transformed so it had full control of its form)~~

src/Engine/UnitActionType.java
```diff
@@ -233,4 +234,35 @@ public String name()
       return "RESUPPLY";
     }
   }
+
+  /**
+   * Effectively a wait, but the unit ends up as a different unit at the end of it.
+   * Is a template action, and thus doesn't get a constant.
```
ThatIsAUsername: template action?

ThislsAUsername: It needs to know what it's transforming *to*, so you can't just make a global TRANSFORM constant because that doesn't make sense.

"Template" isn't 100% technically correct, but it's close to what I mean and it was the first acceptable thing that came to mind.

ThislsAUsername:
OBE due to merge conflicts and merging locally as a result.

PR #194, opened 2019-05-19
Other actions

ThislsAUsername:
I'm not super happy with how I implemented adding DIVE and RISE to the sub versions.
I could make a new action for it, but I'm not sure if that's better (adding essentially duplicate code)

I think how it is now would be fine if we shuffled the unit list construction out to another class, but as noted in the comment, that might get a little ugly too since you need a full list and also a set of buyable lists.

EDIT: wow, okay, I meant to put the map scrolling changes on `thatRequests`. Oops. Hopefully it's all going the same place anyway, so eh?


src/CommandingOfficers/Commander.java
```diff
@@ -132,6 +134,20 @@ public Commander(CommanderInfo info)
...
+    // That has the downfall that there's only one return type from that.
+    // I guess we could have one function that builds the whole unit library, and one that pulls out the buildables and makes the map.
+    // Whatever. This works, for now.
+
+    // Subs gotta dive, man
+    UnitModel sub = getUnitModel(UnitModel.UnitEnum.SUB);
+    UnitModel subsub = new SubSubModel();
+    unitModels.add(subsub);
+    sub.possibleActions.add(new UnitActionType.Transform(subsub, "DIVE"));
+    subsub.possibleActions.add(new UnitActionType.Transform(sub, "RISE"));
```
ThatIsAUsername: I'm still wondering if it would be better to just put the unit type in the `Transform` action, and query the Commander for the specific model to use. You said you're not sure you like that, but is it worse than this?

src/Engine/GameAction.java
```diff
@@ -976,4 +976,56 @@ public UnitActionType getUnitActionType()
...
+    @Override
+    public ActionType getType()
+    {
+      return ActionType.UNIT;
+    }
+    
+    @Override
+    public String toString()
+    {
+      return String.format("[Delete %s in place]", actor.toStringWithLocation());
```
ThatIsAUsername: Why "in place"?

src/Engine/GameAction.java
```diff
@@ -976,4 +976,56 @@ public UnitActionType getUnitActionType()
...
+    {
+      actor = unit;
+      destination = new XYCoord(unit.x, unit.y);
+    }
+
+    @Override
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
+      GameEventQueue eventSequence = new GameEventQueue();
+      eventSequence.add(new UnitDieEvent(actor));
```
ThatIsAUsername: Deleting is the same as killing, eh? Poor saps.

src/Engine/GameInput/SelectUnitAction.java
```diff
@@ -61,13 +61,17 @@ protected OptionSet initOptions()
...
-        if( UnitActionType.ATTACK == chosenSet.getSelected().getUnitActionType() )
+        if( UnitActionType.ATTACK == actionType )
           // We need to select a target to attack.
           next = new SelectActionTarget(myStateData);
-        else if( UnitActionType.UNLOAD == chosenSet.getSelected().getUnitActionType() )
+        else if( UnitActionType.UNLOAD == actionType )
           // We need to select a unit to unload.
           next = new SelectCargo(myStateData);
+        else if( UnitActionType.DELETE == actionType )
+          // We need to select a unit to unload.
```
ThatIsAUsername: Unload... from life!

src/Engine/UnitActionType.java
```diff
@@ -17,15 +17,16 @@
   public static final UnitActionType CAPTURE = new Capture();
   public static final UnitActionType RESUPPLY = new Resupply();
   public static final UnitActionType WAIT = new Wait();
+  public static final UnitActionType DELETE = new Delete();
   public static final UnitActionType LOAD = new Load();
   public static final UnitActionType JOIN = new Join();
 
-  public static final UnitActionType[] GENERIC_ACTIONS = { ATTACK, UNLOAD, CAPTURE, RESUPPLY, WAIT, LOAD, JOIN };
+  public static final UnitActionType[] GENERIC_ACTIONS = { ATTACK, UNLOAD, CAPTURE, RESUPPLY, WAIT, DELETE, LOAD, JOIN };
```
ThatIsAUsername: Nooo, conflicts. I still have this deleted in `ActionReworkMod`. Ah well.

src/Terrain/MapWindow.java
```diff
@@ -191,12 +212,25 @@ public void revealFog(Unit scout, Path movepath)
...
       for( PathNode node : movepath.getWaypoints() )
       {
-        for( XYCoord coord : Utils.findVisibleLocations(this, scout, node.x, node.y) )
+        for( XYCoord coord : Utils.findVisibleLocations(this, scout, node.x, node.y, false) )
         {
-          isFogged[coord.xCoord][coord.yCoord] = false;
-          lastOwnerSeen[coord.xCoord][coord.yCoord] = master.getLocation(coord).getOwner();
+          revealFog(coord, false);
+        }
+        // We need to do a second pass with piercing vision so we can know whether to reveal the units
```
ThatIsAUsername: I'm not a fan of this double-pass approach. Seems like we should be able to just figure it out in one go.

src/Engine/GameAction.java
ThislsAUsername: Because I require the unit to stand still to delete. And because it was copy/pasted from text with a location in it.

src/Engine/UnitActionType.java
ThislsAUsername: Too little, too late.

src/Terrain/MapWindow.java
ThislsAUsername: There doesn't seem to be an easy way to return both sets of info - the places you can reveal cloaked units, and the places you can only see into.

src/CommandingOfficers/Commander.java
ThislsAUsername: I suppose just using the type enum would work well enough, and allow creation in the unit's constructor.
I guess I'm just not a huge fan of CO.getUnitType because we have various reasons for a CO to have multiple models of a single type... but I don't know how to properly fix that.

ThatIsAUsername: I've already nuked the one case where we had that (unless I'm forgetting something). I'm not too keen to bring that back into style, either.

src/Terrain/MapWindow.java
ThatIsAUsername: Isn't the rule for revealing a cloaked unit the same as seeing units under cover in fog?

src/Engine/GameInput/SelectUnitAction.java
ThatIsAUsername: But yeah, that's copypasta.

src/Terrain/MapWindow.java
ThislsAUsername: It is, but whether the cover tile is "fogged" is equal to whether you can see a unit in it.
If a sub is sitting in open water, you'd be able to see the water perfectly well while not seeing the sub.

Also, yes, the rule is the same (and this is a deliberate departure from the carts; piercing vision doesn't help you there vs hidden units)

src/CommandingOfficers/Commander.java
ThislsAUsername: As of the tip of `readResolve`, Muriel and Strong both use CO.`getUnitModel()` for regular functions.
Meridian and MapMaster use it in their constructors.

~~It's also used in Test, but nobody cares about that.~~

ThatIsAUsername: Sure, but using co.`getUnitModel()` isn't the problem. Having more than one valid unit model of a given type is. We don't have that anymore. To be clear, my suggestion is to also use `getUnitModel()` for `Transform`.

src/Units/UnitModel.java
```diff
@@ -41,7 +41,8 @@
   public int idleFuelBurn;
   public int movePower;
   public int visionRange;
-  public boolean visionIgnoresCover = false;
+  public int visionRangePiercing = 1;
```
ThatIsAUsername: All right, so I really don't like this. It absolutely feels like a kludge: fundamentally, units don't have two vision rages, and piercing vision isn't required to see adjacent units (whether cloaked or covered). In fact, on `master`, we don't have any unit or Commander who supports piercing vision at all.

I would go so far as to say we could (and probably should) implement the basics of visibility without needing this flag. Just delete it and get things working that way. If we overload the meaning or use of piercing vision it'll hurt us long-term.

src/CommandingOfficers/Commander.java
ThislsAUsername: Hmm. Okay.
The problem, I suppose, is that we have a whole modifier type for accomplishing... well, exactly that: adding a new unit model type.
And we assume that any given unit model will exist already for any CO, so there's no case where a UnitRemodel can work without creating a two-unit-model problem.

src/Units/UnitModel.java
ThislsAUsername: > fundamentally, units don't have two vision rages, and piercing vision isn't required to see adjacent units (whether cloaked or covered)

That's the thing; *all* units have piercing vision at 1 range or else you wouldn't be able to see into cover. The only difference is that "actual" piercing vision doesn't reveal stealthy units, but I plan to change that.
It's also worth noting that *properties* shouldn't get piercing vision (and they did until this change) (I might be willing to make an exception for units *on* them, but whatever)

> I would go so far as to say we could (and probably should) implement the basics of visibility without needing this flag. Just delete it and get things working that way.

I would do the exact same thing as I did before, but with a hard-coded "1" in the vision logic instead. The problem is that there's two different sight modes, *not* that both sight modes are configurable.

I *could* remove the second path by just making a new object with two collections, or having an in/out collection param, but that will still two passes, just through different collections.
I imagine it would also work to return, say, a `HashMap<XYCoord, bool>`, with the `keyset` being visible tiles and the `bool` determining whether cloaked units should be shown.

In essence, returning what *tiles* are revealed is no longer sufficient information to determine full visibility rules; some other flag is needed.

> If we overload the meaning or use of piercing vision it'll hurt us long-term.

What... what is this supposed to mean?

src/CommandingOfficers/Commander.java
ThatIsAUsername: `UnitRemodelModifier` isn't currently used anywhere. Given that you already have a different way of transforming units, I'm fine with deleting this one.
I'm not really sure why we didn't use a `Map<UnitType, UnitModel>` to start with - that seems more natural than just a list. I don't much like the idea of allowing COs to have multiple unit models for one type.

src/Units/UnitModel.java
ThatIsAUsername: > all units have piercing vision at 1 range or else you wouldn't be able to see into cover

I'd be more inclined to say that piercing vision isn't required to detect units at 1 range. Maybe it's a potato potato thing.

> > If we overload the meaning or use of piercing vision it'll hurt us long-term.
> What... what is this supposed to mean?

Piercing vision allows seeing enemies in cover when we wouldn't normally be able to. Units are normally able to see adjacent enemies, even when they are in cover. Using the piercing attribute to handle a case where piercing vision is not necessary is overloading the use of that attribute. Overloading things like that can lead to unintended dependencies or side effects. It may be clever, it may even work, but it normally means there is a better way to decompose the problem.
On the other hand, if we frame it as "all units have piercing vision at 1 range" then we could argue that it isn't being overloaded to handle this case.
On the other other hand, if all units have piercing vision at 1 range, it seems like we shouldn't need a separate UnitModel attribute for it.

src/CommandingOfficers/Commander.java
ThislsAUsername: Actually, I'm wrong. The UnitRemodel could take a unit between two existing types... but I think that perhaps we should then just feed the UnitRemodelMod a destination type rather than a destination model (as you're proposing we do with Transform).

Okay, I think we've perhaps gotten this settled. Should I just make all these changes on this branch?

src/Units/UnitModel.java
ThislsAUsername: > I'd be more inclined to say that piercing vision isn't required to detect units at 1 range. Maybe it's a potato potato thing.

It's close, but not *quite* the same. Buildings have vision (and maybe piercing vision on top of themselves), but they should not be able to see into cover or detect at 1 range.
Whether you make a piercing vision stat or not, the fact would remain that *units* would have piercing vision at range 1.

> > > If we overload the meaning or use of piercing vision it'll hurt us long-term.
> Piercing vision allows seeing enemies in cover when we wouldn't normally be able to.

I would shorten this. Piercing vision allows seeing enemies in cover. I don't think it's helpful to make a special distinction between the vision you get of adjacent squares and "boosted" vision that lets you do the same feats of detection at higher ranges.

> Units are normally able to see adjacent enemies, even when they are in cover. Using the piercing attribute to handle a case where piercing vision is not necessary is overloading the use of that attribute.

Piercing vision, in my view, is the state of being able to detect hidden units. Units have it at range 1 by default, and a CO could conceivably grant it at greater ranges.
I don't see a good reason to special-case detection at 1 range. It's piercing vision, and works exactly the same as piercing vision.

> Overloading things like that can lead to unintended dependencies or side effects.

The only thing I can think of, here, is that hidden unit detection will *also* happen at whatever piercing vision range you have... but that's a good change, IMO.
Seeing as "piercing vision" and "can see into cover" are exactly equivalent in all respects... it's not overloading, it's looking at a duck and calling it a duck.

> On the other other hand, if all units have piercing vision at 1 range, it seems like we shouldn't need a separate UnitModel attribute for it.

Possibly not, but I don't really want to hardcode a *unit* attribute in some other piece of code.
Something I and others have contemplated in the past, for instance, is giving mobile SAMs piercing vision out to their normal vision range; this would give them a niche in being able to detect cloaked units without taking up attacking space (though that advantage is less with DoR fog rules).

src/CommandingOfficers/Commander.java
ThatIsAUsername: Do eet.

src/Units/UnitModel.java
ThatIsAUsername: Hmmm. I think I will concede (for now, at least). At any rate, this is working now.

ThatIsAUsername:
It might be good to get some test cases and validate that unit hiding/detection works the way we expect it to in all cases, since this stuff isn't super trivial.

src/Units/UnitModel.java
ThislsAUsername: I don't know how relevant this is to you, but I decided to doublecheck the behavior on DoR, and subs aren't revealed by moving past them (at least, outside of fog; flares can't "flare" outside fog either, it seems).

ThatIsAUsername: So flares can't even use their ability if FoW is turned off? Strange. Not that you'd need to, but still.

Not too relevant; I don't have a strong stance on whether hidden/covered visibility conditions should match; it sorta makes sense that they would though.

ThislsAUsername:
> I don't much like the idea of allowing COs to have multiple unit models for one type.

I think I agree with the general sentiment here. I suppose that we'd just have to add hooks for all relevant conditional boosts, then?
If, say, I wanted to have a CO that granted +1 move to vehicles on roads every day, the old unit remodel logic would work; it doesn't work for that anymore.

ThislsAUsername:
....doof, I forgot to change `UnitRemodel`. Fixing now.

ThatIsAUsername:
> If, say, I wanted to have a CO that granted +1 move to vehicles on roads every day, the old unit remodel logic would work; it doesn't work for that anymore.

For that, you don't even need to remodel. Just add 1 to the move attribute.

Weren't you going to remove `UnitRemodelModifier` altogether? I'm pretty sure nobody uses it.

ThislsAUsername:
> > If, say, I wanted to have a CO that granted +1 move to vehicles on roads every day, the old unit remodel logic would work; it doesn't work for that anymore.
> 
> For that, you don't even need to remodel. Just add 1 to the move attribute.

But that would globally affect *all* units. You'd need two models per unit type, one with an extra point of move, one with normal move.

> Weren't you going to remove `UnitRemodelModifier` altogether? I'm pretty sure nobody uses it.

\*cough\*
Seems silly to throw working code away anyhow...

ThatIsAUsername:
> But that would globally affect all units. You'd need two models per unit type, one with an extra point of move, one with normal move.

Not globally; each Commander has their own `UnitModel` for each type. Also, we have a `COMovementModifer` to handle that case. In either case, Even without that, I don't see that it would require multiple `UnitModel`'s of one type.

ThislsAUsername:
> Not globally; each Commander has their own UnitModel for each type.

Bah, I meant "all units for that Commander"

>Also, we have a COMovementModifer to handle that case. In either case, Even without that, I don't see that it would require multiple UnitModel's of one type.

Okay, so I've got a CO named "Bob".
Bob's a NASCAR driver, so he's all about dat speed, but he can't race just anywhere.
So, on any given turn, his units gain +1 move... but they only get the move bonus if they start their turn on a road.

It's the same *general* problem as the Commander's combat modifier hooks solve, but for movement range.

ThatIsAUsername:
Ah. Now I understand what you meant by "a CO that granted +1 move to vehicles on roads every day." - "A CO that grants +1 move to units that start their turn on a road".
So yes, old `UnitRemodelModifier` could do that. New one can't. It _would_ require a separate `UnitModel`, or some kind of hook back into Commander. We are on the same page now, probably.

We could keep `UnitRemodelModifier`, but it does seem duplicative of the `Transform` stuff.

ThislsAUsername:
> So yes, old `UnitRemodelModifier` could do that. New one can't. It _would_ require a separate `UnitModel`, or some kind of hook back into Commander. We are on the same page now, probably.

Seeing as you're against making duplicate models, I suppose we'd want to go for the hook?

ThatIsAUsername:
> Seeing as you're against making duplicate models, I suppose we'd want to go for the hook?

Is this a use case we want to support? I was under the impression that we didn't want this to be a thing. If we do, `UnitRemodelModifier` is probably as good a way as any to make it happen (and would give it a justification for existing alongside `Transform`, possibly), though I'd want to evaluate other options before deciding.

ThislsAUsername:
> Is this a use case we want to support?

I like leaving my options open (I do have a few concepts that that sort of thing, though I'm not sure they're worth adding)

> I was under the impression that we didn't want this to be a thing.

It does cause potential problems, but it can be useful. Overall, hooks might be the better choice because having a 1/1/1 model/type/CO relationship is desirable for keeping things... easy to understand.
I can live either way; I was under the impression you were pushing for the 1/1/1 correspondence.

> I'd want to evaluate other options before deciding.

I think this is probably what it comes down to. We can burn the bridge when we get to it, I guess. (which might be never?)

src/Engine/GameAction.java
```diff
@@ -820,9 +820,14 @@ public GameEventQueue getEvents(MapMaster map)
         // Note that movePath being null is OK for ResupplyAction when it is being re-used.
         if( movePath != null )
         {
-          eventSequence.add(new MoveEvent(unitActor, movePath));
+          // If we should be blocked, don't resupply anything.
+          if( !Utils.enqueueMoveEvent(map, unitActor, movePath, eventSequence) )
+            isValid = false;
```
ThatIsAUsername: If `enqueueMoveEvent()` fails, it doesn't mean the action is invalid; it means the unit got ambushed, so the unit didn't reach its destination.
Like, I think the logic here works, but the usage of `isValid` is no longer consistent with the other actions. I'd rather seethis expressed as `isValid &= (null != movePath);`, and then `if( Utils.enqueueMoveEvent() ) { /*resupply stuff */ }`

ThislsAUsername: My reasoning: it's no longer a valid *resupply* action, and the move's been queued already.

Also, `movePath == null` is a valid action, that's what gets pushed out on turn init.

ThatIsAUsername: Ah, right. I forgot a null move path is valid for resupply. But I think you understand the intent behind what I wrote.
And an attack action that gets preempted is still a valid action, even though the attack itself won't occur. I'm not saying this code won't work, I'm just saying the treatment of `isValid` might be misleading to whoever looks at it in three years.

ThislsAUsername: I mean, I guess I could make a second boolean, but that seems needless.
I could return there, but that's got its own weirdnesses.
Mebbe just another comment, specifying the exceptional case?

ThatIsAUsername: I mean... I might make a second Boolean if it were me, but I lean towards verbose code. I guess it doesn't matter all that much; a comment stating that the unit can still move despite setting `isValid` to false is probably sufficient. I know that's obvious to someone who knows how `enqueueMoveEvent()` works, but that's the kind of thing that won't be obvious if you haven't looked at it in a while.

ThatIsAUsername:
So much for avoiding conflicts. I attempted to deconflict, but since this is your PR you should probs check it over.

PR #195, opened 2019-05-25
Action rework mod

ThatIsAUsername:


src/AI/AIUtils.java
```diff
@@ -63,7 +63,7 @@
   {
     // Create the ActionType-indexed map, and ensure we don't have any null pointers.
     Map<UnitActionType, ArrayList<GameAction> > actionsByType = new HashMap<UnitActionType, ArrayList<GameAction> >();
-    for( UnitActionType atype : UnitActionType.GENERIC_ACTIONS )
+    for( UnitActionType atype : unit.model.possibleActions )
```
ThislsAUsername: Are you sure this change is safe? AIs just assume that, for example, ATTACK is always populated. If one ended up with an APC or T-Copter, they might crash due to pulling a null array from this function's result map.

src/CommandingOfficers/Meridian.java
```diff
@@ -75,7 +75,8 @@ public GameEventQueue initTurn(GameMap map)
     return super.initTurn(map);
   }
 
-  public void receiveUnitTranformEvent(Unit unit, UnitModel oldType)
+  @Override // GameEventListener interface
+  public void receiveUnitTransformEvent(Unit unit, UnitModel oldType)
```
ThislsAUsername: ...whoops

src/Engine/GameAction.java
```diff
@@ -32,19 +32,13 @@
...
-
   /**
    * Returns a GameEventQueue with the events that make up this action. If the action
    * was constructed incorrectly, this should return an empty GameEventQueue.
    */
   public abstract GameEventQueue getEvents(MapMaster map);
   public abstract XYCoord getMoveLocation();
   public abstract XYCoord getTargetLocation();
-  public abstract ActionType getType();
   public abstract UnitActionType getUnitActionType();
```
ThislsAUsername: Huh. I totally figured it would be used. Maybe rename `getUnitActionType()` to avoid verbosity? Then again, maybe we should delay a change like that to avoid merge conflicts for now.

src/Engine/GameAction.java
```diff
@@ -956,7 +896,7 @@ public GameEventQueue getEvents(MapMaster gameMap)
       if( transformEvents.size() > 0 ) // if we successfully made a move action
       {
         GameEvent moveEvent = transformEvents.peek();
-        if (moveEvent.getEndPoint().equals(waitLoc)) // make sure we shouldn't be pre-empted
+        if (moveEvent.getEndPoint().equals(getMoveLocation())) // make sure we shouldn't be pre-empted
         {
           transformEvents.add(new UnitTransformEvent(actor, type.destinationType));
         }
```
ThislsAUsername: I feel like it might make sense to just add a `shouldPreempt()` function to `MoveAction` and base some or all of the other action types off of it to minimize validation duplication.

src/Engine/GameActionSet.java
```diff
@@ -73,6 +73,8 @@ public String toString()
   {
     if( !actionOptions.isEmpty() )
     {
+      // NOTE: Ability and Unit Production actions return null for getUnitActionType(), since they are
+      // not unit actions. However, they are also never put into GameActionSets, so this is probably fine.
       return actionOptions.get(0).getUnitActionType().name();
```
ThislsAUsername: Good point. I hadn't thought of that.

src/AI/AIUtils.java
ThatIsAUsername: Then they had better not make assumptions. I think it's perfectly valid for `getUnitActionsByType` to return a map that doesn't contain entries for unsupported action types.

ThislsAUsername: Then I suppose you'd better fix it?
Or I guess I could?
*shrug*

src/Engine/GameAction.java
ThatIsAUsername: I've never been too worried about verbosity, though I'm fine with changing it. I'm also fine with deferring that change to avoid conflicts.

src/AI/AIUtils.java
ThatIsAUsername: I can see the convenience of guaranteeing that the standard action types will be represented even if the collection is empty, but that also makes it so you have to check for null under certain circumstances, but not others, and I don't like making the interface inconsistent.

Side note: If we do change it back, the second part of this function will break if Muriel tries to move a Meridian tank, because the call to `getAvailableUnitActions` will return a type that doesn't already exist in `actionsByType`.

src/Engine/GameAction.java
ThatIsAUsername: Hmmm, maybe. Out of scope for this PR though.

ThislsAUsername: Made an issue for it.

src/AI/AIUtils.java
ThislsAUsername: Yeah. That's basically what was happening with the save/load with AI bug we discussed.
I don't care too much which way we go, so long as we don't crash or anything.

PR #196, opened 2019-05-26
That requests

ThislsAUsername:
It ain't perfect, but the framework's there now.
Should I add the AI fix mentioned in `actionReworkMod`?

Gotta say, this looks a little weird: 
image


src/Engine/GameInstance.java
```diff
@@ -218,20 +222,43 @@ private String getSaveName()
     return sb.toString();
   }
   
+  public static boolean isSaveCompatible(String filename)
+  {
+    System.out.println(String.format("Checking compatibility of save %s", filename));
```
ThatIsAUsername: If it fails as incompatible, it would be good to print out the version of the save so we can go find the right DefendPeace version if we want.

ThislsAUsername: We do, when we catch the exception.

ThislsAUsername: ...on the other method. Fair point.

src/Engine/GameInstance.java
```diff
@@ -218,20 +222,43 @@ private String getSaveName()
...
     GameInstance load = null;
-    try
+    try (FileInputStream file = new FileInputStream(filename); ObjectInputStream in = new ObjectInputStream(file);)
     {
-      FileInputStream file = new FileInputStream(filename);
-      ObjectInputStream in = new ObjectInputStream(file);
+      verInfo = (GameVersion) in.readObject();
+      if( !new GameVersion().isEqual(verInfo) )
+      {
+        throw new Exception(String.format("Deserializing failed! Save is incompatible version: %s", verInfo.toString()));
```
ThatIsAUsername: Ah, you do print it here. However, we can't get here if the version doesn't match, because `isSaveCompatible` will prevent it from being added to the list.

src/Engine/GameVersion.java
```diff
@@ -0,0 +1,60 @@
...
+public class GameVersion implements Serializable
+{
+  private static final long serialVersionUID = 1L; // This class should always keep the same serialization format.
+  
+  int majorRev, minorRev, hotfix;
+  public GameVersion()
+  {
+    majorRev = 0;
+    minorRev = 1;
+    hotfix = 0;
```
ThatIsAUsername: I feel like we can probably claim credit for something greater than `0:1:0`. What feature set do we want to support to declare 1.0?

That also brings up the more general question of what constitutes a major/minor/hotfix change?
If `majorRev` is supposed to be like the differences between game sequels, then we are (probably) never gonna roll it past 1, and I wonder if it's worth recording.

src/UI/Art/SpriteArtist/SpriteCOSetupArtist.java
```diff
@@ -161,12 +161,14 @@ private static void drawCoInfo(Graphics g, COSetupInfo info, int xCenter, int yC
       g.fillRect(drawX - drawScale, drawY - drawScale, drawW + (2*drawScale), drawH + (2*drawScale));
 
       // draw the CO's color selection
-      BufferedImage colorFrame = SpriteUIUtils.makeTextFrame(c, c.darker(), UIUtils.getColorName(c), 2*drawScale, 2*drawScale);
+      String palette = UIUtils.getPaletteDisplayName(UIUtils.getPaletteName(c), info.getCurrentFaction().name);
+      BufferedImage colorFrame = SpriteUIUtils.makeTextFrame(c, c.darker(), palette, 2*drawScale, 2*drawScale);
```
ThatIsAUsername: Instead of calling `c.darker` repeatedly, you should just call it once and save/reuse it.

ThatIsAUsername: Actually, belay that. No need to change this file. Because reasons.

src/Engine/GameInstance.java
ThatIsAUsername: I think it might be better to check for save compatibility only when actually attempting to load it. It's going to cause confusion of a .svp just doesn't show up in the load menu, and there's no good indicator as to why.

We could even modify the display name in the menu (add a star/version number/whatever) so it's clear we see it, and that it won't work.

ThislsAUsername: Aight, I'll see about doing that.

src/Engine/GameVersion.java
ThislsAUsername: Hmm.
I suppose this also opens the possible can of worms about save compatibility; I think we can probably default everything's version ID and just let GameVersion take over logic to decide whether stuff's compatible.
Under that scheme, I'd propose `majorRev` be "substantial/breaking changes to game logic", while `minorRev` would be non-breaking/UI changes and `hotfix` would be quick bug fixes (they've happened a few times, though obviously I try to make sure they never happen).

For instance, adding a new CO that uses only existing features would be a `minorRev` change since the engine will be backwards-compatible, though saves with the new CO obviously wouldn't be.

ThislsAUsername: Forgot to note: if we go with my above scheme, I'd call it version 3.0.0, probably.

ThatIsAUsername: That all sounds pretty reasonable. Not sure how you get 3:0:0, but fine with me.
So would we like, prompt them if the versions differ, and let them try loading it anyway? Or just always try and notify on failure?

ThislsAUsername: My thinking was that we'd auto fail on `majorRev` difference, but try if there's a `minorRev` difference.
I was gonna say only try to load older saves, but I guess there's no real reason to require that other than "new content that just won't load anyway".

ThislsAUsername: Oh, and I got a `majorRev` of 3 from Jay's previously-imposed versioning. I could like, go back and count, but that sounds like work for no benefit.

ThatIsAUsername: Nah, mostly curious. I think he called the last one 1.2, and was gonna advance to 1.3? Whatever, I'm good with 3:0:0. Should we display the version number somewhere (maybe in the corner of the first menu, a la Dawn of War)? At least log it to the console on startup, I think.

ThislsAUsername: Yeah, he was. Naming it `1.3` goes back to your point of "is it really worth recording a number that will never increment?", though.

My intent was to put it on the title bar, but I forgot. Or maybe I was lazy.

...wait a second. Versioning brings up *another* weirdness. What of the test branch? My assumption was that it was the only one that would be getting "releases", but `master` will have the same versions and that... doesn't seem right, as they're incompatible due to a number of mechanical differences and a lot of added content.

ThatIsAUsername: It seems like the test branch will probably increment versions at about the same rate as master. We could just let them align, though that means master would choke on saves from the test branch.

Could also just rev the minor version for the test branch. That way they are always different, but it's clear they are related.

It might also be good to add a notional idea of the roll criteria for each number as a comment in the file.

src/Engine/GameInstance.java
```diff
@@ -255,7 +260,8 @@ public static GameInstance loadSave(String filename)
       verInfo = (GameVersion) in.readObject();
       if( !new GameVersion().isEqual(verInfo) )
       {
-        throw new Exception(String.format("Deserializing failed! Save is incompatible version: %s", verInfo.toString()));
+        throw new Exception(String.format("Deserializing failed! Save is incompatible version: %s",
+            (null == verInfo) ? "unknown" : verInfo.toString()));
```
ThatIsAUsername: The fact that you have to put this same logic in two places makes me think there's a better way to organize this. I guess this is so you can do the check both when looking for save files and when attempting to load them?

It's also a little odd to have a function named `isSaveCompatible()` and then check compatibility in `loadSave()` without calling that function.

src/Engine/GameVersion.java
```diff
@@ -12,11 +12,12 @@
...
   {
-    majorRev = 0;
-    minorRev = 1;
-    hotfix = 0;
+    majorRev = 3; // For substantial, save-breaking changes to the game logic.
+    minorRev = 0; // For added content or UI updates
+    hotfix   = 0; // For bugfixes (hopefully to never see increment)
   }
   
+  // TODO: Consider more nuanced validation here
```
ThatIsAUsername: We should probably sort that out before merging, since the point of this class is that it isn't supposed to change.

ThislsAUsername: Eh, not really. The class is stored as 3 ints, and the serial version is set.
Unless we actually change how many numbers we're keeping track of, we can do whatever we want without affecting serialization compatibility.

src/Engine/GameInstance.java
ThislsAUsername: ...hah, yeah.
It's a little ugly, for sure, but I don't really want to make a file stream an argument of the function.
I guess I could make overloads of isSaveCompatible, one of which takes a path, and calls the other with the stream.

...then again, that sounds like complicated control flow. Maybe the correct solution is to just delete that logic from the reading function. We'll already have marked the save as incompatible, and we're safe from exceptions, so maybe we should just try what the user asks.

ThatIsAUsername: Yeah, I'd be fine with that approach. If the versions don't match, print a warning, and just try to load it anyway. Catch the exception of course, so it's clear what happened, and they can pick a different one.

src/Engine/GameVersion.java
ThatIsAUsername: Sounds good then.

src/Engine/GameVersion.java
ThislsAUsername: Another possibility: `majorRev` is game branch (so `master` or `test`), `minorRev` would be breaking changes, `hotfix` non-breaking.
Could be argued any way, but I suppose it prooobably doesn't matter all that much right now since the only one being given to users is the test branch, at this point.

ThislsAUsername:
I imagine you should pull the lever on this one, since it's my PR?

Issue #159, opened 2019-01-26
Wally would like some brain enhancements


ThislsAUsername:
Muriel's comparative buying logic is a big one; without it, he's hosed on maps with air.
He also needs to prioritize targets for his N-hit KO algorithm. (Also, throw out single attacks, probably?)
Also, check for possible attacks after attempting to travel safely.
Finally, he definitely needs to stop thinking that using his Neo to hit an infantry with a Md Tank in range of it is "safe".

Wally also needs to understand supplies, and preferably also repairs.

PR #197, opened 2019-05-30
Fix wally

ThislsAUsername:
Squash'd the Wally branch stuff onto the tip of `master` to make my life easier.
Addressed all the review comments; working on making him less stupid at this point.

Since this branch is polluted with `master` changes, pointing the branch at Wally is infeasible.


src/Units/UnitModel.java
```diff
@@ -197,6 +197,26 @@ public boolean hasDirectFireWeapon()
     return hasDirect;
   }
 
+  /**
+   * @return True if this UnitModel has at least one weapon that cannot fire after moving.
+   */
+  public boolean hasImmobileWeapon()
+  {
+    boolean hasSiege = false;
+    if(weaponModels != null && weaponModels.size() > 0)
```
ThatIsAUsername: I still say the second conditional here is optional.

src/AI/AIUtils.java
```diff
@@ -249,6 +250,23 @@ public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Comman
     return shootableTiles;
   }
 
+  /**
+   * @return The range at which the CO in question might be able to attack after moving.
+   */
+  public static int findMaxStrikeWeaponRange(Commander co)
```
ThatIsAUsername: The only case I can think of where this wouldn't return 1 is with your proposed Bship rework... which we don't have.

src/AI/WallyAI.java
```diff
@@ -73,9 +77,40 @@ public WallyAI(Commander co)
...
+        // TODO: figure out how to count non-valid terrain
+        for( int w = 0; w < map.mapWidth; ++w )
+        {
+          for( int h = 0; h < map.mapHeight; ++h )
+          {
+            Environment terrain = map.getLocation(w, h).getEnvironment();
+            if( vroom.canTraverse(terrain) )
+            {
+              validTiles++;
+              totalCosts = vroom.getMoveCost(terrain);
```
ThatIsAUsername: Should this be a `+=`?

src/AI/WallyAI.java
```diff
@@ -97,11 +97,12 @@ private void init(GameMap map)
             if( vroom.canTraverse(terrain) )
             {
               validTiles++;
-              totalCosts = vroom.getMoveCost(terrain);
+              totalCosts += vroom.getMoveCost(terrain);
```
ThatIsAUsername: Ah! Yes it should.

src/AI/WallyAI.java
```diff
@@ -259,6 +260,8 @@ else if( unitQueue.isEmpty() )
           {
             for( Unit target : unitLists.get(co) )
             {
+              if (target.getHP() < 1) // Try not to pick fights with zombies
```
ThatIsAUsername: Does this happen often? Or ever?

src/AI/AIUtils.java
ThislsAUsername: Hey man, I like being prepared.
I suppose I could hardcode in `return 1` for now, but I want the function to exist for possible future shenanigans.

src/AI/WallyAI.java
ThislsAUsername: You can probably find the moment I discovered that, by my lamentations on Discord

src/AI/WallyAI.java
ThislsAUsername: It did. Caused an infinite loop.
It was due to the "AIs can sometimes do the same action twice in a row" bug, also fixed in this PR.
An Arty shot a Recon twice, the Recon sat at 0 HP, and Wally got into an infinite loop.

ThatIsAUsername:
There are a couple of important AI fixes here besides just adding Wallly. What else (besides resolving conflicts with master) do you intend to do on this branch?

ThislsAUsername:
Uhh, I guess it's not strictly necessary to do more, but I was angling to take on the rest of the stuff in issue #159.
I was knee-deep in my own version of adjusting expected threat by your own unit power when the rest of the PRs started getting wrapped up.
The rest of the stuff in there shouldn't be too earth-shattering, I just need to find time to get 'em done.
I think he could also benefit from considering whether an N-hit-KO is *wise*, but I haven't got a solid concept of how I'd want to do that.

Maybe I should just not bother with the rest of the stuff for now, and get to work on the test branch.

There's also some argument to be made for getting the meta-AI framework in before doing further work, once we figure out how we'd want it actually work (and, of course, whether we do in fact want it)

src/AI/WallyAI.java
```diff
@@ -190,6 +190,30 @@ else if( unitQueue.isEmpty() )
...
+        Unit unit = unitQueue.poll();
+        XYCoord position = new XYCoord(unit.x, unit.y);
+
+        if( unit.getCaptureProgress() > 0 )
+        {
+          actions.offer(new GameAction.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, position, gameMap)));
+        }
+        else
+        {
+          tempQueue.offer(unit);
```
ThatIsAUsername: Why not just do a peek, and then do a poll if you create the action instead of shoving them into a temp queue and back again?

src/AI/WallyAI.java
```diff
@@ -647,9 +649,12 @@ private double findAssaultKills(GameMap gameMap, Map<XYCoord, Unit> neededAttack
         {
           neededAttacks.put(xyc, unit);
           double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc.xCoord, xyc.yCoord).defenderHPLoss;
-          thisDamage = findAssaultKills(gameMap, neededAttacks, target, thisDamage);
+
+          if (thisDamage > target.getPreciseHP())
+            continue; // OHKOs should be decided using different logic
 
           log(String.format("  Use %s to deal %sHP?", unit.toStringWithLocation(), thisDamage));
+          thisDamage = findAssaultKills(gameMap, neededAttacks, target, thisDamage);
```
ThatIsAUsername: Won't the printed value of `thisDamage` be inaccurate immediately after it's printed, since you are reassigning the value here?

src/AI/WallyAI.java
```diff
@@ -48,8 +48,21 @@ public String getName()
     {
       return "Wally";
     }
+
+    @Override
+    public String getDescription()
+    {
+      return
+          "Wally thinks he's clever; he knows about artillery, and tries to keep his important units safe until he wants to engage.\n" +
+          "He's a little slow on the uptake, though.";
```
ThatIsAUsername: I feel like your class-level comment was a bit more informative; I adapted that when I did my combined branch.
          "Wally values units based on firepower and the area they can threaten.\n" +
          "He tries to keep units out of harm's way, and to protect expensive units with cheaper ones.";

Calling out his love for artillery is probably a good thing too, but "slow on the uptake" isn't going to mean much to anyone.

src/AI/WallyAI.java
```diff
@@ -144,17 +144,20 @@ public void initTurn(GameMap gameMap)
...
+    AIUtils.queueCromulentAbility(actions, myCo, CommanderAbility.PHASE_TURN_START);
+
     capturingProperties = new ArrayList<XYCoord>();
     for( Unit unit : myCo.units )
     {
       if( unit.getCaptureProgress() > 0 )
       {
         capturingProperties.add(unit.getCaptureTargetCoords());
+        XYCoord position = new XYCoord(unit.x, unit.y);
+        actions.offer(new GameAction.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, position, gameMap)));
```
ThatIsAUsername: I guess this also fixes the temp queue thing too.

src/AI/WallyAI.java
ThislsAUsername: Because I wasn't paying very good attention to the loop condition I copied.
My intent was to queue all such actions at once.

src/AI/WallyAI.java
ThislsAUsername: Yes, but it makes sense to print it here since that's the damage *this* unit can do, as opposed to the total so far.

src/AI/WallyAI.java
ThislsAUsername: Yeah, this is true. I was trying to avoid getting dug into the weeds, but I think you did a better job of also being informative.

ThislsAUsername: I do have a vague plan of how to fix the timidity problem: Set up the threat-area-check to just return a list of threats per square, rather than an aggregate threat value.
That'd let Wally say "oh, this enemy is the only threat here, so I can just shoot it and not worry about reprisal"

The ugly bit is that now you have to iterate weapons and check attack/move ranges for every threatened square you check for each unit, since you have no way of knowing which weapons will reach and how much damage they'll do.
I guess this is my due recompense for overengineering the weapon system, hah.

ThislsAUsername:
Man, I could almost swear Wally's losing to Muriel *worse* than before.

But he's making more smart *individual* choices so... IDK

ThislsAUsername:
Merging to help me stop pulling my hair out on the test branch.
Will leave the branch dangling in case of any desire for review

PR #199, opened 2019-05-31
Add a base UnitActionType so save/load doesn't break static associations

ThislsAUsername:


src/Engine/UnitActionType.java
```diff
@@ -25,9 +25,35 @@
...
+    @Override
+    public boolean equals(Object obj)
+    {
+      if( this == obj )
+        return true;
+      if( obj == null )
+        return false;
+      if( getClass() != obj.getClass() )
+        return false;
+      UnitActionType other = (UnitActionType) obj;
```
ThatIsAUsername: Probably worth checking that `null != other` as well, in case the cast fails.

ThatIsAUsername:
Pros: Simple to implement, works for current cases.
Cons: Potentially fragile. Will fail if anyone ever does a save/load and compares using `==`, and it wont' be obvious why.

src/Engine/UnitActionType.java
ThislsAUsername: Hmm.
We just checked that the classes are equal, and I'm casting to an interface that this class implements.
I can't see a case where that could happen.

ThatIsAUsername: Ah. Right. I blame being awake at 5 am.

PR #200, opened 2019-05-31
Add readResolve() to UnitActionType so the required static items are …

ThislsAUsername:
…used.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -59,7 +59,7 @@ public class Commander extends GameEventListener implements Serializable
   public GameMap myView;
   public ArrayList<Unit> units;
   public ArrayList<UnitModel> unitModels = new ArrayList<UnitModel>();
-  public transient Map<TerrainType, ArrayList<UnitModel>> unitProductionByTerrain;
+  public Map<TerrainType, ArrayList<UnitModel>> unitProductionByTerrain;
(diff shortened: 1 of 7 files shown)


ThatIsAUsername:
Pros: Also simple to implement, and guarantees that `==` will work even after save/load.
Cons: I don't think I see one, besides having to learn about `readResolve`, which is really a pro.

ThislsAUsername:
> Pros: Also simple to implement, and guarantees that `==` will work even after save/load.
> Cons: I don't think I see one, besides having to learn about `readResolve`, which is really a pro.

Hah, yes.
I think I may try to fix the terrains to use this same sort of logic so I can simplify their serialization as well, and I think it may as well be on this PR.
Thoughts?

ThatIsAUsername:
> I think I may try to fix the terrains to use this same sort of logic so I can simplify their serialization as well, and I think it may as well be on this PR.
Thoughts?

Sure, why not?

ThislsAUsername:
> Sure, why not?

Because I'm busy reviewing your code? Heh.

ThatIsAUsername:
> Because I'm busy reviewing your code? Heh.

Oh, cry to me.

ThislsAUsername:
> Oh, cry to me.

I totally am. Life is *hard* when I have to review code instead of writing it ;v

ThislsAUsername:
This is probably the next branch that should be merged.

PR #201, opened 2019-05-31
Game setup rework

ThatIsAUsername:
There are still some warts I plan to address, but it's all functional at this point.

Comments welcome as always.


src/CommandingOfficers/Commander.java
```diff
@@ -25,9 +25,7 @@
 import Terrain.GameMap;
 import Terrain.Location;
 import Terrain.TerrainType;
-import UI.UIUtils;
 import UI.UIUtils.Faction;
-import UI.Art.SpriteArtist.SpriteLibrary;
```
ThislsAUsername: Cleaning imports is a good thing, but it causes merge conflicts more often than not.
The codebase is full of serialization warnings anyway, so it doesn't even bother me to leave them.
...okay, maybe it does bother me. But merge conflicts bother me more.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,270 @@
...
+{
+  private static final Color MENUFRAMECOLOR = new Color(169, 118, 65);
+  private static final Color MENUBGCOLOR = new Color(234, 204, 154);
+  private static final Color MENUHIGHLIGHTCOLOR = new Color(246, 234, 210);
+
+  private static double animHighlightedPlayer = 0;
+
+  private static PlayerSetupController myControl = null;
+
+  private static HashMap<Integer, PlayerPanel> playerPanels = new HashMap<Integer, PlayerPanel>();
```
ThislsAUsername: I have to wonder: why not just an `ArrayList`?

ThislsAUsername:
Okay, wow. I just took this for a spin and it's pretty great.
I think maybe a grid view for the COs would be good as well, but I think having the full art is good too. I imagine we can just see what the users think, lol.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
```diff
@@ -100,10 +100,12 @@ public static void draw(Graphics g, IController controller, Color playerColor)
     myG.fillRect(infoX, infoY, infoW, infoH);
     myG.setColor(new Color(100, 115, 130));
     myG.drawRect(infoX, infoY, infoW, infoH);
-    BufferedImage infoText = SpriteUIUtils.drawTextToWidth(text, aiInfoZoneWidth - nameplateHeight*2);
-    myG.drawImage(infoText, infoX+infoBuffer, infoY+infoBuffer, null);
 
     // Draw the composed image to the window at scale.
     g.drawImage(image, 0, 0, dimensions.width, dimensions.height, null);
+
+    // TODO: Fix drawTextToWidth to not apply scaling automatically.
```
ThislsAUsername: I think this is one of those "hard to document design intent" things, hah.
I suppose it does make sense to only apply scaling when you're actually drawing on the graphics buffer to be shown on screen, though.

ThislsAUsername:
Bug report: if you choose colors on a map with 2 players, then back out and look at a map with more, the extra players will be black.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -43,6 +46,28 @@ public static double calculateSlideAmount(double currentNum, int targetNum)
...
+      coloredCursors = new HashMap<Color, Sprite>();
+    }
+    if( !coloredCursors.containsKey(color) )
+    {
+      Sprite newCursor = new Sprite(SpriteLibrary.getCursorSprites());
+      newCursor.colorize(UIUtils.defaultMapColors[4], color);
+      coloredCursors.put(color, newCursor);
+    }
+
+    // Draw the arrows around the focused player attribute.
```
ThislsAUsername: Is there any reason we can't use this logic for the map cursor?

src/UI/PlayerSetupAiController.java
```diff
@@ -0,0 +1,56 @@
...
+import AI.AIMaker;
+import Engine.IController;
+import Engine.OptionSelector;
+import UI.InputHandler.InputAction;
+
+public class PlayerSetupAiController implements IController
+{
+  ArrayList<AIMaker> aiMakers;
+  OptionSelector aiSelector;
+  private int startingIndex;
```
ThislsAUsername: Might make this final to clarify intent.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Hm. I think this was changed automatically because I committed a temp change in this class.

This branch is probably getting merged after all the others, which I guess means I'll have to fix any conflicts that come up anyway.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
ThatIsAUsername: Yeah, it's much simpler if I can draw in actual pixel-space, and not have to worry about whether a utility function is already scaling for me. Based on how much easier it was to draw these new screens with that philosophy, I want to extend it to all the other screens as well. Not all as part of this PR though, probably.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Uhhh. That's what my brain decided to type. ArrayList would work fine I think, though it would change the implementation slightly. Apparently I like maps.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: That's the plan. I think I want to reify the cursor and animate it a bit before taking that step though.

ThatIsAUsername:
> I think maybe a grid view for the COs would be good as well, but I think having the full art is good too. I imagine we can just see what the users think, lol.

I initially intended a more grid-esque view (along the lines of the image below), but a) that would require having a concept of canonical factions, and b) the linear approach was much faster to implement. Not to mention c) infinitely-scrolling menus are super cool.
image
This grid-like approach will probably be better long-term if we have lots of commanders, because grouping by faction will make it easier to sort through everyone.

src/UI/PlayerSetupCommanderController.java
```diff
@@ -0,0 +1,69 @@
...
+        break;
+      case SEEK:
+        CO_InfoController coInfoMenu = new CO_InfoController(cmdrInfos);
+        IView infoView = Driver.getInstance().gameGraphics.createInfoView(coInfoMenu);
+
+        // Get the info menu to select the current CO
+        for( int i = 0; i < cmdrSelector.getSelectionNormalized(); i++ )
+        {
+          coInfoMenu.handleInput(UI.InputHandler.InputAction.DOWN);
+        }
```
ThislsAUsername: I kind of expected you to read me the riot act over this hack, not re-use it.

ThislsAUsername: I don't really mind, it just amuses me.

ThatIsAUsername: Yeah, I was suppressing my urge to make things "right" just to get something functional quickly. I don't really intend to leave it like this.

ThislsAUsername:
> This grid-like approach will probably be better long-term if we have lots of commanders, because grouping by faction will make it easier to sort through everyone.

I suppose we could do a grid view *without* sorting by faction, but that... wouldn't really make it easier to find the CO you want.
It could be infinitely-scrolling anyway, though.

ThatIsAUsername:
> I suppose we could do a grid view without sorting by faction, but that... wouldn't really make it easier to find the CO you want.

Nope, that would probably be worse. Plus I'm actually pretty pleased with how it turned out as a linear list. Until/unless we get canonical factions sorted, I don't think I see a need to change it from this.

ThislsAUsername:
Another bug/change:
It's no longer possible to set yourself as "no team".
I guess it's not really *strictly necessary*, but I like the idea of an N/A option.
Also, showing the user "team 0" feels weird.

ThatIsAUsername:
> Another bug/change:
> It's no longer possible to set yourself as "no team".
> I guess it's not really _strictly necessary_, but I like the idea of an N/A option.
> Also, showing the user "team 0" feels weird.

I did that on purpose. Everyone is initialized to different teams to start with, so I don't see much need for an explicit N/A option.
Though I do sorta enjoy starting from zero, we can probably change it to display team numbers from 1.

ThislsAUsername:
> I did that on purpose. Everyone is initialized to different teams to start with, so I don't see much need for an explicit N/A option.

IDK, in a free-for-all or such it makes more sense to give everyone "no team" than to specify teams at all.
It's not really functionally different, though, I suppose.

> Though I do sorta enjoy starting from zero, we can probably change it to display team numbers from 1.

Or we could be cart-ish and do letters.

src/UI/PlayerSetupTeamController.java
```diff
@@ -0,0 +1,57 @@
...
+    switch(action)
+    {
+      case ENTER:
+      case BACK:
+        done = true; // Treat Enter/back the same.
+        break;
+      case UP:
+      case DOWN:
+        playerSelector.handleInput(action);
+        teamSelector.setSelectedOption(playerInfos[playerSelector.getSelectionNormalized()].getCurrentTeam());
```
ThislsAUsername: Hmm. It would make more sense *to me* to make an array of team selectors so you don't have to micromanage this one selector like that, but I guess you do you?

src/UI/PlayerSetupController.java
```diff
@@ -0,0 +1,208 @@
...
+  public boolean handleInput(InputAction action)
+  {
+    boolean exitMenu = false;
+    if(null != subMenu)
+    {
+      boolean exitSub = subMenu.handleInput(action);
+      if(exitSub)
+      {
+        // Only apply the change from the sub-menu if it exited via ENTER.
+        if(action == InputAction.ENTER)
```
ThislsAUsername: Hah, you don't even need to reset those controls on BACK, since you check it here.

src/UI/PlayerSetupController.java
```diff
@@ -0,0 +1,208 @@
...
+          if( category == SelectionCategories.TEAM.ordinal() )
+          { /* Do nothing - the sub-menu does the change on its own. */ }
+          if( category == SelectionCategories.AI.ordinal() )
+          {
+            PlayerSetupAiController aiMenu = (PlayerSetupAiController)subMenu;
+            PlayerSetupInfo info = getPlayerInfo(playerSelector.getSelectionNormalized());
+            info.currentAI.setSelectedOption(aiMenu.getSelectedAiIndex());
+          }
+        }
+        subMenu = null;
```
ThislsAUsername: Hmm. This could totally be a switch/case, but I think I like the idea of deleting it all, heh.

src/UI/PlayerSetupController.java
```diff
@@ -0,0 +1,208 @@
...
+            GameInstance newGame = new GameInstance(map);
+  
+            MapView mv = Driver.getInstance().gameGraphics.createMapView(newGame);
+            MapController mapController = new MapController(newGame, mv);
+  
+            // Mash the big red button and start the game.
+            Driver.getInstance().changeGameState(mapController, mv);
+          }
+          exitMenu = true;
+        }
```
ThislsAUsername: I think I like the idea of making this a switch/case.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+      // Figure out the longest-length name.
+      for( int i = 0; i < aiOptions.size(); ++i )
+      {
+        if( aiOptions.get(i).getName().length() > maxAiNameLength ) maxAiNameLength = aiOptions.get(i).getName().length();
+      }
+
+      // Make a string of all spaces of the required length so we can build nameplates of the same length. Soooo hacky, I know.
+      StringBuilder buf = new StringBuilder();
+      for( int n = 0; n < maxAiNameLength; ++n ) buf.append(" ");
```
ThislsAUsername: Why do you say this is hacky? You could do the same by messing with the hbuffer (which would center the names), but this is probably a lot easier and arguably as good or better.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
```diff
@@ -0,0 +1,111 @@
...
+        "I don't though, so all you get is this useless blurb.";
+
+    // Draw a nice frame and populate it with some useful info about the AI.
+    myG.setColor(new Color(255, 255, 255, 80));
+    myG.fillRect(infoX, infoY, infoW, infoH);
+    myG.setColor(new Color(100, 115, 130));
+    myG.drawRect(infoX, infoY, infoW, infoH);
+
+    // Draw the composed image to the window at scale.
+    g.drawImage(image, 0, 0, dimensions.width, dimensions.height, null);
```
ThislsAUsername: Why do you specify width and height? They should be equal to your current dimensions anyway.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -181,6 +205,45 @@ public static void drawMenuFrame(Graphics g, Color bg, Color frame, int x, int y
...
+      height = h;
+      mainColor = main;
+      rimColor = rim;
+      rimIsUp = rimUp;
+      content = SpriteLibrary.createDefaultBlankSprite(width, height);
+      Graphics graphics = content.getGraphics();
+      graphics.setColor(rimColor);
+      graphics.fillRect(0, 0, width, height);
+      int dx = 0, dy = 0;
+      if( rimIsUp ) dy++; else dx++;
```
ThislsAUsername: So, this produces *either* a left border *or* a top border, but not both? That's... just an interesting design choice.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -85,9 +110,8 @@ public static BufferedImage makeTextMenu(Color bg, Color frame, Color focus, Arr
       int hBuffer, int vBuffer)
   {
     // Find the dimensions of the menu we are drawing.
-    int drawScale = SpriteOptions.getDrawScale();
-    int menuTextWidth = SpriteLibrary.getLettersSmallCaps().getFrame(0).getWidth() * drawScale;
-    int menuTextHeight = SpriteLibrary.getLettersSmallCaps().getFrame(0).getHeight() * drawScale;
+    int menuTextWidth = SpriteLibrary.getLettersSmallCaps().getFrame(0).getWidth();
+    int menuTextHeight = SpriteLibrary.getLettersSmallCaps().getFrame(0).getHeight();
     int menuWidth = (( items.isEmpty() ) ? 0 : getMenuTextWidthPx(items, menuTextWidth)) + hBuffer * 2;
     int menuHeight = (( items.isEmpty() ) ? 0 : getMenuTextHeightPx(items, menuTextHeight)) + vBuffer * 2;
```
ThislsAUsername: `getMenuTextHeightPx()` uses the draw scale. Makes the menus unsightly.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,282 @@
...
+      }
+      PlayerSetupInfo info = myControl.getPlayerInfo(highlightedPlayer);
+      BufferedImage playerImage = panel.getImage();
+      int drawX = playerXCenter - (playerImage.getWidth())/2;
+      int drawY = imageHeight / 2 - playerImage.getHeight()/2;
+      SpriteUIUtils.drawCursor(myG, drawX+pane.xPos, drawY+pane.yPos, pane.width, pane.height, info.getCurrentColor());
+    }
+
+    // Finally, draw our rendered image onto the window.
+    g.drawImage(image, 0, 0, dimensions.width, dimensions.height, null);
```
ThislsAUsername: I'm not especially a fan of this, combined with `X = dimensions.X / drawScale`.
It works, sure, but it can distort stuff and it's pretty obvious as you resize the window.

ThislsAUsername: Distortion goes away if you replace it with:
`g.drawImage(image, 0, 0, imageWidth*drawScale, imageHeight*drawScale, null);`

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,282 @@
...
+
+    // Find where the zeroth player CO should be drawn.
+    // Shift from the center location by the spacing times the number of the highlighted option.
+    int playerYCenter = (int)(highlightedPlayerYCenter - (animHighlightedPlayer * panelHeight));
+
+    // Draw all of the player info.
+    for(int i = 0; i < numCOs; ++i, playerYCenter += (panelHeight))
+    {
+      // Only bother to draw it if it is on-screen.
+      if( (playerYCenter > -panelHeight/2) && ( playerYCenter < SpriteOptions.getScreenDimensions().getHeight()+(panelHeight/2) ) )
```
ThislsAUsername: Should be `imageHeight`, not screen dimensions

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,282 @@
...
+      {
+        PlayerSetupInfo playerInfo = myControl.getPlayerInfo(i);
+        Integer key = new Integer(i);
```
ThislsAUsername: Deprecated constructor, you can just use `= i;`

ThislsAUsername:
'nother bug: when you change color but *not* faction, the color name does not update (the color does update)

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,282 @@
...
+      coStrBuf.append("-").append(UIUtils.getPaletteName(info.getCurrentColor())).append(" ").append(info.getCurrentFaction().name);
+
+      boolean cmdrChanged = !info.getCurrentCO().name.equals(commanderName);
+      boolean colorChanged = !UIUtils.getPaletteName(info.getCurrentColor()).equals(colorName);
+      boolean factionChanged = !info.getCurrentFaction().name.equals(factionName);
+      boolean teamChanged = teamNumber != info.getCurrentTeam();
+      boolean aiChanged = !info.getCurrentAI().getName().equals(aiName);
+
+      Graphics g = myImage.getGraphics();
+      if( cmdrChanged || factionChanged )
```
ThislsAUsername: When you change color but not faction, the color name does not update (the color does update)

Also, seems odd to define `coStrBuf` up there, and the individual strings down here.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -623,6 +634,19 @@ else if( Character.isDigit(text.charAt(i)) )
...
+  /**
+   * Returns a BufferedImage containing the contents of `text` rendered on one line, on a transparent background, with no scaling applied.
+   */
+  public static BufferedImage getTextAsImage(String text)
+  {
+    Sprite letters = getLettersSmallCaps();
+    int width = letters.getFrame(0).getWidth() * text.length();
+    int height = letters.getFrame(0).getHeight();
+    BufferedImage textImage = createTransparentSprite(width, height);
+    drawTextSmallCaps(textImage.getGraphics(), text, 0, 0, 1);
```
ThislsAUsername: Why always `SmallCaps`? I feel like stuff is more readable with upper and lower-case letters.
Also in the case of CO names, the capitalization is important for knowing what to name their images, heh.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -0,0 +1,282 @@
...
+        teamPane = new SpriteUIUtils.ImageFrame(65, 23, 28, 23, MENUBGCOLOR, MENUHIGHLIGHTCOLOR, true,
+            SpriteLibrary.getMapUnitHPSprites().getFrame(info.getCurrentTeam()));
+        teamPane.render(g);
+      }
+      if( aiChanged )
+      {
+        aiName = info.getCurrentAI().getName();
+        aiPane = new SpriteUIUtils.ImageFrame(94, 23, 54, 23, MENUBGCOLOR, MENUHIGHLIGHTCOLOR, true, SpriteLibrary.getTextAsImage(info.getCurrentAI().getName()));
+        aiPane.render(g);
+      }
```
ThislsAUsername: Just a note, there are an awful lot of magic numbers here.

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
```diff
@@ -0,0 +1,116 @@
...
+
+    for( int c = 0; c < UIUtils.getCOColors().length; ++c)
+      for(int f = 0; f < UIUtils.getFactions().length; ++f)
+      {
+        // Figure out where this should be drawn.
+        int xOff = c*unitSizePx + unitBuffer*(c) + startX;
+        int yOff = f*unitSizePx + unitBuffer*(f) + startY;
+
+        // Only draw on-screen options.
+        if( (xOff < myWidth) && (yOff < myWidth) && (xOff > -unitSizePx) && (yOff > -unitSizePx))
```
ThislsAUsername: You need width and height, not width and width here.

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
```diff
@@ -0,0 +1,116 @@
...
+
+    // Start by assuming we can center the unit grid.
+    int startX = myWidth / 2 - gridWidth / 2;
+    int startY = myHeight / 2 - gridHeight / 2;
+
+    // Figure out if we need to realign things.
+    int maxX = myWidth - myWidth/3;
+    int maxY = myHeight - myHeight/3;
+    int minX = myWidth/3;
+    int minY = myHeight/3;
```
ThislsAUsername: It's... kinda weird to have a third of your screen potentially empty. I'm thinking shave that down to 1/8th or something, since it feels like a lot of wasted real estate.
Also, the buffers aren't evenly wide since you don't draw the units centered.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -0,0 +1,178 @@
...
+    // Calculate the vertical space each player panel will consume.
+    int panelBuffer = 3;
+    int panelHeight = CommanderPanel.PANEL_HEIGHT+panelBuffer;
+
+    // If we are moving from one option to another, calculate the intermediate draw location.
+//    if( animHighlightedPlayer != highlightedPlayer )
+//    {
+//      double slide = SpriteUIUtils.calculateSlideAmount(animHighlightedPlayer, highlightedPlayer);
+//      animHighlightedPlayer += slide;
+//    }
```
ThislsAUsername: No slide?

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+    // We don't actually want to draw something off the screen, so go forward one again.
+    coToDraw.handleInput(InputAction.DOWN);
+    drawYCenter += panelHeight;
+
+    // Draw all of the visible commander panels.
+    for(; drawYCenter - CommanderPanel.PANEL_HEIGHT/2 < myHeight ; coToDraw.handleInput(InputAction.DOWN), drawYCenter += (panelHeight))
+    {
+      // Only bother to draw it if it is onscreen.
+      if( (drawYCenter > -panelHeight/2) && ( drawYCenter < myHeight+(panelHeight/2) ) )
```
ThislsAUsername: This shouldn't be needed, due to the `for` condition and starting conditions.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -0,0 +1,178 @@
...
+        Integer key = new Integer(coToDraw.getSelectionNormalized());
+
+        // Get the relevant PlayerPanel.
+        if( !coPanels.containsKey(key) ) coPanels.put(key, new CommanderPanel(coInfo, playerColor));
+        CommanderPanel panel = coPanels.get(key);
+
+        // Update the PlayerPanel and render it to an image.
+        BufferedImage playerImage = panel.update(coInfo, playerColor);
+
+        int drawX = cursorSizePx * 2;
```
ThislsAUsername: Kind of a weird way to define spacing. Also it's the only usage of `cursorSizePx`.

ThislsAUsername:
Hokay. Finished reviewing all the jazz.
I still agree with my previously stated sentiment that these controls should all make use of the control/view stack in Driver. (If anything, I feel stronger about it having read the rest of the stuff)

ThislsAUsername:
Hokay. Finished reviewing all the jazz.
I still agree with my previously stated sentiment that these controls should all make use of the control/view stack in Driver. (If anything, I feel stronger about it having read the rest of the stuff)

ThislsAUsername:
That button is just way too easy to hit. Sorry.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Left and bottom, or top and right, yes. Lets me put a lighter top rim on dark panels and a darker bottom rim on lighter panels. Gives em that nice slightly-raised look.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: I think I mostly just did that since you used small caps for the current CO selection menu. No reason we can't allow both.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Yeah. I could defend that by saying it's a rigidly-defined format that isn't expected to change, and isn't any different than defining a spec for a binary data format, but really I was just trying to get a working prototype quickly so I could iterate from there.
I may go back and un-magicify things, unless I conveniently forget.

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
ThatIsAUsername: whups

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
ThatIsAUsername: Yeah, I initially had it top-left justified, but it felt too out-of-the way putting it that far off. Making it semi-centered felt a lot more natural, but this is definitely tweakable.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThatIsAUsername: I definitely intend to re-add sliding. Here and everywhere.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThatIsAUsername: Hm. I might have been using it more places initially. At any rate, it guarantees enough space to draw the cursor, plus an appropriate buffer. Made sense to me at the time.

src/UI/PlayerSetupCommanderController.java
ThatIsAUsername: Fixed to take starting index in constructor.

src/UI/PlayerSetupController.java
ThatIsAUsername: Deleted it all.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
ThatIsAUsername: Because `image` is at pixel-space size, and `dimensions` is at screen-space size. If I take that out it won't scale the image to match the screen.

src/UI/Art/SpriteArtist/SpriteMiniMapArtist.java
```diff
@@ -87,15 +102,29 @@ private static BufferedImage generateMapImage(MapInfo mapInfo)
...
+      {
+        colorsDiffer = true;
+        break;
+      }
+    }
+
+    return colorsDiffer;
+  }
+
+  private static boolean colorsDiffer(Color c1, Color c2)
```
ThislsAUsername: I could be wrong, but doesn't the `.equals()` method do this?

src/UI/PlayerSetupInfo.java
```diff
@@ -6,130 +6,74 @@
 import AI.AIMaker;
 import CommandingOfficers.Commander;
 import CommandingOfficers.CommanderInfo;
-import Engine.OptionSelector;
-import UI.InputHandler.InputAction;
 import UI.UIUtils.Faction;
 
 /**
- * Stores state to setup Commanders with during game setup.
+ * Stores a player's info with during game setup.
```
ThislsAUsername: "with"

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Hm. It does distort things a bit, doesn't it? I'll take your fix. I think technically that could leave us up to `drawScale` pixels shy of the window's edge, but we don't actually draw this stuff to the window's edge.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Yes

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Well, yeah. I think I did that in other places. My hands just kinda do what they want some times.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Scope narrowed.

src/UI/Art/SpriteArtist/SpriteMiniMapArtist.java
ThatIsAUsername: So it does.

src/UI/PlayerSetupInfo.java
ThatIsAUsername: sigh

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThislsAUsername: This is still an issue
image

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
ThislsAUsername: I still think I want like 1/7th or 1/8th or something in that ballpark. The margins feel a little too large still IMO.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Yes. Still working, man. Patience.

ThislsAUsername:
I'm not super duper thrilled with how the new SlidingValue works, but I'm not sure how to actually express why.

I think it's probably fair to say most of your lazy instantiation logic doesn't really need to be there, though; static assignment is perfectly fine and will reduce complexity.

ThatIsAUsername:
Really? It seems a lot cleaner than what I was doing before. I guess let me know if you figure out how to express it.

The lazy instantiation does feel awkward, but if I statically create it, it'll do an odd jump to the starting position the first time you get to that screen. I'd be open to suggestions though.

src/UI/Art/SpriteArtist/PlayerSetupTeamArtist.java
```diff
@@ -52,22 +55,23 @@ public static void draw(Graphics g, MapInfo mapInfo, IController controller)
...
-//      animHighlightedPlayer += slide;
-//    }
-
     // Find where the zeroth player CO should be drawn.
     // Shift from the center location by the spacing times the number of the highlighted option.
-    int playerYCenter = (int)(playerZoneYCenter - (highlightedPlayer * panelHeight));
+    int target = (int)(playerZoneYCenter - (highlightedPlayer * panelHeight));
+    if( null == player0YCenter ) player0YCenter = new SlidingValue(target);
+    long callTime = System.currentTimeMillis();
+    if( callTime - lastCallTime > 64 )
```
ThislsAUsername: Is there any reason you couldn't just sub in the "Is `controller` equal to the controller I had before?" logic that a lot of the other artists have?

ThislsAUsername:
> Really? It seems a lot cleaner than what I was doing before. I guess let me know if you figure out how to express it.

Hmm. I guess it just feels like it's storing more state than it needs to.
I think it could be pared down to:
Store start/end values and starting timestamp
`get()` becomes something like:
```
long td = System.currentTimeMillis() - startTime;
return startNum + (endNum - startNum) * Math.pow(0.7, td / UPDATE_DELAY_MS);
```

> The lazy instantiation does feel awkward, but if I statically create it, it'll do an odd jump to the starting position the first time you get to that screen. I'd be open to suggestions though.

When you check if the new controller is the same as the one you had before, just do a `snap()`.

Unless there's something I'm missing here, the stuff I've proposed here should work.
I admit I haven't tested it though, and contrary to popular belief, I am *not* actually a compiler.

src/UI/Art/SpriteArtist/PlayerSetupTeamArtist.java
ThatIsAUsername: I'm not finding examples of that right now, but I don't think it would work anyway. If you hit `back` then `down` then `enter`, the controller won't have changed, and you'll see the team panels snap into the new position anyway (assuming you increase 64 by an order of magnitude or so).

ThatIsAUsername:
>I think it could be pared down to:
Store start/end values and starting timestamp

Almost. You have to store the intermediate value as well, otherwise it looks weird when you hit up/down quickly. It'll start sliding one way, and if you hit the other way before it reaches the target it'll snap back instantly instead of sliding. You are right that the logic here could still be simplified some though.

ThislsAUsername:
> Almost. You have to store the intermediate value as well, otherwise it looks weird when you hit up/down quickly. It'll start sliding one way, and if you hit the other way before it reaches the target it'll snap back instantly instead of sliding. You are right that the logic here could still be simplified some though.

Nah, you can just set the `startValue` to `get()` in `set()`. That should work just fine.

src/UI/Art/SpriteArtist/PlayerSetupTeamArtist.java
ThatIsAUsername: Whoops, that is all false. Never mind. Yeah, that's doable.

ThatIsAUsername:
> > Almost. You have to store the intermediate value as well, otherwise it looks weird when you hit up/down quickly. It'll start sliding one way, and if you hit the other way before it reaches the target it'll snap back instantly instead of sliding. You are right that the logic here could still be simplified some though.
> 
> Nah, you can just set the `startValue` to `get()` in `set()`. That should work just fine.

I tried that. I think it caused some side effects, but maybe I had something else going on too. I can take another look.

src/UI/Art/SpriteArtist/PlayerSetupTeamArtist.java
ThislsAUsername: If you were hoping I would forget, I have not.
I'm definitely in favor of avoiding basing logic on some arbitrary number of milliseconds

src/UI/SlidingValue.java
```diff
@@ -9,84 +9,73 @@
...
-    return visualValue;
+
+    // Otherwise, figure out how much closer to get and move in.
+    if( currentValue != targetValue )
+    {
+      double slide = calculateSlideAmount();
+      currentValue += slide;
+    }
+    return currentValue;
   }
```
ThislsAUsername: Seeing as my implementation is simpler still than this simplification, I think I'd still like to see mine happen.

If you're dead set on having "snap to" logic (which isn't really needed IMO), just keep a bool to record whether you've arrived yet, and just return the destination without any other logic needed if it's true.

src/UI/Art/SpriteArtist/PlayerSetupTeamArtist.java
ThatIsAUsername: What? No. Why would I think you forgot? There's a whole comment chain right here to remind you.

Patience.

src/UI/SlidingValue.java
ThatIsAUsername: I like "snap to" because it prevents tons of unnecessary calculations and the single-pixel shift 1.5 seconds after you thought you were already there.

ThatIsAUsername:
All right, now's the time to let me know if I missed anything.

Which might be likely considering it's after 0300.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -44,32 +45,37 @@ public static void draw(Graphics g, IController controller, ArrayList<CommanderI
...
     }
-    // We don't actually want to draw something off the screen, so go forward one again.
-    coToDraw.next();
-    drawYCenter += panelHeight;
+    // We don't actually want to draw something off the screen, so go forward until we are on-screen again.
+    while( drawYCenter + CommanderPanel.PANEL_HEIGHT/2 < 0 )
+    {
+      coToDraw.next();
+      drawYCenter += panelHeight;
+    }
```
ThislsAUsername: Do these *both* need to be loops?

ThatIsAUsername: Yes, because I'm using the CO selector's absolute position, which means my original offset could be very negative.

I can't use the normalized position because then the whole list snaps back in the other direction when you go past the end instead of scrolling forever.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
```diff
@@ -11,19 +11,25 @@
...
 import UI.PlayerSetupAiController;
+import UI.UIUtils;
 
 public class PlayerSetupAiArtist
 {
   private static HashMap<Integer, BufferedImage> aiNameplates = new HashMap<Integer, BufferedImage>();
   private static int maxAiNameLength = -99;
 
+  private static IController myControl;
+  private static SpriteCursor spriteCursor = new SpriteCursor(0, 0, 0, 0, UIUtils.getCOColors()[0]);
```
ThislsAUsername: Might stand to reason to make this just the default constructor with no args.

src/UI/UIUtils.java
```diff
@@ -193,20 +192,15 @@ public static String getPaletteName(Color colorKey)
...
+      return "Red Shirts";
     if ("red".equalsIgnoreCase(palette) && "star".equalsIgnoreCase(faction))
-      return "Orange";
+      return "Orange Star";
     if ("maroon".equalsIgnoreCase(palette) && "fire".equalsIgnoreCase(faction))
-      return "Red";
-    return palette;
+      return "Red Fire";
+    return palette + ' ' + faction;
   }
```
ThislsAUsername: Still looks weird
image

ThislsAUsername:
> All right, now's the time to let me know if I missed anything.

Well, the cursor on the map's still an orange square.

src/AI/SpenderAI.java
```diff
@@ -39,6 +39,14 @@ public String getName()
     {
       return "Spender";
     }
+
+    @Override
+    public String getDescription()
+    {
+      return
+          "All Spender knows is that money in the bank doesn't help on the field, so it tries to spend all funds as quickly as possible.\n" +
+          "This can sometimes result in building units that are not useful.";
```
ThislsAUsername: I might replace "it" with "he", but this is more concise than mine would have been.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -209,7 +209,7 @@ public static int getMenuTextHeightPx(ArrayList<String> menuOptions, int scaledC
   {
     // Height of the letters plus 1 (for buffer between menu options), times the number of entries,
     // minus 1 because there is no buffer after the last entry.
-    return (scaledCharHeightPx + SpriteOptions.getDrawScale()) * menuOptions.size() - SpriteOptions.getDrawScale();
+    return (scaledCharHeightPx + 1) * menuOptions.size() - 1;
```
ThislsAUsername: This doesn't give quite enough height, I think.
image

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -63,6 +63,8 @@ public static void draw(Graphics g, MapInfo mapInfo, PlayerSetupController contr
         PlayerSetupAiArtist.draw(g, subMenu, control.getPlayerInfo(control.getHighlightedPlayer()).getCurrentColor());
       }
     }
+    // Start preloading infantry sprites in the background so the ColorFaction screen doesn't freeze on first entry.
+    PlayerSetupColorFactionArtist.preloadOneInfantrySprite();
```
ThislsAUsername: I feel like this is unorthodox, but it works.

src/UI/UIUtils.java
```diff
@@ -82,10 +82,10 @@ private static void initCosmetics()
 
       // Throw some color names in there for the defaults
       // toString() is not user-friendly
-      paletteNames.put(Color.PINK, "salmon");
+      paletteNames.put(Color.PINK, "rose");
       paletteNames.put(Color.CYAN, "cyan");
-      paletteNames.put(Color.ORANGE, "citrus");
-      paletteNames.put(PURPLE, "sparking");
+      paletteNames.put(Color.ORANGE, "orange");
```
ThislsAUsername: Oh, goody
image

ThislsAUsername:
Aight, Done reviewing the new changes. Looks good overall.

src/UI/Art/SpriteArtist/PlayerSetupAiArtist.java
ThatIsAUsername: Sensible.

src/AI/SpenderAI.java
ThatIsAUsername: Okedoke

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Yeah. But this way we get the same effect, and we don't have to mess with actual threading.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Yeah, I noticed that. Unfortunately, fixing that bleeds changes into... everything. I was semi-planning to make a separate PR to remove all the drawScale factors except when drawing to the window.

src/UI/UIUtils.java
ThatIsAUsername: You just love these mirror matchups, don't you?

ThatIsAUsername:
> Well, the cursor on the map's still an orange square.

Yeah, fixing that requires reworking `SpriteMapArtist` with the new "no scaling until we draw to the window" mindset (or allowing the cursor to draw with scaling, but I don't want to do that).

I want to call that out of scope for the current effort, and make a separate PR to purge the scale factors and fix the map cursor (and I'd rather not create more PRs until we merge the current mess).

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: What if I just extend the transparent sprite for the text by one line for now, and fix it when we properly address draw-scaling?

ThislsAUsername: If we're gonna be fixing the draw scale stuff soonishly (as in, before the next release), I think it's reasonable to just leave it as is.

src/UI/UIUtils.java
ThislsAUsername: I dunno, it's just really weird to me.

ThislsAUsername:
> I want to call that out of scope for the current effort, and make a separate PR to purge the scale factors and fix the map cursor (and I'd rather not create more PRs until we merge the current mess).

I'm on board with this.
I think I'm ready to push the big red button on this.

ThatIsAUsername:
> EDIT: Belay that. The text on the save select menu doesn't scale properly.

It does if you pull first.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
ThatIsAUsername: Depends on when the next release is maybe, but yeah, soon (tm) ishly.

ThislsAUsername:
> > EDIT: Belay that. The text on the save select menu doesn't scale properly.
> 
> It does if you pull first.

Foiled again.

PR #203, opened 2019-06-17
Cinder Mug

ThislsAUsername:
I'm a little worried the descriptive text is too verbose (check it out at default size), but Strong's is longer so I'm not *too* worried yet.

I used single quotes 'cause we don't display doublequotes yet.

I think I copied your art style well enough, lemme know if that ain't the case.


src/CommandingOfficers/CommanderCinder.java
```diff
@@ -29,8 +29,9 @@ public instantiator()
     {
       super("Cinder");
       infoPages.add(new InfoPage(
-          "Commander Cinder trades unit health for a better action economy.\n" +
-          "An efficient and merciless commander.\n"));
+          "Commander Cinder uses this callsign at all times. Her given name, according to her, is 'not important'.\n" + 
+          "'Cinder' is not a name, but designates time-warpers from Grey Sky's super-soldier program.\n" +
```
ThatIsAUsername: This is slightly confusing. It sounds like you are saying "Cinder" is both her specific call-sign, and also the generic designation for time-warpers.

It's also a little odd to emphasize time-warping - the name "Cinder" isn't obviously related to time-warping, and her gameplay doesn't have an obvious connection to time-travel either. If we want to make time-manipulation a primary aspect of her character, then it should be clear how that is reflected in her gameplay.

ThatIsAUsername:
> I'm a little worried the descriptive text is too verbose (check it out at default size), but Strong's is longer so I'm not _too_ worried yet.

Yeah, Strong's is too long - I want to shorten that up a bit. Maybe quite a bit.

> I think I copied your art style well enough, lemme know if that ain't the case.

Hah. "My" art style is just me trying poorly to approximate AW's art style.

With the recent UI changes, it's no longer necessary to make the portrait the full window size, so the first image could be made smaller.
She's also facing the wrong way - she's always going to be looking away from her opponents, which is odd.

src/CommandingOfficers/CommanderCinder.java
ThislsAUsername: > This is slightly confusing. It sounds like you are saying "Cinder" is both her specific call-sign, and also the generic designation for time-warpers.

Yes, that's exactly the case. Supposed to give off a "doesn't really care anymore" vibe.

> her gameplay doesn't have an obvious connection to time-travel

How exactly do you think she can get 1 unit to do 10 actions in a single turn?
Also relevant is that, in AW terms, she effectively builds her units the *turn before* she actually built them, since she gets to use them immediately and the industry is ready to build that turn.
The idea is that the injury is a side-effect of messing with time to expedite her troops.

ThislsAUsername:
> She's also facing the wrong way - she's always going to be looking away from her opponents, which is odd.

AW has peeps facing both ways, and Cinder in particular is supposed to be... sort of unnerving.

src/CommandingOfficers/CommanderCinder.java
ThatIsAUsername: > How exactly do you think she can get 1 unit to do 10 actions in a single turn?

I think most people would assume the answer to this question is "whips" before turning to time manipulation. I know your interpretation because you've told me before, but if you hadn't I might be wondering how time manipulation ties in.

ThislsAUsername: > I think most people would assume the answer to this question is "whips" before turning to time manipulation.

And this is why I call out time shenanigans in the fluff and make her ability names fire-themed.
If you have suggestions on how to make the linkage more clear, I'm all ears.

ThatIsAUsername: Hah. I made up words. It's even longer than what you had, and maybe you'll hate it.

We should probably figure out a sensible character limit for these descriptions.

```
'Cinder' is the designation of a member of Grey Sky's super-soldier program who can disrupt the flow of time to gain initiative in battle.
They are called 'cinders' due to the thermal blowback caused by temporal shifting that has ended many a soldier's career.
Having adopted this designation as her name, Commander Cinder continues to employ these dangerous tactics with the troops under her command.
```
Or I can shorten it a bit, so it's about the same length again:
```
'Cinders' are members of Grey Sky's super-soldier program who disrupt the flow of time to gain initiative in battle.
They are so-called due to the thermal blowback caused by their dangerous temporal shifting tactics.
Having taken this title as her name, Commander Cinder now employs these methods with the troops she commands.
```

ThislsAUsername: Agh, the code blocks make it hard to read since you gotta scroll back and forth.

How about:
'Cinder's are products of Grey Sky's super-soldier program who gain initiative in battle by warping time.
Unpredictable thermal surges result from their temporal meddling; thus the name.
Having taken this title as her name, Commander Cinder now employs these methods with the troops she commands.

ThislsAUsername: > We should probably figure out a sensible character limit for these descriptions.

I feel like it's more about the lines, with lines that will split at default length counting as 2. But yes.

ThatIsAUsername: I don't like putting the apostrophe before the 's' in `'Cinder's` - it looks possessive. 

For the second sentence, maybe:
"They are named for the unpredictable thermal surges that result from their temporal meddling."

For the third sentence, I think I prefer to keep the ability tie-in. the current second half feels a bit cliché, or maybe just underwhelming. "Almost" everyone? "Even" her enemies? Yawn.

ThislsAUsername: > For the third sentence, I think I prefer to keep the ability tie-in. the current second half feels a bit cliché, or maybe just underwhelming. "Almost" everyone? "Even" her enemies? Yawn.

I'm not about to claim that *Kanbei* fears any CO that's in the vague realm of balance. (And Flak should fear most everyone, but he's too stupid, so that's a second reason why you can't make *everyone* afraid.)

"Even" her enemies is supposed to be an ability tie-in, but perhaps a little too vague?

ThatIsAUsername: > I'm not about to claim...

Right. Which is why I just stick to facts, and let the user decide how menacing that is. Otherwise it ends up being either "she's the most scariest evar"--which is silly and overdone--or, "some people think she's scary... other people don't", which says approximately nothing.

> "Even" her enemies is supposed to be an ability tie-in, but perhaps a little too vague?

Too vague for me, at least, even after knowing it's supposed to be a tie-in.

ThislsAUsername: Okay, how about:
'Cinders' are products of Grey Sky's super-soldier program who gain initiative in battle by warping time - they're named for the unpredictable thermal surges resulting from their temporal meddling.
Having taken this title as her name, Commander Cinder brings fire and unpredictable strikes to the battlefield.

ThatIsAUsername: Fire - I know she's fire-themed and all, but the phrase "fire and unpredictable strikes" makes me expect an explicit in-game representation of fire in some way.
Maybe just
"... brings unpredictable strikes to the battlefield.", or even
"... is feared for her unpredictability on the battlefield."

Or just what you have is probs fine too, really.

I also prefer "they are named for the ..." over " ... thus the name", since the latter feels more awkward. Maybe even throw in "dangerous" alongside "unpredictable" or something to underscore that as the cause of the harm her soldiers take.
I'd also say "from" or "caused by" over "follow" (their temporal medding), but that's also nitpick-level stuff.

ThislsAUsername: > the phrase "fire and unpredictable strikes" makes me expect an explicit in-game representation of fire in some way.

There's nothing saying we can't make that her ability trigger animation.

'Cinders' are products of Grey Sky's super-soldier program who gain initiative in battle by warping time - they're named for the unpredictable thermal surges caused by their temporal meddling.
Having taken this title as her name, Commander Cinder's blazing speed dominates the battlefield.

ThatIsAUsername: Sounds good, ship it.

PR #210, opened 2019-06-28
Added option to disable in-game animations.

ThatIsAUsername:
Useful when watching AI battles.


src/UI/Art/SpriteArtist/SpriteOptions.java
```diff
@@ -276,19 +286,20 @@ public GraphicsOption(String name, int min, int max, int defaultValue)
       minOption = min;
       optionName = name;
       setSelectedOption(defaultValue);
-      storedValue = defaultValue;
```
ThislsAUsername: ???

src/UI/Art/SpriteArtist/SpriteOptions.java
```diff
@@ -299,6 +310,7 @@ public int getSelectionNormalized()
     public void setSelectedOption(int value)
     {
       super.setSelectedOption(value - minOption);
+      storedValue = getSelectionNormalized();
```
ThatIsAUsername: !!!

ThislsAUsername: "I see", said the blind man, as he took up his hammer and saw.

ThislsAUsername:
What if I'm a fan of the resupply animation but not so much the battle/move animations?

ThatIsAUsername:
> What if I'm a fan of the resupply animation but not so much the battle/move animations?

Then I guess you just have to figure out if you hate the battle/move animations more than you like the resupply animations, I guess? You want a three-state switch? Or a state-per-animation-type switch?

ThislsAUsername:
> > What if I'm a fan of the resupply animation but not so much the battle/move animations?
> 
> Then I guess you just have to figure out if you hate the battle/move animations more than you like the resupply animations, I guess? You want a three-state switch? Or a state-per-animation-type switch?

I don't think the carts let you turn off the resupply animation at all.
I can see how it'd be troublesome to make an option for each type of animation, though. I can think of resupply/repair, movement, combat, powers, capture, load/unload, transform, and maybe even custom animations for custom actions... yeah.

ThatIsAUsername:
> I can see how it'd be troublesome to make an option for each type of animation, though

I mean, we could create tiers of animations, I suppose. So at Tier X, any animation Tier X or less would be included. That way we wouldn't need an option per category. But splitting animations into categories / tiers would take some effort as well.
Not sure I want to mess with that at this point. We have few animations, and none that take long, so I'm fine with an all-or-none approach at this point.

ThatIsAUsername:
This commit was apparently included in #212.

PR #212, opened 2019-06-30
Game options

ThatIsAUsername:


src/UI/Art/SpriteArtist/MapArtist.java
```diff
@@ -45,19 +45,9 @@ public MapArtist(GameInstance game, MapView view)
...
-      HIGHLIGHT_COLOR = new Color(255, 255, 255, 160); // white
-    }
-    else
-    {
-      FOG_COLOR = new Color(255, 255, 255, 200); // white
-      HIGHLIGHT_COLOR = new Color(128, 128, 255, 112); // bluish
-    }
+    // Choose colors for fog and highlighted tiles.
+    FOG_COLOR = new Color(72, 72, 96, 200); // dark blue
+    HIGHLIGHT_COLOR = new Color(255, 255, 255, 160); // white
```
ThislsAUsername: Got a solid decision on that, did you?
Works for me.

ThatIsAUsername: White fog looks too much like snow.

src/UI/Art/SpriteArtist/SpriteOptions.java
```diff
@@ -81,7 +77,7 @@ private static void initialize()
     // Calculate the size of the longest item panel needed.
     for( int i = 0; i < allOptions.length; ++i )
     {
-      String[] allItems = allOptions[i].optionList;
+      String[] allItems = (String[])allOptions[i].optionList;
```
ThislsAUsername: I'm not a super fan of this cast.

ThatIsAUsername: I normally try to avoid casts, yeah. I'm considering what it would take to make `GameOption` generic, and hopefully avoid the need for casting everywhere.

src/UI/Art/SpriteArtist/MapArtist.java
ThislsAUsername: Did you check rain against the blue?

src/UI/GameOption.java
```diff
@@ -6,27 +6,33 @@
...
     optionName = name;
     setSelectedOption(defaultValue);
     optionList = Options;
   }
-  public GameOption(String name, int min, int max, int defaultValue)
+  public GameOption(String name, int min, int max, int stride, int defaultValue)
   {
-    super(max - min);
-    minOption = min;
+    super(1 + (max - min)/stride);
```
ThislsAUsername: I think we really need to document that `(max - min)/stride` must be integer.
Things could get... wacky.

src/UI/Art/SpriteArtist/MapArtist.java
ThatIsAUsername: Yeah. It's also not great, but it's not as bad as the white/snow, and eventually rain will be animated; the current approach is just a placeholder.

src/UI/Art/SpriteArtist/SpriteOptions.java
ThatIsAUsername: Done in new PR.

src/UI/GameOption.java
ThatIsAUsername: It's all integer division, so it can't not be an integer. If you put in 0 to 50k with a stride of 235, you'll get 213 values out of it, exactly as you would expect.

src/UI/GameOption.java
```diff
@@ -6,27 +6,33 @@
...
+    for( int i = 0; i < numOptions; i++ )
     {
-      optionList[i] = "" + (min + i);
+      int value = min + (i*stride);
+      System.out.println(String.format("Storing %s at index %s", value, i));
+      optionList[i] = "" + value;
+      if( value == defaultValue )
+      {
+        setSelectedOption(i);
+      }
```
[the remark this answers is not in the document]
ThatIsAUsername: Not sure where the start of this sentence is.

src/UI/GameOption.java
```diff
@@ -6,27 +6,33 @@
...
   {
-    super(max - min);
-    minOption = min;
+    super(1 + (max - min)/stride);
+    minOption = 0;
     optionName = name;
-    setSelectedOption(defaultValue);
-    optionList = new String[1 + max - min];
-    for( int i = 0; i <= max - min; i++ )
+    int numOptions = 1 + (max - min) / stride;
```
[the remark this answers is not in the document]
ThatIsAUsername: Yes, but that only works after the constructor has been executed. Which it won't have been, in the new PR.

src/UI/GameOptionSetupController.java
```diff
@@ -0,0 +1,107 @@
...
+import Engine.IController;
+import Engine.OptionSelector;
+import Terrain.Environment.Weathers;
+import UI.InputHandler.InputAction;
+
+public class GameOptionSetupController implements IController
+{
+  private GameOption fowOption = new GameOption("Fog of War", false);
+  private GameOption startingFundsOption = new GameOption("Starting Funds", 0, 50000, 1000, 0);
+  private GameOption incomeOption = new GameOption("Income", 500, 20000, 500, 1000);
```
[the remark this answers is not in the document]
ThatIsAUsername: ehh, why not

src/CommandingOfficers/Commander.java
```diff
@@ -135,7 +133,7 @@ public Commander(CommanderInfo info)
     modifiers = new ArrayList<COModifier>();
     units = new ArrayList<Unit>();
     ownedProperties = new HashSet<XYCoord>();
-    money = DEFAULTSTARTINGMONEY;
+    money = 0;
```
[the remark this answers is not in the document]
ThatIsAUsername: OK

src/UI/GameOption.java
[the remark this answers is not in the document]
ThislsAUsername: Oh, that was the thing about (max - min) should be divisible by `stride`.
I suppose it doesn't really need documentation; you can get weird results, but it shouldn't crash and these things should be fairly obvious.

src/UI/GameBuilder.java
```diff
@@ -23,4 +30,21 @@ public void addCO(Commander co)
...
+    GameInstance newGame = null;
+
+    // Build the CO list and the new map and create the game instance.
+    Commander[] cos = commanders.toArray(new Commander[commanders.size()]);
+    MapMaster map = new MapMaster( cos, mapInfo );
+    if( map.initOK() )
+    {
+      newGame = new GameInstance(map, isFowEnabled, defaultWeather, startingFunds);
+
+      for(Commander co: commanders) co.incomePerCity = incomePerCity;
```
ThislsAUsername: On second inspection, I'm realizing this violates how I had it in my head that COs should be set up.
My thought was that starting funds, income, ability energy exchange rates, and similar should be set at the time the CO is constructed, and not changed after.
Then you can put D2D effects in the constructor of a given CO and not worry about anybody else overwriting you.

ThislsAUsername: My initial idea was that those things would be stored statically, set during game init, and not changed... but that is a rather non-obvious way to design constructor arguments.
My counterproposal: GameBuilder gets the list of COInfos, and becomes an argument to the CO's constructor so game settings can be pulled in during init of the base CO.

ThatIsAUsername: I'm not sure I like the idea of passing `GameBuilder` to `Commander` since that creates a new dependency for no tangible benefit.
If we really want these CO attributes to be set at construction and not changed after, we would have to make them `private` (or at least `protected`) and probably `final` as well.
Frankly, I'd be more inclined to make `incomePerCity` an attribute of `GameInstance` instead (or of a `Scenario` class that I think we will want eventually), since income is set per-game (not per-CO) anyway.

ThislsAUsername: > I'm not sure I like the idea of passing `GameBuilder` to `Commander` since that creates a new dependency for no tangible benefit.

In fact, these changes already break Bear&Bull.

> If we really want these CO attributes to be set at construction and not changed after, we would have to make them `private` (or at least `protected`) and probably `final` as well.

`protected` is fine with me (Though it will break our UI. We'll need a getter). `final` is not.
Bear&Bull set their income and unit costs to 90% in the constructor, but GameBuilder overwrites the income with what everyone else gets... except they *also* keep track of the original income they had (from the constructor), so they'll go back to 1k-based income as soon as they use an ability.
I suppose we could move to having an `init` function to call after the GameBuilder sets up the variables, but that's a code smell if the peeps on StackOverflow and such are to be believed.

> Frankly, I'd be more inclined to make `incomePerCity` an attribute of `GameInstance` instead (or of a `Scenario` class that I think we will want eventually), since income is set per-game (not per-CO) anyway.

And then COs could freely play with an `incomeAdjustment` property? That should work just fine... with the caveat that now the COs need access to that object at construction (Bear&Bull), turn init (everyone, unless we move the income logic), and events (Patch).

PR #213, opened 2019-06-30
GameOptions is now generic.

ThatIsAUsername:
New PR to avoid muddying the water in the first PR.


src/UI/Art/SpriteArtist/SpriteOptions.java
```diff
@@ -24,9 +27,9 @@
   private static Dimension dimensions = new Dimension(WINDOWWIDTH_DEFAULT * drawScale, WINDOWHEIGHT_DEFAULT * drawScale);
 
   // Set up configurable options.
-  private static GameOption drawScaleOption = new GameOption("Draw Scale", 1, 6, 1, DRAWSCALE_DEFAULT);
-  private static GameOption animationsOption = new GameOption("Animations", true);
-  private static GameOption damageSystemOption = new GameOption("Damage System", Weapon.stratDescriptions, 0);
+  private static GameOption<Integer> drawScaleOption = new GameOptionInt("Draw Scale", 1, 6, 1, DRAWSCALE_DEFAULT);
+  private static GameOptionBool animationsOption = new GameOptionBool("Animations", true);
```
ThislsAUsername: Mild inconsistency here

ThislsAUsername:
I can't help but think we're missing a file or two here, defining `GameOptionInt` and `GameOptionBool` constructors....

src/UI/Art/SpriteArtist/SpriteOptions.java
ThatIsAUsername: Yay flexibility!

ThatIsAUsername:
> 
> 
> I can't help but think we're missing a file or two here, defining `GameOptionInt` and `GameOptionBool` constructors....

Whoops. Fixed.

src/UI/Art/SpriteArtist/SpriteOptions.java
ThislsAUsername: Yay flexibility?

src/UI/GameOptionBool.java
```diff
@@ -0,0 +1,17 @@
...
+{
+  public GameOptionBool(String name, boolean defaultValue)
+  {
+    super(name, new Boolean[] { false, true }, (defaultValue ? 1 : 0));
+  }
+
+  @Override
+  public String getCurrentValueText()
+  {
+    boolean value = optionList.get(getSelectionNormalized()-minOption);
```
ThislsAUsername: The `minOption` feels unnecessary here, seeing as you control what that is (to always be zero)

ThatIsAUsername: True. Removed.

src/UI/Art/SpriteArtist/SpriteOptions.java
ThatIsAUsername: `GameOptionInt` is the same as `GameOption<Integer>`; use whichever you fancy.

I guess I can make it uniform if you really want.

src/Engine/Utils.java
```diff
@@ -103,6 +102,12 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
   {
     ArrayList<XYCoord> reachableTiles = new ArrayList<XYCoord>();
 
+    if( null == unit || unit.x < 0 || unit.y < 0 )
+    {
+      System.out.println("WARNING! Finding destinations for ineligible unit!");
+      return reachableTiles;
+    }
```
ThislsAUsername: ...oops

src/Engine/GameAction.java
```diff
@@ -99,7 +99,6 @@ public GameEventQueue getEvents(MapMaster gameMap)
         isValid &= (gameMap.getLocation(attackLocation).getResident() == defender);
         isValid &= (null != defender) && attacker.canAttack(defender.model, attackRange, moved);
         isValid &= (null != defender) && attacker.CO.isEnemy(defender.CO);
-        isValid &= (null == moveLocation.getResident()) || (attacker == moveLocation.getResident());
```
ThislsAUsername: Infinite recursion?

ThatIsAUsername: Yeah. I hit a case multiple times with Muriel vs Spender in FoW where each one would be trying to perform an attack through FoW, but the move tiles were both hidden. This check failed, so the action was cancelled. If just one AI hits the case it's OK(ish); it just ends the turn (though it should ambush the unit instead). If both hit the case, they ping-pong back and forth without generating any actual events to execute. Execution goes from `MapController.animationEnded` to `SpriteMapView.animate` and back again repeatedly, with occasional trips to `MapController.startNextTurn` and back to `animate` until it runs out of stack.

ThatIsAUsername: I'm OK just removing this check, because this is exactly what `Utils.enqueueMoveEvent` handles.

ThislsAUsername: Reading this again, yes, the check is actually wrong and makes the program not work correctly.
I believe it would also prevent a human player from attempting to attack from a square that an enemy is on.
Might be worth throwing checks for this in the fog test/making sure this isn't checked in other events..

src/Engine/Utils.java
ThislsAUsername: Y'know, I feel like this is probably triggered by the same bug I fixed on Wally, with the double-actions.

Should we account for issues that appear when the AI goes on the fritz somehow, or just make sure the AI doesn't do weird stuff? Or both?

src/CommandingOfficers/Bear_Bull.java
```diff
@@ -83,7 +84,7 @@ public void swapD2Ds(boolean setIncome)
     {
       isBull = true;
       if( setIncome )
-        incomePerCity = (int) (incomeBase * BULL_MOD);
+        incomeAdjustment = (int) (incomeBase * BULL_MOD) - gameRules.incomePerCity;
```
ThislsAUsername: I woulda just made the mod variables -0.1 and +0.2, then did just 1 + mod on the unit cost, but this is fine too.

ThatIsAUsername: Hm. Didn't occur to me. Ah well, too late now.

src/AI/Muriel.java
```diff
@@ -586,7 +586,7 @@ private void queueUnitProductionActions(GameMap gameMap)
 
       // If we have a lot of cash on hand, don't worry about cost effectiveness - just maximize damage instead.
       // If we ever collect more than twice our income in funds, we just aren't spending fast enough. Fix that.
-      int incomePerTurn = myCo.ownedProperties.size() * myCo.incomePerCity;
+      int incomePerTurn = myCo.ownedProperties.size() * myCo.gameRules.incomePerCity;
```
ThislsAUsername: I think we want some sort of "true income" getter for this, because this is just a bug breeding ground

src/Engine/GameInstance.java
```diff
@@ -198,7 +201,11 @@ public GameEventQueue turn()
 
     // Set the cursor to the new CO's last known cursor position.
     setCursorLocation(playerCursors.get(activeCoNum).xCoord, playerCursors.get(activeCoNum).yCoord);
-    
+
+    // Handle income and any other scenario-specific events.
+    events.addAll(gameScenario.initTurn(gameMap));
```
ThislsAUsername: Does `initTurn()` need the current CO or something, so it can calculate and dispense income?
Then again, maybe it's the better part of valour to not move stuff around too much yet?

ThatIsAUsername: Might put that there eventually. Not yet though.

src/CommandingOfficers/Commander.java
```diff
@@ -428,4 +417,19 @@ private void readObject(ObjectInputStream stream) throws IOException, ClassNotFo
...
+   */
+  public int getIncomePerTurn()
+  {
+    int count = 0;
+    for( XYCoord coord : ownedProperties )
+    {
+      // Re-check ownership just because.
+      if( myView.getLocation(coord).getOwner() == this && myView.getLocation(coord).isProfitable() ) ++count;
+    }
+    return (count * gameRules.incomePerCity) + incomeAdjustment;
```
ThislsAUsername: These parentheses are on the wrong pair of values.

ThislsAUsername:
I'm happy with merging everything here (minus the parentheses thing), and we may as well use squash since I haven't based anything on it.

PR #215, opened 2019-07-21
Refactor boilerplate COModifier code into a superclass

ThislsAUsername:
Seems like a good idea to get rid of duplicated logic

### DIFF
--- a/src/CommandingOfficers/Modifiers/CODamageModifier.java
+++ b/src/CommandingOfficers/Modifiers/CODamageModifier.java
@@ -3,11 +3,11 @@ package CommandingOfficers.Modifiers;
 import java.util.ArrayList;
 
 import CommandingOfficers.Commander;
+import CommandingOfficers.Modifiers.COModifier.GenericCOModifier;
 import Units.UnitModel;
 
-public class CODamageModifier implements COModifier
+public class CODamageModifier extends GenericCOModifier
 {
-  ArrayList<UnitModel> modelsToModify;
   private int attackModifier = 0;
 
   /**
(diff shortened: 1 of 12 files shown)


src/CommandingOfficers/Modifiers/CODamageModifier.java
```diff
@@ -20,36 +20,12 @@
...
-    else
-    {
-      System.out.println("Attempting to add null model to COMovementModifier!");
-      throw new NullPointerException(); // Make sure this oversight doesn't go unnoticed.
-    }
   }
 
   @Override
-  public void apply(Commander commander)
+  public void applyChanges(Commander commander, ArrayList<UnitModel> models)
```
ThatIsAUsername: `applyChanges` and `revertChanges` are `protected` in `GenericCOModifier`. I'm assuming you meant for them to be protected here too.

As an aside, I would almost rather have the public-facing method be named `applyChanges`, and make `apply` the internal-only one. Small change, but it might make it more clear just that it's not attaching the modifier to the Commander itself, and so the changes won't be undone automatically.

ThatIsAUsername:
It's a little strange to implement `apply` for some modifiers and `applyChanges` for others. Since `GenericCOModifier` just manages the set of units to change, maybe rename it (and the apply function) to reflect that. Like `GenericCoUnitModifier` and `applyUnitChanges` or whatever, and then it'll be clear they go together. Making `apply` and `revert` does help with that already though, so it may be fine as is.

I think this looks pretty good overall. Feel free to apply my suggestions or not and merge.

PR #217, opened 2019-07-22
Add fancy units

ThislsAUsername:
I dunno if you/we actually want all of this jazz, but I figured I'd make this based on `master` so we can take it if we like it.


src/Engine/UnitActionType.java
```diff
@@ -410,7 +412,7 @@ public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor,
       XYCoord moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
       if( ignoreResident || map.isLocationEmpty(actor, moveLocation) )
       {
-        return new GameActionSet(new GameAction.ExplodeAction(actor, movePath, this), false);
+        return new GameActionSet(new GameAction.ExplodeAction(actor, movePath, this), true); // We don't really need a target, but I want a confirm dialogue
```
ThislsAUsername: This feels like a bit of a hack.
I'm thinking we probably want to move whether an action is targeted to the UnitActionType itself?
I almost did it, but then there was too many ripples and I'm lazy.

src/Engine/GameAction.java
```diff
@@ -828,6 +831,100 @@ public UnitActionType getUnitActionType()
...
+        Location moveLocation = gameMap.getLocation(moveCoord);
+        isValid &= (null == moveLocation.getResident()) || (benefactor == moveLocation.getResident());
+      }
+
+      if( isValid )
+      {
+        if( Utils.enqueueMoveEvent(gameMap, benefactor, movePath, repairEvents) )
+        {
+          // No surprises in the fog.
+          repairEvents.add(new HealUnitEvent(beneficiary, true));
```
ThatIsAUsername: Hah, I've had a `UnitHealEvent` locally for a while, but it never got added (I think because you objected to separating funds transfer from healing). Ah well.

src/Engine/GameAction.java
```diff
@@ -828,6 +831,100 @@ public UnitActionType getUnitActionType()
...
+    @Override
+    public XYCoord getTargetLocation()
+    {
+      return repairCoord;
+    }
+
+    @Override
+    public String toString()
+    {
+      return String.format("[Heal at %s after moving to %s]",
```
ThatIsAUsername: Howsabout "Heal %s at %s after moving to %s"?

src/Engine/GameAction.java
```diff
@@ -828,6 +831,100 @@ public UnitActionType getUnitActionType()
...
+      Unit beneficiary = null;
+
+      if( (null != gameMap) && (null != startCoord) && (null != repairCoord) &&
+          gameMap.isLocationValid(startCoord) && gameMap.isLocationValid(repairCoord) )
+      {
+        benefactor = gameMap.getLocation(startCoord).getResident();
+        beneficiary = gameMap.getLocation(repairCoord).getResident();
+        isValid &= benefactor != null && !benefactor.isTurnOver;
+        isValid &= (null != beneficiary) && !benefactor.CO.isEnemy(beneficiary.CO);
+        isValid &= (null != gameMap) && (gameMap.isLocationValid(repairCoord));
```
ThatIsAUsername: Shouldn't you make sure the location is valid before getting the resident on line 875? The `null == gameMap` check is redundant, and the other should probably also go in the `if` condition.

src/Engine/GameAction.java
```diff
@@ -828,6 +831,100 @@ public UnitActionType getUnitActionType()
...
+      }
+    }
+
+    @Override
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
+      // Repair actions consist of
+      //   MOVE
+      //   HEAL
+      //   REPAIR
```
ThatIsAUsername: HEAL and REPAIR?

src/Engine/GameEvents/HealUnitEvent.java
```diff
@@ -0,0 +1,73 @@
...
+import Engine.XYCoord;
+import Terrain.MapMaster;
+import UI.MapView;
+import UI.Art.Animation.GameAnimation;
+import Units.Unit;
+
+public class HealUnitEvent implements GameEvent
+{
+  private Unit unit;
+  public final int repairPower = 1;
```
ThatIsAUsername: Oh. So you aren't using this for turn-init healing. Are you planning to?

src/Engine/GameAction.java
ThislsAUsername: Is there a good reason to separate them? Seems like it would add complexity for little gain.

src/Engine/GameAction.java
ThislsAUsername: bah, shoulda been heal then resupply

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: I built the logic to handle arbitrary healing, but I wasn't gonna rewrangle everything in a branch that was possibly throwaway.

ThislsAUsername: So in a word, yes, now that I know that the branch is liable to be merged.

src/Engine/GameAction.java
ThislsAUsername: The entire line is redundant, actually. The other is already in the `if`.
At least this copy-paste just wastes cycles, haha

src/Engine/GameAction.java
ThislsAUsername: "Heal unit at %s from %s after moving to %s"?

ThislsAUsername: The interacting units aren't available for... *reasons*.

src/Engine/GameEvents/MassDamageEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    for (Unit victim : victims.keySet())
+    {
+      int starting = victim.getHP();
+      if( lethal && damage > starting )
+      {
+        victim.damageHP(-damage);
+        victims.put(victim, starting);
+
+        // Guess he's not gonna make it.
+        // TODO: Is there a better way to do this?
```
ThatIsAUsername: This kind of logic (determining whether to create a given event) should probably be done at the `GameAction.getEvents()` level, not on the fly within a different event's execution.

It seems like it should be pretty straightforward to loop over the victims in `ExplodeAction` to see which ones will be killed, and adding `UnitDieEvent` for each. Maybe only use the `MassDamageEvent` for the ones that don't. I'd consider calling it just `UnitDamageEvent` for consistency and generality.

src/Engine/UnitActionType.java
```diff
@@ -17,20 +17,24 @@ public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor)
   }
   public abstract GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor, boolean ignoreResident);
   public abstract String name();
+  public boolean shouldConfirm = false;
```
ThatIsAUsername: Public?

src/Engine/UnitActionType.java
```diff
@@ -303,6 +307,55 @@ private Object readResolve()
...
+      {
+        ArrayList<GameAction> repairOptions = new ArrayList<GameAction>();
+        ArrayList<XYCoord> locations = Utils.findLocationsInRange(map, moveLocation, 1, 1);
+
+        // For each location, see if there is a friendly unit to repair.
+        for( XYCoord loc : locations )
+        {
+          // If there's a friendly unit there who isn't us, we can repair them.
+          Unit other = map.getLocation(loc).getResident();
+          if( other != null && other.CO == actor.CO && other != actor &&
```
ThatIsAUsername: Should we allow repairing allies?

src/Units/CarrierModel.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private static final MoveType moveType = new FloatHeavy();
+  private static final UnitActionType[] actions = UnitActionType.COMBAT_VEHICLE_ACTIONS;
+  private static final WeaponModel[] weapons = { new CarrierMissiles() };
+
+  public CarrierModel()
+  {
+    super("Carrier", UnitEnum.CARRIER, ChassisEnum.SHIP, UNIT_COST, MAX_FUEL, IDLE_FUEL_BURN, VISION_RANGE, MOVE_POWER, moveType, actions, weapons);
+    holdingCapacity = 2;
+    UnitEnum[] carryable = { UnitEnum.B_COPTER, UnitEnum.T_COPTER, UnitEnum.FIGHTER, UnitEnum.BOMBER, UnitEnum.STEALTH, UnitEnum.BBOMB };
```
ThatIsAUsername: Hm. Probs out of scope for this, but maybe carryables should be defined in terms of chassis, rather than explicit unit types?

src/Units/PiperunnerModel.java
```diff
@@ -0,0 +1,27 @@
...
+  private static final int MAX_FUEL = 99;
+  private static final int IDLE_FUEL_BURN = 0;
+  private static final int VISION_RANGE = 4;
+  private static final int MOVE_POWER = 9;
+
+  private static final MoveType moveType = new MoveType();
+  private static final UnitActionType[] actions = UnitActionType.COMBAT_VEHICLE_ACTIONS;
+  private static final WeaponModel[] weapons = { new PipeGun() };
+
+  public PiperunnerModel()
```
ThatIsAUsername: Still not a fan of Piperunners. If we are gonna add them, we might wanna rethink the "pipe" sprite we have too.

src/Units/Unit.java
```diff
@@ -211,7 +211,7 @@ public double alterHP(int change)
     double before = HP;
     // Change the unit's health, but don't grant more
     // than 10 HP, and don't drop HP to zero.
-    HP = Math.max(1, Math.min(10, getHP() + change));
```
ThatIsAUsername: Map damage is now up to 9% more effective?

src/Units/Weapons/PipeGun.java
```diff
@@ -0,0 +1,11 @@
+package Units.Weapons;
+
+public class PipeGun extends WeaponModel
+{
+
+  public PipeGun()
+  {
+    super(WeaponType.PIPEGUN, 9, 2, 5);
```
ThatIsAUsername: Instead of all these magic numbers, it's easier for future maintainers if you just declare some class-level variables, and pass those into `super`. That way I know what '9' means without going spelunking.

src/Units/Weapons/StealthPews.java
```diff
@@ -0,0 +1,11 @@
+package Units.Weapons;
+
+public class StealthPews extends WeaponModel
```
ThatIsAUsername: ... Really?

src/Units/Weapons/WeaponModel.java
```diff
@@ -54,29 +58,35 @@ public int getIndex()
...
-/* CRUISERTORPEDOES */{   0,    0,     0,    0,       0,       0,   0,         0,       0,        0,         0,       0,      0,        0,        0,          0,       0,      0,  90,      90 }, // CRUISERTORPEDOES
-/* CRUISERMGUN      */{   0,    0,     0,    0,       0,       0,   0,         0,       0,        0,         0,      55,     65,      115,      115,          0,       0,      0,   0,       0 }, // CRUISERMGUN
-/* SUBTORPEDOES     */{   0,    0,     0,    0,       0,       0,   0,         0,       0,        0,         0,       0,      0,        0,        0,         55,      25,     95,  55,      55 }  // SUBTORPEDOES
+// defenders:      INFANTRY, MECH, RECON, TANK, MD_TANK, NEOTANK,           APC, ARTILLERY, ROCKETS,             ANTI_AIR, MOBILESAM, FIGHTER, BOMBER,                        B_COPTER, T_COPTER,                        BATTLESHIP, CRUISER, LANDER, SUB, SUB_SUB
+// defenders:      INFANTRY, MECH, RECON, TANK, MD_TANK, NEOTANK, MEGATANK, APC, ARTILLERY, ROCKETS, PIPERUNNER, ANTI_AIR, MOBILESAM, FIGHTER, BOMBER, STEALTH, STEALTH_HIDE, B_COPTER, T_COPTER, BBOMB, CARRIER, BBOAT, BATTLESHIP, CRUISER, LANDER, SUB, SUB_SUB
```
ThatIsAUsername: Did you mean to (mostly) double up the "defenders" line here?

ThatIsAUsername: Did you mean to (mostly) double up the "defenders" line here?

src/Engine/UnitActionType.java
ThatIsAUsername: That sounds sensible in the 10 seconds I've considered it so far. Out of scope for this PR though, yes.

src/Engine/GameAction.java
ThatIsAUsername: If we combine them, I think `UnitRepairEvent` makes more sense than `Heal`. Separating them could make sense because it allows them to happen independently, and allows tracking/listening to cases where one happens but not the other.

src/Engine/GameAction.java
ThatIsAUsername: "That thing that will never happen"?

Currently, Attack, Capture, Wait, Load, Unload, Join, Resupply, Transform, Explode, and Delete track the interacting unit(s). So, every action type except this one and Production.

"Heal Tank at (3, 4) after moving to (3, 3)" seems better than
"Heal at (3, 4) after moving to (3, 3)" or
"Heal unit at (3, 4) from (3, 3) after moving to (3, 3)"

src/Engine/GameAction.java
ThislsAUsername: With `Heal`, I was following your convention with `CreateUnitEvent`. I don't think of "this costs money" with either "heal" or "create" unit, but adding the option in the constructor would make it plenty clear.

I suppose we could make a "MoneyChange(Delta?)Event" that would give or remove cash from a player, but it feels like that's the only use case where there's just money changes happening and nothing else. Except it could be a "TurnInitEvent" just as well.

As for tracking/listening, I think it makes more sense to add whether there was a money cost in the published values. One more optional if-check is less added complexity than at least a new class and definitely at least one more function.
Plus, splitting them into separate events means you no longer know for sure *why* the money is changing.

src/Engine/GameAction.java
ThislsAUsername: Netplay.
It's occurred to me since that it's desirable even if we never get netplay, for replays.
We could just store the starting gamestate and play back the GameActions one by one, with an "AI" that just polls the queue of available actions.

I'm thinking we don't want to store the units, because the units will be instantiated on the fly and will have different IDs than the serialized events... Unless maybe the build action kept a reference to the unit it will produce? (Seems hokey to me)

That logging is nicer, yes. We could probably compromise on it and make the unit references transient... but then the thing you're logging might end up uninitialized after deserialization.


For reference, I think replays are the next feature on my general radar, since that'll be really nice for both "spectators" and the actual players to know what's going on.

I've been thinking the "AI that just pipes out GameActions in the right order" is the way to go here, but I've been looking at some [interesting](https://randomascii.wordpress.com/2013/07/16/floating-point-determinism/) [articles](http://yosefk.com/blog/consistency-how-to-defeat-the-purpose-of-ieee-floating-point.html) on this sort of "lockstep" behavior, and it appears floating-point math is liable to mess everything up without a lot of work. Perhaps just move to using integer percents everywhere?

Another approach would be to store the events themselves. I didn't like that idea because the events just can't get away from storing the interacting units, but it probably wouldn't be a problem if, say, the Create event actually kept a reference to the unit it already built and only ever instantiated it once.
This bears a certain amount of cost in that we'll have to make sure all of the events daisy-chain properly after serializing/deserializing.
This *also* has two advantages:
There's no new "AI" to build, you just load the events in the queue and off you go (assuming we can trigger "end turn" from there)
The lack of floating point determinism shouldn't be an issue, if we avoid events-spawning-events, since unit death will be 100% determined by "is there a death event in the queue?". This probably works if we just always require insertion into the event queue before execution and save events off as they're executed.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: Fixed in the `mod` branch

src/Engine/GameEvents/MassDamageEvent.java
ThislsAUsername: This was made under the (highly mistaken) impression that `GameAction.getEvents()` uses a generic `GameMap` rather than a `MapMaster`. I'd suggest we make this easier to remember somehow, but I don't think you can fix stupid.

It would be straightforward, for sure.
I think the `MassDamageEvent` should still apply to the killed units, since the point of publishing the event is to notify listeners of who was affected by the damage.

...I guess I can't argue too hard with `UnitDamageEvent`, but I put the "Mass" in there to specify that it's affecting more than one unit at a time.

src/Engine/UnitActionType.java
ThislsAUsername: The plan was vaguely to pull the "Is this targeted" dealie out of `GameActionSet`, but I didn't end up doing that.

Ideally also it'd be `final`, but that would mean copying it into every constructor, heh.

src/Engine/UnitActionType.java
ThislsAUsername: I think *probably* what we should do is allow allies to repair themselves on your property. I think that's how DoR does it (might be wrong though)

src/Units/CarrierModel.java
ThislsAUsername: The thought did cross my mind.
The only time I can think of that'd be different from what we have now is with Piperunners not being able to load into Landers, but who even cares since they can't normally interact :P 
But yes, probably out of scope.

src/Units/PiperunnerModel.java
ThislsAUsername: Hah, perhaps.
We could add some train-tracks terrain and let them move on those. ~~and maybe store units like the Pain Train from SFW can~~

It occurs to me that Piperunners also conflict with the idea of making pipeseams just a "unit" you place on the map.
Maybe locations can/should have a "this unit" sort of variable that they pass out instead of null when calling `getResident()`? Except *that* might mean you can't actually stand on them... This probably calls for more functions with more clear and specific use cases.

src/Units/Unit.java
ThislsAUsername: Matches the carts.

src/Units/Weapons/PipeGun.java
ThislsAUsername: Hmm. That does sound nice, but it's also a lot of typing I don't much feel like doing.

src/Units/Weapons/StealthPews.java
ThislsAUsername: *YES*

src/Units/Weapons/WeaponModel.java
ThislsAUsername: ...That was a guide for adding the new rows. Which I then forgot to remove.
That knowledge might help you review the chart changes, if you care to?

Planning to delete in the `mod` branch.

src/Engine/GameEvents/MassDamageEvent.java
ThislsAUsername: ...On that note, I guess I ought to add [defeat] to both the damage event and the delete event, since they can both kill the last surviving unit.

ThatIsAUsername: I could see how `MassDamageEvent` might be more clear in the CO event listener, since I think battle-damage charge is normally handled elsewhere. Maybe we can consolidate those and reduce ambiguity even more?

ThislsAUsername: I don't think we can consolidate these. The charge for combat takes into account both funds damage dealt and taken (with different multipliers) and the amount of HP removed from the enemy you attacked.
I guess you could just use the same listening function but with only the "attacked" unit filled out, but that would break a lot of assumptions and be really ugly IMO.

src/Units/PiperunnerModel.java
ThatIsAUsername: I don't think I'd make pipe seams a unit. I mean, we haven't addressed destructible terrain yet, but that doesn't seem like the first avenue I'd go down.

src/Units/Weapons/StealthPews.java
ThatIsAUsername: _ugh_

src/Engine/GameAction.java
ThatIsAUsername: All right. Well. I haven't properly ruminated on all this yet, but my gut reflex is that we do it one way for 90% of cases, so we should probably follow suit in this one. Any change of this sort is gonna be large anyway, and it will be simple to change this one with all the others when we figure out how to do it.
I'd rather have the more informative message for the time in between now and whenever we decide and figure out how to change things in 2025.

src/Engine/GameAction.java
ThatIsAUsername: Hmm. Yeah, with CO powers that spawn units for no cost, we may have to tweak `CreateUnitEvent` or make a new one. Ah well, out of scope for this PR.
I don't object to this approach, I was just amused that you also created a healing event.

src/Engine/GameAction.java
ThatIsAUsername: Also at one point I wanted to figure out how to make our AI deterministic, but I didn't go to far because then all the AI matches would be fixed. Still not sure why it isn't deterministic - maybe the natural ordering of units in the queue varies from run to run or something. Have to figure that out if we want replays to be a thing.

src/Units/PiperunnerModel.java
ThislsAUsername: Mm. Wargroove made destructible terrain a unit, and it's the easiest naive approach to the problem.
Though yeah, destructible terrain being its own thing is probably more desirable (especially if we want to do stuff like the fancy devices)
That'd require adding a whole new "system" and modifying our targeting logic, so I'm not super enthused... but honestly it'll probably be less complicated than I think.

src/Engine/GameAction.java
ThislsAUsername: > All right. Well. I haven't properly ruminated on all this yet, but my gut reflex is that we do it one way for 90% of cases, so we should probably follow suit in this one. Any change of this sort is gonna be large anyway, and it will be simple to change this one with all the others when we figure out how to do it.

That's probably fair.

> I'd rather have the more informative message for the time in between now and whenever we decide and figure out how to change things in 2025.

I did say replays was my next big feature goal (though destructible terrain may take precedence, we'll see).
I suppose it makes sense to use the current scheme, since it's nicer right now and we're not sure which way the replays should be set up yet anyhow.

> Also at one point I wanted to figure out how to make our AI deterministic, but I didn't go to far because then all the AI matches would be fixed.

IDK, it'd make debugging issues with them a lot easier.

> Still not sure why it isn't deterministic - maybe the natural ordering of units in the queue varies from run to run or something.

Pretty sure that's it exactly. The coordinate and/or location and/or unit IDs won't all be the same from run to run, so builds and units aren't considered in exactly the same order each time. To fix it, we'd need to make sure to use deterministic-order collections.
...which may or may not be annoying to do, depending on how ties are broken. Could maybe just have an "Age" var on everything to do that, with instantiation order determining it and being deterministic?

> Have to figure that out if we want replays to be a thing (unless we just store all actions directly instead of re-running the AI, of course).

Just storing the actions is exactly what I had in mind. IMO it's totally valid to use RNG in the AI, since the AI's calculations should only ever be done once for any given game.
(In my vague ideal end-state, there'd be some way to add new AI at run-time to allow stuff like the Starcraft AI tournaments, so I really don't want to constrain AI behavior.)

ThatIsAUsername: Yeah, I'm not gonna try to bite off plugin-based AI modules just yet.

Still want to change this action message though, even if it's "temporary".

src/Engine/GameEvents/MassDamageEvent.java
ThatIsAUsername: Yeah, I agree that `MassDamageEvent` should apply to killed units.

Which you call the event is up to you, I think either works well.

Yes, mass damage/delete should both have game-ending potential.

Yeah, I guess we don't have to consolidate the damage stuff.

src/Engine/UnitActionType.java
ThatIsAUsername: Yes. In DoR, you repair units on allies' properties at the start of your turn, just like your own.

Not sure why we wouldn't allow allies to heal each other too (though in that case, the one doing the action should pay, unlike the passive turn-init healing), though I'm not sure if AWDS allows that.

src/Engine/UnitActionType.java
ThatIsAUsername: This works well enough for now, anyway.

src/Units/PiperunnerModel.java
ThatIsAUsername: I like the idea of trains at some point, maybe. I was considering making pipes an electric/plasma fence, which would look cool and be appropriate for the job, but y'know, time.

In general though, if you want something to fill a new kind of role, it should be a new kind of thing. The League of Legends folks have an in-joke whenever unexpected behavior occurs that "it's coded as minions" - things like player-generated walls are literally just a row of minions with different graphics, and don't always act as much like a wall as you'd want.

ThatIsAUsername:
Just to consolidate:
- [ ] Improve heal action message
- [ ] Mass Damage / Unit Deletion should be able to trigger defeat
- [ ] Delete the extra commented header line in WeaponModel
- [ ] Seriously consider the fact that "StealthPews" is just like, the worst

ThislsAUsername:
>     * [x]  Mass Damage / Unit Deletion should be able to trigger defeat

Done, but also technically unnecessary and adds a warning to suppress.

src/Engine/GameAction.java
```diff
@@ -1029,6 +1030,7 @@ public ExplodeAction(Unit unit, Path path, UnitActionType.Explode pType)
       actor = unit;
     }
 
+    @SuppressWarnings("unused") // Unused code has been tested and may be useful at some point
```
ThatIsAUsername: Eclipse is complaining that `ExplodeAction.getEvents()` is unused?

ThatIsAUsername: Ah. Never mind. It's because `isLethal` is set to false and never changed anywhere ever.
I normally just wouldn't include it until we had a use case for lethal exploding things, and avoid adding suppression flags.

ThislsAUsername: Should I delete the code, then?

ThatIsAUsername: I probably would, yeah. It'll be simple enough to add again if we ever support that. Until then, it just looks like a YAGNI violation.

ThatIsAUsername:
Shouldn't we also remove `isLethal` from the `Explode` `UnitActionType`, since it never changes (and could cause confusion/bugs later, being public and all)?

PR #218, opened 2019-07-23
Fancy units mod

ThislsAUsername:
I did some mildly(?) sketchy stuff here, so I made a separate PR for GPs

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import AI.AIController;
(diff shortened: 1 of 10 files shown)


src/CommandingOfficers/Commander.java
```diff
@@ -226,7 +228,13 @@ public GameEventQueue initTurn(GameMap map)
       aiController.initTurn(myView);
     }
 
-    return new GameEventQueue();
+    GameEventQueue events = new GameEventQueue();
```
ThatIsAUsername: Why move unit init here?

One difference (which may or may not matter) is that units will now register events all at once instead of sequentially.

ThislsAUsername: Makes more sense, to me, to have it here, and it added unnecessary complexity to MapController IMO.

That difference would matter if the UnitHealEvent pre-calculated anything, but it doesn't, so it doesn't.

ThislsAUsername: I guess it would also matter if someone listened to UnitHealEvents and published events afterwards. That'd be weird though.

ThislsAUsername:
...that "revert" button is also too easy to press

PR #220, opened 2019-08-06
Let me see yer eye(d)s

ThislsAUsername:
Look ma, no warnings

...aside from the fact that I apparently don't have a 1.7 JRE. *That* might explain a few things...

### DIFF
--- a/src/Engine/GameAction.java
+++ b/src/Engine/GameAction.java
@@ -94,7 +94,7 @@ public interface GameAction
       isValid &= (movePath != null) && (movePath.getPathLength() > 0);
       if( isValid )
       {
-        Location moveLocation = gameMap.getLocation(moveCoord);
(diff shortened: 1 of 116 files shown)


src/Engine/GameAction.java
```diff
@@ -94,7 +94,7 @@ public GameEventQueue getEvents(MapMaster gameMap)
       isValid &= (movePath != null) && (movePath.getPathLength() > 0);
       if( isValid )
       {
-        Location moveLocation = gameMap.getLocation(moveCoord);
+        isValid &= gameMap.isLocationValid(moveCoord);
```
ThatIsAUsername: This should probably move up before the `if` statement.

ThatIsAUsername:
Sweet. I this looks mergeable then. Did you say you want to not squash this one? Eh, I'll just let you do it.

PR #223, opened 2020-01-13
Complaint magnet

ThislsAUsername:
This is probably already unreasonably big, and it's not even properly done yet.
Big items:
Consolidated unit and weapon definitions to a single file each (per scheme).
Added the ability to have multiple entire sets of units.
Removed the Weapon class, since its only relevant state was ammo count.
Added the DoR unit roster (missing Carrier's Launch and Rig's temp-building jazz)

I think I want some feedback on what I've done so far before I try adding something as complicated as Launch, or the weird edge-case deal that Temp Ports have (they block non-friendly non-transports)


src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+    awbwModels.unitModels.add(subsub);
+    UnitModel sneaky = new StealthHideModel();
+    stealth.possibleActions.add(new UnitActionType.Transform(sneaky, "HIDE"));
+    sneaky.possibleActions.add(new UnitActionType.Transform(stealth, "APPEAR"));
+    awbwModels.unitModels.add(sneaky);
+
+    return awbwModels;
+  }
+
+  public enum AWBWUnitEnum
```
ThatIsAUsername: Wasn't one of your goals to ditch `UnitEnum`? Now commanders don't have a way to find a specific unit type, and `UnitEnum` is duplicated per unit set. Which is maybe unavoidable if we want multiple unit sets, but hey.
It looks like it's only used for indexing into the damage chart though - if only we had a way to avoid needing that...

src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+    for (UnitModel um : factoryModels)
+      awbwModels.unitModels.add(um);
+    for (UnitModel um : seaportModels)
+      awbwModels.unitModels.add(um);
+    for (UnitModel um : airportModels)
+      awbwModels.unitModels.add(um);
+
+    // Handle transforming units separately, since we don't want two buy-entries
+    UnitModel subsub = new SubSubModel();
+    sub.possibleActions.add(new UnitActionType.Transform(subsub, "DIVE"));
```
ThatIsAUsername: Why are you adding unit-specific actions here instead of in the `UnitModel` class itself?

src/Test/TestCOModifier.java
```diff
@@ -101,24 +101,24 @@ private boolean testProductionModifier()
     ArrayList<UnitModel> factoryModels = patch.unitProductionByTerrain.get(FC);
     
     // Define a type we want to add, and verify it isn't already buildable.
-    UnitModel bship = patch.getUnitModel(UnitModel.UnitEnum.BATTLESHIP);
-    testPassed &= validate( !factoryModels.contains(bship), "    Factory can build Battleship by default. Malformed test!" );
+    UnitModel ship = patch.unitProductionByTerrain.get(TerrainType.SEAPORT).get(3);
```
ThatIsAUsername: Ahhh, magic numbers

ThatIsAUsername: Also this is significantly more unwieldy than just asking for a CO's Battleship model, even if we replace the magic number with an enum or whatever.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -314,24 +314,24 @@ public static SpriteSetKey instance(TerrainType terrain, Commander co)
...
     }
 
-    public static UnitSpriteSetKey instance(UnitModel.UnitEnum unitType, Faction faction, Color color)
+    public static UnitSpriteSetKey instance(String unitType, Faction faction, Color color)
     {
       UnitSpriteSetKey key = null;
       for( int i = 0; i < instances.size(); ++i )
       {
-        if( instances.get(i).unitTypeKey == unitType && instances.get(i).factionKey == faction && instances.get(i).colorKey == color)
+        if( instances.get(i).unitTypeKey.equalsIgnoreCase(unitType) && instances.get(i).factionKey == faction && instances.get(i).colorKey == color)
```
ThatIsAUsername: Do you not need run `unitType` through `standardizeID` to ensure we are comparing equal things?

src/Units/UnitModelScheme.java
```diff
@@ -0,0 +1,62 @@
...
+  public abstract GameReadyModels getGameReadyModels();
+
+  // Holds the data for a single Commander's unit selection.
+  public static class GameReadyModels
+  {
+    public HashMap<TerrainType, ArrayList<UnitModel>> shoppingList = new HashMap<TerrainType, ArrayList<UnitModel>>();
+    public ArrayList<UnitModel> unitModels = new ArrayList<UnitModel>();
+  }
+  
+  public static UnitModel getModelFromString(String pName, ArrayList<UnitModel> models)
```
ThatIsAUsername: If we are really going to use Strings for the canonical ID/name of units, we might consider throwing an exception if we fail to match it to a model.
Granted, if we get a `null` model back, it'll probably cause an exception later on, but may as well fail faster.

src/Units/UnitModelScheme.java
```diff
@@ -0,0 +1,62 @@
...
+  {
+    String name = UnitModel.standardizeID(pName);
+    UnitModel model = null;
+    ArrayList<String> coreRoles = UnitModel.getUnitRoleIDs();
+    if (coreRoles.contains(name))
```
ThatIsAUsername: Instead of calling `contains` and `indexOf`, I'd probably just call `indexOf` and compare the return to `-1` to avoid iterating the collection twice. On the other hand, this isn't performance-critical code, and this is arguably more idiomatic.

src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+import Units.MoveTypes.Tread;
+
+public class AWBWUnits extends UnitModelScheme
+{
+  private static final long serialVersionUID = 1L;
+
+  @Override
+  public String toString()
+  {
+    return "AWBW" + (schemeValid ? "" : "!");
```
ThatIsAUsername: I might prepend the '!' so it looks less like punctuation. As an added bonus, '!' means "not" when prepended in most programming languages.

src/Units/MoveTypes/TiresRugged.java
```diff
@@ -0,0 +1,23 @@
...
+{
+  private static final long serialVersionUID = 1L;
+
+  public TiresRugged()
+  {
+    super();
+    
+    setMoveCost(Weathers.CLEAR, TerrainType.GRASS, 1);
+
+    setMoveCost(Weathers.RAIN, TerrainType.GRASS, 3); // I get to make stuff up, so why not?
```
ThatIsAUsername: `TiresRugged` is made up?

src/Units/UnitModel.java
```diff
@@ -6,88 +6,115 @@
...
     TROOP, TANK, AIR_LOW, AIR_HIGH, SHIP, SUBMERGED
   };
 
   public String name;
-  public UnitEnum type;
+  public UnitRoleEnum role;
   public ChassisEnum chassis;
   private int moneyCost = 9001;
   public int moneyCostAdjustment = 0;
+  public double customStarValue = 0;
```
ThatIsAUsername: This is set but never used anywhere. Also, DefendPeace doesn't have any notion of "stars" - you must be thinking of a completely different, entirely unrelated game.

src/Units/UnitModelScheme.java
```diff
@@ -0,0 +1,62 @@
...
+  private static final long serialVersionUID = 1L;
+  
+  /** Indicates whether, for the given map/game, all the units can be instantiated */
+  public boolean schemeValid = true;
+
+  /**
+   * When the UI wants to show a unit type, it should use this string.
+   * Typically the name of whatever passes for infantry, but can be anything.
+   */
+  public abstract String getIconicUnitName();
```
ThatIsAUsername: This function is overridden, but never actually used anywhere.

src/Engine/GameAction.java
```diff
@@ -1106,4 +1110,128 @@ public XYCoord getTargetLocation()
       return destination;
     }
   } // ~UnitDeleteAction
+
+  public static class UnitProduceAction implements GameAction
```
ThatIsAUsername: Missing class comment header/javadoc to match other GameActions.

src/Engine/GameAction.java
```diff
@@ -1106,4 +1110,128 @@ public XYCoord getTargetLocation()
...
+    {
+      type = pType;
+      actor = unit;
+      destination = new XYCoord(unit.x, unit.y);
+    }
+
+    @Override
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
+      GameEventQueue eventSequence = new GameEventQueue();
```
ThatIsAUsername: Are there no preconditions? Funds, unit type, etc?

src/Engine/GameAction.java
ThatIsAUsername: Also, the name is awful close to `UnitProductionAction` - might want to try to differentiate those more.

src/Engine/UnitActionType.java
```diff
@@ -469,4 +470,81 @@ public String name()
...
+      {
+        return new GameActionSet(new GameAction.UnitProduceAction(this, actor), false);
+      }
+      return null;
+    }
+
+    @Override
+    public String name()
+    {
+      return "BUILD";
```
ThatIsAUsername: Should this identify the type being built?

src/Engine/GameAction.java
```diff
@@ -1106,4 +1110,128 @@ public XYCoord getTargetLocation()
       return destination;
     }
   } // ~UnitDeleteAction
+
+  public static class UnitProduceAction implements GameAction
+  {
+    final UnitProduce type;
```
ThatIsAUsername: Should this (and Flare) be top-level `UnitActionType`s since they aren't CO-specific/custom actions?

src/Units/DoRUnits.java
```diff
@@ -0,0 +1,791 @@
...
+    private static final WeaponModel[] weapons = { new DoRWeapons.MobileSAMWeapon() };
+
+    public MobileSAMModel()
+    {
+      super("Mobile SAM", DoRUnitEnum.MOBILESAM, UnitRoleEnum.ANTI_AIR, ChassisEnum.TANK, UNIT_COST, MAX_AMMO, MAX_FUEL, IDLE_FUEL_BURN, VISION_RANGE, MOVE_POWER,
+          moveType, actions, weapons, STAR_VALUE);
+    }
+  }
+
+  public static class RigModel extends DoRUnitModel
```
ThatIsAUsername: This guy needs a Material, right?

src/Units/DoRUnits.java
```diff
@@ -0,0 +1,791 @@
...
+      holdables = new Vector<ChassisEnum>(Arrays.asList(ChassisEnum.AIR_LOW, ChassisEnum.AIR_HIGH));
+    }
+
+    /** DoR Carriers re-supply and repair their cargo at the beginning of every turn. Make it so. */
+    @Override
+    public GameEventQueue getTurnInitEvents(Unit self, MapMaster map)
+    {
+      GameEventQueue events = new GameEventQueue();
+      for( Unit cargo : self.heldUnits )
+        if( !cargo.isFullySupplied() )
```
ThatIsAUsername: We only heal onboard units if they are also in need of resupply?

src/Test/TestUnitMovement.java
```diff
@@ -156,23 +156,23 @@ private boolean testFuelCosts()
...
+    testPassed &= validate(multiPath.getFuelCost(testCo1.getUnitModel(UnitRoleEnum.RECON), testMap) == 7,
         "    Recon movecost is wrong for road, grass, forest, or city.");
-    testPassed &= validate(grassPath.getFuelCost(testCo1.getUnitModel(UnitEnum.TANK), testMap) == 7,
+    testPassed &= validate(grassPath.getFuelCost(testCo1.getUnitModel(UnitRoleEnum.ASSAULT), testMap) == 7,
         "    Tank does not charge 1 fuel per space of grass.");
-    testPassed &= validate(multiPath.getFuelCost(testCo1.getUnitModel(UnitEnum.TANK), testMap) == 5,
+    testPassed &= validate(multiPath.getFuelCost(testCo1.getUnitModel(UnitRoleEnum.ASSAULT), testMap) == 5,
         "    Tank movecost is wrong for road, grass, forest, or city.");
-    testPassed &= validate(multiPath.getFuelCost(testCo1.getUnitModel(UnitEnum.CRUISER), testMap) == 396,
+    testPassed &= validate(multiPath.getFuelCost(testCo1.unitModels.get(15), testMap) == 396,
```
ThatIsAUsername: magic number again

src/CommandingOfficers/Commander.java
```diff
@@ -95,60 +68,11 @@ public Commander(CommanderInfo info, GameScenario.GameRules rules)
...
-      unitModels.put(um.type, um);
-    for (UnitModel um : airportModels)
-      unitModels.put(um.type, um);
-
-    UnitModel subsub = new SubSubModel();
-    unitModels.put(subsub.type, subsub); // We don't want a separate "submerged sub" build option
-    UnitModel stealthy = new StealthHideModel();
-    unitModels.put(stealthy.type, stealthy);
+    // Fetch our fieldable unit types from the rules
+    GameReadyModels GRMs = rules.unitModelScheme.getGameReadyModels();
```
ThatIsAUsername: Well, this got cleaner. Nice.

src/CommandingOfficers/Commander.java
```diff
@@ -254,10 +178,20 @@ public boolean isEnemy(Commander other)
...
-  // Leaving this for now, to avoid merge conflicts
-  public UnitModel getUnitModel(UnitEnum unitType)
+  public UnitModel getUnitModel(UnitRoleEnum unitType)
   {
-    return unitModels.get(unitType);
+    UnitModel um = null;
+
+    for( UnitModel iter : unitModels )
+    {
+      if( iter.role == unitType )
```
ThatIsAUsername: Since this just returns the first unit with the matching role, we can't get specific units anymore. Hrm.

src/Units/UnitModel.java
```diff
@@ -6,88 +6,115 @@
...
 {
   private static final long serialVersionUID = 1L;
 
-  public enum UnitEnum
+  /**
+   * Defines the core types of units we expect to be in any given unit set.
+   * Unit sets should have no less than one of each.
+   * Unit sets should order models such that the most "sensible" type (generally cheapest, land) is the first type of a given role.
+   */
+  public enum UnitRoleEnum
```
ThatIsAUsername: `UnitRoleEnum`s seem to have some overlap. `MECH`s are pretty clearly also `INFANTRY`, Stealths are `AIR_ASSAULT` but could also be `AIR_SUPERIORITY`, etc. Would it make sense to turn this into a bitmap? Could include things like ability to cloak/capture in there too.

src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+    }
+    public AWBWUnitModel(String pName, AWBWUnitEnum pType, UnitRoleEnum pRole, ChassisEnum pChassis, int cost, int pAmmoMax, int pFuelMax, int pIdleFuelBurn, int pVision,
+        int pMovePower, MoveType pPropulsion, ArrayList<UnitActionType> actions, ArrayList<WeaponModel> weapons, double starValue)
+    {
+      super(pName, pRole, pChassis, cost, pAmmoMax, pFuelMax, pIdleFuelBurn, pVision, pMovePower, pPropulsion, actions, weapons, starValue);
+      type = pType;
+    }
+
+    @Override
+    public UnitModel clone()
```
ThatIsAUsername: Would it be possible to push `clone` up to `UnitModel` so it doesn't have to be re-implemented in each child?

src/Terrain/MapInfo.java
```diff
@@ -40,4 +46,40 @@ public int getNumCos()
...
+          {
+            schemeInvalid = true; // This unit isn't supported in the scheme, so drop the scheme
+            break;
+          }
+        }
+        if( schemeInvalid )
+          break;
+      }
+
+      scheme.schemeValid = !schemeInvalid;
```
ThatIsAUsername: Why not `scheme.schemeValid = false`? In theory, it will always be true when passed in, but still.

src/Units/Unit.java
```diff
@@ -27,37 +24,27 @@
...
   public boolean isTurnOver;
   public boolean isStunned;
   private double HP;
-  public ArrayList<Weapon> weapons;
 
   public Unit(Commander co, UnitModel um)
   {
     CO = co;
     model = um;
+    ammo = model.maxAmmo;
```
ThatIsAUsername: This appears to assume that a Unit can only have one weapon with limited ammo.

src/AI/Muriel.java
```diff
@@ -292,17 +292,12 @@ public GameAction getNextAction(GameMap gameMap)
         shouldResupply = true;
       }
       // If we are out of ammo.
-      if( unit.weapons != null && unit.weapons.size() > 0 )
+      if( unit.ammo == 0 )
```
ThatIsAUsername: Won't this be true for APCs? You have to also check that the unit has weapons before worrying about reloading.

src/CommandingOfficers/Modifiers/CODefenseModifier.java
```diff
@@ -22,7 +22,7 @@ protected final void modifyUnits(Commander commander, ArrayList<UnitModel> model
   {
     for( UnitModel um : models )
     {
-      if( um.weaponModels != null )
+      if( um.weapons != null )
```
ThatIsAUsername: This modifier only applies to units with weapons? I feel like we copied this from AttackModifier and forgot to remove a condition.

src/CommandingOfficers/Modifiers/UnitRemodelModifier.java
```diff
@@ -46,17 +42,11 @@ public void applyChanges(Commander commander)
...
-        // Store off the unit's weapons.
-        unitWeapons.put(unit, unit.weapons);
-
         // Swap the unit's identity and store it for later.
         modelSwapBacks.put(unit, unit.model);
-        unit.model = unit.CO.unitModels.get(modelSwaps.get(unit.model.type));
-
-        // Change out weapons.
-        doWeaponSwap(unit, unit.model);
+        unit.model = modelSwaps.get(unit.model);
```
ThatIsAUsername: There is no longer any weapon/ammo handling here. Is that just unnecessary now that ammo is part of Unit?

src/Engine/GameEvents/BattleEvent.java
```diff
@@ -61,13 +61,13 @@ public void performEvent(MapMaster gameMap)
     // Apply the battle results that we calculated previously.
     Unit attacker = battleInfo.attacker;
     Unit defender = battleInfo.defender;
-    battleInfo.attackerWeapon.fire(); // expend ammo
+    battleInfo.attackerWeapon.fire(attacker); // expend ammo
```
ThatIsAUsername: This syntax feels a bit convoluted. I might prefer `attacker.fire(battleInfo.attackerWeapon)`.

src/Units/WeaponModel.java
```diff
@@ -0,0 +1,90 @@
...
+        System.out.println("WARNING: fired an empty gun!");
+    }
+  }
+
+  /**
+   * @return returns its base damage against defender if the unit is in range
+   */
+  public double getDamage( UnitModel defender, int range )
+  {
+    if( defender != null )
```
ThatIsAUsername: Is a null `defender` an expected case? If not, maybe an exception is in order.

src/Units/DoRWeapons.java
```diff
@@ -0,0 +1,386 @@
...
+    RECONMGUN, FLAREMGUN, ANTI_AIRMGUN,
+    TANKCANNON, TANKMGUN, MD_TANKCANNON, MD_TANKMGUN, WARCANNON, WARMGUN,
+    ARTILLERYCANNON, ANTITANKCANNON, ROCKETS, MOBILESAM,
+    FIGHTERMISSILES, BOMBERBOMBS, SEAPLANESHOTS,
+    DUSTERMGUN, COPTERROCKETS, COPTERMGUN,
+    GUNBOATGUN, CRUISERTORPEDOES, CRUISERMGUN, 
+    SUBTORPEDOES, CARRIERMGUN, BATTLESHIPCANNON,
+  };
+  
+  protected final static boolean MAX_AMMO = true;
```
ThatIsAUsername: Should this be named `USE_AMMO`?

src/Engine/UnitActionType.java
```diff
@@ -469,4 +470,81 @@ public String name()
...
+    final UnitModel typeToBuild;
+    public UnitProduce(UnitModel type)
+    {
+      typeToBuild = type;
+    }
+
+    @Override
+    public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor, boolean ignoreResident)
+    {
+      XYCoord moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
```
ThatIsAUsername: We really need a `Path.getEndCoord` convenience function.

src/Engine/GameAction.java
```diff
@@ -1106,4 +1110,128 @@ public XYCoord getTargetLocation()
...
+          flareEvents.add(event);
+        }
+      }
+      return flareEvents;
+    }
+
+    @Override
+    public String toString()
+    {
+      return String.format("[%s will launch a flare to %s from %s]", actor.toStringWithLocation(), launchLocation, moveCoord);
```
ThatIsAUsername: This does not follow the action-description format convention. Command first.

src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+
+    private static final MoveType moveType = new Flight();
+    private static final UnitActionType[] actions = UnitActionType.BASIC_ACTIONS;
+    private static final WeaponModel[] weapons = {};
+
+    public BBombModel()
+    {
+      super("BBomb", AWBWUnitEnum.BBOMB, UnitRoleEnum.AIR_ASSAULT, ChassisEnum.AIR_HIGH, UNIT_COST, MAX_AMMO, MAX_FUEL, IDLE_FUEL_BURN, VISION_RANGE, MOVE_POWER,
+          moveType, actions, weapons, STAR_VALUE);
+      possibleActions.add(0, new UnitActionType.Explode(5, 3));
```
ThatIsAUsername: Magic 0. Minor instance, but I will still normally try to use a named temp.

src/Units/AWBWUnits.java
```diff
@@ -0,0 +1,778 @@
...
+    private static final MoveType moveType = new FloatLight();
+    private static final UnitActionType[] actions = UnitActionType.TRANSPORT_ACTIONS;
+
+    public BBoatModel()
+    {
+      super("BBoat", AWBWUnitEnum.BBOAT, UnitRoleEnum.TRANSPORT, ChassisEnum.SHIP, UNIT_COST, MAX_AMMO, MAX_FUEL, IDLE_FUEL_BURN, VISION_RANGE, MOVE_POWER, moveType,
+          actions, new WeaponModel[0], STAR_VALUE);
+      holdingCapacity = 2;
+      holdables = new Vector<ChassisEnum>(Arrays.asList(ChassisEnum.TROOP));
+      possibleActions.add(0, UnitActionType.REPAIR_UNIT);
```
ThatIsAUsername: Magic 0. Minor instance, but still. I still will normally try to use a named temp.

ThatIsAUsername:
Overall this looks like a pretty reasonable way to implement multiple unit sets. There are still some rough edges, but I know it's also still WIP.

The DoR faction names/colors don't really seem to fit in with the others. The buildings in particular look kinda washed out.

src/Units/AWBWUnits.java
ThislsAUsername: `UnitEnum` is now an implementation detail of any given unit set. You don't need the enum since you don't need a damage chart - only a way to generate a base damage number (which you can calculate by accessing custom variables for your unit set).

I think the damage chart with enum indexing is about as good a system as you can get, for most purposes.
It makes targeted unit damage balance changes trivial, and doesn't require complex logic.

src/Units/AWBWUnits.java
ThislsAUsername: Because it requires another `UnitModel` to create the action, since it's a transformation.
I suppose I could build `sub` inside `subsub`, but there's no interface currently for generating required *related* `UnitModel`s.

src/Test/TestCOModifier.java
ThislsAUsername: I do entirely agree, but we can't guarantee there *is* a Battleship for any given unit set.

The magic number is fine IMO since this is test code.

src/Units/AWBWUnits.java
ThislsAUsername: I should probably standardize between the save file and this, too.
Do we want `~` or `!`, and should they be prepended or appended?

src/Units/MoveTypes/TiresRugged.java
ThislsAUsername: DoR weather doesn't change terrain move costs, and the movetype only exists in DoR.

src/Units/UnitModel.java
ThislsAUsername: I wanted to get the variable in now, since it'd be a massive pain to get it in later.

How about I change it to `customPowerValue`?

src/Units/UnitModelScheme.java
ThislsAUsername: I originally was using it to set the unit used for picking your palette/faction choice, but then I realized how awful an idea it was since the non-DoR factions would all default to the same bike sprite.

src/Engine/GameAction.java
ThislsAUsername: Good point. Should fix that.

Yeah, I noticed the name after I made it... but they're gonna end up being synonyms.
I guess we could do `ProduceUnitAction` and `UnitProduceUnitAction`, heh.

src/Engine/GameAction.java
ThislsAUsername: `UnitProduceEvent` should fail properly on its own if we don't have the funds, and unit type is defined... by the production event type, so we can't really check it meaningfully.

src/Engine/UnitActionType.java
ThislsAUsername: It could. It doesn't in DoR, so I aped that for a start.

src/Engine/GameAction.java
ThislsAUsername: Flare, maybe, since we don't have a use-case yet for making Flares with different ranges/AoEs.
I don't want to have to make a new action type for every new unit type we want to have a unit build, though.

src/Units/DoRUnits.java
ThislsAUsername: Sure, if and when we get around to temp air/ports.

src/Test/TestUnitMovement.java
ThislsAUsername: Test code, again.

src/Units/DoRUnits.java
ThislsAUsername: Good point. Should fix.

src/CommandingOfficers/Commander.java
ThislsAUsername: That's sort of intentional, since I'd like the rest of the game to work regardless of unit set, so long as the unit set follows our assumptions.

src/Units/UnitModel.java
ThislsAUsername: That could make sense, yeah.
I'd have to consider the implications on our existing use cases (and where we wanna go with things) a bit more before going for that, though.

src/Units/AWBWUnits.java
ThatIsAUsername: I like having it up front, since that is a stronger demand for attention. Either works, but maybe '!', for the same reason.

src/Units/AWBWUnits.java
ThislsAUsername: That sounds lovely, but the children need to be able to clone their extra values.
Do you have a way to solve that?

src/Units/UnitModel.java
ThatIsAUsername: This is the amount of power a Commander gains when they lose a unit of that type, right? I'm not sure that "custom" is the best word, since every unit gets their own value anyway. Maybe just `abilityPowerValue`?

src/Units/UnitModelScheme.java
ThatIsAUsername: I mean, if you are playing with the DoR unit set, then non-DoR factions _would_ have bikes. I can go either way on this.

src/Terrain/MapInfo.java
ThislsAUsername: They're different flags. One's local, and can be true or false depending on whether the map's units fit within the unit set.

src/Units/Unit.java
ThislsAUsername: Au contraire: it abstracts ammo from all weapons into one pool.
It seems counter to the abstract design to try to manage multiple separate stores of ammo per unit.

src/AI/Muriel.java
ThislsAUsername: IIRC, they're initialized to -1.
Either way, it's immaterial at the moment because Muriel doesn't build them.

src/CommandingOfficers/Modifiers/CODefenseModifier.java
ThislsAUsername: ...whoops?
Clearly, refactoring this was a good way of catching existing bugs.

src/CommandingOfficers/Modifiers/UnitRemodelModifier.java
ThislsAUsername: Yep. `Weapon`s don't exist, only `WeaponModel`s... which are owned by the `UnitModel`.
Simplifying logic is great.

Also, not requiring a 1:1 mapping between any enums and `UnitModel`s means you could have a CO with multiple forms of a given unit type (say, balancing unit spam by giving them awful attack/defense).

src/Units/DoRWeapons.java
ThislsAUsername: ...yes, probably.

src/Engine/UnitActionType.java
ThislsAUsername: Yeeeees.

src/Units/AWBWUnits.java
ThislsAUsername: It's just slapping it to the start of the array.
If you don't know what index 0 means, you probably shouldn't be messing around in the codebase lol

src/Units/UnitModelScheme.java
ThislsAUsername: That's true, but then you can't tell which faction is which... which is still important, since your infantry are still different.
I guess we could figure out whether the given faction has its own sprite for the iconic unit, then pick INFANTRY as a default if it doesn't?

ThislsAUsername: Clearly, the correct solution is to check if the iconic type is available, then iterate through the UnitRoleEnum looking for a unique unit image.
...I was initially joking, but that actually doesn't sound half-bad.

src/Engine/GameAction.java
ThislsAUsername: I'm starting to feel like GameAction.java is getting a little too big to manage comfortably.
How do you feel about splitting each class out like GameEvent has?

src/Units/UnitModelScheme.java
ThislsAUsername: Throwing an exception would make the MapMaster constructor uglier than it needs to be: if getModelFromString fails, you just... don't get the unit.
So long as we let users play with a unit set that doesn't match the map, this isn't an exceptional execution path... and it's not exceptional anyhow, since it's used for checking map/scheme compatibility in the first place.

src/Engine/UnitActionType.java
```diff
@@ -44,7 +44,7 @@ public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor)
     @Override
     public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor, boolean ignoreResident)
     {
-      XYCoord moveLocation = new XYCoord(movePath.getEnd().x, movePath.getEnd().y);
+      XYCoord moveLocation = movePath.getEndCoord();
```
ThislsAUsername: I feel like maybe this should be moved into a parent function or something, so it isn't copy-pasted everywhere?
*shrug*

src/Units/UnitModelScheme.java
ThatIsAUsername: Uhhh. I mean. You do you, man. I feel like the UnitScheme might be a more definitive thing than the faction. Don't bikes have to be available to the non-DoR factions anyway?

src/Engine/GameAction.java
ThatIsAUsername: Hah. I did those differently partly just to see which one we liked more in the long run. Though they aren't completely comparable. The `UnitModelScheme` subclasses also have many class definitions in one file - what makes `GameAction` more worthy of splitting?

To be clear, I think I'm fine with doing that as long as we organize the files in a sane way.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Hm. What happens when we make campaign missions where three battleships appear 10 turns into the match? Or do you just want to cross that bridge when we get to it?

src/Units/AWBWUnits.java
ThatIsAUsername: Hrm. Could just have the parent `clone` handle the common attributes. The child `clone` can call the parent one and then finish off the extra things. I think that should work.

src/Terrain/MapInfo.java
ThatIsAUsername: Right. Silly me.

src/Units/Unit.java
ThatIsAUsername: Uhh. OK. True. So if I make a new `MechWarrior` unit that has both an AA gun and a cannon, they will both fire the same bullets? Not sure how I feel about that.

src/AI/Muriel.java
ThatIsAUsername: Sigh. True enough.

src/Units/AWBWUnits.java
ThatIsAUsername: I mean, I know what that line does, I just don't like magic numbers. They are slightly more forgivable on standard library functions I guess, which is why I said it's a "minor instance".

The 5 and 3 are also magic, and that's not a standard library call.

src/UI/MainUIController.java
```diff
@@ -118,7 +118,7 @@ private boolean handleMainMenuInput(InputHandler.InputAction action)
                     saves.add(new SaveInfo(filepath, filename, prettyName));
                   else
                     // Throw an tilde in there to tell the user "yeah, we see it, and it ain't gonna work."
```
ThatIsAUsername: Throw a what in there?

src/Units/DoRUnits.java
```diff
@@ -734,11 +734,11 @@ public GameEventQueue getTurnInitEvents(Unit self, MapMaster map)
     {
       GameEventQueue events = new GameEventQueue();
       for( Unit cargo : self.heldUnits )
+      {
+        events.add(new HealUnitEvent(cargo, self.CO.getRepairPower(), self.CO)); // Event handles cost logic
```
ThatIsAUsername: Now we are healing even when they don't need it? I mean, it works I guess.

src/Units/WeaponModel.java
```diff
@@ -0,0 +1,75 @@
...
+      return getDamage(defender);
+    return 0;
+  }
+  /**
+   * @return returns its base damage against that unit type
+   */
+  public double getDamage(UnitModel defender)
+  {
+    if( defender == null )
+      throw new IllegalArgumentException("Stare not into the void. It just may stare back.");
```
ThatIsAUsername: Now you're just trying to get a reaction. You know what? It's a unique message that we can use it to find the problem. I'm fine with this. I probably shouldn't do code review past midnight.

src/Engine/UnitActionType.java
ThatIsAUsername: _shrug_

src/Units/UnitModelScheme.java
ThislsAUsername: All of what you said is true... but it's extremely un-useful to just see row on row of identical bikes, with no indication of what factions they actually are.
When you're picking the faction, it kinda makes sense to show imagery based on *that* faction rather than falling through to whatever their default is.

I think I'll play around with this a bit and see what happens..

src/Engine/GameAction.java
ThislsAUsername: The metric I'm using is "are there enough lines to make it annoying to scroll around and try to find what I'm looking for?"
Also important is how big the individual classes are, though I guess we could break up a file by defining a few well-defined chunks of classes.

DoRUnits and AWBWUnits are both around 800 lines, which is definitely getting unwieldy... but they're not really going to grow and each individual class is pitifully trivial (and having them right next to each other is also nice for comparison and bulk-edit stuff)

UnitActionType is getting up there, at 551 lines, but at least I can still mouse-roller scroll through the whole file pretty quickly.

GameAction is 12k lines, with enough boilerplate per class to make them longer than e.g. `GameInputState`s are.

I think an aggregated "GameEvent.java" with everything in it would be even longer, since we have a number of GameEvents with no associated GameAction.

I think there's also a point where we'd want to start putting groups of files into subfolders (which I ended up doing with the Tester branch's COs because... y'know).
I think GameEvent is pushing towards having enough classes to make that valuable. If I had to pick now, I think I'd do "generic unit" (move/heal/resupply/fight/capture/die/join), "special unit" (flare/repair/produce/transform/load[?]/unload[?]), and "non-unit" (mass damage/weather/defeat/ability/build unit [as a unit isn't the actor]) events.

src/Units/AWBWUnits.java
ThislsAUsername: Fair enough, I suppose.

You definitely have a point on the Explode constructor there, though.

src/Units/DoRUnits.java
ThislsAUsername: The event's made to handle healing units that don't need healing.

ThislsAUsername:
> The DoR faction names/colors don't really seem to fit in with the others. The buildings in particular look kinda washed out.

As for faction/palette names, I was just going with the source material here.
"Army" is super generic, but allows for amusing things like "Red Army".
"Wolves" is actually pretty legit IMO.
"Brenner's" is jank, but I don't have a better name on me. (That combo could just be a special one like OS is)
"IDS", "New Rubinelle", and "Lazurian" are sort of in the same boat as "Brenner's", though not as bad.
"IDS Army" could just be special cased to just "IDS", I guess?

I definitely just grabbed the palettes from the spritesheet and plopped 'em down. I agree they don't mesh super well, but that's what happens when you mix art styles.
If you want to refine particularly the building palettes that for our sprites, you can... but it doesn't offend my sensibilities enough for me to care much (which would make it hard for me to fix it to your satisfaction).

src/Units/Unit.java
ThislsAUsername: I agree that's a little awkward.
I'd probably propose just making one of them infinite ammo, since that's the typical approach for secondary weapons - in this case, I'd go for the AA gun (since it has precedent in the Cruiser).

src/CommandingOfficers/Commander.java
ThislsAUsername: We do have unit-fetching based on string lookup right now, which is what we use to deal with map init. This doesn't necessarily need to be different.

I will say that such missions would likely only be balanced with one unit set in mind, since trying to set up both DoR and standard Battleships (and my design, when that happens) to be the same strength in the same place at the same time would be... interesting.

If the missions are defined as external/text files, it's going to be exactly the same as map import.
If the missions are defined in code, we have even more flexibility.

src/Units/UnitModel.java
ThislsAUsername: Okay. I took a look at it.
I'm actually quite liking the whole flags idea.

It does occur to me that it's definitely overlapping with our whole "chassis" concept, and that maybe the chassis should *also* be folded into this new role bitmap?
Then again, that means we need to decide what sort of semantics we want to define for how e.g. transports deal with multiple-chassis units. If a Bike is both TROOP and TANK, or an amphibious tank is both TANK and SUB, should APCs and Landers still load them (respectively)?

There's one incompatibility with current usage, and I'm not sure it's all that important based on the other comment thread about the Battleships:
I wanted it to be possible to make completely unit-set-agnostic maps by specifying unit roles rather than unit names (particularly for FTA counter inf purposes)... but I think it's fair to say that high-quality maps *in general* are balanced around a particular unit set.

This might matter a little If we do end up renaming infantry to riflemen, but then again I can't see that being too world-shattering.

ThislsAUsername: Counterpoint to the chassis thing:
The overlap I was seeing was primarily that a CO (or even AI) could be interested in using chassis or role for filtering unit types. This effect is particularly strong in `INFANTRY`/`TROOP`.
That particular case is quite strongly overloaded, but the other chassis types don't really align quite so easily.

It's also relevant that the chassis enum mixes both morphology and location in it: it would seem to be pretty incoherent for something to be `TROOP`, `SUB`, and `AIR_HIGH` all at once (and no other chassis types). (I guess you could make an argument about a Protoss Colossus standing in water, or maybe an Avatar of Khaine standing in water... but still.)

src/Units/Unit.java
ThatIsAUsername: Yeah, perhaps. It's fine for now anyway, since no existing units have more than one ammo-bound weapon anyway.

src/Units/UnitModelScheme.java
ThatIsAUsername: If you want a better indication of what the factions actually are, then maybe we just need an overlay with the faction/color name as you scroll around. That way you always know what combination you're getting.

ThislsAUsername: I think that's a good idea, but I also think it's fair to only show a bike if the faction in question has a custom bike.
If the penguins don't have penguins on bikes, then maybe just show a normal penguin instead.
That way you don't have to manually check each faction with a duplicate/default unit.

ThislsAUsername:
I think my solution for enabling temp ports' weird logic (i.e. they block non-friendly ships) will involve refactoring the movement code some, so I think it's worth making a new PR for the more advanced features.
That sound fair?

src/Units/UnitModelScheme.java
ThatIsAUsername: Every faction is supposed to have custom infantry though, right? That's how they've always been distinguished. I'm actually not sure why we need to consider other unit types at all.

src/Engine/GameAction.java
ThatIsAUsername: I'm still fine with splitting this up if you care to.

ThatIsAUsername:
I ain't got time to mess with sprite sheet colors, and they swap out our buildings for the AW ones anyway, so that's not a priority.
Special-casing seems like a decent plan. Rename "Brenner's" to "Chartreuse" or whatever the color is normally, then just fix it for the correct combo. Could do something similar for the others too, I bet.
You will have to modify movement code, yes. A new PR seems appropriate for that.

Is there anything else that still needs my input?

src/Units/UnitModelScheme.java
ThislsAUsername: Fair enough.

ThislsAUsername:
> Special-casing seems like a decent plan. Rename "Brenner's" to "Chartreuse" or whatever the color is normally, then just fix it for the correct combo. Could do something similar for the others too, I bet.

Having let this percolate a little bit more, I honestly like "Brenner's" as a name. The country names are a little weird still, but I don't really feel it's a problem.

> Is there anything else that still needs my input?

Whether or not to turn UnitRoleEnum into a bit field (and if so, whether to include chassis)? I looked at the implications and talked through them a bit on the thread about that.

src/Units/UnitModel.java
ThatIsAUsername: > If a Bike is both TROOP and TANK, or an amphibious tank is both TANK and SUB, should APCs and Landers still load them (respectively)

Yes. If APCs load troops, then they should be able to load anything that has that flag.

> The overlap I was seeing

I mean, infantry and mechs are both troops. If we introduce engineers and armored suits, they would presumably also be troops. I see no problem with AIs/COs filtering by role or by chassis.

> it would seem to be pretty incoherent for something to be TROOP, SUB, and AIR_HIGH

Well, yeah. Are you just worried about people setting multiple bits that seem to contradict each other? Lets just not do that. Alternatively, if we ever decide to build an Iron Man suit, now we can.

PR #224, opened 2020-03-03
Code rejiggering

ThislsAUsername:
This PR basically does the refactoring discussed on the `complaintMagnet` review.
I figured I'd make it separate since that one was already full of stuff.


src/Units/DoRUnits.java
```diff
@@ -135,24 +135,11 @@ public DoRUnitModel(String pName, DoRUnitEnum pType, UnitRoleEnum pRole, Chassis
...
-        weapons = new ArrayList<WeaponModel>();
-        for( WeaponModel weapon : weapons )
-        {
-          weapons.add(weapon.clone());
-        }
-      }
-
       // Create a new model with the given attributes.
-      DoRUnitModel newModel = new DoRUnitModel(name, type, role, chassis, getCost(), maxAmmo, maxFuel, idleFuelBurn, visionRange, movePower,
+      DoRUnitModel newModel = new DoRUnitModel(name, type, role, chassis, moneyCost, maxAmmo, maxFuel, idleFuelBurn, visionRange, movePower,
```
ThatIsAUsername: Why change `getCost()` to `moneyCost` here but not in `AWBWUnits.java`?

src/Units/UnitModel.java
```diff
@@ -21,45 +20,39 @@
...
+  public static final long LAND            = 1 <<  8;
+  public static final long AIR_LOW         = 1 <<  9;
+  public static final long AIR_HIGH        = 1 << 10;
+  public static final long SEA             = 1 << 11;
+  public static final long SUBSURFACE      = 1 << 12;
+  // Role; Unit sets are expected to have at least one of each
+  public static final long MECH            = 1 << 16; // Footsoldier equipped against hardened targets
+  public static final long RECON           = 1 << 17; // Scout
+  public static final long ASSAULT         = 1 << 18; // Fast unit that can deal with hardened ground targets
+  public static final long SIEGE           = 1 << 19; // Unit with heavy strike capabilities at range
```
ThatIsAUsername: So Snipers would not count as `SIEGE`?

src/Units/UnitModel.java
```diff
@@ -123,9 +116,9 @@ private UnitModel(String pName, UnitRoleEnum pRole, ChassisEnum pChassis, int co
 
     for( TerrainType terrain : TerrainType.TerrainTypeList )
     {
-      if( (((chassis == ChassisEnum.AIR_HIGH) || (chassis == ChassisEnum.AIR_LOW)) && terrain.healsAir()) ||
-          (((chassis == ChassisEnum.TANK) || (chassis == ChassisEnum.TROOP)) && terrain.healsLand()) ||
-          (((chassis == ChassisEnum.SHIP) || (chassis == ChassisEnum.SUBMERGED)) && terrain.healsSea()) )
+      if( (((role & (AIR_LOW | AIR_HIGH)) > 0) && terrain.healsAir()) ||
```
ThatIsAUsername: We should put these kinds of messy comparisons in a function; `hasRole(AIR_LOW | AIR_HIGH)` or some such.

src/CommandingOfficers/Meridian.java
```diff
@@ -179,7 +179,7 @@ protected void perform(MapMaster gameMap)
       // Lastly, all land vehicles are refreshed and able to move again.
       for( Unit unit : COcast.units )
       {
-        if( unit.model.chassis == UnitModel.ChassisEnum.TANK )
+        if( ((unit.model.role & UnitModel.TANK) > 0) )
```
ThatIsAUsername: Messy syntax should be hidden in a function.

src/Units/DoRUnits.java
```diff
@@ -31,7 +31,7 @@ public String toString()
   @Override
   public String getIconicUnitName()
   {
-    return "Bike";
+    return "Infantry";
```
ThatIsAUsername: You added support for loading different unit types, and then changed the iconic type back to infantry?

src/Engine/GameEvents/GameEventListener.java
```diff
@@ -71,12 +74,12 @@ public final void unregister()
...
   public void receiveCommanderDefeatEvent(CommanderDefeatEvent event){};
-  public void receiveLoadEvent(LoadEvent event){};
+  public void receiveLoadEvent(LoadLifecycle.LoadEvent event){};
   public void receiveMoveEvent(MoveEvent event){};
-  public void receiveUnitJoinEvent(UnitJoinEvent event){};
+  public void receiveUnitJoinEvent(JoinLifecycle.JoinEvent event){};
   public void receiveResupplyEvent(ResupplyEvent event){};
   public void receiveUnitDieEvent(UnitDieEvent event){};
-  public void receiveUnloadEvent(UnloadEvent event){};
+  public void receiveUnloadEvent(UnloadLifecycle.UnloadEvent event){};
```
ThatIsAUsername: Why do you use <action>`Lifecycle for some but not others?

Also, these functions probably shouldn't actually take the event anyway; they should instead accept the relevant components (like Battle, Capture, MassDamage, etc).

src/Units/DoRUnits.java
ThislsAUsername: Because I'm a fool.

src/Units/UnitModel.java
ThislsAUsername: Nah, I don't think so. Meridian doesn't want tanks transforming into snipers, anyhow.
I'm... not sure snipers really have enough range to qualify for 2 range, and they aren't proper indirect-fire, so I dunno that I'd want to bring them in to a unit set with similar assumptions as the existing ones.

src/Units/DoRUnits.java
ThislsAUsername: Yeah. Killing the unit role enum entirely meant I couldn't get it from the enum directly, so basing it on the iconic unit was the next logical step. It might make sense to instead just create a set of `UnitModel`s from the UnitSet and just use that, but that feels... hacky?

Based on our previous discussion, it makes sense to use infantry since that's what differentiates most factions anyhow.

src/Engine/GameEvents/GameEventListener.java
ThislsAUsername: I dunno, it's how it turned out when I was moving things around and renaming.
Some of them seemed to require it, others didn't.

Guess I should try to standardize it?

src/Units/UnitModel.java
ThatIsAUsername: Yeah, well I'm also not super-comfortable with having Meridian's ability key off of unit role alone - seems fragile if we ever do add new unit options. Can she not choose those units by name?

src/Units/DoRUnits.java
ThatIsAUsername: If we always use Infantry as the iconic type--and it sounds like you don't intend to change this--why keep `getIconicUnitName` and related things?

src/Engine/GameEvents/GameEventListener.java
ThatIsAUsername: Maybe. I generally change the listener functions to accept the necessary components rather than the whole event once I actually use it somewhere, so it may make sense to just do that up front.

I didn't bother before though, so it's clearly not a priority.

src/Units/UnitModel.java
ThislsAUsername: I would argue that basing it off of unit name makes it *more* fragile.
If I, say, wanted to steal Wargroove's units, the existing implementation would match up Archers and Knights.

Also, my comment above this block does address the fragility somewhat - Unit models should be ordered such that fetching the first unit of a given role is a sensical operation in most cases.

ThatIsAUsername: Well, fair enough. Different approaches will leave us vulnerable to different kinds of change. I guess I won't quibble too much over this until we have a real use case that forces our hand.

PR #225, opened 2020-03-20
Moved PlayerSetupColorFactionArtist sprite preloading onto a separate thread

ThatIsAUsername:
This works, but:
We could just as easily initialize the sprite preloader thread statically, e.g.
`private static InfantrySpritePreloader isp = new InfantrySpritePreloader();`

This would cause the thread to kick off as soon as the owning class was loaded (when first referenced at run-time). We could also kick off the thread earlier - no reason to wait until we are almost about to access the screen to start preparing.

At any rate, this is still an improvement, since the preloading process would create jitter when we were running it on the UI thread.

### DIFF
--- a/src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
+++ b/src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
@@ -25,6 +25,14 @@ public class PlayerSetupColorFactionArtist
   private static SlidingValue gridY = new SlidingValue(0);
   private static SpriteCursor spriteCursor = new SpriteCursor(0, 0, SpriteLibrary.baseSpriteSize, SpriteLibrary.baseSpriteSize, UIUtils.getCOColors()[0]);
 
+  private static boolean donePreloading = false;
+  private static InfantrySpritePreloader isp;
+  public static void startSpritePreloader(String name)
+  {
+    if( !donePreloading && (null == isp) )
+    isp = new InfantrySpritePreloader(name);
+  }
+
   public static void draw(Graphics g, IController controller)
   {
     PlayerSetupColorFactionController control = (PlayerSetupColorFactionController)controller;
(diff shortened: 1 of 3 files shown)


src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
```diff
@@ -127,25 +141,47 @@ private static void initialize()
...
-      if( nextFac >= UIUtils.getFactions().length )
+      myThread = new Thread(this);
+      myThread.start();
+      myUnitName = unitName;
+    }
+
+    @Override
+    public void run()
+    {
+      while( !donePreloading )
```
ThislsAUsername: Kind of a weird loop structure. Why not just use a nested for?

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
```diff
@@ -84,6 +92,12 @@ public static void draw(Graphics g, IController controller)
     BufferedImage image = SpriteLibrary.createTransparentSprite(myWidth, myHeight);
     Graphics myG = image.getGraphics();
 
+    if( !donePreloading )
+    {
+      isp.join();
+      isp = null;
+    }
```
ThislsAUsername: If I'm reading this right:
If we aren't done pre-loading before this gets here, the game just freezes until we are.
If we *are* done pre-loading, then isp never becomes null.

Is that accurate?

ThislsAUsername:
I think, if I had my 'druthers, pre-loading would be kicked off only once in SpriteEngine.getMainUIView()
The issue there is that we don't know what our iconic unit type is at that point.

If you're happy to bolt our UI to "infantry" as the one true basic unit type, then maybe we can/should get rid of all this iconic unit shenanigans.

Another worthwhile idea is that maybe we should separate the unit *display* name from the unit image resource file name.

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
ThatIsAUsername: I just copied/moved what had been done before. A simple double loop would be more straightforward though.

src/UI/Art/SpriteArtist/PlayerSetupColorFactionArtist.java
ThatIsAUsername: Mmmm, yes. Not exactly what was intended though.

ThatIsAUsername:
> I think, if I had my 'druthers, pre-loading would be kicked off only once in SpriteEngine.getMainUIView()

Works for me.

I'm still not sure I ever caught the vision for having varied iconic unit types, so I'm fine with consolidating with Infantry as the standard.

ThislsAUsername:
Looks good to me. If you're happy, I'm happy.

PR #226, opened 2020-03-23
Preload thread

ThatIsAUsername:
Should we go ahead and remove the IconicUnit stuff since it isn't used, or keep the door ajar in case we want it in the future?


ThislsAUsername:
I think, if nothing else, hard-coding in fewer places is better.

I can't say I'm particularly inspired to chase making the iconic unit stuff better at the moment, but I won't say I never will be.

ThatIsAUsername:
I normally don't like to leave nonfunctional stuff around, but eh. It shouldn't hurt anything.

PR #227, opened 2020-04-10
Controls

ThatIsAUsername:
Functional, but not super pretty. I may have violated some best practices. Actually, I should wait for you to tell me that.
Probably need to save/load settings to a config file before this hits real users.

### DIFF
--- a/src/UI/Art/SpriteArtist/ControlOptionsSetupArtist.java
+++ b/src/UI/Art/SpriteArtist/ControlOptionsSetupArtist.java
@@ -32,7 +32,7 @@ public class ControlOptionsSetupArtist
     myControl = control;
 
     // Set up some initial parameters.
-    int spacing = kbps.get(0).getImage().getHeight()*2;
+    int spacing = kbps.get(0).getKeysImage().getHeight()*2;
     int xDraw = spacing;
     int yDraw = spacing/2;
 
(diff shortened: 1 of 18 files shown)


src/UI/Art/SpriteArtist/ControlOptionsSetupArtist.java
```diff
@@ -0,0 +1,177 @@
...
+   */
+  private static class KeyBindingPanel
+  {
+    InputHandler.InputAction myAction;
+
+    // A couple of helper quantities.
+    private static final int textBufferPx = 4;
+    private static final int keyBufferPx = 8;
+
+    // The composed TeamPanel image.
```
ThislsAUsername, line 84: A team of keybindings? :stuck_out_tongue:

src/UI/InputHandler.java
```diff
@@ -3,11 +3,16 @@
 import java.awt.event.KeyEvent;
 import java.util.ArrayList;
 import java.util.Arrays;
-public class InputHandler
+import java.util.HashMap;
+
+import Engine.IController;
+import Engine.OptionSelector;
+import UI.InputHandler.InputAction;
+public class InputHandler implements IController
```
ThislsAUsername, line 17: Hmm. I can't say I expected to see this, but It does make a certain amount of sense as a way to keep input-specifics in this one class to abstract it away from the rest of the codebase.

ThislsAUsername:
Some general UI feedback, screenshot for reference
image
It'd be nice to x-align the first key panel for each input item.
The fact that the action types have up-right shadows and `ADD` has down-left shadows looks a little weird and inconsistent to me. A good alternative, I think, would be to just make both light (or one light/one dark) to distinguish them.

src/UI/Art/SpriteArtist/ControlOptionsSetupArtist.java
ThatIsAUsername: Of course. A single bound key wouldn't be worth a whole lot in this game.

src/UI/InputHandler.java
ThatIsAUsername: Well, I didn't expect to do this, but then it kinda happened.

ThatIsAUsername:
Yeah, I did the up-right shadow on the left column because the vertical line on the right provides some separation and differentiation for those panels. It does look a tad wonky though.

ThislsAUsername:
Yeah. That looks good to me.

src/UI/InputHandler.java
ThislsAUsername: Hmm. I feel like the static and non-static portions aren't particularly closely-bound.
It is amusing/novel to leave them together, though, so I guess I don't care much either way.

ThislsAUsername:
Works for me. Did you want to make any other changes?

src/UI/InputHandler.java
ThatIsAUsername: I started out by making a separate class, but it did nothing but make calls to InputHandler anyway.
I imagine I could find a more structured way to compose it if I tried, but it's simple enough and works well enough now that it doesn't seem worth the effort to try and overthink it at this point.

ThatIsAUsername:
There are any number of changes that _could_ be done, like putting in a "restore defaults" button somewhere, or changing the sub-menu on the main page to not stand out so much (since I'm still not a huge fan of that), but this is quite usable, so I think I'm fine with it.

PR #228, opened 2020-04-19
Dor actions

ThislsAUsername:
This PR includes some bugfixes made on the test branch because otherwise I'd forget to ever port them over.


ThatIsAUsername:
Aaaaah, you broke transports! Now you have to add a test case.

Rules are rules.

ThatIsAUsername:
Join is broken too...

src/Engine/GameInput/SelectLaunchable.java
```diff
@@ -0,0 +1,50 @@
...
+    launchables.toArray(cargoArray);
+    return new OptionSet(cargoArray);
+  }
+
+  @Override
+  public GameInputState<?> select(Unit option)
+  {
+    GameInputState<?> next = this;
+
+    myStateData.unitActor = option;
```
ThatIsAUsername, line 38: Should we confirm that `option` is actually on this transport before assuming all is well? Probably not critical, but c.f. `SelectCargo`, which does this.

src/Engine/Utils.java
```diff
@@ -92,140 +92,136 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
     }
   }
 
+  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap, boolean) findPossibleDestinations()} **/
```
ThatIsAUsername, line 95: This link doesn't go to a real function.

src/Engine/Utils.java
```diff
@@ -92,140 +92,136 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
...
+  public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gameMap)
+  {
+    return findPossibleDestinations(new XYCoord(unit.x, unit.y), unit, gameMap);
+  }
+  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap, boolean) findPossibleDestinations()} **/
```
ThatIsAUsername, line 100: This one either.

src/Engine/Utils.java
```diff
@@ -527,8 +520,8 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
     if( map.isLocationValid(x, y) )
     {
       int range = (piercing)? viewer.model.visionRangePiercing : viewer.model.visionRange;
-      // if it's a surface unit, give it the boost the terrain would provide
-      if( viewer.model.isSurfaceUnit() )
+      // if it's a surface unit, give it the boost the terrain would provide, so long as it's not adjacent-only vision
+      if( (!piercing || range > 1) && viewer.model.isSurfaceUnit() )
```
ThatIsAUsername, line 524: Why `!piercing || range > 1`? Are there any units with piercing vision range > 1? Why should they get the piercing vision boost?

src/Engine/Utils.java
```diff
@@ -92,140 +92,136 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
...
-    if( null == unit || unit.x < 0 || unit.y < 0 )
+    if( null == fff || null == start || start.xCoord < 0 || start.yCoord < 0 )
     {
       System.out.println("WARNING! Finding destinations for ineligible unit!");
       return reachableTiles;
     }
 
-    // set all locations to false/remaining move = 0
-    int[][] costGrid = new int[gameMap.mapWidth][gameMap.mapHeight];
+    // set all locations to unreachable/remaining move = 0
```
ThatIsAUsername, line 121: Nothing is being set equal to zero here (except `i` and `i`, pedant).

src/Engine/MapController.java
```diff
@@ -397,9 +387,10 @@ public void onStateChange()
         contemplatedAction.aiming = false;
         break;
       case PATH_SELECT:
-        if( null != contemplatedAction.actor )
+        XYCoord coord = myGameInputHandler.getUnitCoord();
+        if( null != coord )
         {
-          myGame.setCursorLocation(contemplatedAction.actor.x, contemplatedAction.actor.y);
+          myGame.setCursorLocation(coord.xCoord, coord.yCoord);
```
ThatIsAUsername, line 393: This could just call `myGame.setCursorLocation(coord)`.

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
```diff
@@ -0,0 +1,212 @@
...
+    {
+      launcher = actor;
+      launchee = passenger;
+      cargoAction = passengerAction;
+    }
+
+    @Override
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
+      // UNLOAD actions consist of
```
ThatIsAUsername, line 105: This appears to be pasta'd from UNLOAD.

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
```diff
@@ -0,0 +1,212 @@
...
+      // Generate events.
+      if( isValid )
+      {
+        launchEvents.add(new LaunchEvent(launcher, launchee));
+        GameEventQueue cargoEvents = cargoAction.getEvents(gameMap);
+        launchEvents.addAll(cargoEvents);
+
+        XYCoord firstEndPoint = cargoEvents.peekFirst().getEndPoint();
+        if (null != firstEndPoint && firstEndPoint.equals(launcher.x, launcher.y))
+          // If the first action ends on top of the launcher, assume it's a move action and re-load the unit to avoid weird states
```
ThatIsAUsername, line 127: You explicitly disallow launching to the start point elsewhere - why special-case it here?

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public XYCoord getTargetLocation()
+    {
+      return new XYCoord(launcher.x, launcher.y);
+    }
+
+    @Override
+    public String toString()
+    {
+      return String.format("[Launch from %s]", launcher.toStringWithLocation());
```
ThatIsAUsername, line 148: I feel like this could be more informative.

src/Engine/UnitActionLifecycles/TerraformLifecycle.java
```diff
@@ -0,0 +1,128 @@
...
+import Engine.GameEvents.GameEventQueue;
+import Engine.GameEvents.MapChangeEvent;
+import Terrain.Environment;
+import Terrain.GameMap;
+import Terrain.MapMaster;
+import Terrain.TerrainType;
+import Terrain.Environment.Weathers;
+import Units.Unit;
+
+public abstract class TerraformLifecycle
```
ThatIsAUsername, line 17: One step closer to building a proper Alpha Centauri clone. Yesssss...

src/Engine/UnitActionLifecycles/TerraformLifecycle.java
```diff
@@ -0,0 +1,128 @@
...
+    final int captureAmount;
+    final int priorCaptureAmount;
+
+    public TerraformEvent(Unit u, XYCoord loc, TerrainType endType)
+    {
+      super(loc, Environment.getTile(endType, Weathers.CLEAR));
+      unit = u;
+      XYCoord unitXY = new XYCoord(u.x, u.y);
+      priorCaptureAmount = (unitXY.equals(loc) ? unit.getCaptureProgress() : 0);
+      captureAmount = unit.getHP(); // TODO: Apply CO buffs.
```
ThatIsAUsername, line 114: Do Rigs build more slowly when damaged on the carts too? I don't think I ever noticed that.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -154,99 +154,50 @@ private static void createTerrainSpriteSet(SpriteSetKey spriteKey)
...
     }
     else if( terrainType == TerrainType.SHOAL )
     {
-      ss = new TerrainSpriteSet(terrainType, formatString, w, h);
       ss.addTileTransition(TerrainType.SEA, String.format(transitionFormatString, terrainType.toString().toLowerCase(), "%s", "%s"), w, h);
       ss.addTileTransition(TerrainType.GRASS, String.format(transitionFormatString, terrainType.toString().toLowerCase(), "%s", "%s"), w, h);
       ss.addTileTransition(TerrainType.BRIDGE, String.format(transitionFormatString, terrainType.toString().toLowerCase(), "%s", "%s"), w, h);
       ss.addTileTransition(TerrainType.RIVER, String.format(transitionFormatString, terrainType.toString().toLowerCase(), "%s", "%s"), w, h);
     }
-    else
```
ThatIsAUsername, line 246: Why is this unnecessary?

src/Units/MoveTypes/MoveTypeFey.java
```diff
@@ -0,0 +1,53 @@
...
+import Units.Unit;
+import Engine.FloodFillFunctor;
+import Engine.XYCoord;
+import Engine.FloodFillFunctor.BasicMoveFillFunctor;
+import Terrain.GameMap;
+
+/**
+ * A variant MoveType that is barred from entering non-friendly ownable healing terrain
+ */
+public class MoveTypeFey extends MoveType
```
ThatIsAUsername, line 13: Interesting name - why?

ThatIsAUsername:
Units that resupply while in cargo holds show the resupply animation in the top-left corner of the map. This is weird.

When launching a seaplane, if you move to a possible destination and then hit `BACK`, the menu for selecting a unit to launch is centered over the new location instead of over the launching unit. Hitting `BACK` again leaves the action-select menu centered over the new point.

After selecting a location, hitting `BACK` instead of `WAIT` will center the cursor back on the launching unit (though maybe it shouldn't until `BACK` is hit again?). This one is probably out of scope to change in this PR, it just occurred to me.

src/Engine/GameInput/SelectLaunchable.java
ThislsAUsername: If I recall correctly, I was the one who wrote that logic initially, and I did it to avoid casting `Object` to `Unit` rather than any desire to validate that it's actually in the transport.
IMO it's not that useful to validate everywhere: at this point, it won't get here without being valid... and if it goes through here invalid... it will still fail safely when the action/event gets generated.

src/Engine/Utils.java
ThislsAUsername: Where did these links come from, anyhow. Does Eclipse have a thing for making them?

src/Engine/Utils.java
ThislsAUsername: `!piercing` because that's normal vision. If I got a rocket (vision 1) onto a mountain I'd expect to see 3 spaces.
`range > 1` because everybody gets range 1 piercing vision and they don't get a piercing boost from mountains; by contrast, when your whole vision is piercing then you *do* get a piercing vision boost.

src/Engine/Utils.java
ThislsAUsername: It's an unnecessary clarification to "unreachable", i.e. "You don't have any move here"

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
ThislsAUsername: Bah. Coulda sworn I caught that one.
Probably ended up ctrl-z'ing it and didn't notice.

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
ThislsAUsername: I'd like to defend against the board getting in weird states. Say we move into fog and somehow get ambushed with no empty spaces to move into -- then this logic would be useful.
Also, I was *initially* planning to allow acting on top of the launcher, but then I realized there was no easy way to allow any action in place without letting you plan to e.g. attack from on top of a *different* unit.

src/Engine/UnitActionLifecycles/TerraformLifecycle.java
ThislsAUsername: They totally do. You probably noticed because they're high-priority targets :stuck_out_tongue:

src/Units/MoveTypes/MoveTypeFey.java
ThislsAUsername: I was initially planning to make just `DoRFloat` but that's super lame and uninformative.
This is generic *and* has a really evocative/informative name.

ThislsAUsername:
> Units that resupply while in cargo holds show the resupply animation in the top-left corner of the map. This is weird.
Pretty sure that's not new, lol.
That does remind me I ought to get around to making things die when they hit negative fuel. And have actual unit info UI.

Sigh.

> When launching a seaplane, if you move to a possible destination and then hit `BACK`, the menu for selecting a unit to launch is centered over the new location instead of over the launching unit. Hitting `BACK` again leaves the action-select menu centered over the new point.
> 
> After selecting a location, hitting `BACK` instead of `WAIT` will center the cursor back on the launching unit (though maybe it shouldn't until `BACK` is hit again?). This one is probably out of scope to change in this PR, it just occurred to me.

Sounds like issue-fodder?

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: It happens frequently, since there is no longer a case for every individual terrain type - the `TerrainSpriteSet` is generated unconditionally up above, so the "error" being reported is pretty irrelevant anyhow.

src/Engine/UnitActionLifecycles/TerraformLifecycle.java
ThatIsAUsername: Nah, I noticed because I was reviewing code. I feel like I a) don't often build/see rigs, and b) rarely see them damaged without being quickly destroyed since they aren't front-line units.

src/Engine/UnitActionLifecycles/LaunchLifecycle.java
ThatIsAUsername: I might put that exceptional case in a comment here, just so it's immediately clear how that could happen.

src/Engine/Utils.java
ThatIsAUsername: I don't think Eclipse auto-generates them (or how it would know where to point on its own) - it's just javadoc format to reference another part of the code. That way if you generate html docs or hover the symbol you get a nice hyperlink to follow.

src/Engine/GameInput/SelectLaunchable.java
ThatIsAUsername: May wanna check that memory (`git blame` can help), and also check `SelectMoveLocation` that does a similar check despite not needing to cast.
And you are right, it's probably not critical since the UI enforces particular behavior. I guess I tend to err on the side of caution, because simple checks cost almost nothing, and fast failing is better than slow failing or erroneously appearing to work.

ThatIsAUsername:
Carriers can build up to four seaplanes, but can only hold three. Is that intentional?
Also, the option to build a seaplane doesn't show price.

src/Engine/Utils.java
```diff
@@ -622,4 +615,11 @@ public static boolean enqueueMoveEvent(MapMaster gameMap, Unit unit, Path movePa
     }
     return tilesInRange;
   }
+
+  public static void trimFullLocations(GameMap gameMap, ArrayList<XYCoord> destinations)
```
ThatIsAUsername, line 619: It would be good to have a javadoc comment to explain what this does.

src/Engine/Utils.java
```diff
@@ -92,140 +92,136 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
...
+  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap, boolean) findPossibleDestinations()} **/
+  public static ArrayList<XYCoord> findPossibleDestinations(XYCoord start, Unit unit, GameMap gameMap)
+  {
+    return findFloodFillArea(start, unit.getMoveFunctor(), Math.min(unit.model.movePower, unit.fuel), gameMap);
+  }
   /**
    * Returns the list of XYCoords in gameMap reachable by unit this turn.
    * @param unit The unit to evaluate.
    * @param gameMap The map to search over.
    * @param includeOccupiedSpaces If true, will include spaces occupied by a friendly unit, if some action could end on this space (e.g. LOAD, JOIN).
```
ThatIsAUsername, line 109: Docs here are no longer correct.

src/AI/AIUtils.java
```diff
@@ -169,10 +169,11 @@ public static GameAction moveTowardLocation(Unit unit, XYCoord destination, Game
 
     // Find the full path that would get this unit to the destination, regardless of how long. 
     Path path = Utils.findShortestPath(unit, destination, gameMap, true);
-    if( path.getPathLength() > 0 ) // Check that the destination is reachable at least in theory.
+    ArrayList<XYCoord> validMoves = Utils.findPossibleDestinations(unit, gameMap); // Find the valid moves we can make.
+    Utils.trimFullLocations(gameMap, validMoves);
```
ThatIsAUsername, line 173: It feels a little awkward to have `findPossibleDestinations` return destinations that aren't possible, and then have to trim them as a separate step. Hrm.

src/AI/Muriel.java
```diff
@@ -139,7 +139,7 @@ private UnitMatchupAndMetaInfo getUnitMatchupInfo(Unit myUnit, Unit otherUnit)
     if( null != umami ) return umami;
 
     double myDamage = 0;
-    WeaponModel myWeapon = myUnit.chooseWeapon(otherUnit.model, 1, false);
+    WeaponModel myWeapon = myUnit.chooseWeapon(otherUnit.model, 1, myUnit.model.hasMobileWeapon());
```
ThatIsAUsername, line 142: This is a fundamental change to how Muriel calculates unit matchups.

ThatIsAUsername: Was this to accommodate antitanks? Won't Muriel just ignore them either way?
I think this might only make a difference for BShips, actually.

src/Engine/Path.java
```diff
@@ -245,11 +245,11 @@ public void snip(int newLastPoint)
 
   public void snipCollision(GameMap map, Unit unit)
   {
-    for( int i = 0; i < waypoints.size(); i++)
+    for( int i = waypoints.size()-1 ; i > 0; i--)
     {
       PathNode point = waypoints.get(i);
       Unit obstacle = map.getLocation(point.x, point.y).getResident();
-      if( null != obstacle && unit.CO.isEnemy(obstacle.CO) )
+      if( null != obstacle )
```
ThatIsAUsername, line 252: Why remove the enemy-CO condition?

src/Engine/UnitActionLifecycles/LoadLifecycle.java
```diff
@@ -99,6 +99,9 @@ public GameEventQueue getEvents(MapMaster gameMap)
         }
       }
 
+      // Make it so the passenger actually executes the move
+      movePath.snipCollision(gameMap, passenger);
```
ThatIsAUsername, line 103: The seaplanes now do a move animation to the space adjacent to the transport instead of onto it.

src/Engine/Utils.java
ThatIsAUsername: Are there units with only piercing vision? Also, doesn't this exclude the possibility of units with only piercing vision that can see one space?

src/Units/MoveTypes/MoveTypeFey.java
ThatIsAUsername: Evocative, yes. Informative I might dispute.

ThatIsAUsername:
> Pretty sure that's not new, lol.

Nope, but it also wasn't a case we encountered before.

> Sounds like issue-fodder?

UI for fuel/ammo and making things die on empty yes.
Not moving the cursor when hitting `BACK` while selecting an action after choosing a move space, yes.
Correctly centering menus when backing out of a launch command? That might be within scope here.

src/Engine/GameInput/SelectLaunchable.java
ThislsAUsername: Pretty sure the GameInput rework happened while we were both at the parents' house and we passed code back and forth a bit. `blame` wouldn't know that, and I'm not sure it copes with squash commits either.
I guess it's also possible you wrote it and I generated that reasoning during review.

Works for me.

src/Units/MoveTypes/MoveTypeFey.java
ThislsAUsername: Everybody knows the Fey can't enter your house without your permission.
It's like, a staple of western folklore.
In any case, I *also* put that meaning in the javadoc.

src/AI/AIUtils.java
ThislsAUsername: That's the thing. They *are* possible.
You could quibble that e.g. I should know when a LOAD or JOIN is possible and exclude unit-destinations that aren't eligible for that, but I don't think encoding that knowledge into `findPossibleDestinations` is a great idea either... especially since it wouldn't solve the problem as non-Wally AIs will still just hardcode a WAIT and then be surprised (read: loop infinitely on that action) when it doesn't work.

I imagine it could be fair to build that functionality into `findPossibleDestinations` somehow, but I'm not sure how best to structure it... and I don't want to grow our list of overloads more, either.

src/AI/Muriel.java
ThislsAUsername: It is to accommodate antitanks. I made the smallest possible change to avoid crashing from null ref exception.

This is why, years ago, I asked if `hasMobileWeapon` was what you really wanted when you wrote `hasDirectWeapon`. You said no.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThislsAUsername: Right. Because otherwise there was *no* animation IIRC, since the move failed to validate to execute.
I don't like having validation failures as part of the happy path.

src/Engine/Path.java
ThislsAUsername: Bah, I just realized now that my change probably breaks fog ambushes in some cases.

So. My intent with this change was to snip off anything that would invalidate a move, for its usage in LoadLifecycle below (which should probably also appear in join...)

I guess the big issue here is that *most* of the time, we only want to move into empty spaces... but there are significant and critical exceptions to that logic, and we haven't really codified that anywhere, it's just sorta spread out through the codebase with a bunch of fairly brittle hacks to get stuff to work how it's expected to.

ThislsAUsername:
> Carriers can build up to four seaplanes, but can only hold three. Is that intentional?
> Also, the option to build a seaplane doesn't show price.

Carriers can only hold *two*; you're testing with Strong. I only copied how they function in the sauce material.
I guess I could add a price indicator, but IDK how to do that nicely.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThatIsAUsername: There was an animation, and it went from unit to transport.
There was _also_ a `WARNING! Unable to move Infantry to (6, 6)`

src/AI/AIUtils.java
ThislsAUsername: Note: before the AI-fix commit, Wally would too, I just fixed him differently than I fixed the others 'cause I'm a silly boy.

src/AI/Muriel.java
ThatIsAUsername: I meant it: `hasDirectWeapon` is what I intended. I deliberately kept Muriel from knowing about ranged attacks until I could decide how exactly to handle them.

src/Engine/Utils.java
ThislsAUsername: Not at the moment, no. I guess that's a possibility.
Sonja/Lin are implemented to make piercing == normal after the buff.

It does exclude the possibility of units that *should* be buffed with complete piercing vision that only have 1 vision, but in practice, piercing vision only comes with a general vision buff.

src/AI/AIUtils.java
ThatIsAUsername: They are only possible if an open transport of the right type is in the space. This didn't cause infinite-loop actions before because we checked for that condition.

src/Engine/Path.java
ThatIsAUsername: > So. My intent with this change was to snip off anything that would invalidate a move

If you move into a transport with an adjacent unit, the move animation goes onto the adjacent unit and you still get the validation failure/WARNING message.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThatIsAUsername: But the current implementation can also trigger that warning by moving over a unit onto a transport.

src/Engine/Path.java
ThislsAUsername: Right, which makes it not-a-very-good change.
Which is true largely because of my last paragraph there.

I feel like we need a DRY solution for this.

src/AI/Muriel.java
ThislsAUsername: I'm well aware, but then you did not guard against *buying* antitanks... and in any case, you could start with them on the map and you'd still crash without this change.

src/AI/AIUtils.java
ThislsAUsername: If so, I don't know where that check is. Every AI had this infinite-loop problem.

ThatIsAUsername: We didn't check for it in AI. We checked for it when finding destinations. It just wasn't possible for the AI to choose an invalid move space. The walkable spaces with friendly units on them didn't get highlighted like they do now.

src/Engine/Utils.java
```diff
@@ -92,21 +92,25 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
     }
   }
 
-  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap, boolean) findPossibleDestinations()} **/
+  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap) findPossibleDestinations()} **/
   public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gameMap)
   {
     return findPossibleDestinations(new XYCoord(unit.x, unit.y), unit, gameMap);
```
ThatIsAUsername, line 98: Is there any reason not to just make the call straight to `findFloodFillArea` instead of taking a detour through an overload?

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThislsAUsername: Fair enough.

src/Engine/Utils.java
ThislsAUsername: Both overloads are used and it prevents having to duplicate the logic to find `initialFillPower`.

src/AI/AIUtils.java
ThislsAUsername: Wow, okay. I don't know how I missed that some invocations did actually pass in false to `findPossibleDestinations`.

I'm gonna build that check into FloodFillFunctor.

src/Engine/Utils.java
ThatIsAUsername: Eh, good enough I guess.

src/AI/Muriel.java
ThatIsAUsername: Well, maybe I need to review that as part of whatever upgrades I ever find time for one day.

We can leave it like this for now; I don't think the change will affect much, practically speaking.

src/AI/AIUtils.java
```diff
@@ -26,19 +26,25 @@
 
 public class AIUtils
 {
+  /** Overload for {@link #getAvailableUnitActions(Unit, GameMap, boolean)} */
```
ThatIsAUsername, line 29: Comment should probably state or `@note` or something that is assumes `includeOccupiedDestinations==false`.

ThatIsAUsername: Actually, hold up. Why does this need an overload at all? The prior implementation just assumed `includeOccupiedDestinations==true` in `getAvailableUnitActions` -- if the unit can't board transports, those actions won't be found anyway.

src/AI/AIUtils.java
```diff
@@ -170,11 +175,11 @@ public static GameAction moveTowardLocation(Unit unit, XYCoord destination, Game
 
     // Find the full path that would get this unit to the destination, regardless of how long. 
     Path path = Utils.findShortestPath(unit, destination, gameMap, true);
-    if( path.getPathLength() > 0 ) // Check that the destination is reachable at least in theory.
+    ArrayList<XYCoord> validMoves = Utils.findPossibleDestinations(unit, gameMap, false); // Find the valid moves we can make.
```
ThatIsAUsername, line 178: Why not move this down to just above line 183 to where it's actually used? No reason to do this calculation if we don't need to. I know you check that `validMoves.size() > 0`, but that is trivially true if `path.getPathLength() > 0`.
I also don't like to pass in magic values; I'd declare a named temporary and pass that in just so it's obvious by reading what `false` means in this context.

src/AI/AIUtils.java
```diff
@@ -227,7 +232,7 @@ public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Comman
   {
     XYCoord origin = new XYCoord(unit.x, unit.y);
     Map<XYCoord, Double> shootableTiles = new HashMap<XYCoord, Double>();
-    ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, false);
+    ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, true);
```
ThatIsAUsername, line 235: This is your function, but you changed the default value when you put the boolean back in.
Also it's a magic value blah blah.

src/AI/InfantrySpamAI.java
```diff
@@ -184,24 +185,12 @@ public GameAction getNextAction(GameMap gameMap)
...
-        // and build a GameAction to move to the closest one.
-        boolean includeTransports = false;
-        ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeTransports);
-        Utils.sortLocationsByDistance(goal, destinations);
-        XYCoord destination = destinations.get(0);
-        Path movePath = Utils.findShortestPath(unit, destination, gameMap);
-        GameAction move = new WaitLifecycle.WaitAction(unit, movePath);
-        actions.offer(move);
-        break;
+        GameAction move = AIUtils.moveTowardLocation(unit, goal, gameMap);
```
ThatIsAUsername, line 188: ISAI wasn't using this already? Silly me.

src/AI/WallyAI.java
```diff
@@ -400,7 +399,7 @@ else if( (null == resident) && myCo.unitProductionByTerrain.containsKey(gameMap.
         boolean foundAction = false;
 
         // Find the possible destinations.
-        ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, true);
+        ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, false);
```
ThatIsAUsername, line 402: This `false` used to be `true`, FYI.

src/Engine/Utils.java
```diff
@@ -93,17 +93,17 @@ public static void highlightLocations(GameMap map, ArrayList<XYCoord> locations)
   }
 
   /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap) findPossibleDestinations()} **/
-  public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gameMap)
+  public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gameMap, boolean includeOccupied)
   {
-    return findPossibleDestinations(new XYCoord(unit.x, unit.y), unit, gameMap);
+    return findPossibleDestinations(new XYCoord(unit.x, unit.y), unit, gameMap, includeOccupied);
```
ThatIsAUsername, line 98: Just as an aside, `includeOccupiedSpaces`, though slightly more verbose, is definitely more descriptive.

src/AI/AIUtils.java
ThislsAUsername: JOIN is also an action that exists. For a bit there at least, Wally would use it to move units out of the way of other units' attacks.

src/AI/AIUtils.java
ThislsAUsername: I'm not sure that it *is* trivially true. If I'm trapped in a sea of friendly units, I'll see a way out with infinite movement but won't actually be able to make any progress.

src/AI/WallyAI.java
ThislsAUsername: Only ATTACK and CAPTURE are considered in that block, though. I guess I should do the temp var thing anyway.

src/AI/InfantrySpamAI.java
ThislsAUsername: ISAI was probably where the logic was written first, lol

src/Engine/UnitActionLifecycles/LoadLifecycle.java
```diff
@@ -168,6 +168,7 @@ public void performEvent(MapMaster gameMap)
         gameMap.removeUnit(unitCargo);
         unitCargo.x = -1;
         unitCargo.y = -1;
+        unitCargo.isTurnOver = true;
```
ThislsAUsername, line 171: This is why I probably did the adjacent-units thing before...

src/AI/AIUtils.java
ThislsAUsername: By the way: Eclipse *does* have a function to generate them automatically.
If you enter # while in a javadoc, it will autoprompt/complete signature links for you.

ThatIsAUsername: Isn't the existence of JOIN just another argument to include occupied destinations when finding unit actions?

src/Engine/MapController.java
```diff
@@ -301,8 +301,8 @@ private void handlePathSelect(InputHandler.InputAction input)
         }
         break;
       case BACK:
-        myGameInputHandler.back();
         contemplatedAction.movePath = null;
+        myGameInputHandler.back();
```
ThatIsAUsername, line 305: Moving this doesn't appear to change anything.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThatIsAUsername: haks r badd

src/Engine/Utils.java
```diff
@@ -571,10 +570,13 @@ public static void sortLocationsByTravelTime(Unit unit, ArrayList<XYCoord> mapLo
...
-    for( PathNode point : path.getWaypoints() )
+    boolean includeOccupiedSpaces = true; // Shouldn't matter, as we don't invoke canEnd()
+    FloodFillFunctor fff = unit.getMoveFunctor(includeOccupiedSpaces);
+    for( int i = 1; i < path.getPathLength(); ++i)
     {
-      Unit obstacle = map.getLocation(point.x, point.y).getResident();
-      if( null != obstacle && unit.CO.isEnemy(obstacle.CO) )
+      XYCoord from = path.getWaypoint(i-1).GetCoordinates();
+      XYCoord to   = path.getWaypoint( i ).GetCoordinates();
+      if( 0 > fff.getRemainingFillPower(map, unit.model.movePower, from, to) )
```
ThatIsAUsername, line 579: Do we really need to monitor move power here? This function is just checking if the path goes through something the unit can't go through. Actually, we just call `getRemainingFillPower` without ever updating fill power, so we aren't even monitoring move distance. Is that intentional? If so, it's misleading anyway.
In fact, since `includeOccupiedSpaces` is true, it appears that this wouldn't detect trying to walk through enemies.

src/Engine/GameInput/SelectLaunchable.java
ThatIsAUsername: Yeah, maybe. It may also be necessary since we want to validate UI input wherever it's input, but we also want to check action/event validity since the AI doesn't use the UI to build them. I don't want to break the existing working paradigm without properly thinking through the consequences. It may not be as elegant as it could be, but it works.

src/Engine/Utils.java
ThatIsAUsername: All right, sustained. Mostly because I know it would probably be a silly amount of work to force it to behave consistently, and as you say this should be fine in practice.

src/AI/AIUtils.java
ThatIsAUsername: All righty.

src/AI/AIUtils.java
ThislsAUsername: Right, but I didn't feel like touching all of the invocations and none of our present AIs are actually interested in joining or loading.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThislsAUsername: Hacks are bad, yes.
I suppose the *correct* fix would be to make MoveEvent unconditionally do the turn-over, fuel-burn, and fog-reveal?
At that point you'd get everything but the actual change in position, which is obviated by the Join or Load that will be right after it anyhow.

src/Engine/Utils.java
ThislsAUsername: We're not monitoring move power, really. Yeah, I probably should add a comment.
`getRemainingFillPower()` is being used to determine traversability... which includes being blocked by enemy units in the usual case (but we could totally add a new movetype that goes through enemies and it would work just fine).
`includeOccupiedSpaces` is *only* for the "can I validly end up here" check, not the traversability check (and I am open for suggestions on better nomenclature).

src/Engine/MapController.java
ThislsAUsername: I check if `movePath` is `null` in `onStateChange()`, and use that to determine whether (on a backwards input) I should put the cursor at the end of the movepath or on the acting unit.
Go figure.

ThislsAUsername:
...There was a funny looking button, so I clicked it. Go figure.

src/AI/AIUtils.java
ThatIsAUsername: I'm not understanding something, or maybe you aren't. The function is called `getAvailableUnitActions`, but you have modified it to ignore unit actions that can end on friendly units for no reason I can discern. I don't even know why this function needs an overload to allow specifying that we *don't* want to include those results; it didn't use to have one, and nobody uses it.
The 3-param overload is only ever invoked by the 2-param overload, and the 2-param overload is only ever called from ISAI. "Touching all the invocations" would just involve removing the superfluous overload and changing the other to assume `includeOccupiedDestinations==true`.
I don't want to leave this function returning different results than advertised, when it did what was expected before and nothing in this PR depends on it changing functionality.

src/Engine/MapController.java
ThatIsAUsername: I still can't elicit a difference in behavior based on this code change Does it change anything for you?
While messing with this though, I did discover that if you select a unit, then deselect it, then go into the stats screen and back out; the cursor will center on the unit you selected, even though there is no reason it should.

src/Engine/Utils.java
ThatIsAUsername: If we aren't monitoring move power, why bring in FFF at all? The new formulation for this path-collision check is more verbose and less clear in intent than the old formulation.

ThislsAUsername: The reasoning is that FFF is now the 'canonical' way of finding move cost, which creates a new consistent branch point for extensibility.
If I wanted to make a unit type that can move through other units right now, I could. It would work transparently through these functions because they no longer explicitly check for enemy units, but work through the FFF.

If you'd like the FFF function signatures to change in some way to make things more clear, that's fine, but ideally the MoveType should no longer be used directly for practical purposes.
Wally doesn't use it for his analysis, but I'd have to do a lot of meddling to make that work well anyhow.

ThatIsAUsername: grumble grumble fine.

src/Engine/GameEvents/MoveEvent.java
```diff
@@ -44,34 +43,30 @@ public void performEvent(MapMaster gameMap)
...
-
-        // reveal fog as applicable
-        for( Commander co : gameMap.commanders )
-        {
-          co.myView.revealFog(unit, unitPath);
-        }
+      // reveal fog as applicable
+      for( Commander co : gameMap.commanders )
+      {
+        co.myView.revealFog(unit, unitPath);
```
ThatIsAUsername, line 58: Shouldn't we wait until we know where the unit is moving before clearing fog?

src/Engine/Path.java
```diff
@@ -164,14 +163,18 @@ public int getPathLength()
   /**
    * @return the amount of fuel it would cost to travel this path with the given unit type 
   **/
-  public int getFuelCost(UnitModel model, GameMap map)
+  public int getFuelCost(Unit unit, GameMap map)
```
ThatIsAUsername, line 166: Can we change `FloodFillFunctor` to operate on a `UnitModel` instead of needing an actual unit? Maybe have an optional `movePower` param int the constructor to make it place nice. I don't like requiring information that isn't needed for the task being performed, which in this case (and the other one we are bickering over) causes unneeded complication.
I know you want FFF to be *the* way that moves are evaluated, but this is another case where the old formulation looks much more straightforward.

src/Test/TestUnitMovement.java
```diff
@@ -161,23 +161,23 @@ private boolean testFuelCosts()
     multiPath.addWaypoint(2, 5);
 
     // Make sure the action didn't actually execute.
-    boolean testPassed = validate(grassPath.getFuelCost(testCo1.getUnitModel(UnitModel.TROOP), testMap) == 7,
+    boolean testPassed = validate(grassPath.getFuelCost(new Unit(testCo1, testCo1.getUnitModel(UnitModel.TROOP)), testMap) == 7,
```
ThatIsAUsername, line 164: Not a huge fan of this either, especially since it was precipitated by a change that also feels unnecessary.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
```diff
@@ -168,7 +168,6 @@ public void performEvent(MapMaster gameMap)
         gameMap.removeUnit(unitCargo);
         unitCargo.x = -1;
         unitCargo.y = -1;
-        unitCargo.isTurnOver = true;
```
ThatIsAUsername: Didn't you just add this? Why are you removing it again?

src/Engine/GameEvents/MoveEvent.java
ThislsAUsername: We already have the full `unitPath`, though?
If you're talking about preemption, that's done during event construction.

src/Engine/Path.java
ThislsAUsername: >`UnitModel`
No. It needs to know the affiliation of the unit, so it can be blocked by opposing forces.
In addition, it needs to know the *exact* unit that is moving, so it can always allow ending on that unit (regardless of whether you're allowing it to end on *other* units).

>`movePower`
Not a bad idea; I guess it'd just be used for an overload that doesn't take `initialFillPower`?
Removing it from the function inputs entirely would be annoying since `Utils` is using it.

>old fomulation is more straightforward
It wouldn't be unreasonable to do a cost-output function.

src/Test/TestUnitMovement.java
ThislsAUsername: Can't argue with that.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThislsAUsername: You said `haks r badd`.
So I fixed the thing the hack was bandaging.

The reason I added it was because the MoveEvent was failing and not ending the turn of units that Load... which meant air units could immediately Launch the same turn.

ThatIsAUsername: OK, now I'm confused. I thought you said you had done this to fix the thing that forced you to move adjacent-then load or something, though in retrospect I'm not sure why that would affect it.
It seems perfectly logical to have turn-ending as a part of a load action; I'm not sure why this would be a hack.
What am I missing (on both sides of this change)?

src/Engine/GameEvents/MoveEvent.java
ThatIsAUsername: Yes, and based on the pre-PR version of this function, the event should be ready to fire. You added a new check below to ensure `fuelBurn <= unit.model.movePower && etc.` before actually executing the move.
Either we need that check (in which case we should wait to clear fog until we know for sure we are moving) or we don't need that check.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
ThislsAUsername: At first, there weren't any particular changes to ensure `unitCargo.isTurnOver == true` around the move->load sequence. This allowed boarding and Launching from a carrier on the same turn.
Then I added a call to snip the move event in load actions (and fiddled with the path snip function in a bad way). This fixed the issue by pointing the loading unit to go adjacent to the transport, and all was well. Then you pointed out that you'd the the same behavior by just moving through a friendly, so I reverted the commit and added the above line (which did work).
(Then I re-updated the snipping in a better way so it should be resilient to pre-emptions on invalid spaces)
Then I made sure `MoveEvent` always sets `unitCargo.isTurnOver = true`, so I reverted this line as now unnecessary.

For the record, the `unitCargo.x/y = -1;` are there because `removeUnit()` won't set them that way if it fails to remove the unit. (this can happen if I e.g. Launch from a Carrier into a Cruiser) I tried changing that in `removeUnit()`, and it made some test cases fail for no apparent reason, so I stopped poking it :P

src/Engine/Path.java
ThatIsAUsername: > No. It needs to know... In addition, it needs to know

That's my point though. `getFuelCost` *doesn't* need to know any of that stuff. You are placing more constraints on this function and making it harder to use (as evidenced by the test file changes) for no reason other than that you want to use FFF to accomplish the task. That's enough reason for me to object even if the change didn't make the function more verbose and harder to read as well.
I get that you may want to have a move-cost function that requires a `from`/`to` pair so we can disallow certain transitions (at least, that's the only reason I can think of to not just have a `to` field), but this just feels like a case of "I have a new hammer, and what are screws?".

src/Engine/GameEvents/MoveEvent.java
ThislsAUsername: I believe the move execution function does validate that you can do the actual move too, so I don't think the check is actually needed -- there are safeguards earlier up the chain that should prevent bad things from happening.
I was optimizing for the case (as detailed below) in which we fire a `MoveEvent` onto an occupied space (i.e. Load or Join). Logically, the fuel cost/end turn/fog reveal *should* still occur in that case.

src/Engine/Path.java
ThislsAUsername: > You are placing more constraints on this function and making it harder to use (as evidenced by the test file changes) for no reason other than that you want to use FFF to accomplish the task.

That's accurate, but I want to use FFF because to do otherwise could generate inconsistent results.
Let's say I add a hovercraft unit, which can go from `SEA` to `GRASS` at a cost of 4 move (but otherwise moves at 1 cost on either). The *actual* fuel cost would be drastically lower than it should be, since this function would be used to calculate it, in a world where FFF was not used.

> I get that you may want to have a move-cost function that requires a from/to pair so we can disallow certain transitions (at least, that's the only reason I can think of to not just have a to field), but this just feels like a case of "I have a new hammer, and what are screws?".

I suppose part of it is that I've had this sort of capability gnawing at the back of my brain for longer than we've even *had* this game.
As for "I have a new hammer, and what are screws?", I would argue the screws are actually functionally inappropriate for what I have planned.

ThatIsAUsername: >  because to do otherwise could generate inconsistent results

And that's fair. But I should not need a hovercraft to be able to ask "How much fuel does it take for a hovercraft to climb from `SEA` to `GRASS`? The fact that we *do* need one now feels like a regression.

> I would argue the screws are actually functionally inappropriate for what I have planned.

Aye, and I don't doubt that's true. But hammers and nails aren't called for here.

src/Engine/GameEvents/MoveEvent.java
ThatIsAUsername: Well, then--and write this down, because I don't say this often--I think you should remove the check so that it doesn't look like our logic is inconsistent and cause questions when we look at the function in six months.

src/Engine/Path.java
ThislsAUsername: > And that's fair. But I should not need a hovercraft to be able to ask "How much fuel does it take for a hovercraft to climb from `SEA` to `GRASS`? The fact that we _do_ need one now feels like a regression.

What alternative would you propose? I guess I could add null checks so it doesn't *need* the unit field, so you could leave it null if you want?
My intent is to centralize control and knowledge of where a unit can go and how, to the greatest feasible extent.

src/AI/AIUtils.java
ThislsAUsername: > I'm not understanding something, or maybe you aren't. The function is called `getAvailableUnitActions`, but you have modified it to ignore unit actions that can end on friendly units for no reason I can discern.

I believe I made the initial modification after observing the AIs get stuck in an infinite loop trying to send a unit to an occupied space. This "use case" is actually not as bad anymore, since the MoveEvent *will* end the offending unit's turn even if the move is invalid (so the game can continue).

> I don't even know why this function needs an overload to allow specifying that we _don't_ want to include those results; it didn't use to have one, and nobody uses it.

In addition to ISAI, it's invoked by `getAvailableUnitActionsByType()`, which is used by Muriel.

> I don't want to leave this function returning different results than advertised, when it did what was expected before and nothing in this PR depends on it changing functionality.

I guess I don't have any strong objections now that an infinite loop of failed Wait actions is no longer possible with a single unit.

src/Engine/MapController.java
ThislsAUsername: Get a DoR Carrier and put something in it.
Select Launch, and move the cursor around. Hit BACK.
With this swap, the cursor centers on the Carrier. Without it, it stays where it was (and renders the launch select menu there).

It's not even just the stats screen; just select/deselect, open the meta actions menu, and back out. What's happening is that when you *open* the meta actions menu, you hit `MapController` line 383 because your `GameInputState`'s holding onto the last unit you selected. You just don't see the result until you *close* the menu.
In any case, I just fixed it by pushing up a change to line 407, which *also* fixes the last unit you selected being hyper.

src/Engine/FloodFillFunctor.java
```diff
@@ -0,0 +1,69 @@
...
+    public BasicMoveFillFunctor(Unit mover, boolean includeOccupied, boolean canTravelThroughEnemies)
+    {
+      unit = mover;
+      this.includeOccupiedDestinations = includeOccupied;
+      this.canTravelThroughEnemies = canTravelThroughEnemies;
+    }
+
+    public int getRemainingFillPower(GameMap map, int initialFillPower, XYCoord from, XYCoord to)
+    {
+      int cost = findMoveCost(from, to, map);
```
ThatIsAUsername, line 40: Shouldn't we check that the map location is valid before we try to find the move cost?
It seems like we'd want
```
if( !map.locationValid(to) )
{...}
else if( !canTravel ... )
{...}
else
  cost = findMoveCost()
```

src/Engine/Path.java
ThatIsAUsername: > What alternative would you propose?

I would change the `getFuelCost` function to take a `UnitModel` again.

> Ninjas 

Sure, and I'm sure we could come up with yet more additional possible features that would require more feature creep (e.g. Pirates who can board/commandeer enemy boats, which could require the ability to end a move on an enemy). Right now though, we don't have that use case, and when (if) we get there we could easily add an overload that takes map residents and traveler allegiance into account. I don't want to raise the barrier to asking simple questions.

src/Engine/FloodFillFunctor.java
```diff
@@ -0,0 +1,69 @@
...
+ * Provides a stateless interface for determining traversability
+ * Used in {@link #findShortestPath(XYCoord, Unit, int, int, GameMap, boolean) findShortestPath()} and {@link #findPossibleDestinations(XYCoord, Unit, GameMap, boolean) findPossibleDestinations()}
+ */
+public interface FloodFillFunctor
+{
+  /**
+   * Determines whether the Location (x, y), can be traveled through.
+   * @param ignoreUnits If set, this function will ignore enemy-unit presence.
+   */
+  int getRemainingFillPower(GameMap map, int initialFillPower, XYCoord from, XYCoord to);
```
ThatIsAUsername, line 17: Why `getRemainingFillPower`? That kind of makes it sounds like FFF is managing the fill power internally. It seems more natural to use something like `getTransitionCost`, and then we don't even need to pass in a starting fill power.
`newPower = getRemainingFillPower(oldPower)`
would become
`newPower = oldPower -= getTransitionCost()`
which frankly reads more naturally to me.

src/AI/AIUtils.java
ThatIsAUsername: > AIs get stuck in an infinite loop... MoveEvent will end the offending unit's turn

If AIs are trying to execute `WAIT` commands derived from `JOIN` or `LOAD` actions, then I would think that is the thing to fix, and modifying a MOVE action to always end the unit's turn no matter what is fine I guess, but is a superficial patch fix. That's beyond scope for this PR though, other than to maybe record the issue.
Also, badly-constructed actions should just end the AI's turn altogether (so maybe that's another issue to fix).

> In addition to ISAI, it's invoked by getAvailableUnitActionsByType(), which is used by Muriel.

Touchè, mon frère.

src/Engine/MapController.java
ThatIsAUsername: Cool. That silly hyper-unit thing has been around for a while.
Although... If/when we add missiles to the game, I expect they will also use `FREE_TILE_SELECT`, so that's something to consider if that state assumes it's the default non-acting state.
I may be OK with leaving that problem to future us though.

src/Test/TestGameEvent.java
```diff
@@ -244,11 +244,7 @@ boolean testMoveEvent()
     testPassed &= validate(7 == mech.x && 6 == mech.y, "    Mech should think he is at (7, 6) after moving.");
     testPassed &= validate(testMap.getLocation(7, 6).getResident() == mech, "    Mech is not at (7, 6) after moving.");
 
-    path.addWaypoint(7, 0); // New endpoint over water.
-    new MoveEvent(mech, path).performEvent(testMap); // This should not execute. Water is bad for grunts.
-    testPassed &= validate(7 == mech.x && 6 == mech.y, "    Mech does not think he is at (7, 6), but should.");
-    testPassed &= validate(testMap.getLocation(7, 6).getResident() == mech, "    Mech is not still at (7, 6), but should be.");
-    testPassed &= validate(testMap.getLocation(7, 0).getResident() == null, "    Location (7, 0) should still be empty.");
+    // Test moving into water removed since MoveEvent no longer validates that the unit can actually get there.
```
ThatIsAUsername, line 247: Huh. I misunderstood when we were talking before. I thought you were saying we had a test case that was *accidentally* trying to move a unit over water. I thought the fix for the "broken test case" would be to change the destination so it was a valid move.
Since this test is doing it on purpose to validate that you can't do that, we should probably keep it around (and re-add the corresponding validity checks to `MoveEvent`).
I can do that if you want since I'm the one who inadvertently told you to tear it out.

ThislsAUsername: I don't really care that much; as noted, the check not being there doesn't really break anything (as MoveEvents are generated by a process which *does* validate)

ThislsAUsername: I figure an equivalent test could also be constructed using `enqueueMoveEvent()`

ThatIsAUsername: Sure, but it's also possible to construct a MoveEvent without going through `enqueueMoveEvent()`.
I mean, we probably can live without this test and the checks that make it pass, but I hate to remove a valid test case.

ThatIsAUsername:
It looks like it's possible to LAUNCH a seaplane and LOAD it back into the same carrier in the same turn. It then takes up two cargo slots.
It also looks like the RESUPPLY animation still triggers half off-screen the turn after building a seaplane.

src/Engine/UnitActionLifecycles/LoadLifecycle.java
```diff
@@ -168,6 +168,7 @@ public void performEvent(MapMaster gameMap)
         gameMap.removeUnit(unitCargo);
         unitCargo.x = -1;
         unitCargo.y = -1;
+        unitTransport.heldUnits.remove(unitCargo); // Make sure we can't double-load
```
ThatIsAUsername, line 171: Why is this needed? Shouldn't the unload happen as part of the `LAUNCH` event?
I'm still not even sure why we allow launching back to the launching ship.

ThislsAUsername: Yes, actually. However, the `LAUNCH` assumes that moves ending on its space are *not* followed by a `LOAD`, so it queues another `LOAD` at the end.
The double `LOAD` is why you ended up seeing two spaces taken up. This fixes that, though most paths for that to happen are now closed off (You could have an AI manually construct a `LAUNCH` like that, but you can't do it in the UI and `LaunchFactory` isn't gonna do it for you)

ThislsAUsername:
Are we happy with this branch? I think I am.

PR #243, opened 2020-05-05
Re-added MoveEvent test case and validation.

ThatIsAUsername:


src/Engine/GameEvents/MoveEvent.java
```diff
@@ -45,22 +45,25 @@ public void performEvent(MapMaster gameMap)
       Path.PathNode endpoint = unitPath.getEnd();
       int fuelBurn = unitPath.getFuelCost(unit.model, gameMap);
 
-      unit.fuel = Math.max(0, unit.fuel - fuelBurn);
+      boolean includeOccupiedSpaces = false;
+      if( fuelBurn <= unit.fuel && fuelBurn <= unit.model.movePower
+          && unit.getMoveFunctor(includeOccupiedSpaces).canEnd(gameMap, endpoint.GetCoordinates()))
```
ThislsAUsername, line 50: This breaks the thing I was originally fixing by moving stuff *out* of the check.
If you LOAD, the movement should expend your turn and fuel, and reveal fog. This goes double for transports with LAUNCH, as obviously that gets exploity.

ThatIsAUsername: Ach, true. OK, should be fixed. I had to slightly weaken the validation (I think it would validate even when moving onto enemies now), but no existing tests broke, and I prevented the warning-on-load that would normally occur.
I feel like we have a sloppy abstraction with how we handle movement destinations that we could probably iron out if we tried.

ThatIsAUsername: I actually kind of want t just remove `includeOccupiedSpaces` altogether; it's almost never set to `false`. If we do that and redefine the conditional in `BasicMoveFillFunctor.canEnd()` to be
```java
  if( obstacle == null || obstacle == unit
      || obstacle.hasCargoSpace(unit.model.role)
      || (obstacle.model == unit.model && obstacle.getHP() < obstacle.model.maxHP) )
```
then it would just recognize that `LOAD`/`JOIN` destinations do in fact represent valid moves, which seems like it would simplify quite a bit. The only reason I didn't is because Wally actually uses `includeOccupiedSpaces==false` when traveling, and I didn't want to make changes that might affect his behavior.

ThislsAUsername: I really don't like the idea of hard-coding notions of what units are valid destinations supported directly into this code, though. Not every unit is guaranteed to be able to join or load; those are actions.
It could, perhaps, query the unit to check if it has any valid actions there.

ThislsAUsername: As for `includeOccupiedSpaces==false`, `Path.snipCollision()` also uses it to deal with finding a proper space to stop in after an ambush when moving over spaces that you can't stop in.
(The carts don't actually run into this issue in practice because there are no traversible-but-un-endible terrain types, and all units you can move through share their piercing vision with you.)

Both Spender and Wally use it; Spender uses it because he doesn't want to think about join/load, Wally uses it because the rest of the logic would have him join/load when attempting to travel... which isn't necessarily wrong, just weird.

Finally, `AIUtils.moveTowardLocation()` uses it (which ISAI and Muriel use).

ThatIsAUsername: > Not every unit is guaranteed to be able to join or load; those are actions.

Right, but the change accounts for that and won't count destinations if the unit can't join/load. Unless you are proposing future unit types that break the mold.

> there are no traversible-but-un-endible terrain types

Do we have those?

But point sustained, at any rate, which is why I didn't change FFF this way.
Do you have any issues with the MoveEvent change?

ThislsAUsername: > Right, but the change accounts for that and won't count destinations if the unit can't join/load. Unless you are proposing future unit types that break the mold.

I was referring to possible future unit types that break the mold, yes.
Mostly, I just don't want to have the conditions for specific types of actions in code that's only designed to handle movement; if I add a fun new action type that involves moving onto another unit, then I'll have to add *that* to the conditions here as well... and it probably won't be universal, so that makes the logic even more questionable.

> > there are no traversible-but-un-endible terrain types
> Do we have those?

No, but the system now supports them.
It may be worth noting that Wargroove sea units treat bridges this way. Their solution for ambushes is to just insta-gib the unit if it ends on a non-stayable, and I'm not a super fan of that lol.
> Do you have any issues with the MoveEvent change?

Uhhh. Seems fine?

ThatIsAUsername: >  I just don't want to have the conditions for specific types of actions in code that's only designed to handle movement

Seems reasonable enough.

PR #245, opened 2020-05-09
Added cost to seaplan build option.

ThatIsAUsername:


src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
```diff
@@ -43,7 +43,7 @@ public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor,
     @Override
     public String name()
     {
-      return "BUILD " + typeToBuild.toString();
+      return String.format("BUILD %s (%d)", typeToBuild.toString(), typeToBuild.getCost());
```
ThislsAUsername, line 46: That's *really* wide. I guess it doesn't matter that much, but it kind of hurts my sensibilities lol

ThatIsAUsername: It's as wide as it needs to be. Plus it probably looks wider just because there aren't many other options so it's short.
Pretty sure nobody's going to complain that the menu's are proportioned as pleasingly as they would prefer, but I *would* anticipate complaints if folks don't know how much they are spending.

PR #247, opened 2020-05-10
Combat math refactor

ThislsAUsername:
"I'm pretty much done with coding this," I said
"I'll just do a little testing and put up the PR," I said

Testing has a way of revealing problems/possible improvements, heh.

I kinda want to make `Unit.getPreciseHP()` return an int in [1,100], but I feel like trying to juggle HP values in more than one value scale across our whole code base is a mistake.
I thought about making the health an object, but I don't think it'd make it any better.

Thoughts?


src/Engine/Combat/CombatEngine.java
```diff
@@ -25,10 +34,74 @@ public static BattleSummary calculateBattleResults( Unit attacker, Unit defender
...
+    int battleRange = Math.abs(attackerX - defenderX) + Math.abs(attackerY - defenderY);
+    WeaponModel attackerWeapon = attacker.chooseWeapon(defender.model, battleRange, attackerMoved);
+    WeaponModel defenderWeapon = null;
+    // Only attacks at point-blank range can be countered
+    if( 1 == battleRange )
+      defenderWeapon = defender.chooseWeapon(attacker.model, battleRange, false);
+
+    CombatContext context = new CombatContext(map, attacker, attackerWeapon, defender, defenderWeapon, battleRange, attackerX, attackerY);
+    
+    // If the attacker and defender get swapped, we want to still give a coherent BattleSummary.
```
ThatIsAUsername, line 61: Does this happen often?

src/Engine/Combat/CombatEngine.java
```diff
@@ -25,10 +34,74 @@ public static BattleSummary calculateBattleResults( Unit attacker, Unit defender
...
+      // New battle instance with defender counter-attacking.
+      double counterHP = Math.ceil(context.defender.getPreciseHP() - defenderHPLoss); // Account for the first attack's damage to the now-attacker.
+      BattleParams defendInstance = StrikeParams.getCounterAttack(context, counterHP);
+
+      attackerHPLoss = defendInstance.calculateDamage();
+      unitDamageMap.put(context.defender, new AbstractMap.SimpleEntry<WeaponModel,Double>(context.defenderWeapon, attackerHPLoss));
+      if( !isSim && attackerHPLoss > context.attacker.getPreciseHP() ) attackerHPLoss = context.attacker.getPreciseHP();
+    }
+    
+    // Calculations complete. We are setting up our BattleSummary, so go back to using known-accurate info
```
ThatIsAUsername, line 89: Are we using potentially-inaccurate info leading up to this?

src/Engine/Combat/CombatEngine.java
ThislsAUsername: Venge SCOP

src/Engine/Combat/CombatEngine.java
ThislsAUsername: Depends on what you mean by "accurate".
In my case, I meant that it hasn't been mucked with by the COs so your attacker/defender are based on who started the fight rather than who got firststrike

src/Engine/Combat/CombatEngine.java
```diff
@@ -25,10 +34,74 @@ public static BattleSummary calculateBattleResults( Unit attacker, Unit defender
...
+    return new BattleSummary(attacker, unitDamageMap.get(attacker).getKey(),
+                             defender, unitDamageMap.get(defender).getKey(),
+                             map.getEnvironment(attackerX, attackerY).terrainType,
+                             map.getEnvironment(defenderX, defenderY).terrainType,
+                             unitDamageMap.get(defender).getValue(), unitDamageMap.get(attacker).getValue());
   }
 
+  public static BattleParams buildSimpleAttack( Unit attacker, int battleRange, Unit defender, GameMap map, int terrainStars, boolean attackerMoved )
```
ThatIsAUsername, line 97: Is the attack "simple" just because it doesn't account for the counterattack?
I'd almost prefer to call this `calculateAttackDamage` or something, and just have it return just the pain rather than the whole `BattleParams` (since nobody uses that at this point anyway).

src/Engine/Combat/StrikeParams.java
```diff
@@ -0,0 +1,110 @@
...
+      this.terrainStars = terrainStars;
+      defenderHP = defender.body.getHP();
+
+      // Apply any last-minute adjustments.
+      attacker.body.CO.buffAttack(this);
+      defender.body.CO.buffDefense(this);
+    }
+
+    @Override
+    public double calculateDamage()
```
ThatIsAUsername, line 104: I'm not a super-fan of breaking the calculation out over two different functions like this, and I don't think I see the need for using inheritance here. It just feels like it adds complexity for no tangible benefit.

ThislsAUsername: I suppose I could just make `StrikeParams` have defense info, and then it's just defaulted to 100 until `BattleParams` fiddles with it?

ThislsAUsername: Thinking about it a bit more, I don't really like that solution since it puts all the calculations into a class that doesn't know about what actually is getting hit.
At some point in the future, I'd like to support multiple formulas for determining `overallDefense`.
Then again, you could argue that that calculation could be put in an object under the `StrikeParams`, and the defending CO could just override the default if necessary.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
```diff
@@ -218,7 +221,7 @@ public Unit getAttacker()
     }
     public boolean attackerDies()
     {
-      return battleInfo.attacker.getPreciseHP() - battleInfo.attackerHPLoss <= 0;
+      return (int) ((battleInfo.attacker.getPreciseHP() - battleInfo.attackerHPLoss) * 10) <= 0;
```
ThatIsAUsername, line 224: You are storing HP as an integer, converting it to a double to pass here, then converting back to an int and truncating. This feels more complicated; does it actually grant better accuracy (or better adherence to cart behavior if you are optimizing for that)? I guess truncating this will technically deliver a different result for values in the range (0, 0.1), but those are all precluded on the `Unit` side.

src/Test/TestGameEvent.java
```diff
@@ -108,7 +108,7 @@ private boolean testCaptureEvent()
     captureEvent.performEvent(testMap);
     testPassed &= validate(infA.getCaptureProgress() == 10, "    Infantry capture progress is not 10.");
 
-    // Hurt the unit so he won't captre as fast.
+    // Hurt the unit so he won't capture as fast.
```
ThatIsAUsername, line 111: Solid addition.

src/Units/AWBWWeapons.java
```diff
@@ -13,7 +13,7 @@
     CARRIERMISSILES, BATTLESHIPCANNON, CRUISERTORPEDOES, CRUISERMGUN, SUBTORPEDOES
   };
   
-  protected final static boolean USE_AMMO = true;
+  protected final static boolean USE_AMMO = false;
```
ThatIsAUsername, line 16: Maybe change this to `INFINITE_AMMO` so it matches the function it's passed into? Probably the wrong value was used because they are named to disagree.

src/Units/Unit.java
```diff
@@ -187,28 +187,32 @@ public void fire(WeaponModel weapon)
...
   public double alterHP(int change)
   {
-    double before = HP;
-    // Change the unit's health, but don't grant more
-    // than 10 HP, and don't drop HP to zero.
-    HP = Math.max(0.1, Math.min(model.maxHP, getHP() + change));
-    return HP - before;
+    int before = health;
+    // Change the unit's health, but don't
+    // go over maxHP, and don't drop HP to zero.
```
ThatIsAUsername, line 213: We should probably put this kind of info into javadoc comments so it's clear how `damageHP` and `alterHP` differ.

src/Engine/Combat/CombatEngine.java
```diff
@@ -1,9 +1,21 @@
...
+import Engine.Combat.StrikeParams.BattleParams;
 import Terrain.GameMap;
 import Terrain.MapMaster;
 import Units.Unit;
+import Units.WeaponModel;
 
+/**
+ * CombatEngine provides COs with two levels of control of what goes on in combat:
+ * changeCombatContext(), which takes in a CombatContext and allows the COs to fiddle with just about anything.
+ * applyCombatModifiers(), which is for directly modifying a unit's firepower in a given attack.
```
ThatIsAUsername, line 17: These comments don't seem to be accurate anymore.

src/CommandingOfficers/Venge.java
```diff
@@ -120,22 +121,22 @@ public void changeCombatContext(CombatContext instance)
...
-  public void applyCombatModifiers(BattleParams params, boolean amITheAttacker)
+  public void buffStrike(StrikeParams params)
   {
-    if( amITheAttacker )
-    {
       if( counterAtFullPower && params.isCounter )
       {
         // counterattack as if the unit had not taken damage.
-        params.attackerHP = params.attacker.getHP();
+        params.attackerHP = params.attacker.body.getHP();
```
ThatIsAUsername, line 129: Any reason not to give `Combatant` a `getHP()` just to avoid this extra level of indirection?

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThislsAUsername: You're right, it is more complicated.
Cart behavior uses exclusively int math, I think, which would require mucking with how the combat math is set up. It's something I'm keeping generally in mind, but not really going for broke on it.

The reason I am converting it to a double at all is because it keeps the `getHP()` and `getPreciseHP()` functions producing the same denomination of health.
I guess it's not out of the question to make a third function for the integer state (or I guess I could just make `health` public), but I'd really like it to be obvious from the accessing syntax that these are not directly-comparable quantities.

The reason I added this cast right here is because I was seeing a few zombie (0 hp) units pop up in testing, so that logic was necessary to sync up with how the unit would actually store the result.

src/Units/AWBWWeapons.java
ThislsAUsername: If I called it `INFINITE_AMMO`, I'd have to put exclamation points in front of all of its current usages.
I'm just glad I decided to test running out of ammo, heh.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: Just the function names, I think, but good catch.

src/CommandingOfficers/Venge.java
ThislsAUsername: Not really.

src/CommandingOfficers/Commander.java
```diff
@@ -96,11 +97,14 @@ protected void addCommanderAbility(CommanderAbility ca)
...
-  /**
-   * Allows a Commander to inject modifications that must be evaluated at combat time
-   * @param striking - lets the Commander know if it's dealing or taking damage 
-   */
-  public void applyCombatModifiers(BattleParams params, boolean amITheAttacker)
+  /** Called any time you are making a weapon attack */
+  public void buffStrike(StrikeParams params)
+  {}
+  /** Called any time you are attacking a unit, always after {@link #buffStrike(StrikeParams)} */
+  public void buffAttack(BattleParams params)
```
ThatIsAUsername, line 104: Why do we have both a `buffStrike` and a `buffAttack`, and why use these names instead of `applyCombatModifiers`, which feels more descriptive (especially since modifications could be nerfs).
There also needs to be better info here about exactly when each of these is invoked in the combat calculation process so users know when each could/should be used.

src/Engine/Combat/StrikeParams.java
```diff
@@ -0,0 +1,110 @@
...
+{
+  public final Combatant attacker;
+  public final GameMap map; // for reference, not weirdness
+
+  // Stuff inherited for reference from CombatContext
+  public final int battleRange;
+
+  public double baseDamage;
+  public double attackerHP;
+  public double attackPower;
```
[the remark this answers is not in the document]
ThislsAUsername: It is.
The nomenclature made more sense to me, as `Factor` implies it's a multiplier, when it's really hundredths-of-a-multiplier.
Strong *adds* 15, aka a 15% buff for 115A vs infantry or 130A if it's inf vs inf

src/CommandingOfficers/Commander.java
ThislsAUsername: `buffStrike` exists so that when I slot in attacks vs terrain, we can do conditional buffs that don't require information about exactly what you're shooting.
`buffAttack` exists so you can do conditional buffs that *do* require that info.
`CombatModifier` sounds like a class of object to me.
I thought I *did* put pretty good info in there. It explicitly states strike -> attack -> defense

src/AI/Muriel.java
```diff
@@ -138,24 +141,10 @@ private UnitMatchupAndMetaInfo getUnitMatchupInfo(Unit myUnit, Unit otherUnit)
...
     // Now go the other way.
-    double otherDamage = 0;
-    WeaponModel otherWeapon = otherUnit.chooseWeapon(myUnit.model, 1, myUnit.model.hasMobileWeapon());
-    if( null != otherWeapon )
-    {
-      BattleInstance.BattleParams params = new BattleInstance.BattleParams(otherUnit, otherWeapon,
-          myUnit, Environment.getTile(TerrainType.ROAD, Weathers.CLEAR), false, null);
-      otherDamage = params.calculateDamage();
-    }
+    double otherDamage = CombatEngine.buildSimpleAttack(otherUnit, 1, myUnit, myCo.myView, 0, false).calculateDamage();
```
ThatIsAUsername, line 147: This is certainly cleaner (except or magic numbers aaargh). Nice.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: You can push changes whenever. I've looked over the whole first batch of changes.

src/Units/AWBWWeapons.java
ThatIsAUsername: ... or you could change the default value back, and enjoy your newfound sense of alignedness.

src/CommandingOfficers/Commander.java
ThatIsAUsername: The existing comments specify the order in which they are applied, but not at which phase. They don't give me any idea when I should use one over the other (and less idea of what to use if I want to apply a nerf...).

src/AI/Muriel.java
ThislsAUsername: Exactly as many magic numbers as before!

src/Units/AWBWWeapons.java
ThislsAUsername: I named it `USE_AMMO` to be idiomatic in the `WeaponModel` constructors. I tell it to `USE_AMMO`, or I don't and it doesn't. Matching the declaration of the constructor wasn't really a priority.
I'm not sure what you mean by that last comment.

src/CommandingOfficers/Commander.java
ThislsAUsername: What do you mean by "phase"?
> They don't give me any idea when I should use one over the other

Au contraire: `buffStrike` is for all weapon attacks, `buffAttack` is only for weapon attacks directed at units. `buffDefense` is for when *your* unit is being targeted. That's all there in the comments.
As for nerfing, yeah, it's a little weird nomenclature-wise. I guess I could call them `modifyStrike/Attack/Defense`?

src/CommandingOfficers/Venge.java
ThislsAUsername: Actually, nevermind that.
Combatant is composed of `final` fields, and I think I want it to stay that way. If I was to put a `getHP()` on Combatant, I'd expect it to replace the instance in `params`, and I'm not gonna do that because then it'd have to be mutable.

src/Engine/Combat/StrikeParams.java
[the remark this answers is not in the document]
ThatIsAUsername: Follow-up question: is there any reason to have two separate classes here?

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: Ah, right. I'd be a little more descriptive in the comment then to clarify that they may be swapped *in the context object*, but our original variables will be fine.

To someone who know the system well this is obvious, but I never write comments for people who know the system well.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: Same here as above. You say just enough here to confuse someone who isn't already intimately familiar with the combat system, and not enough to clarify what you mean.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThatIsAUsername: > but I'd really like it to be obvious from the accessing syntax that these are not directly-comparable quantities

We could just go all strong-OO and make a `UnitHP` object/datatype with accessors and whatnot, but actually let's not.
Maybe just keep it as the 0-100 percentage value at all times except when displaying it? That's closer to what I expected you meant when you talked about using integer calculations. The tricky bit there is that we'd have to get used to thinking in one scale despite seeing in another.

src/Units/AWBWWeapons.java
ThatIsAUsername: > I named it USE_AMMO to be idiomatic in the WeaponModel constructor calls

What? The variable name in `WeaponModel`'s constructor is called `infiniteAmmo`. The member variable is named `hasInfiniteAmmo`. Renaming this one to match would (aligning them) would make it clear exactly what we are doing.
Right now, `USE_AMMO==false` corresponds to `infiniteAmmo==true`, which adds an extra layer of confusion re-aligning the concepts in your head. Making it matchy-matchy makes it easier.
As for being idiomatic, I'd argue renaming this to match would be *more* idiomatic, for two reasons:
1) A new dev who looked at this wouldn't necessarily connect the variable name `USE_AMMO` to the concept of "having a limited quantity", whereas a variable named `HAS_INFINITE_AMMO` is exceedingly clear.
2) `USE_AMMO` is set to false. Anybody who sees a Boolean value set to false is going to assume that means the thing we pass it to is being told not to "use ammo". If they manage to connect it to the concept of "having infinite ammo", then the weapon is clearly being told to not use ammo, meaning it should have infinite ammo; the opposite is true.

There's a reason naming things is one of the two Hard Problems in computer science.

src/CommandingOfficers/Venge.java
ThatIsAUsername: What? I'm confused. `getHP()` is not a mutator, and `params` doesn't have a `getHP` call.

src/Engine/Combat/StrikeParams.java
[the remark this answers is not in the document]
ThatIsAUsername: Fair enough.

src/CommandingOfficers/Commander.java
ThatIsAUsername: For two functions that are named so similarly, more care must be made in delineating the difference, especially since we don't have a handy domain-language dictionary to clarify that "strike" now (as of this PR) just means any targeted weapon discharge, but "attack" means it's directed against a specific *kind* of target (which is currently the only kind of target).
Otherwise it just looks like `buffAttack` is for when attacking a unit, and `buffStrike` is... also for attacking (implicitly: a unit, because what else is there to attack? Even when we get pipe seams/etc, that's not the normative case for weapons employment).

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThislsAUsername: Hmm. You make a good point. Does that sound like it's in scope for this PR?

src/Units/AWBWWeapons.java
ThislsAUsername: If I call it `HAS_INFINITE_AMMO`, my pea-brain will look at the `WeaponModel` constructors and panic because it "says right there" that the tank cannon gets infinite ammo.
I suppose I could meet you in the middle with `DISABLE_INFINITE_AMMO`?

> Right now, `USE_AMMO`==false corresponds to `infiniteAmmo`==true, which adds an extra layer of confusion re-aligning the concepts in your head.

That's not true at all.
`USE_AMMO` *being present* corresponds with `infiniteAmmo`==true.
`USE_AMMO` being *not* there corresponds with `infiniteAmmo`==false.
Honestly, I don't care all that much about aligning concepts in this case. I just want to not have to look at it :stuck_out_tongue:

> A new dev who looked at this wouldn't necessarily connect the variable name `USE_AMMO` to the concept of "having a limited quantity",

Perhaps a better term would be `REQUIRES_AMMO`, then?

> USE_AMMO is set to false. Anybody who sees a Boolean value set to false is going to assume that means the thing we pass it to is being told not to "use ammo". If they manage to connect it to the concept of "having infinite ammo", then the weapon is clearly being told to not use ammo, meaning it should have infinite ammo; the opposite is true.

That's almost exactly the opposite of how my brain works.
If I see a weapon constructor, and it says `USE_AMMO`, brain says "yay good, we use ammo".
If I decide to look deeper and see that that boolean happens to be false, I'm going to look at where it's used and quickly arrive at the right conclusion.

src/CommandingOfficers/Venge.java
ThislsAUsername: Yes, but `Combatant` exists as a handy summary-object of how a given unit will participate in a fight, for use in `Strike`/`BattleParams`.
You'll note I moved the `Unit`s, their `WeaponModel`s, and their positions into it and made them final to enforce not mucking about with that stuff.

By contrast, `Strike`/`BattleParams` has an HP value for each `Unit`, and it's important that it be mutable. If I put knowledge of HP state into the `Combatant`, I would expect that to represent the HP that would actually be used to calculate effective stats for the combat (and would also remove those members from the `params`).
By making the user go to the unit itself to fetch the HP, I'm making him acknowledge that he's looking at the *current board state* rather than what the combat will be calculated with.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThatIsAUsername: Well, given the name of this branch and your initial comments when discussing the change I thought that was the main point of this PR; at this point I'm not sure what the scope is supposed to be anymore.
So uh, sure.

src/CommandingOfficers/Commander.java
ThislsAUsername: > but "attack" means it's directed against a specific kind of target (which is currently the only kind of target).
> Otherwise it just looks like buffAttack is for when attacking a unit, and buffStrike is... also for attacking (implicitly: a unit, because what else is there to attack? Even when we get pipe seams/etc, that's not the normative case for weapons employment).

...and this is kind of why I was wondering whether I should split these changes into two PRs.
I feel like you wouldn't be asking this question if I *had* destructible terrain already implemented here.

I dunno what you want from me. I put the functions right next to each other. I said "use this one to affect attacks on units" and "use that one to affect all attacks".
They have different parameter types (one of which, notably, carried absolutely no data about the thing being targeted until you asked me to re-combine the damage calculation).
They have an explicit call order, too.

src/Units/AWBWWeapons.java
ThatIsAUsername: > That's not true at all.

Hahaha, except it is; you just reiterated what I said. observe:
> USE_AMMO being present corresponds with infiniteAmmo==true.

If `USE_AMMO` is *present*, it is false, which corresponds with having infinite ammo. Thanks for helping to make my point about the names being confusing.

> I suppose I could meet you in the middle with DISABLE_INFINITE_AMMO

Ah, I hadn't looked closely enough to notice you only pass `USE_AMMO` to weapons with limited ammo. Yes, I can meet you in the middle there. Or `REQUIRES_AMMO` or `LIMITED_AMMO` or `EATS_BULLETS` or whatever (except not that last one please).

> That's almost exactly the opposite of how my brain works.

Aaaaand, this is why naming things is hard. You have to be anticipate potential misconceptions and methodically destroy them, leaving a single unambiguous interpretation. If you only write code for yourself, nobody else will be able to read it.

ThatIsAUsername: > You have to be anticipate potential misconceptions and methodically destroy them

Actually, I should clarify that this is *not* the only approach. Another approach I've encountered is to use perfectly ambiguous names that can grant no preconceptions. E.g. if you change the variable to be named `SILLY_KANGAROO`, then nobody would be able to guess at what it meant, and they'd have to look at how it was used.
A side effect of this style of naming things is that conversations about the code can get weird. "OK, we've got a new unit, and it's got a new type of weapon." "Is it a silly kangaroo weapon, or a normal one?" etc.

So uh, yeah.

ThislsAUsername: > If `USE_AMMO` is _present_, it is false, which corresponds with having infinite ammo. Thanks for helping to make my point about the names being confusing.

It's only confusing if you look exactly one level deep. If you look zero or two levels deep, the meaning is abundantly clear.
0: "the gun says it uses ammo"
1: "the gun says uses ammo=false, so it must have infinite"
2: "the constructor parameter is called infinite ammo, so it's actually *disabling* infinite ammo"

> Ah, I hadn't looked closely enough to notice you only pass `USE_AMMO` to weapons with limited ammo. Yes, I can meet you in the middle there. Or `REQUIRES_AMMO` or `LIMITED_AMMO` or `EATS_BULLETS` or whatever (except not that last one please).

What would be the point of defining a value that's constant for *all* weapons if I was going to pass it to every single weapon's constructor? At that point, I'd just remove it from the constructor and make it an implied parameter.
...you should know by now to not give me ideas like that.

> Aaaaand, this is why naming things is hard. You have to be anticipate potential misconceptions and methodically destroy them, leaving a single unambiguous interpretation. If you only write code for yourself, nobody else will be able to read it.

I can't anticipate every misconception. Even if I could, it'd be a lot harder than actually getting the code working anyhow.
I've actually revised the name of that variable more than once already in pursuit of clarity.

> Actually, I should clarify that this is _not_ the only approach. Another approach I've encountered is to use perfectly ambiguous names that can grant no preconceptions. E.g. if you change the variable to be named `SILLY_KANGAROO`, then nobody would be able to guess at what it meant, and they'd have to look at how it was used.

You're just piling more evidence on the docket for why we should use `EATS_BULLETS`.

src/CommandingOfficers/Venge.java
ThatIsAUsername: > I'm making him acknowledge that he's looking at the current board state

Hm. I'm not sure that distinction is as clear as you imagine it to be.

> You'll note I moved the Units, their WeaponModels, and their positions into it and made them final 

I hadn't. And if I'm just looking for an accessor (How much health does this guy have, anyway), I'm not going to look closely enough to see that the unit is declared `final` and then make generalizations about what that means about the state of the world.

Maybe add the utility method with a name like `get[Starting|PreFight]HP()` just to make that meaning clear (though that's the meaning I would have naturally inferred).

I feel like this PR is bringing out some major differences in how we approach/interpret things.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThislsAUsername: I wanted to do two things, mostly:
Get rid of `BattleInstance`. Every time I went to look at the combat code, I'd click on `CombatEngine` and be bitterly disappointed.
Introduce a split of at least two-fold for dynamic attack modifiers, to support dynamic attack buffs vs non-units.

I also changed the unit HP storage since I figured "hey, we're fiddling with stuff, why not?"... and then I realized it wasn't quite as easy to do as I had thought. (Funny, I had the exact same problem when I was initially drafting the `Unit` class and stuff.)

src/CommandingOfficers/Venge.java
ThislsAUsername: > > I'm making him acknowledge that he's looking at the current board state
> 
> Hm. I'm not sure that distinction is as clear as you imagine it to be.

Well. I'm looking at a reference to a unit that is (theoretically) on the map somewhere, waiting to act.
My first assumption on talking to that unit is that it has zero knowledge that a combat calculation is taking place. It knows the current board state, and nothing else.
I guess you could make the assumption that the unit is *cloned* for the purposes of combat calculation, but we avoid cloning unless it's necessary.

> if I'm just looking for an accessor (How much health does this guy have, anyway), I'm not going to look closely enough to see that the unit is declared `final` and then make generalizations about what that means about the state of the world.

That's entirely fair. but if you're looking for an accessor, you'll go `params.atta` and Eclipse will happily pop up and point you to `attackerHP`, and all will be right with the world.
Then you say, "Wait, what's the *actual* HP of the unit as it is on the board right now?", so you go to the unit reference.

> Maybe add the utility method with a name like `get[Starting|PreFight]HP()` just to make that meaning clear (though that's the meaning I would have naturally inferred).

I guess I could, but we clearly don't mind having call-chains (see: `unit.model.thing` littered all throughout the codebase)

> I feel like this PR is bringing out some major differences in how we approach/interpret things.

It may also be worth noting that when I move stuff around, I typically have to change a few lines in Muriel. Wally is forced to change much less frequently.
I guess what I'm saying is I already was noticing there are some differences, lol.

src/CommandingOfficers/Commander.java
ThatIsAUsername: > I feel like you wouldn't be asking this question if I had destructible terrain already implemented here.

I would. Attacking terrain is definitely something that only even occurs to me in situations where it is needed; it's an exception to the rule of directing bullets at enemy units.
Even if I had taken note of the fine distinction in the two comments (ooh, one includes the word "units" and the other doesn't), that isn't sufficient evidence to determine that `buffStrike` is intended for non-unit-targeted attacks--it's just as likely that you didn't happen to feel like typing "units" at the time you wrote the description. Knowing that both functions are invoked, and in some order, tells me there *should* be a difference, but any inference I'd make based on those javadocs feels like reading tea leaves.

> I dunno what you want from me

Clarity of intent. These function names sound like synonyms; even after being told what they are I only know the difference by rote. At minimum, the javadoc comments should be updated to clarify exactly how their intended use differs--omitted information that might lead to an inference isn't good enough. Just spell it out like the 5-year-old neighbor kid will be the next one using these functions. "This function is intended to be used for any weapon attacks against anything, including destructible map terrain", "This function is intended for adjusting unit firepower when attacking units only, or because the modification depends on the unit being attacked", "This function is called for modifying defense for a unit being attacked, and will be evaluated after `bufStrike` is called for the attacker's Commander, so any firepower adjustment for the guy shooting us is already accounted for in `params` when this function is called" or whatever.

For that matter, is there anything to prevent a Commander from modifying defense in the attack function and vise versa? Maybe we should trim down the functions to take in only the data they need (and maybe also to return the power adjustment instead of modifying `params` directly).

Also also; you say that all three are called, but it's implied that one of them does not apply if we aren't striking a unit. Will `buffAttack` not be called in pipe seam attacks?

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThatIsAUsername: > Every time I went to look at the combat code, I'd click on CombatEngine and be bitterly disappointed

Yeah, I'm pretty sure I did the same. Putting logic where it sounds like it should be is a Good Thing.

> hey, we're fiddling with stuff

Well, you're already fiddling with it; may as well do it proper.

src/CommandingOfficers/Venge.java
ThatIsAUsername: >  if you're looking for an accessor, you'll go params.atta

Well, maybe, or maybe I'll assume that accessors were named with the normal Java-style `get` prefix and start there. And then I'll be confused because I know units are combatants, so why doesn't this have unit's functions? And then I'll notice that it's a wrapper around a unit and try to type `params.attacker.getUnit`, which isn't there, and then `params.attacker.unit` which still isn't there, and then notice that it's actually been helpfully named `body` for some reason, and throw my keyboard at the ceiling.

ThislsAUsername: > Well, maybe, or maybe I'll assume that accessors were named with the normal Java-style `get` prefix and start there.

I don't think that assumption is all that valid. We use bare member variables in most any place that we aren't looking to specifically gain something from using functions.

> And then I'll be confused because I know units are combatants, so why doesn't this have unit's functions? And then I'll notice that it's a wrapper around a unit and try to type `params.attacker.getUnit`, which isn't there, and then `params.attacker.unit` which still isn't there, and then notice that it's actually been helpfully named `body` for some reason, and throw my keyboard at the ceiling.

So what you're saying is you don't like my creative naming schemes?
IDK, something about naming variables the same thing as their types rubs me the wrong way. Seems like you should be naming the variable for the usage *in this instance*, not just aping whatever the type name is.
I don't feel super strong about it, though, since standardization is also good.

src/CommandingOfficers/Commander.java
ThislsAUsername: > Knowing that both functions are invoked, and in some order, tells me there should be a difference, but any inference I'd make based on those javadocs feels like reading tea leaves.

Hmm. Maybe I feel differently because I spend so much time being a pedant.
There are two differences, other than ordering, in those descriptions.
Strike mentions weapons, Attack does not.
Attack mentions hitting a unit, Strike does not.
Seeing as typically, you hit units with weapons, that leaves "hitting a unit" as the most significant difference.
At the same time, Strike *does not rule out* attacking units, ergo it would seem reasonable to assume Strike applies to a strict superset of what Attack applies to.

> omitted information that might lead to an inference isn't good enough

Particularly in writing, I tend to have intent behind the amount and kind of detail I provide. Destructible terrain is one possible use case of `buffStrike`, but I wouldn't want to rule out other potential usecases down the line. I suppose I could just update the doc at that time, but I don't think it makes sense to try to exhaustively account for all use-cases of a given piece of code in the documentation for that piece of code.
The reason the documentation is here instead of there is twofold:
`CombatEngine` itself has a bunch of stuff to keep track of
Someone looking to use these functions will likely care about the detail of general purpose and order of operations.

> For that matter, is there anything to prevent a Commander from modifying defense in the attack function and vise versa?

No. My intent (such as it was) was to use strike/attack/defense as actions units can participate in, rather than the statistics used for that participation *per se*.

> Maybe we should trim down the functions to take in only the data they need

IMO, that's a lot of work for not much gain. If we had 50 people working on the project, I'd be miles more interested in getting the compiler to enforce the separation of concerns/behaviors I'm interested in keeping.
We'd have read-only interfaces for all model objects, and ideally use the compiler to prevent modification of game state anywhere but GameEngine/MapController/GameEvent, and study any that happened in the former two with a critical eye.

> (and maybe also to return the power adjustment instead of modifying params directly).

This kills an existing use case. Venge's Iron Will adjusts his unit's HP value so that he can counterattack as if he hadn't taken damage.
(To tell the truth, I made Venge specifically as a way to effectively stake out and defend the extent of capabilities I wanted our combat calculations to have.)

> Also also; you say that all three are called, but it's implied that one of them does not apply if we aren't striking a unit. Will `buffAttack` not be called in pipe seam attacks?

Actually, only one applies if we aren't striking a unit. `buffStrike` is called unconditionally for all weapon attacks.
IFF the attack targets a unit, `buffAttack` is called... then finally `buffDefense` is called to give the defending unit's CO a look in.
In the case of shooting up the scenery, there's no combatant for `buffAttack` to be interested in, and there's no defending CO to want to call `buffDefense`.

src/CommandingOfficers/Venge.java
ThatIsAUsername: > I don't think that assumption is all that valid. We use bare member variables in most any place that we aren't looking to specifically gain something from using functions.

Yeah, we have done a terrible job maintaining proper object-oriented encapsulation or any kind of consistent naming scheme. Which is fine for a two-developer side project, but wouldn't fly in the real world. I mean, actually it would because real-world code review practices are often not great, but it certainly doesn't help with handing off code.

> something about naming variables the same thing as their types rubs me the wrong way.

Must be tough being you, because naming variables after object types (typically modified by an adjective, e.g. `myUnit` or `attackingUnit` or `unitReference`) is standard practice almost everywhere. 

> Seems like you should be naming the variable for the usage `in this instance`

Which makes sense, and is what the adjective is for. Naming the `Combatant`'s `Unit` "body" doesn't though, because DP doesn't have a concept of unit bodies, and it just seems like you got bored while typing and decided to be clever. As much as I appreciate creativity, and as much as software design is a creative art, there is still great value in sticking with standard/predictable naming conventions.
I'm not saying you have to change it, because it's clear enough in this small class which is only used in a narrow context, but with something that is used in a larger context it could become confusing.

src/CommandingOfficers/Commander.java
ThatIsAUsername: > I don't think it makes sense to try to exhaustively account for all use-cases of a given piece of code in the documentation

Agreed. And I agree with documenting the intended use of the functions here, since Commanders are the ones who use them. I don't need exhaustive examples for all use-cases, but I do need enough information to clarify intent for *how* or *when* it should be used. Otherwise I'm left wondering why I have two functions that have synonymous names and take two data types which inherit from each other; it looks like someone added a second version of a function without realizing someone had already implemented it. The comment pointing out the order in which they are called indicates it actually *is* intentional (probably), but doesn't tell me why/how they differ.

> Particularly in writing, I tend to have intent behind the amount and kind of detail I provide.

Nobody who reads your code will read that much into your wording. You probably don't read that much into the wording of other people's code. I don't *want* people trying to read that much into what I write about my code, because as we've (firmly) established, different people think differently, and their natural guess as to what you mean will most likely differ from your intended meaning unless you spell it out.

> Strike mentions weapons, Attack does not.
Attack mentions hitting a unit, Strike does not.
Seeing as typically, you hit units with weapons, that leaves "hitting a unit" as the most significant difference.

Tea leaves. Don't just say one is for hitting units. Clarify that the other is for hitting non-units.

> IMO, that's a lot of work for not much gain

Yeah, probably true.

> Actually, only one applies if we aren't striking a unit. buffStrike is called unconditionally for all weapon attacks.
IFF the attack targets a unit, buffAttack is called... then finally buffDefense is called to give the defending unit's CO a look in.
In the case of shooting up the scenery, there's no combatant for buffAttack to be interested in, and there's no defending CO to want to call buffDefense

Ah, I see. Given that I *still* apparently didn't quite grok this even after all this chatter, it seems like excellent info to include in some documentation...

src/Units/AWBWWeapons.java
```diff
@@ -13,7 +13,7 @@
     CARRIERMISSILES, BATTLESHIPCANNON, CRUISERTORPEDOES, CRUISERMGUN, SUBTORPEDOES
   };
   
-  protected final static boolean USE_AMMO = false;
+  protected final static boolean EATS_BULLETS = false;
```
ThatIsAUsername, line 16: ... of course you did.

src/CommandingOfficers/Venge.java
ThislsAUsername: > Yeah, we have done a terrible job maintaining proper object-oriented encapsulation or any kind of consistent naming scheme.

Naming schemes, I'll give you.
Encapsulation exists to prevent access to implementation details. Just adding getters and setters for every variable isn't any better encapsulation than allowing raw member access.
HP, for instance, doesn't have raw access because I wanted to be explicit about exactly what "form" of it you were accessing, and to enforce certain semantics on exactly how it is modified.
CO str/def doesn't have raw access because I wanted to force changes to those stats to use dead reckoning: "I move it up 20, then back down 20" as opposed to "I set it to 120, then back to 100", since the second doesn't play well with multiple modifiers.

I won't say we've done a perfect job of keeping everything hidden that *should* be, but most of the things that are public *are* public because we have reasons to be modifying them.

src/Engine/Combat/StrikeParams.java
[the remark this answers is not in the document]
ThislsAUsername: > Follow-up question: is there any reason to have two separate classes here?

One is agnostic to target type, the other is not.

src/Engine/GameEvents/HealUnitEvent.java
```diff
@@ -35,29 +35,19 @@ public void sendToListener(GameEventListener listener)
   @Override
   public void performEvent(MapMaster gameMap)
   {
+    int maxHP = unit.model.maxHP;
+
     if (null == payer)
-      unit.alterHP(repairPower);
-    else
+      unit.alterHP(repairPowerHP);
+    else if( unit.getPreciseHealth() < maxHP * 10 )
```
ThatIsAUsername, line 42: This feels an awful lot like implementation-specific information living outside of `Unit`.

I don't like that health and HP are now distinct concepts; I should be able to compare `unit.getWhichever()` and compare that to `unit.model.maxHP` and have it be correct.
It might not hurt to also have a `unit.isHurt` just to avoid having to do a comparison.

src/Engine/GameEvents/HealUnitEvent.java
```diff
@@ -35,29 +35,19 @@ public void sendToListener(GameEventListener listener)
...
-        unit.alterHP(repairedHP);
+      int affordableHP = payer.money / costPerHP;
+      int actualRepair = Math.min(repairPowerHP, affordableHP);
 
-        // Top off HP if there's excess power but we hit the HP cap
-        if (repairedHP < repairPower && unit.getHP() == maxHP)
-          unit.alterHP(1);
-      }
+      int deltaHP = unit.alterHP(actualRepair);
+      payer.money -= deltaHP * costPerHP;
```
ThatIsAUsername, line 50: This revised logic is a lot cleaner and more concise. I approve.

src/Units/Unit.java
```diff
@@ -185,43 +191,65 @@ public void fire(WeaponModel weapon)
...
+    return health;
   }
-  public double getPreciseHP()
+  public int getHP()
   {
     return healthToHP(health);
   }
-  private static double healthToHP(int input)
+  /** @return health value == its corresponding HP value * 10 */
+  public static int effectiveHealth(int input)
```
ThatIsAUsername, line 209: This... is not quite what I envisioned when I suggested keeping the units consistent everywhere except when displaying the value.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > This feels an awful lot like implementation-specific information living outside of `Unit`.

I suppose I could make a getter for "HP times 10", but I don't foresee this implementation detail ever changing.

> I don't like that health and HP are now distinct concepts; I should be able to compare `unit.getWhichever()` and compare that to `unit.model.maxHP` and have it be correct.

Right. That's why I had health stored as a floating point, initially.
Fundamentally, HP and health *aren't* the same thing. Health is just how the game tracks whether a unit should be dead or not.
HP is a related quantity that determines firepower scaling and such. I initially tried to get rid of HP as a separate concept, but I had to do a *lot* more hacking around in a similar vein to `maxHP * 10` (but more complicated, because I had to deal with rounding errors *everywhere*).
The example that comes to mind is `alterHP`. If `maxHP` is in range [1-100], then suddenly I can have a maxHP of 95... so I had to check the maximum bound *after rounding to HP*. I forgot I made the change and left it in. I think I'll back it out, since it's unnecessary.
It was a lot harder for me to keep track of what I was doing, since I didn't have clear terminology delineating what's in range [1-100] and what's in range [1-10], and it seems like most applications in the game logic care more about HP than health.

> It might not hurt to also have a `unit.isHurt` just to avoid having to do a comparison.

That's fair.

src/Units/Unit.java
ThislsAUsername: The problem is that HP is *actually* more important most of the time.
Attack/defense scaling, healing, and anything to with funds don't actually interact with the actual health mechanic.

ThatIsAUsername: Can we not adjust those calculations to work with the [0-100] health values instead?
I _really_ don't like the idea of having HP =/= health.

src/Engine/GameEvents/HealUnitEvent.java
ThatIsAUsername: > I don't foresee this implementation detail ever changing

Famous last words... especially given that you just changed it.

> then suddenly I can have a maxHP of 95

How? Because you are rounding to the nearest 1 instead of 10? I'm not sure I understand.

src/Units/Unit.java
ThislsAUsername: Technically, yes, but practically no.
Specifically, attack/defense/etc all work on [0-10], rounded up. So either we provide an accessor on Unit to get that value (if nothing else, doing the rounding up), then we *have* to do that calculation manually everywhere else.
Also worth noting is that when I think about HP in CO design/etc, I'm thinking about [1-10]. I don't want to remember/care about there being a precise health concept; it exists almost entirely to broker when fractional bits of damage should add up to a full HP.
If we do provide that accessor, we need good terminology to separate the two concepts or we (or at least I) will get constantly confused. ("effective health" IMO isn't good enough, at least not without *also* linking it to HP)

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > Famous last words... especially given that you just changed it.

It matches cart behavior as-is. They store health as [0-100] integers.
Seeing as this is a clone and exists to extend the original system, I don't think it's within scope for *this project* to change that.
If I'd had the certainty I have today a) about the implementation details and b) about floating point (in)consistency, I probably would have arrived at this solution as the least evil when I first wrote this stuff.

> How? Because you are rounding to the nearest 1 instead of 10? I'm not sure I understand.

That's exactly why.
`maxHP` is a public, non-final field of `UnitModel` for a reason. I don't want to support fractional HP in `maxHP`, as it makes no sense from the standpoint of the player.
I suppose I could also enforce that constraint via making a getter/setter for it, but I would much rather enforce it by datatype. I could make a `getMaxHealth()` if you want.

ThatIsAUsername: This doesn't clarify how we end up with `maxHP==95`. We aren't setting different `maxHP`s for different units.

> I had to check the maximum bound after rounding to HP

Don't we convert to HP by doing a `ceil(health/10)`? No rounding involved. Certainly nothing that could end up going outside the valid HP range anyway.
I must be having a slow day again, because I am not tracking here.

src/Units/Unit.java
ThatIsAUsername: > Technically, yes, but practically no

Technically===practically. 

> attack/defense/etc all work on [0-10]

Then these are the calculations I'm suggesting we modify; your answer seems to presuppose that we can't do that for some reason.

> So either we provide an accessor on Unit to get that value

We do not. We only hand out the One True [0-100] value, and the UI divides by 10 and ceilings the result before displaying.
(Also you didn't provide an _or_ option here)

> Also worth noting is that when I think about HP in CO design/etc, I'm thinking about [1-10].

Yes; that was the one downside of this approach I mentioned previously. I'm willing to live with that though if it makes life easier in other ways. Weapon-damage values etc are already stored this way, so it seems like a sensible/straightforward thing to do.

src/Units/Unit.java
```diff
@@ -201,10 +201,19 @@ public int getPreciseHealth()
...
+  public boolean isHurt()
+  {
+    return health < healthFromHP(model.maxHP);
+  }
   public int getHP()
   {
     return healthToHP(health);
   }
+  /** @return value in range [0-1.0]; represents the unit's current effectiveness */
+  public double getHPRatio()
```
ThatIsAUsername, line 213: I feel like `getPercentHealth` might be more clear; `getHPRatio` sounds like it's comparing two HP values rather than an HP and a max HP.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > This doesn't clarify how we end up with `maxHP==95`. We aren't setting different `maxHP`s for different units.

...yet.
I don't have specific plans to make a CO that changes maxHP, but I don't want to rule it out.

> Don't we convert to HP by doing a `ceil(health/10)`? No rounding involved. Certainly nothing that could end up going outside the valid HP range anyway.
> I must be having a slow day again, because I am not tracking here.

Okay. So.
In `alterHealth`, we add the current health and the change.
Then, if it is a positive alteration, we set the HP to its maximum value.

Let's say I have 81 health, with a cap of 95 health because my CO is a sadist.
A BBoat heals me.
That's +10 health, so I end up at 91 (or 10 HP)
Because that alteration was healing, we round up to 10 HP and re-assign health (this is cart behavior), and end up at 100 health.
Obviously, I must now apply the health cap and shift down to 95 health.
If, however, `maxHP` can only be an integer value, the rounding-up could not break anything so long as the HP cap had already been enforced.

src/Units/Unit.java
ThislsAUsername: > > Technically, yes, but practically no
> 
> Technically===practically.

That's... literally not true, but I think I can see what you're going for.

> > attack/defense/etc all work on [0-10]
> 
> Then these are the calculations I'm suggesting we modify; your answer seems to presuppose that we can't do that for some reason.

That's cart behavior, and I subjectively prefer it greatly because it makes combat easier to reason about (Wargroove uses the exact health percentage, so I'm not just guessing when I say I like it better).
I'm not changing it. You can make a fork if you really care that much.

> > So either we provide an accessor on Unit to get that value
> 
> We do not. We only hand out the One True [0-100] value, and the UI divides by 10 and ceilings the result before displaying.

For the reasons mentioned above... I'm not gonna go for that.

> (Also you didn't provide an _or_ option here to match with your _either_)

...whoops. The "then" after the comma should have been the "or".

src/Units/Unit.java
ThislsAUsername: But "percent health" is what `health` already is.
Words are hard.

src/Engine/GameEvents/HealUnitEvent.java
ThatIsAUsername: > but I don't want to rule it out.

... Why? damage values are determined based on a table; there's no reason to give a specific model more HP instead of just changing the damage it takes.

> ... because my CO is a sadist

I mean. I _guess_. But even still:

> Let's say I have 81 health...

This whole situation only arises due to converting to/from HP. If Health is the "native units" of the quantity, and we only convert to HP numbers for display, then there's no problem.

src/Units/Unit.java
ThatIsAUsername: Then is `HPRatio` ever going to be different from `health/100` (barring the case where `maxHP!=100`, which I still see as spurious)?

ThislsAUsername: I don't think so? It's just a convenience function, as you can see in the diff.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > ... Why? damage values are determined based on a table; there's no reason to give a specific model more HP instead of just changing the damage it takes.

More HP isn't exactly the same as taking less damage, nor is it exactly the same as just having more defense.
One of the custom COs I've seen proposed has overheal on powers, which makes affected units extremely durable without changing their defense or damage chart... and can persist across turns.

> This whole situation only arises due to converting to/from HP. If Health is the "native units" of the quantity, and we only convert to HP numbers for display, then there's no problem.

That's a big "if", and one I will not be conceding any time soon.
HP is the "real" value. Health is used only to track when less-than-a-whole-HP damage should add up to a whole HP.

src/Units/Unit.java
ThatIsAUsername: Ah, touché. I'm still not a fan of "ratio" though. Maybe "getHPAsPercent" then? I dunno.

src/Engine/GameEvents/HealUnitEvent.java
ThatIsAUsername: > More HP isn't exactly the same as taking less damage, nor is it exactly the same as just having more defense.

True, but for separate unit types that doesn't really matter.

> One of the custom COs I've seen proposed has overheal on powers

Sure, but now we are in the land of trying to predict everything we could ever want, and I try to avoid YAGNI violations. I also try not to limit flexibility arbitrarily, but at some point you do need to have ground rules.

> HP is the "real" value.

Well, can we go the other way then? Store it as a [0-100] int in `Unit`, but nobody else ever gets to see anything but the [0-10] float value? I just feel like we are building a space lander with both a `getRangeToGround()` and `getAltitude()` which return meters and feet, or is it feet and meters, and we are gonna lose track and create a 5 billion dollar red dust cloud (though I guess the stakes are technically a bit lower here).
For example, `CombatEngine.calculateOneStrikeDamage()` doesn't specify which one it's returning, and I'm pretty sure it's not the same as what was available before.

src/Units/Unit.java
ThatIsAUsername: > Wargroove uses the exact health percentage, so I'm not just guessing when I say I like it better

I wasn't suggesting to change any gameplay behavior. The functions themselves could stay exactly the same, since it's easy enough to pull the unit's health and `ceil(number/10)` before plugging it into the existing calculation. That should be a local implementation detail as part of the formula though, not a system-wide complexity magnifier.

Besides, if the carts store the HP as a [0-100] value, that's basically what they have to be doing anyway, right? Actually, it looks like you _are_ essentially doing that already. You get the health, convert it to HP, and then run the calculations. Cool. Well then. I guess we managed to disagree while saying the same thing here. Fantastic.

I'm just trying to avoid littering the `Unit` interface with excessive inscrutable accessors, and avoid multiple representations of the same value (DRY, etc).
Right now we have:
`getHP()`
`getHPFactor()`
`getPreciseHealth()`
`getEffectiveHealth()`
`effectiveHealth()` (static)
`healthToHP()`      (static)
`healthFromHP()` (static)

This is a bit much, and some of these are only used in one or two places. I made a suggestion in the other head, so I won't bother repeating it here.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > True, but for separate unit types that doesn't really matter.

...what?

> Well, can we go the other way then? Store it as a [0-100] int in `Unit`, but nobody else ever gets to see anything but the [0-10] float value?

That's what I did initially, but you said "you're messing with stuff already, so do it right".
It's kind of a weird situation all around.

> I just feel like we are building a space lander with both a `getRangeToGround()` and `getAltitude()` which return meters and feet, or is it feet and meters, and we are gonna lose track and create a 5 billion dollar red dust cloud (though I guess the stakes are technically a bit lower here).

Yeah. That's why I've tried to be careful in labeling.

> For example, `CombatEngine.calculateOneStrikeDamage()` doesn't specify which one it's returning, and I'm pretty sure it's not the same as what was available before.

It specifies health. It's... unfortunate that "health" and "HP" are synonyms, but I don't have better terminology on me.

src/Units/Unit.java
ThislsAUsername: > I guess we managed to disagree while saying the same thing here. Fantastic.

You did say you didn't like HP and health not being the same thing, at least initially.
Fair enough, though.

src/Engine/GameEvents/HealUnitEvent.java
ThislsAUsername: > I'm just trying to avoid littering the `Unit` interface with excessive inscrutable accessors, and avoid multiple representations of the same value (DRY, etc).
> Right now we have:
> `getHP()`
> `getHPFactor()`
> `getPreciseHealth()`
> `getEffectiveHealth()`
> `effectiveHealth()` (static)
> `healthToHP()` (static)
> `healthFromHP()` (static)
> 
> This is a bit much, and some of these are only used in one or two places. I made a suggestion in the other [t]h[r]ead, so I won't bother repeating it here.

There's a balance to be struck between clean interfaces and duplicating logic, to be sure.
`getHP()` is probably the most important
`getHPFactor()` I made mostly just to de-duplicate code. I'm not gonna die on this hill either way
`getPreciseHealth()` is almost as important as `getHP()`, since it's the direct getter for `health`
`getEffectiveHealth()` is used for combat-calculation-adjacent things, where it needs to be compared to a damage number (We could re-standardize to a double for damage, but that means adding those rounding error adjustments back in)
`effectiveHealth()` acts as the DRY source for the round-up-to-HP operation on health
The last two are just converters.

Yeah, it seems like a bit much, I'll grant, but they all have a solid reason to exist.
To be fair, we would OBE most of them by going back to "always HP, sometimes floating point"... I just dunno that that's more desirable than what we have.

ThatIsAUsername: > >    True, but for separate unit types that doesn't really matter.
> ...what?

I think I a brained my word or something. Nothing to see here.

> but you said "you're messing with stuff already, so do it right"

Immediately before that, I said:
> Maybe just keep it as the 0-100 percentage value at all times except when displaying it? That's closer to what I expected you meant when you talked about using integer calculations. The tricky bit there is that we'd have to get used to thinking in one scale despite seeing in another.

Which seems both very clear and very different than what happened. It would have been good to mention that you had a different fix in mind; then we could have pushed it to a new PR. On the other hand, if you think this implementation matches that description, then we are indeed speaking different languages.

> Yeah. That's why I've tried to be careful in labeling

Adding shades of meanings in a half-dozen overloads can quickly become more overwhelming than clear. Maybe you can keep them all straight, but I won't. I think it will be better if we can simplify the `Unit` interface to just the HP versions and only store it as the `health` value at rest.

> It specifies health. It's... unfortunate that "health" and "HP" are synonyms, but I don't have better terminology on me.

So let's consolidate. Please. I don't care if we call it FluffyTamarin, I just want it to be one thing, at least outside of `Unit`. Maybe even call it `healthPercent` inside `Unit`, since that's what you said it really is (We are talking about allowing different `maxHP` values, not different `health` values, right?).

src/Units/Unit.java
ThatIsAUsername: I don't like them not being the same thing. I would just call the down-conversion to HP an implementation detail of the combat calculations, and not give it a separate label at all (at least in the world where we decided [0-100] is the True Value).
But if we decide [0-10] is the Real Thing, then we can hide the [0-100] dealio inside `Unit` and call it whatever we want: `health`, `healthPercent`, `inverseClosenessToDying`, `nobodyActuallyLikesCheeseWhiz`, whatever.

src/Engine/GameEvents/HealUnitEvent.java
ThatIsAUsername: > "always HP, sometimes floating point"... I just dunno that that's more desirable than what we have.

I will vote in the affirmative.

ThislsAUsername: > Which seems both very clear and very different than what happened. It would have been good to mention that you had a different fix in mind; then we could have pushed it to a new PR. On the other hand, if you think this implementation matches that description, then we are indeed speaking different languages.

Well. I *started* to do what you suggested, then said "wait, that's really hard to keep track of" because the important thing is HP.

> > "always HP, sometimes floating point"... I just dunno that that's more desirable than what we have.
> 
> I will vote in the affirmative.

Okay. Doesn't thrill me, but I'll live.

ThislsAUsername:
Probably the easiest to review this next commit's diff is to compare with commit `EATS_BULLETS`/b5a712250ba46952bb55f8b1f64b6fcff16df60c, since that just leaves the intervening commits' changes.

src/Units/Unit.java
```diff
@@ -191,79 +191,59 @@ public void fire(WeaponModel weapon)
...
+  private static double healthToHP(int input)
   {
-    return effectiveHealth(input) / 10;
+    return ((double)input)/10;
   }
-  public static int healthFromHP(int input)
+  private static int healthFromHP(double input)
   {
-    return input * 10;
+    return (int) (input * 10);
```
ThatIsAUsername, line 218: You mentioned wanting to leave open the possibility of units with `maxHP != 10`. Shouldn't this 10 and the one above be a `model.maxHP` reference? I know these are static right now, but maybe they shouldn't be, unless we'd rather pass in the model.

ThislsAUsername: Nope. This is a 100% static conversion factor - it defines 1 HP == 10 health, irrespective of how much HP you *can* have.
So if I have 110 health and I divided by maxHP, I would report 10 HP instead of 11 HP.

ThislsAUsername:
I believe I'm satisfied with this PR at this point.

PR #248, opened 2020-05-19
Meteors

ThislsAUsername:
I'm a silly goose


src/AI/Muriel.java
```diff
@@ -331,18 +330,18 @@ public GameAction getNextAction(GameMap gameMap)
...
 
-          // Calculate the cost of the damage we can do.
-          double attackDamage = CombatEngine.calculateOneStrikeDamage(unit, 1, target, gameMap, environment.terrainType.getDefLevel(), unit.model.hasMobileWeapon());
+                if( shouldAttack(unit, results.defender, gameMap) )
+                  return (results.defender.model.getCost() / 10) * hpDamage;
 
-          double hpDamage = Math.min(attackDamage, target.getPreciseHP());
-          double damageValue = (target.model.getCost()/10) * hpDamage;
+                return 0.;
+              }, (terrain, params) -> 0.01); // Attack terrain, but don't prioritize it over units
```
ThatIsAUsername, line 341: This gives attacking terrain too much importance. You think I'm kidding, but it totally derails units who have more important things to do and shouldn't be messing with meteors in the first place. I think I'd rather Muriel just deny the existence of meteors than do this.

src/Engine/Combat/CombatEngine.java
```diff
@@ -42,11 +63,12 @@ public static BattleSummary simulateBattleResults( Unit attacker, Unit defender,
...
    * @return A BattleSummary object containing all relevant details from this combat instance.
    */
-  public static BattleSummary calculateBattleResults(Unit attacker, Unit defender, GameMap map, int attackerX, int attackerY, boolean isSim)
+  public static BattleSummary calculateBattleResults(Unit attacker, Unit defender, GameMap map,
+                                                     boolean attackerMoved, int attackerX, int attackerY,
+                                                     boolean isSim)
   {
     int defenderX = defender.x; // This variable is technically not necessary, but provides consistent names with attackerX/Y.
     int defenderY = defender.y;
-    boolean attackerMoved = attacker.x != attackerX || attacker.y != attackerY;
```
ThatIsAUsername, line 49: Why move this into the parameter list?

src/Terrain/TerrainType.java
```diff
@@ -52,134 +54,142 @@ private TerrainType(int attributeFlags, int defense, Color mainColor, String nam
...
+  public TerrainType getBaseTerrain() { return (null == mBase)? SEA : mBase; }
   
   ////////////////////////////////////////////////////////////////////////////////////////////
   //// Publicly-accessible terrain Flyweight instances.
-  private static final int AIRPORT_FLAGS = LAND | CAPTURABLE | PROFITABLE | PROVIDES_COVER | HEALS_AIR;
-  private static final int AIRPORT_DEFENSE = 3;
-  private static final Color AIRPORT_COLOR = new Color(125, 125, 125); // TODO: define unique color for each building type?
-  private static final String AIRPORT_NAME = "AIRPORT";
-  public static final TerrainType AIRPORT = new TerrainType( AIRPORT_FLAGS, AIRPORT_DEFENSE, AIRPORT_COLOR, AIRPORT_NAME);
+  private static final int SEA_FLAGS = WATER;
```
ThatIsAUsername, line 61: Why reorder all of these?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -262,6 +262,11 @@ public void render(Graphics g)
       unitArtist.drawUnitIcons(mapGraphics, currentActor, actorCoord.xCoord, actorCoord.yCoord, animIndex);
     }
 
+    for( DamagePopup popup :
```
ThatIsAUsername, line 265: Why move when damage previews are drawn?

src/Units/DoRWeapons.java
```diff
@@ -353,35 +363,35 @@ public CarrierMGun()
...
-/* SUBTORPEDOES     */{   0,    0,     0,     0,     0,        0,    0,       0,        0,         0,        0,       0,         0,  55,       0,      0,        0,      0,        0,        0,     120,      20,  55,      55,     110,         80,     85 }, /* SUBTORPEDOES     */
-/* CARRIERMGUN      */{   0,    0,     0,     0,     0,        0,    0,       0,        0,         0,        0,       0,         0,  55,      35,     35,       40,     40,       45,       55,       0,       0,   0,       0,       0,          0,      0 }, /* CARRIERMGUN      */
-/* BATTLESHIPCANNON */{  75,   70,    70,    70,    70,       65,   65,      50,       40,        70,       55,      75,        75,  65,       0,      0,        0,      0,        0,        0,      95,      65,  65,      50,       0,         45,     75 }  /* BATTLESHIPCANNON */
+// defenders:      INFANTRY, MECH,  BIKE, RECON, FLARE, ANTI_AIR, TANK, MD_TANK, WAR_TANK, ARTILLERY, ANTITANK, ROCKETS, MOBILESAM, RIG, FIGHTER, BOMBER, SEAPLANE, DUSTER, B_COPTER, T_COPTER, GUNBOAT, CRUISER, SUB, SUB_SUB, CARRIER, BATTLESHIP, LANDER, METEOR
```
ThatIsAUsername, line 366: AWBW just uses NeoTank defense, and DoR has specific values? Interesting.

ThatIsAUsername:
Overall good stuff here. The targetable interface is a nice abstraction for working with variable target types. Not so much a fan of the new damage-popups pass-through call-chain though, since that encodes information for a specific action type in the GameAction interface - Why not add a target array to `ContemplatedAction` instead?
Also, FYI, Java 8 supports default method implementations in interfaces, so `GameAction` technically still doesn't have to be an abstract class.

ThatIsAUsername:
Oh, and there's no animation when attacking terrain. Should fix that too.

src/AI/Muriel.java
ThislsAUsername: Okay.

src/Terrain/TerrainType.java
ThislsAUsername: I needed to do some reordering because e.g. SEA has to be constructed before anything else can use it as a parameter in their own constructors.
I was originally planning to try to impose some sort of logical ordering, but realized I didn't know exactly what kind I wanted :P

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: We can now show damage popups for things that aren't targeted.
I added BBombs as an example.

src/Units/DoRWeapons.java
ThislsAUsername: Yeah, and IIRC AW2 actually uses Md Task defense.
Go figure.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: I wanted to use path info to do it in `BattleLifecycle`, since that's a better indicator than whatever the unit's *current* position is (transports and launch and stuff).
I could apply the change more broadly, though that'd be a lot of fiddling with logic that doesn't really need to change yet AFAIK.

ThislsAUsername:
> Not so much a fan of the new damage-popups pass-through call-chain though, since that encodes information for a specific action type in the GameAction interface

I wasn't a super fan of making a call-chain, but I figured it was better than rejiggering what's visible where.
As for the "specific action type" thing, I disagree. It's just a vehicle for showing arbitrary string/color popups on the map when the action is contemplated.

> Why not add a target array to `ContemplatedAction` instead?

That sound like it might be simpler, but the simplest solution I saw was the one I picked. Mind elaborating?

> Also, FYI, Java 8 supports default method implementations in interfaces, so `GameAction` technically still doesn't have to be an abstract class.

Hmm. I think GameEventListener could benefit from that, but I don't feel like backing out that refactor until there's a reason to do it.

src/Units/DoRWeapons.java
ThatIsAUsername: Then--and I don't care either way--why use neo defense here?

ThatIsAUsername:
> As for the "specific action type" thing, I disagree. It's just a vehicle for showing arbitrary string/color popups on the map when the action is contemplated.

The name of the call chain is specific to a particular action type. If it's a vehicle for arbitrary info popups, the name should reflect that.

> That sound like it might be simpler, but the simplest solution I saw was the one I picked. Mind elaborating?

`MapController` has a private `ContemplatedAction` class with info about the action being considered. `SpriteMapView` already calls `getContemplated<blah>` to retrieve the actor, its location, and whether it is currently targeting an action. `ContemplatedAction` could have a list of targets instead of an `isAiming` Boolean. This would probably require building the damage preview boxes on the UI side again, but since they are a UI affordance that seems sensible anyway.

src/Units/DoRWeapons.java
ThislsAUsername: `AWBWUnits.java`, really.
The immediate use case is AWBW players who want to playtest stuff, so that's the first thing I'm gonna support.

ThatIsAUsername: What? Somehow your explanations often leave me more confused. `AWBWUnits.java` has stats for Md Tanks as well.

ThislsAUsername:
> The name of the call chain is specific to a particular action type. If it's a vehicle for arbitrary info popups, the name should reflect that.

That's harder to make a name for, especially a pithy one.
"Damage popup" also has the advantage of evoking more clearly what it should look like on the UI, though I suppose you could call that a disadvantage depending on viewpoint.

> `MapController` has a private `ContemplatedAction` class with info about the action being considered. `SpriteMapView` already calls `getContemplated<blah>` to retrieve the actor, its location, and whether it is currently targeting an action. `ContemplatedAction` could have a list of targets instead of an `isAiming` Boolean. This would probably require building the damage preview boxes on the UI side again, but since they are a UI affordance that seems sensible anyway.

You lose important information with that, though.
The UI has no idea how much damage an EXPLODE action can do, and it doesn't even know (nor, in fact, does `MapController`) that an EXPLODE action is currently being considered.

src/Units/DoRWeapons.java
ThislsAUsername: Yes, but it's called `AWBWUnits.java`, not `AW2Units.java`.
If I'm gonna simulate something, I might as well do it right ~~unlike AWBW~~

ThatIsAUsername:
> The UI has no idea how much damage an EXPLODE action can do

True; maybe `ContemplatedAction` should actually have the action in question once one is ready.

ThislsAUsername:
> True; maybe `ContemplatedAction` should actually have the action in question once one is ready.

This would imply we need a function *very similar* to `Select`, except it outputs a `GameAction` instead of changing our `GameInputState`.
Sounds like a good idea, if we have a good enough way to discriminate. I guess `getContemplatedAction<?>(? input)` would work?

ThatIsAUsername:
> This would imply we need a function very similar to Select, except it outputs a GameAction instead of changing our GameInputState

Maybe. I guess `getInputType()` and `getReadyAction()` won't work because they aren't useful until committing to the action. I'm not sure I would like what adding a `getActionPreview()` or similar would do to that interface.

> EDIT: Call me crazy

You crazy. That might work though, I dunno.

ThislsAUsername:
> You crazy. That might work though, I dunno.

I feel like it makes sense to put off a rework of our UI-based action construction system for another PR. How about you?

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: > I wanted to use path info to do it in BattleLifecycle, since that's a better indicator than whatever the unit's current position is (transports and launch and stuff).

Isn't `Path` just an equal indicator of the unit's current position? The `BattleEvent` will always be created (and the movement status/battle results calculated) before the unit actually moves, and this check works for units at (-1, -1) or whatever. It feels like duplicate information.

ThatIsAUsername:
> UI-based

Not sure what you mean by that, but at any rate, I know "eventually" often means probably never (or at least until it gets painful enough that we can't ignore it). I agree trying to move map-highlight logic or whatever inside the InputHandlers is a larger chunk to bite off and way beyond scope for this PR, but how hard would it be to just add either the targets or the Damage Previews to the `ContemplatedAction` just to keep the interface cleaner?
If that's not gonna work for whatever reason we can just hold off until it gets painful enough that we can't ignore it.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: The `Path` will always have exactly correct position information (whereas the unit might be off-map)
You're right, realistically.
However, I have now realized the *real* reason: `calculateOneStrikeDamage()` needs it to be a param because Muriel does her combat assessments with fake units.

ThislsAUsername:
> > UI-based
> 
> Not sure what you mean by that

Most actions are constructed by Unit directly. What I meant was just "through user interaction".

> I know "eventually" often means probably never (or at least until it gets painful enough that we can't ignore it)

Hah, you might be right about that.

> I agree trying to move map-highlight logic or whatever inside the InputHandlers is a larger chunk to bite off and way beyond scope for this PR, but how hard would it be to just add either the targets or the Damage Previews to the `ContemplatedAction` just to keep the interface cleaner?

You'd need the same call chain regardless, I think, unless we wanted `MapController` to calculate which action it's looking at itself. That might be a bit annoying to do, dunno.
Also, storing it in the ContemplatedAction makes it a copy of the state that needs to be managed, and I don't really feel like assessing where all that needs to be managed right now.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: Then `calculateOneStrikeDamage()` needs it. Why here?

ThatIsAUsername:
> What I meant was just "through user interaction"

Welp, that's not what I had assumed you meant, and also still doesn't tell me what you actually mean. Are you talking about completely gutting the current Unit/Action paradigm to create actions in a wholly different way? Because that sounds crazy.

> Hah, you might be right about that.

Which is why I normally try to absorb the pain up-front and not put myself into technical debt to begin with. I don't like doing things that work "for now" when I already know a better way to do it.

> MapController to calculate which action it's looking

`SpriteMapView` was doing the battle simulations before; the only reason it can't still do that is because BBomb doesn't behave like other units.

> storing it in the ContemplatedAction makes it a copy

Yes. It's basically cliff-notes of the current `InputHandler` state. Maybe `InputHandler` just needs a way to access the state data or something.

Also:
Should we make `GameAction` an interface again, or just leave it for now? Presumably whenever we get rid of the whole damage preview thing it can go back to being one anyway.
Also the fact that we are talking about getting rid of a feature of the pull request before it's even merged makes me think we should get rid of it before it's merged.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: Ah, now I remember.
I was contemplated a more limited version of LAUNCH that would cause the unit to unload, act in place, and immediately re-load.
In that particular case, a 1-length Path would be the only way to know (before you actually *execute* the move) that the unit hasn't moved.

ThatIsAUsername: Is that... something we want to allow? I don't think AW allows it, and you were just waxing eloquent about source accuracy.
At any rate, I'd prefer to add affordances for new capabilities along with those capabilities; otherwise, the needed hooks often end up different from what I anticipated anyway.

ThislsAUsername:
> Welp, that's not what I had assumed you meant, and also still doesn't tell me what you actually mean. Are you talking about completely gutting the current Unit/Action paradigm to create actions in a wholly different way? Because that sounds crazy.

What? No. Just moving the input-handling logic into the `GameInputState` machine from `MapController`.

> Which is why I normally try to absorb the pain up-front and not put myself into technical debt to begin with. I don't like doing things that work "for now" when I already know a better way to do it.

Sure, but this isn't at all a small bug-fix. It belongs in its own PR, IMO.
I'm not sure I believe you on the "later means never", either.

> `SpriteMapView` was doing the battle simulations before; the only reason it can't still do that is because BBomb doesn't behave like other units.

Right, but IMO it's cleaner to drive it by the action rather than having the UI *assume* that a given `MapController` state implies a certain type of action is being planned.

> Yes. It's basically cliff-notes of the current `InputHandler` state. Maybe `InputHandler` just needs a way to access the state data or something.

Sounds reasonable.

> Should we make `GameAction` an interface again, or just leave it for now?

I was planning to leave it. I can't think of any reason a GameAction needs to inherit from something else :P

> Presumably whenever we get rid of the whole damage preview thing it can go back to being one anyway.

I feel like you and I have different definitions of what technical debt is.

> Also the fact that we are talking about getting rid of a feature of the pull request before it's even merged makes me think we should get rid of it before it's merged.

But... we're *not* talking about removing it? Maybe you are, but I'm not. IMO, having the action handle damage previews directly is the solution that most closely maps the domain to the code.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: > Is that... something we want to allow? I don't think AW allows it, and you were just waxing eloquent about source accuracy.

Not on LAUNCH, no. But I have ideas for things that would work quite nicely with that feature.
Imagine a transport that can cart a Battleship up onto the land so it can snipe from a better position :stuck_out_tongue: 

> At any rate, I'd prefer to add affordances for new capabilities along with those capabilities; otherwise, the needed hooks often end up different from what I anticipated anyway.

That's a fair take. I feel like I have a pretty good track record on that, but I also somewhat frequently contemplate stuff off and on for months before I actually code it lol

ThatIsAUsername:
> I feel like you and I have different definitions of what technical debt is.

We seem to have different ideas of lots of things.

> But... we're not talking about removing it?

Depends on what "it" is, I suppose. I was referring to the damage preview callback, specifically, which I though you were planning to magic away as part of the proposed InputHandler change, which I agree is beyond scope for here.

I'm fine with waiting to see if and how this becomes painful, since hopefully it'll be you who has to clean it up.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: > Imagine a transport that can cart a Battleship up onto the land so it can snipe from a better position

I prefer to save my nightmares for when I'm actually asleep, thank you very much.

ThislsAUsername: > I prefer to save my nightmares for when I'm actually asleep, thank you very much.

You're welcome?

PR #249, opened 2020-05-24
Qol fixes

ThatIsAUsername:
Muriel no longer walks in range of enemy units quite so readily.
Also fixed some issues with fuel and ammo.
Just getting this in now to pave the way for more improvements once things stabilize.

### DIFF
--- a/src/AI/Muriel.java
+++ b/src/AI/Muriel.java
@@ -405,8 +405,30 @@ public class Muriel implements AIController
           UnitMatchupAndMetaInfo umami = getUnitMatchupInfo(unit, target);
           if( umami.costEffectivenessRatio < COST_EFFECTIVENESS_MIN ) continue;
 
+          // Find locations that would be dangerous for us so we can avoid sauntering into enemy fire.
+          HashSet<XYCoord> noGoZone = new HashSet<XYCoord>();
+          final int MAX_RELEVANT_DISTANCE = 27; // ~3x the move distance of the fastest units. Up for tweaking.
+          for(int j = 0; j < enemyLocations.size(); ++j)
+          {
+            XYCoord threatCoord = enemyLocations.get(j);
+            Unit threat = gameMap.getLocation(threatCoord).getResident();
+            XYCoord unitCoord = new XYCoord(unit.x, unit.y);
+            if( Utils.findManhattanDistance(unitCoord, threatCoord) <= MAX_RELEVANT_DISTANCE )
+            {
+              // If we, in the enemy's place, would attack `unit` with `threat`, then we should not let them attack us.
+              UnitMatchupAndMetaInfo invUmami = getUnitMatchupInfo(threat, unit);
+              if( invUmami.costEffectivenessRatio > COST_EFFECTIVENESS_MIN && threat.getHP() >= 6 )
+              {
+                // Add coordinates that `threat` could target to our "no-go" list.
+                Map<XYCoord, Double> threatMap = AIUtils.findThreatPower(gameMap, threat, unit.model);
+                noGoZone.addAll(threatMap.keySet()); // Ignore the valueMap of the return; we have already decided `threat` is dangerous.
+              }
+            }
+            else break; // Don't bother considering far-away baddies for our no-go zone.
+          }
+
           // Try to move towards the enemy, but avoid blocking a factory.
-          move = AIUtils.moveTowardLocation(unit, coord, gameMap);
+          move = AIUtils.moveTowardLocation(unit, coord, gameMap, noGoZone);
           if( null != move && (gameMap.getLocation(move.getMoveLocation()).getEnvironment().terrainType != TerrainType.FACTORY))
           {
             log(String.format("  Found %s", gameMap.getLocation(coord).getResident().toStringWithLocation()));
--- a/src/AI/Muriel.java
+++ b/src/AI/Muriel.java
@@ -316,6 +316,10 @@ public class Muriel implements AIController
           }
         }
         if( !queuedActions.isEmpty() ) break; // Break so we don't inadvertently plan two actions for this unit.
+        else
+        {
+          log("  Cannot find an available resupply station.");
+        }
       }
 
       // Find all the things we can do from here.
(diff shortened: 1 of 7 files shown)


ThislsAUsername:
...It's gonna be really annoying if there are conflicts with `meteors`.

src/AI/Muriel.java
```diff
@@ -405,8 +405,30 @@ public GameAction getNextAction(GameMap gameMap)
...
+              // If we, in the enemy's place, would attack `unit` with `threat`, then we should not let them attack us.
+              UnitMatchupAndMetaInfo invUmami = getUnitMatchupInfo(threat, unit);
+              if( invUmami.costEffectivenessRatio > COST_EFFECTIVENESS_MIN && threat.getHP() >= 6 )
+              {
+                // Add coordinates that `threat` could target to our "no-go" list.
+                Map<XYCoord, Double> threatMap = AIUtils.findThreatPower(gameMap, threat, unit.model);
+                noGoZone.addAll(threatMap.keySet()); // Ignore the valueMap of the return; we have already decided `threat` is dangerous.
+              }
+            }
+            else break; // Don't bother considering far-away baddies for our no-go zone.
```
ThislsAUsername, line 432: I'd change these to `if x break;` and `if y continue;`, since I find that construct more readable than enclosing stuff in braces.

src/AI/Muriel.java
```diff
@@ -405,8 +405,30 @@ public GameAction getNextAction(GameMap gameMap)
...
+            XYCoord unitCoord = new XYCoord(unit.x, unit.y);
+            if( Utils.findManhattanDistance(unitCoord, threatCoord) <= MAX_RELEVANT_DISTANCE )
+            {
+              // If we, in the enemy's place, would attack `unit` with `threat`, then we should not let them attack us.
+              UnitMatchupAndMetaInfo invUmami = getUnitMatchupInfo(threat, unit);
+              if( invUmami.costEffectivenessRatio > COST_EFFECTIVENESS_MIN && threat.getHP() >= 6 )
+              {
+                // Add coordinates that `threat` could target to our "no-go" list.
+                Map<XYCoord, Double> threatMap = AIUtils.findThreatPower(gameMap, threat, unit.model);
+                noGoZone.addAll(threatMap.keySet()); // Ignore the valueMap of the return; we have already decided `threat` is dangerous.
```
ThislsAUsername, line 429: This does ignore the effect of terrain and the multiple-weapons-with-different-ranges thingie.
That's not really important at the moment, though.

Might be worth breaking `findThreatPower()` into two parts: "where can this threaten" and "how threatening, really?"
Then again, that would be difficult to do right since you'd have to reverse-engineer the conditions under which the unit in question could threaten the tile you'd like to diagnose.
It's interesting looking at my past design decisions and going "this could be better" and then "wait nevermind, past me has a point"

src/Engine/Utils.java
```diff
@@ -446,6 +446,12 @@ public int compare(SearchNode o1, SearchNode o2)
     return industries;
   }
 
+  /** Calculate the Manhattan distance between the two coordinates (number of moves for a flying unit). */
+  public static int findManhattanDistance(XYCoord c1, XYCoord c2)
+  {
+    return Math.abs(c1.xCoord - c2.xCoord) + Math.abs(c1.yCoord - c2.yCoord);
+  }
```
ThislsAUsername, line 453: You uh... you do realize we have a function on XYCoord for this, right?

src/Units/Unit.java
```diff
@@ -184,6 +190,8 @@ public WeaponModel chooseWeapon(UnitModel targetType, int range, boolean afterMo
     double maxDamage = 0;
     for( WeaponModel weapon : model.weapons )
     {
+      if( !weapon.loaded(this) ) continue; // Can't shoot with no bullets.
```
ThislsAUsername, line 193: Hah! Unforeseen consequence of killing `Weapon`.
Good catch.

ThislsAUsername:
I'm gonna bet Muriel stomps Wally pretty hard now.

src/AI/Muriel.java
ThatIsAUsername: > effect of terrain

One day I intend to account for that. I have ideas. Not today.

And yeah, it can be hard to break logic apart without forcing yourself to repeat some of it.

src/Engine/Utils.java
ThatIsAUsername: I thought I had made one of those at one point, but I couldn't find it when I went looking.

ThatIsAUsername:
> I'm gonna bet Muriel stomps Wally pretty hard now.

It's not as big a change as you might think. She will still happily do attacks that take units into enemy range, so once units start mixing it kinda looks mostly the same.

src/AI/Muriel.java
ThatIsAUsername: Not sure what you mean. This line has no braces, and there is no `continue` here. Do you mean to invert the conditional above and do
```java
if( !( Utils.findManhattanDistance(unitCoord, threatCoord) <= MAX_RELEVANT_DISTANCE ) )
  break;
else
{
  // Do the logic
}
```

This way makes sense to me. If conditions are met, do all this logic, otherwise just skip it.

ThislsAUsername: More like:
```java
if( unitCoord.getDistance(threatCoord) >= MAX_RELEVANT_DISTANCE )
  break; // No more relevant threats
if( !threat.canAttack(unit.model) && shouldAttack(threat, unit, gameMap) )
  continue; // This threat isn't one
// Do logic
```
Either way makes sense, but fewer syntactically relevant lines (assuming you aren't arbitrarily grouping things) makes the code more readable IMO.
In this case, you no longer have an `if...else` to keep track of and two fewer scopes to keep track of.

ThatIsAUsername: Ah, that makes sense. I'll do things like that for longer functions sometimes, but for this I didn't bother. Maybe I'll change it if I have to go back in and add/refactor stuff later.

ThislsAUsername:
Looks good to go to me.

PR #250, opened 2020-05-30
Input rejiggering

ThislsAUsername:
I found a good name.

### DIFF
--- a/src/Engine/GameInput/ConfirmUnitAction.java
+++ b/src/Engine/GameInput/ConfirmUnitAction.java
@@ -1,10 +1,7 @@
 package Engine.GameInput;
 
-import java.util.Collection;
-
-import Engine.XYCoord;
+import java.util.ArrayList;
 import Engine.Combat.DamagePopup;
-import Terrain.GameMap;
 
 public class ConfirmUnitAction extends GameInputState<ConfirmUnitAction.ConfirmActionEnum>
 {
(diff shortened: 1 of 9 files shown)


ThatIsAUsername:
Highlighting between different attack options doesn't update/move the damage previews. Looks like FreeTile and ConstrainedTile select functions are missing a `consider` call.

Comment at SelectUnitAction:73 refers to unit deletion, but BBombs also require confirmation (Only semi-related, but I noticed it while reading through).

BBomb damage popups can get hidden behind confirmation menu. Not sure what the best fix is here, but it's a thing.

ThislsAUsername:
> Highlighting between different attack options doesn't update/move the damage previews. Looks like FreeTile and ConstrainedTile select functions are missing a `consider` call.

Huh. I don't know how I missed that. Coulda sworn I tested it.

> Comment at SelectUnitAction:73 refers to unit deletion, but BBombs also require confirmation (Only semi-related, but I noticed it while reading through).

`EXPLODE` is just an extra-fancy `DELETE` :stuck_out_tongue: 

> BBomb damage popups can get hidden behind confirmation menu. Not sure what the best fix is here, but it's a thing.

Yeah. Damage popups for *attacks* can also get hidden under each other and stuff too (say you target an adjacent unit and one of them does over 100%)
I guess we could make a "hide menu" key for double-checking stuff, but that wouldn't help with the popups themselves overlapping.

PR #251, opened 2020-06-01
Move anims

ThislsAUsername:
Definitely still needs more animations done, but I don't feel like messing with that quite yet.
Fixes the grey-while-attacking thing, too, so that's cool.

If you have strong feelings on the file naming convention, speak now.

This may conflict with my other PR and probably conflicts with yours, but c'est la vie.


src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -246,7 +248,17 @@ public void render(Graphics g)
 
     // Get a reference to the current action being built, if one exists.
     Unit currentActor = mapController.getContemplatedActor();
-    XYCoord actorCoord = mapController.getContemplationCoord();
+    AnimState actorAnimState = AnimState.IDLE;
+    XYCoord actorDrawCoord = mapController.getContemplationCoord();
```
ThatIsAUsername, line 252: How did we end up with `getContemplatedActor` and `getContemplationCoord`?

src/UI/Art/Animation/GameAnimation.java
```diff
@@ -13,4 +28,23 @@
    * Allows the caller to tell this animation to end early.
    */
   public void cancel();
+
+  public default Unit getActor()
```
ThatIsAUsername, line 32: These changes break the interface. `GameAnimation` should be usable for CO abilities, in-game events (e.g. volcanic eruptions), and whatever else. Pushing unit-specific (and worse, action-type-specific) information to the top level ruins the abstraction. Eventually, we'll just have to keep adding things to support more stuff for different animation types.

src/UI/Art/Animation/SimpleMoveAnimation.java
```diff
@@ -0,0 +1,93 @@
+package UI.Art.Animation;
+
+import java.awt.Graphics;
+import Engine.Path;
+import Engine.XYCoord;
+import Units.Unit;
+
+public class SimpleMoveAnimation implements GameAnimation
```
ThatIsAUsername, line 8: Why "Simple"? Are you also going to make a `ComplexMoveAnimation`?

src/UI/Art/SpriteArtist/Sprite.java
```diff
@@ -63,9 +63,12 @@ public Sprite(BufferedImage spriteSheet, int spriteWidthPx, int spriteHeightPx)
 
   /**
    * Sprite copy-constructor. Perform a deep-copy on each of the other sprite's frames.
-   * @param other
    */
   public Sprite(Sprite other)
+  {
+    this(other, false);
+  }
+  public Sprite(Sprite other, boolean hFlip)
```
ThatIsAUsername, line 71: Why create a flipped copy instead of just drawing the image flipped (as is currently done for the idle sprites)?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -3,119 +3,95 @@
...
-   * 
-   * @param spriteSheet The image containing the frames to be used in this sprite set.
-   * @param width The width of each frame in the sprite.
-   * @param height The height of each frame in the sprite.
+   * Fetch a sprite sheet for each animation state and colorize it
+   * @param fileFinder Mapping from action type to file name
+   * @param coColors What colors the end sprites should use
    */
-  public UnitSpriteSet(BufferedImage spriteSheet, int width, int height, ColorPalette coColors)
+  public UnitSpriteSet(Function<AnimState, String> fileFinder, ColorPalette coColors)
```
ThatIsAUsername, line 25: Why bust out a lambda function when you could just pass in a templated string?

I know we have officially moved to Java 8, but that doesn't mean we need to pull in as many shiny new toys as possible, especially when they involve distributing pointers to private member functions.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -3,119 +3,95 @@
...
-
-      System.out.println("WARNING: RasterFormatException in UnitSpriteSet constructor.");
+      System.out.println("WARNING: Exception hit in UnitSpriteSet constructor:" + e);
       System.out.println("WARNING:   Attempting to continue.");
-      for( ; action <= ACTION_DIE; ++action )
-      {
+    }
+
+    // Handle the case of having one sideways move direction and not the other defined
+    if( null == sprites[AnimState.MOVEEAST.ordinal()] && null != sprites[AnimState.MOVEWEST.ordinal()] )
```
ThatIsAUsername, line 48: Where did you get the move sprites? I'm gonna infer that they all game in west-facing format, and you just decided to future-proof that variable.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -3,119 +3,95 @@
...
+      sprites[AnimState.MOVEEAST.ordinal()] = new Sprite(sprites[AnimState.MOVEWEST.ordinal()], true);
+
+    if( null == sprites[AnimState.MOVEWEST.ordinal()] && null != sprites[AnimState.MOVEEAST.ordinal()] )
+      sprites[AnimState.MOVEWEST.ordinal()] = new Sprite(sprites[AnimState.MOVEEAST.ordinal()], true);
+
+    // Fill out any missing images
+    Sprite defaultSprite = sprites[AnimState.IDLE.ordinal()];
+    // Use the IDLE action if it exists, otherwise we are going with a black rectangle.
+    if( null == defaultSprite )
+      new Sprite(null, SpriteLibrary.baseSpriteSize, SpriteLibrary.baseSpriteSize);
```
ThatIsAUsername, line 58: Shouldn't there be an assignment here?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -241,4 +207,18 @@ public void drawUnitIcons(Graphics g, Commander[] COs, Unit u, int animIndex, in
...
+
+  public static boolean isStateFlippable(AnimState state)
+  {
+    switch(state)
+    {
+      case MOVENORTH:
+      case MOVEEAST:
+      case MOVESOUTH:
+      case MOVEWEST:
+        return false;
```
ThatIsAUsername, line 219: Why not east/west?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -259,8 +271,10 @@ public void render(Graphics g)
     // Draw the currently-acting unit so it's on top of everything.
     if( null != currentActor )
     {
-      unitArtist.drawUnit(mapGraphics, currentActor, actorCoord.xCoord, actorCoord.yCoord, fastAnimIndex);
-      unitArtist.drawUnitIcons(mapGraphics, currentActor, actorCoord.xCoord, actorCoord.yCoord, animIndex);
+      SpriteLibrary.getMapUnitSpriteSet(currentActor)
```
ThatIsAUsername, line 274: Why are we bypassing `UnitArtist` now?
Also this won't draw unit icons when a unit is selected anymore.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -318,20 +320,24 @@ public static UnitSpriteSet getMapUnitSpriteSet(String type, Faction faction, Co
...
   private static String getMapUnitSpriteFilename(String unitType, String faction)
-  {
-    StringBuffer spriteFile = new StringBuffer();
-    spriteFile.append("res/unit/faction/").append(faction).append("/");
-    spriteFile.append(UnitModel.standardizeID(unitType)).append("_map.png");
-    return spriteFile.toString();
+  { return getMapUnitSpriteFilename(unitType, faction, AnimState.IDLE); }
+  private static String getMapUnitSpriteFilename(String unitType, String faction, AnimState state)
+  {
+    final String format = "res/unit/faction/%s/%s_map%s.png";
```
ThatIsAUsername, line 336: Ah, so you have both lambdas *and* templated strings.I still feel like this could be simplified.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
```diff
@@ -264,7 +265,7 @@ public BufferedImage update(PlayerSetupInfo info)
       if( factionChanged || colorChanged )
       {
         UnitSpriteSet inf = SpriteLibrary.getMapUnitSpriteSet(myControl.getIconicUnit(), info.getCurrentFaction(), info.getCurrentColor());
-        BufferedImage infSprite = inf.sprites[inf.ACTION_IDLE].getFrame(0);
+        BufferedImage infSprite = inf.sprites[AnimState.IDLE.ordinal()].getFrame(0);
```
ThatIsAUsername, line 268: This is the part that will cause conflicts with the `tech` branch. Incidentally, why change this to use `ordinal()`?

src/UI/Art/Animation/SimpleMoveAnimation.java
```diff
@@ -0,0 +1,93 @@
...
+    this.path = path;
+    startTime = System.currentTimeMillis();
+    tilesPerMs = actor.model.movePower / (double) maxTime;
+    endTime = (long) ((path.getPathLength() - 1) / tilesPerMs);
+  }
+
+  @Override
+  public boolean animate(Graphics g)
+  {
+    long animTime = System.currentTimeMillis() - startTime;
```
ThatIsAUsername, line 31: This... doesn't actually do the animating. Why?

src/UI/Art/SpriteArtist/UnitArtist.java
```diff
@@ -36,12 +38,13 @@ public void drawUnitIcons(Graphics g, Unit unit, double x, double y, int animInd
    */
   public void drawUnit(Graphics g, Unit unit, int x, int y, int animIndex)
   {
-    // Convert "real" game-model location to a draw-space location.
     int drawX = (int) (myView.getTileSize() * x);
     int drawY = (int) (myView.getTileSize() * y);
+    boolean tired = unit.isStunned || (unit.isTurnOver && unit.CO == myGame.activeCO);
+    AnimState state = (tired) ? AnimState.TIRED : AnimState.IDLE;
```
ThatIsAUsername, line 44: This seems duplicative-yet-different of `SimpleMoveAnimation.getAnimState()`.

src/UI/Art/Animation/SimpleMoveAnimation.java
```diff
@@ -0,0 +1,93 @@
...
+  private final Path path;
+  private final double tilesPerMs;
+
+  public SimpleMoveAnimation(Unit actor, Path path)
+  {
+    this.actor = actor;
+    this.path = path;
+    startTime = System.currentTimeMillis();
+    tilesPerMs = actor.model.movePower / (double) maxTime;
+    endTime = (long) ((path.getPathLength() - 1) / tilesPerMs);
```
ThatIsAUsername, line 25: This makes it finish immediately for wait-in-place actions. Much better.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: I blame me. I like to say function names in my head, and those flow nicer than standardizing.

src/UI/Art/Animation/GameAnimation.java
ThislsAUsername: The `MapView` needs to know whether/where to draw the current acting unit.
`MapView` doesn't have a pre-existing concept of volcano eruptions, so it won't need instructions to modify how it draws any particular one. Units, OTOH, are always being drawn.
This is a special case that's broadly applicable and (because of default implementations) won't require animations that don't need it to put code in to support it.

The only fix I can see is making move and combat animations draw the unit themselves, and still report `getActor()` so the `MapView` knows not to draw the actor itself.

src/UI/Art/Animation/SimpleMoveAnimation.java
ThislsAUsername: Is teleportation not a form of movement?

src/UI/Art/SpriteArtist/Sprite.java
ThislsAUsername: Because I didn't want to put code in to figure out which direction of sprite (EAST/WEST) we have and draw that flipped in the animation code.
Much simpler to just make sure we have all relevant directions and just use the right one.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: Handing out a pointer to a private member is more private than just making the member public, which is the only thing I could do other than copy/pasting code.
...and besides, I like lambdas.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: A spritesheet. And yeah.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: See above where I mentioned it's simpler to just have all the directions than figure it out at runtime.
I suppose we could just standardize and hold one copy for east/west, and just *always* flip the other direction?

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Didn't want to change `UnitArtist` to take draw-space coordinates. (Tried it, and it was a little ugly having an overload that take the same basic info of two functions)

Second point is an intended change, though I guess I could revert that.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: /shrug

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThislsAUsername: Because it was type `int` before, now it's type `Enum`

src/UI/Art/Animation/SimpleMoveAnimation.java
ThislsAUsername: Because that means you have to duplicate drawing logic for each animation that wants to have the unit not get greyed out. (movement, combat, probably repair/supply?)
I guess we could make a "UnitAnimation" base class, but that's still questionable.

Also also, Animations are in a really weird spot that sits between being fully `SpriteArtist` specific and fully generic. Having logic to draw the units directly should also be accompanied by moving the animations in question into the `SpriteArtist` namespace.

src/UI/Art/SpriteArtist/UnitArtist.java
ThislsAUsername: I pulled it up out of the `UnitSpriteSet` drawing function, since it no longer needs to know the unit's state when the caller can tell it exactly what to draw.

...and it's not really duplicated logic, they're different functions using the same interface :P

src/UI/Art/Animation/SimpleMoveAnimation.java
ThislsAUsername: Pretty happy about how animations have a reasonable (and hard-limit) max time, and faster units animate faster.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: Yeah, that could work. You have to do the check when you first read the file anyway, so you could just do the flip there (if needed) and simplify the logic everywhere else.

ThislsAUsername: Not having to have flipping code in the draw function would be cool, but we technically store sprites relative to faction/colors rather than CO instance or HQ, so we'd have to change that in order to do it properly.
I feel like faction/colors is also the *right* level to store them, too, as that means we don't have to keep multiple copies around for different armies with the same faction/colors over the life of the application.

ThatIsAUsername: Hm, I guess I wasn't clear enough in how I said things. Let's try again, because I thought I was agreeing with you before, but your latest response implies otherwise.
Currently, before this PR, all sprites are stored on disk as right/east facing. We read/store it that way, and when drawing we flip it if needed to face left/west.
In this PR as it stands, it checks for east and west, reads either/both, and copies one to the other if one is missing. This means it independently stores two sets of sprites when one set (plus flipping on draw) would suffice (as in the era before this PR).
I'm not suggesting to remove flipping code from the draw function, just to avoid storing two sets of east/west sprites; the old/existing logic would suffice to draw flipped or not without needing the extra storage.
This would be easy to implement because you already have to know which way it's facing when you load it from disk; you could just flip west sprites to face east at load and assume everything is east-facing from there on (though you'd still have to handle north/south).

Or, in short:
> I suppose we could just standardize and hold one copy for east/west, and just always flip the other direction?

Right?

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThatIsAUsername: Ah. Don't care much either way, but why change it?

src/UI/Art/Animation/SimpleMoveAnimation.java
ThatIsAUsername: It would probably make sense to move the concrete animation classes into the `SpriteArtist` package - that was probably the original intent, e.g. `FillRectDeathAnim` vs `SpriteDeathAnim`, except now there is no `FillRect`. The whole animation package can probably go under `SpriteArtist` for all I care (and `SpriteArtist` could probably be more generic too), but that is all way out of scope for this.

> Because that means you have to duplicate drawing logic for each animation that wants to have the unit not get greyed out.

I don't think I understand. We only create greyed-out sprites for `IDLE`, right? It literally isn't possible to draw greyed-out sprites for other animations right now.

I think the only catch is that `SpriteMapView` no longer knows not to draw the acting unit, since the contemplated actor is nulled as soon as the action begins executing, but there are other ways to fix that.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: > flip west sprites to face east at load and assume everything is east-facing from there on (though you'd still have to handle north/south).

Idle/tired is "must flip if applicable"
Move north/south is "maybe flip"
Move east is "never flip"
Move west is "always flip"

I feel like that's more complex, not simpler.

src/UI/Art/SpriteArtist/Sprite.java
ThatIsAUsername: I mean. We already do that for IDLE.

src/UI/Art/Animation/SimpleMoveAnimation.java
ThatIsAUsername: Depends on which sci-fi nerd you ask, but we'll just say yes. Moving along the ground is the default though, and naming the animation to match the action is convenient.
Teleportation is a way of relocating a unit, but nobody is going to assume teleportation if you say you want to "move a unit here" (unless that's the only/most reasonable way that can happen in context).
I mean, it works, it just a) doesn't match the action, and b) seems to point to a broader picture that doesn't exist.

src/UI/Art/Animation/GameAnimation.java
ThatIsAUsername: > The only fix I can see

Hold my Kool-aid.

src/UI/Art/Animation/SimpleMoveAnimation.java
ThislsAUsername: > It would probably make sense to move the concrete animation classes into the `SpriteArtist` package - that was probably the original intent, e.g. `FillRectDeathAnim` vs `SpriteDeathAnim`, except now there is no `FillRect`.

Fair.
> The whole animation package can probably go under `SpriteArtist` for all I care (and `SpriteArtist` could probably be more generic too), but that is all way out of scope for this.

Nah. The `GameAnimation` interface at least needs to stay outside of `SpriteArtist`, as it's used in the generic `MapView` interface.
> > Because that means you have to duplicate drawing logic for each animation that wants to have the unit not get greyed out.
> 
> I don't think I understand. We only create greyed-out sprites for `IDLE`, right? It literally isn't possible to draw greyed-out sprites for other animations right now.

Right, but you have two choices:
1) Don't take over drawing the unit. `MapView` draws it for you, and it draws it `TIRED` during combat/etc because its turn is over as soon as the move event completes.
2) Do take over drawing the unit. You have flexibility, but potentially duplicate code.

I'm thinking maybe the right move is to push the "active unit drawing logic" into a base `UnitAnimation` class, and then just have the `MapView` use that to draw when there's no other action happening?
> I think the only catch is that `SpriteMapView` no longer knows not to draw the acting unit, since the contemplated actor is nulled as soon as the action begins executing, but there are other ways to fix that.

Other ways like what? The `GameEvent`/`GameAnimation` are the only ones who know what unit is acting (or if a unit is acting), since `MapController` doesn't have any visibility into AI-generated actions.

src/UI/Art/SpriteArtist/Sprite.java
ThislsAUsername: And the circumstances are different, because we don't know at `UnitSpriteSet` creation time which way IDLE should face. We *do* know for movement.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: All right, let's approach this a different way. You said:
> I suppose we could just standardize and hold one copy for east/west, and just always flip the other direction?

Clearly I have no idea what you mean by it, because every time I attempt to explain what it means you say I'm wrong. Pretend I'm three. What are you trying to say?

ThislsAUsername: You interpreted my statement correctly.
It doesn't make the logic everywhere else simpler, though.

src/UI/Art/SpriteArtist/Sprite.java
ThatIsAUsername: By that logic, we should just create idleEast and idleWest copies of the sprite.

ThislsAUsername: That's not a terrible idea, though I feel like it would necessitate some code changes that don't really need to be made.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: I suspect that the logic used was shaped by the available resources, and that it wouldn't be particularly more complex (move east/west can be the same anim state, and the `flip` parameter gets used), but I also don't think I care enough given the probably-trivial amount of RAM it would save.
Accepting either east/west but then only using east if it's there _is_ a bit odd, to be fair, but I'm not sure I see us making non-ambidextrous sprites (unless you do it just so I'm wrong here).

src/UI/Art/Animation/SimpleMoveAnimation.java
ThatIsAUsername: Imma resolve this because it's moved to the other PR.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: > I suspect that the logic used was shaped by the available resources

Not particularly? I made the logic to accept either, then realized that I'd set up the images wrong and renamed them to face the right direction :facepalm: 
> it wouldn't be particularly more complex (move east/west can be the same anim state, and the `flip` parameter gets used)

IMO the `flip` parameter should always match whatever the CO says.
> Accepting either east/west but then only using east if it's there _is_ a bit odd, to be fair, but I'm not sure I see us making non-ambidextrous sprites (unless you do it just so I'm wrong here).

You know me too well.

ThatIsAUsername: > IMO the flip parameter should always match whatever the CO says.

Ahh, that kinda hurts my head. I'd want to rename it to `coFacing` or something, so that it's clear it doesn't apply in all cases. Would it be better to have an overload that just doesn't take in that parameter instead of accepting and not using it?

ThislsAUsername: > Ahh, that kinda hurts my head

Huh. Why's that?
> I'd want to rename it to coFacing or something, so that it's clear it doesn't apply in all cases.

That's super cool with me. Still in the range of pithy, with a significant clarity advancement.
> Would it be better to have an overload that just doesn't take in that parameter instead of accepting and not using it?

My worry in that case that we'd use it for IDLE/TIRED, and get dumb results.
I guess we could like, `assert` that it's not being used for those cases, but that honestly feels even more questionable.
Maybe what we should do is just bring back `Commander` into the param list, and let that function fetch the facing for that commander in case it's needed? That should bring the invocations of that logic down to 1.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: I feel like most of the time I've seen people use lambdas it's because they think they are clever, and not because they add any value. This... doesn't seem to be an exception.
It adds complexity in order to reduce comprehensibility and increase coupling. That's a three-fer, and not in the right direction. Why not either a) partially fill out the string and pass it to `UnitSpriteSet` or b) just make `UnitSpriteSet` the owner of the string in the first place, since it's loading the actual file at this point anyway. Actually I like that better; nobody else needs to know the string or file exists at all. I may go ahead and add yet one more PR to our absurd chain here.

src/UI/Art/SpriteArtist/PlayerSetupArtist.java
ThislsAUsername: Because an int variable doesn't have a useful `.ToString()`, which I kind of depend on to read in the movement sprites.

ThislsAUsername:
Are we good here, aside from merging your beverage of choice and adding the rest of the sprites? (some of which it might be reasonable to postpone, seeing as they've not been made by the AWBW folks yet)

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: But now they aren't consistent aaaargh. I mean whatever, it works. If we didn't have an IDE with auto-complete I'd object more strongly.

ThatIsAUsername:
> Are we good here

I think so.

ThatIsAUsername:
So uh, why change `Rig` to `APC`? Won't the DoR crowd riot?

ThislsAUsername:
> So uh, why change `Rig` to `APC`? Won't the DoR crowd riot?

*shrug*
There's two versions of the DoR unit names anyway: stuff like Duster->Fighter and Fighter->Interceptor
(That's also where we got the two different power names for the DoR conversions)

If it's really that big of a deal, we can add separate concepts of "name" and "sprite name"

ThatIsAUsername:
> There's two versions of the DoR unit names anyway: stuff like Duster->Fighter and Fighter->Interceptor
> (That's also where we got the two different power names for the DoR conversions)

I hadn't known that before. Interesting. You also didn't actually answer my (first) question.

ThislsAUsername:
> I hadn't known that before. Interesting. You also didn't actually answer my (first) question.

I shrugged.
Don't really know, don't really care, not really worried about it.

ThatIsAUsername:
> I shrugged.

That's an answer to the second question, not the first.

ThislsAUsername:
> That's an answer to the second question, not the first.

About the actual renaming? Standardization. It does the exact same job.
If you don't mean that, then I'm gonna need a better hint, here.

ThatIsAUsername:
Yeah, the "why" for the renaming. It doesn't do the "exact" same job, but eh, whatever.

ThislsAUsername:
> Yeah, the "why" for the renaming. It doesn't do the "exact" same job, but eh, whatever.

It has an extra function, I suppose, but it's single-use and takes multiple turns.
It still resupplies and transports footsoldiers as its primary utility.

PR #253, opened 2020-06-08
Kool aid

ThatIsAUsername:
I was going to try and work on Tech today, but ah well.

This removes subclass-specific information from the animation interface and provides a flexible way to toggle drawing map units which can be applied to more than just the currently-acting one. It also provides the logic to draw the flipped/non-flipped version of the east/west sprite.

This does technically put UI information into a non-UI class, but as you said it's a pretty pervasive use case. We could probably avoid that by giving `SpriteMapView` a list of things to not draw or something, but that feels like it would be overdoing it at this point.

Thoughts?


src/UI/Art/Animation/SimpleMoveAnimation.java
```diff
@@ -3,6 +3,9 @@
 import java.awt.Graphics;
 import Engine.Path;
 import Engine.XYCoord;
+import UI.Art.SpriteArtist.SpriteLibrary;
+import UI.Art.SpriteArtist.UnitSpriteSet;
+import UI.Art.SpriteArtist.UnitSpriteSet.AnimState;
 import Units.Unit;
 
 public class SimpleMoveAnimation implements GameAnimation
```
ThislsAUsername, line 11: May as well rename it too, since you wanted to.

ThislsAUsername:
My trouble with putting a variable called `isDrawable` on `Unit` is that "this unit is currently being used in an animation" isn't the only usecase where you shouldn't draw the unit.
That information is encoded by `MapWindow` (wow, that's actually a terrible name) not returning the `Unit` in question, but you'd have to fiddle around and look at how components interact to figure that out.

I think I'd rather have a special case in `MapController` where it figures out if a unit is acting in a given event and caches that for return to `MapView`.
We could then further (as I mentioned before) push the draw logic into a "UnitAnimation" class, which `MapView` outsources its drawing of the currently-selected unit to (and other animations can inherit from it/steal its logic)

Your fix as-is, for instance, should make units go back to being grey during an attack.

ThatIsAUsername:
> My trouble with putting a variable called isDrawable on Unit is that "this unit is currently being used in an animation"

I mean, we could call is `isAnimating` or something to make it more clear. I think what `MapWindow` returns though is visibility, not drawability.

> where it figures out if a unit is acting in a given event and caches that for return to MapView

I guess it could have a `getLastActor()` or somewhat, and just set it based on the last event to be sent off to the animator. Or something, I dunno.

> We could then further (as I mentioned before) push the draw logic into a "UnitAnimation"

That would still require knowledge of the subclass for users of the `GameAnimation` interface, right?

> Your fix as-is, for instance, should make units go back to being grey during an attack.

Gah, you're right. Of course it does.

ThislsAUsername:
Also, you didn't get rid of the `isStateFlippable` check, so your passing in of the `flip` parameter doesn't actually do anything.
...and also that check should probably be inside the function you're calling, as "how the unit should be drawn, given a particular animation state" should be encoded within that function rather than whatever happens to *call* that function.

ThatIsAUsername:
Why did you... edit my comment instead of replying?

> Also, you didn't get rid of the isStateFlippable check, so your passing in of the flip parameter doesn't actually do anything.

Right, but it would if the east/west move sprites were flippable, and I'm not sure why they shouldn't be.

> and also that check should probably be inside the function you're calling

It could be, yeah. The function takes that as a parameter though, so that's what I did.

ThatIsAUsername:
> My trouble with putting a variable called isDrawable on Unit is that "this unit is currently being used in an animation" isn't the only usecase where you shouldn't draw the unit.

Right, and being in an animation doesn't necessarily mean it shouldn't be drawn. BattleAnimation depends on it being drawn by `MapView`. Which is maybe part of the problem. Deciding that "acting units don't get auto-drawn" and forcing the battle animator to draw the unit would fix the greyness too.
Trying to think if there are any downsides - it would keep us from arbitrarily making other units disappear if we did that instead of a `Unit` flag, but that's not a huge downside.

ThislsAUsername:
Gonna try restoring your original comment because I'm an idiot, so reposting my edit-reply-thing

> I mean, we could call is `isAnimating` or something to make it more clear. I think what `MapWindow` returns though is visibility, not drawability.
 
Would work. Feels wasteful, but would work.
It does determine visibility, sure, but you also shouldn't draw things you can't see :P

> I guess it could have a `getLastActor()` or somewhat, and just set it based on the last event to be sent off to the animator. Or something, I dunno.

Sounds about right. I'd call it `getActingUnit()`, though.

> That would still require knowledge of the subclass for users of the `GameAnimation` interface, right?

No, actually. `MapView`, at that point, would only need to know which unit to _not_ draw (info that can be provided through other means).
The only other affordance would be that `MapView` could/should build a copy of the base `UnitAnimation` while the player is contemplating a unit action.

ThislsAUsername:
> Why did you... edit my comment instead of replying?

Uhhhhhhh. Huh. I can do that?
Neat. Oops.

> Right, and being in an animation doesn't necessarily mean it shouldn't be drawn. BattleAnimation depends on it being drawn by `MapView`. Which is maybe part of the problem. Deciding that "acting units don't get auto-drawn" and forcing the battle animator to draw the unit would fix the greyness too.

I've proposed stuff based on that idea on the other PR a few times now. It's a sensible solution.
> Trying to think if there are any downsides - it would keep us from arbitrarily making other units disappear if we did that instead of a `Unit` flag, but that's not a huge downside.

There's the logic duplication, but I've proposed ways to get around that (use inheritance for the unit-based events, and `MapView` uses the base class to animate the unit being contemplated).

The only thing I can think of is if we want to have both participants in a combat do an animation (e.g. like FE map battles where they kind of bop at each other)... but again, that's less a downside and more "hey look here's a feature we don't actually care about yet".

ThatIsAUsername:
> but I've proposed ways to get around that (use inheritance for the unit-based events, and MapView uses the base class to animate the unit being contemplated).

Wanna take a stab at that as a mod to my mod to my mod to your PR?

> The only thing I can think of is if we want to have both participants in a combat do an animation (e.g. like FE map battles where they kind of bop at each other)... but again, that's less a downside and more "hey look here's a feature we don't actually care about yet".

Yes and yes. We could change the `MapController` interface to return a set of acting/animating units instead of just one if we ever want to go that way.

ThislsAUsername:
> Wanna take a stab at that as a mod to my mod to my mod to your PR?

I dunno why you made a mod branch on your mod branch, but sure. I'll try to remember to take a stab at it tonight.
> Yes and yes. We could change the `MapController` interface to return a set of acting/animating units instead of just one if we ever want to go that way.

Again, I feel like it actually makes *more* sense to put that on the `GameAnimation` side.
What if, in our fancy new `ThreeDeeArtist`, we decide to have all friendly units within X tiles of a killing blow applaud the attacker?
`MapController` might be able to figure out that we're doing combat and tell `MapView` about the two combatants, but it has no way of knowing we went off the deep end in the silliness department... nor does the `GameEvent`, since it also is agnostic to what display engine we're using.

ThislsAUsername:
I think we're good to go here.

PR #254, opened 2020-06-09
More kool

ThatIsAUsername:
Took a quick stab at tracking current unit instead of using `isDrawable`. Shouts of rage appreciated.

### DIFF
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -24,7 +24,6 @@ public class MapController implements IController, GameInputHandler.StateChanged
 
   // A menu to display options to the player.
   private InGameMenu<? extends Object> currentMenu;
-  private Unit currentActor = null;
 
   // A GameInputHandler to convert inputs into player actions, and
   // a reference to the current GameInputState's OptionSelector.
@@ -382,6 +381,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
         {
           executeGameAction(myGameInputHandler.getReadyAction());
         }
+        myGameInputHandler.reset(); // Reset the input handler to get rid of stale state
         break;
       case PATH_SELECT: // no special handling
         break;
@@ -448,7 +448,6 @@ public class MapController implements IController, GameInputHandler.StateChanged
       if( events.size() > 0 )
       {
         actionOK = true; // Invalid actions don't produce events.
-        currentActor = myGameInputHandler.getActingUnit();
         // Send the events to the animator. They will be applied/executed in animationEnded().
         changeInputMode(InputMode.ANIMATION);
         myView.animate(events);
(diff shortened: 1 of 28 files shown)


src/UI/Art/Animation/NobunagaBattleAnimation.java
```diff
@@ -30,6 +37,11 @@ public boolean animate(Graphics g)
   {
     long animTime = System.currentTimeMillis() - startTime;
 
+    // Draw the attacker in position.
+    int spriteIndex = 0; // No need to be fancy.
+    boolean flip = attackerX > defenderX;
+    actorSpriteSet.drawUnit(g, UnitSpriteSet.AnimState.IDLE, spriteIndex, attackerX * tileSize, attackerY * tileSize, flip );
```
ThislsAUsername, line 43: So the attacker always faces east, except when east of the defender.
But the defender doesn't turn to face the attacker in any case.
I think I'd prefer just using the CO's facing as we did in the past.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -258,14 +258,14 @@ public void render(Graphics g)
...
     {
       unitArtist.drawUnit(mapGraphics, currentActor, actorCoord.xCoord, actorCoord.yCoord, fastAnimIndex);
       unitArtist.drawUnitIcons(mapGraphics, currentActor, actorCoord.xCoord, actorCoord.yCoord, animIndex);
-    }
 
-    for( DamagePopup popup : mapController.getDamagePopups() )
-      drawDamagePreview(mapGraphics, popup);
+      for( DamagePopup popup : mapController.getDamagePopups() )
+        drawDamagePreview(mapGraphics, popup);
+    }
```
ThislsAUsername, line 268: Solid change.

src/Engine/MapController.java
```diff
@@ -450,6 +448,7 @@ private boolean executeGameAction(GameAction action)
       if( events.size() > 0 )
       {
         actionOK = true; // Invalid actions don't produce events.
+        currentActor = myGameInputHandler.getActingUnit();
```
ThislsAUsername, line 451: Pretty sure this still doesn't handle AIs

ThislsAUsername:
This is more or less what I had in mind, yeah.

src/Engine/MapController.java
ThatIsAUsername: It should - AI actions still go through `executeGameAction()`.

src/UI/Art/Animation/NobunagaBattleAnimation.java
ThatIsAUsername: Yeah, I agree. I didn't have that here though, since `SpriteMapView` owns it. Probably fixable if we like the direction of this change otherwise.

src/Engine/MapController.java
ThislsAUsername: Right, but AIs _don't_ go through `myGameInputHandler`.
We need a way to harvest the acting unit a) from the AI directly, b) from the event, or c) from the animation. (Also, I'm curious why adding a "get acting unit if applicable" doesn't make sense for `GameAnimation`, but it makes sense to put a drawing-logic variable in `Unit`)
Test it. There's 'ghosting'.

PR #255, opened 2020-06-10
Cyanide

ThislsAUsername:
Haven't done the inheritance thing yet, but eh, it works.

### DIFF
--- a/src/UI/Art/Animation/BaseUnitAnimation.java
+++ b/src/UI/Art/Animation/BaseUnitAnimation.java
@@ -45,7 +45,7 @@ public class BaseUnitAnimation implements GameAnimation
       AnimState state = AnimState.MOVEEAST; // Draw the unit running in place
       if( SpriteMapView.shouldFlip(actor) )
         state = AnimState.MOVEWEST;
-      drawUnit(g, actor, state, actor.x, actor.y);
+      drawUnit(g, actor, state, actorCoord.xCoord, actorCoord.yCoord);
     }
 
     return animTime > duration;
--- a/src/UI/Art/Animation/BaseUnitAnimation.java
+++ b/src/UI/Art/Animation/BaseUnitAnimation.java
@@ -6,6 +6,7 @@ import java.util.ArrayList;
 import Engine.XYCoord;
 import UI.Art.SpriteArtist.SpriteLibrary;
 import UI.Art.SpriteArtist.SpriteMapView;
+import UI.Art.SpriteArtist.UnitSpriteSet;
 import UI.Art.SpriteArtist.UnitSpriteSet.AnimState;
 import Units.Unit;
 
@@ -17,20 +18,20 @@ public class BaseUnitAnimation implements GameAnimation
   protected int tileSize;
 
   protected Unit actor;
+  private UnitSpriteSet actorSpriteSet;
   protected XYCoord actorCoord;
 
   public BaseUnitAnimation(int tileSize, Unit actor, XYCoord actorCoord)
   {
-    startTime = System.currentTimeMillis();
-    this.tileSize = tileSize;
-    this.actor = actor;
-    this.actorCoord = actorCoord;
+    update(tileSize, actor, actorCoord);
   }
   public BaseUnitAnimation update(int tileSize, Unit actor, XYCoord actorCoord)
   {
     startTime = System.currentTimeMillis();
     this.tileSize = tileSize;
     this.actor = actor;
+    if( null != actor )
+      this.actorSpriteSet = SpriteLibrary.getMapUnitSpriteSet(actor);
     this.actorCoord = actorCoord;
     return this;
   }
@@ -64,7 +65,8 @@ public class BaseUnitAnimation implements GameAnimation
    */
   public void drawUnit(Graphics g, Unit actor, AnimState state, int spriteIndex, double realX, double realY)
   {
-    SpriteLibrary.getMapUnitSpriteSet(actor).drawUnit(g, actor, state, spriteIndex, (int) realX * tileSize, (int) realY * tileSize );
+    if( null != actor && null != actorSpriteSet )
+      actorSpriteSet.drawUnit(g, actor, state, spriteIndex, (int) (realX * tileSize), (int) (realY * tileSize));
   }
 
   @Override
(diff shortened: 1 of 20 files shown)


src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -231,41 +196,42 @@ public void render(Graphics g)
     if( drawY < 0 )
       drawY = 0;
 
+    // Get a reference to the current action being built, if one exists.
+    Unit currentActor = mapController.getContemplatedActor();
+    XYCoord actorCoord = mapController.getContemplationCoord();
+    boolean notifyOnAnimEnd = true;
+    if( null != currentActor && null == currentAnimation ) // Draw the currently-acting unit so it's on top of everything.
+    {
+      currentAnimation = new BaseUnitAnimation(drawMultiplier, currentActor, actorCoord);
```
ThatIsAUsername, line 205: This creates a new animation and destroys it every frame - I feel like there's a better way to handle this.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -275,13 +241,14 @@ public void render(Graphics g)
...
-        mapController.animationEnded(eventsToAnimate.poll(), eventsToAnimate.isEmpty());
+        if( notifyOnAnimEnd ) // ...but only if we're animating events
+          mapController.animationEnded(eventsToAnimate.poll(), eventsToAnimate.isEmpty());
 
         // Get the next event animation if one exists.
         loadNextEventAnimation();
       }
     }
-    else if( getCurrentGameMenu() == null )
+    if( getCurrentGameMenu() == null )
```
ThatIsAUsername, line 251: Hm. Won't this draw the cursor even during animations? I kinda like having the cursor disappear whenever animations happen. Just another cue that the user can't use the cursor. Guess it probably wouldn't be the end of the world.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Yeah, probably.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Yeah.

ThatIsAUsername:
It looks like none of the animations inherit from `BaseUnitAnimation` (which you noted) and `BaseUnitAnimation` is only used to draw the selected unit running in place. Given that the base class doesn't seem to be needed for the other animatinos, and is a horribly inefficient way of doing the one thing it does -- do we actually need it?

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: It appears to be destroying performance. I can barely run the game on my laptop as soon as I create a unit.

ThislsAUsername: Wait, seriously?
That doesn't make any sense.
First, it should only activate once you select a unit. Second, I removed the `new` anyhow.

src/UI/Art/SpriteArtist/Sprite.java
```diff
@@ -111,9 +111,7 @@ public void addFrame(BufferedImage sprite)
 
   public BufferedImage getFrame(int index)
   {
-    // Normalize the index if needed.
-    for(; index >= spriteImages.size() || index < 0; index += spriteImages.size() * ((index < 0)? 1:-1));
-    return spriteImages.get(index);
+    return spriteImages.get(Math.abs(index) % spriteImages.size());
```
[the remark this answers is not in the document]
ThislsAUsername: Good point. I should fix that.
This loop caused horrible lag on my machine because the `index` coming in was very large (due to the change in how we calculate animation indices)

(This would have happened anyway previously, it would have just taken a while to get the index that big)

ThislsAUsername:
> Given that the base class doesn't seem to be needed for the other animatinos, and is a horribly inefficient way of doing the one thing it does -- do we actually need it?

Is it really that inefficient? I found an inefficiency (the one mentioned above), but I don't see how new'ing something or calling a function is *that* much of a performance hit.

As for the inheritance, yeah, I haven't gotten around to that yet :P

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: OK, this probably wasn't it, given that it's still an issue with the new changes you made. Something changed in `cyanide` though that is creating huge lag once a unit exists.

ThatIsAUsername: Looks like it was that loop you indicated. Which prompts me to ask - why did you change how the animation index is calculated?

src/UI/Art/Animation/BaseUnitAnimation.java
```diff
@@ -0,0 +1,85 @@
...
+  protected long startTime = 0;
+  protected long duration = -42;
+
+  protected int tileSize;
+
+  protected Unit actor;
+  private UnitSpriteSet actorSpriteSet;
+  protected XYCoord actorCoord;
+
+  public BaseUnitAnimation(int tileSize, Unit actor, XYCoord actorCoord)
```
ThatIsAUsername, line 24: This class isn't used for unit idling. Maybe it should be `BaseUnitActionAnimation` for clarity?

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: I changed it so it could be a static method, so I could use it without daisy-chaining the state info it's based on around. (and also because there's no particularly great reason IMO for holding effectively-meaningless state info)

I'd have thrown the mod logic in the anim index getter function, but that could cause animation timings to be off because we have no enforced animation length.

ThislsAUsername:
On the performance note: it *does* appear this branch does burn marginally more CPU mojo while animating a contemplated unit than `MoreKool` does, but I've spent like 2 hours trying to get a profiler set up that will give me method-level info and having a miserable time of it so I might quit lol.

ThatIsAUsername:
Is this still drafty, or are the windows shuttered yet?

ThislsAUsername:
I think the only real asterisk is showing the cursor during animations. I think I'll give fixing that a shot real quick.

PR #257, opened 2020-06-20
Tech

ThatIsAUsername:
I still need to actually add the new unit type and touch up the drop animation, but it's otherwise close enough for commenting now.
I am considering one more tweak for the drop-zone selection algorithm - for equally-valued spaces, choose whichever is _farthest_ from your own HQ, to encourage landing behind the enemy vanguard when possible. That may not always be desirable (and frankly would probably not matter all that often).


src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,452 @@
...
+      super("Tech");
+      infoPages.add(new InfoPage(
+          "Tech is a first-rate grease monkey who like nothing better than to build new machines and set them loose.\n"));
+      infoPages.add(new InfoPage(
+          "Passive:\n" +
+          "- Tech has some of the best mechanics around, allowing her units to repair 3 HP per turn instead of just 2.\n"));
+      infoPages.add(new InfoPage(
+          TECHDROP_NAME + " (" + TECHDROP_COST + "):\n" +
+              "Deploys " + TECHDROP_NUM + " BattleMech to the front lines.\n" +
+              "All mechanical units gain " + TECHDROP_BUFF + "% attack.\n" +
```
ThislsAUsername, line 44: You're buffing both attack and defense, though.

ThatIsAUsername: Read the next line?

src/Engine/GameEvents/TeleportEvent.java
```diff
@@ -0,0 +1,154 @@
...
+    {
+      unitDies = true;
+      new UnitDieEvent(unit).performEvent(gameMap);
+
+      // Our unit died; check if we are defeated.
+      if( unit.CO.units.size() == 0 )
+      {
+        // CO is out of units. Too bad.
+        new CommanderDefeatEvent(unit.CO).performEvent(gameMap);
+      }
```
ThislsAUsername, line 134: I've been thinking about our scheme of generating all possible events immediately, and I feel like maybe win/loss conditions in particular should just be checked whenever the event queue becomes free.
What say you?

src/CommandingOfficers/Tech.java
ThislsAUsername: ...good job, me.

src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,452 @@
...
+      // Bump up our power level.
+      myCommander.addCOModifier(damageBuff);
+      myCommander.addCOModifier(defenseBuff);
+
+      // Perform any events we generated in getEvents().
+      while(!abilityEvents.isEmpty())
+      {
+        abilityEvents.poll().performEvent(gameMap);
+      }
+    }
```
ThislsAUsername, line 210: I still think it'd be better if `perform()` just returned a `GameEventQueue`. It makes sense from an animation standpoint and a replays/netplay perspective.

src/Engine/GameEvents/TeleportEvent.java
ThatIsAUsername: That could be difficult with certain events. Like, _losing_ an HQ is a lose condition, but _not having_ an HQ is not a lose condition. Not sure how to divorce evaluating lose conditions from the events that caused them.

ThislsAUsername: Hmm. I suppose what we could do is have `GameInstance` listen for and record stuff like "CO x lost a unit", and clear that state on having the loss-condition-check function called?
This also puts that logic right next to our `GameRules` in case we ever want to support fiddling with that.

ThatIsAUsername: Putting that logic in `GameRules` is probably the long-term solution, since that will allow us to support alternate win conditions for scenario missions, but I don't think we are there yet.

src/CommandingOfficers/Tech.java
ThatIsAUsername: And that event queue would have to be passed all the way back up to `MapController.animationEnded`? I guess there's no reason we couldn't do that.
Wouldn't `MapController` just have to call `performEvent` on each event in the returned queue anyway, before publishing it to the animator?
I could see it being useful to allow events to create/return sub-events (e.g. `TeleportEvent` can result in a `UnitDieEvent` and possible a `CommanderDefeatEvent`), but I'm sure there would be other ways to handle that case.

ThislsAUsername: > And that event queue would have to be passed all the way back up to `MapController.animationEnded`? I guess there's no reason we couldn't do that.

No? We could just have the ability action call `perform()` directly and return the events itself.
We could support both the existing method and the new scheme by making the "official" interface be `getEvents()`, and have that default to producing our current ability event that calls `perform()`.

> Wouldn't `MapController` just have to call `performEvent` on each event in the returned queue anyway, before publishing it to the animator?

Yes? Unless I'm missing something (which I probably am, seeing as I'm not even close to finished reviewing), you're just executing the events and not feeding them to `MapController` to be animated at all.

> I could see it being useful to allow events to create/return sub-events (e.g. `TeleportEvent` can result in a `UnitDieEvent` and possible a `CommanderDefeatEvent`), but I'm sure there would be other ways to handle that case.

We already have a pattern for handling that: the `GameAction` figures out if any of those are needed, and puts them all in the queue in sequence.
There's probably an argument to be made that e.g. a `TeleportEvent` might want to integrate the obstacle's death into the teleport animation itself, so maybe we have good reasons for supporting both patterns?

ThatIsAUsername: > No? We could just have the ability action call perform() directly and return the events itself.

Yeah, we could make abilities return all/most effects as events and call `getEvents()` in `AbilityAction.getEvents()` instead of just returning a single `AbilityEvent`. I've no problem with that; it would probably be better than my current approach of fetching/handling those extra events underneath `AbilityEvent`.
We can't do this by changing `perform()` though, since it is only called after the animation is over. We could move all functionality into events though, and it would effectively go away.

> Wouldn't MapController just have to call performEvent on each event in the returned queue anyway

I got the order backward here. Events are always animated before being executed and published.

> We already have a pattern for handling that: the GameAction figures out

That allows Actions to generate sub-events, not events to generate sub-events. Right now for example, there is no `TeleportAction`, but `TeleportEvent` is generating the `UnitDieEvent` and `CommanderDefeatEvent` if needed. It keeps track of the sub-events so it can send them to the listeners when it sends itself.
Hm. Maybe I should just create a TeleportAction and have `TechDrop` use that.

> both patterns

To be explicit, they are: requiring all events via `getEvents()` vs. allowing events to generate sub-events at execution time?

ThislsAUsername: > Yeah, we could make abilities return all/most effects as events and call `getEvents()` in `AbilityAction.getEvents()` instead of just returning a single `AbilityEvent`.

Right, and we could combine this with making the *default* implementation of Ability's `getEvents()` just return an `AbilityEvent`.
This could result in kind of a weird system where some abilities just never call `perform()` while others use it exclusively, but I feel like I might be okay with that.

> That allows Actions to generate sub-events, not events to generate sub-events. Right now for example, there is no `TeleportAction`, but `TeleportEvent` is generating the `UnitDieEvent` and `CommanderDefeatEvent` if needed. It keeps track of the sub-events so it can send them to the listeners when it sends itself.

Right, and to me that feels like it might be unnecessarily complicated.

> Hm. Maybe I should just create a TeleportAction and have `TechDrop` use that.

Sounds good to me. ~~and then we can make a `TeleportActionFactory` and suddenly we have Swap Mech~~

> > both patterns
> 
> To be explicit, they are: requiring all events via `getEvents()` vs. allowing events to generate sub-events at execution time?

Yessir. The argument I can see for the second one is to allow a single animation to integrate all the effects instead of e.g. "mech falls onto screen... \*beat\*... infantry explodes"
There may be a better solution for that too, though.
(The first idea that pops to mind: have a flag on events to make them not animate, so you can have [`TeleportEvent`, `UnitDieEvent`, CODefeat] in your queue, and the death doesn't need to get animated because it's animated as part of `TeleportEvent`)

ThatIsAUsername: > Right, and to me that feels like it might be unnecessarily complicated.

I can't disagree. This was the first way I got it working, but I'm sure that whole sequence can be streamlined. Should I try to clean that up in this PR, or make a new one?



I've been toying with the idea of allowing animations to overlap - define a "done but still animating" point at which the next animation in the queue can start. This would make e.g. abilities that buff every unit go faster (though they could potentially be combined into one large animation that handles all that).
Completely up for debate though; the "do/don't animate" flag could also work, though I feel like its usage would be a bit niche. I think I may try decomposing `TeleportEvent` further and see how things look then.

ThatIsAUsername:
Argh, making a new unit model is a pain.
Also, Wally crashes because `threatMap` doesn't have an entry for BattleMechs for some reason.
Anyway, let me know when it's safe to push up more changes without ruining your groove.

src/CommandingOfficers/Tech.java
ThislsAUsername: > I've been toying with the idea of allowing animations to overlap - define a "done but still animating" point at which the next animation in the queue can start. This would make e.g. abilities that buff every unit go faster (though they could potentially be combined into one large animation that handles all that).

I do kind of like this idea, since it does have slightly broader application, I think.
That said, we'd have to have a *list* of active animations, track when all relevant ones are done, and do something about the possibility of multiple animations drawing the same unit (even if it's "don't let it happen").

> Completely up for debate though; the "do/don't animate" flag could also work, though I feel like its usage would be a bit niche. I think I may try decomposing `TeleportEvent` further and see how things look then.

It's for things that have knock-on effects like unit death. One that comes to mind would be Witchfire - have a buff animation (with a special version for unit death) and don't separately animate the death.

ThislsAUsername:
> Argh, making a new unit model is a pain.

You can just... clone the one you want the in-bound damage chart from, and cherry-pick the weapons you need and tweak as needed?

> Also, Wally crashes because `threatMap` doesn't have an entry for BattleMechs for some reason.

It should be in there, if you added `BattleMech` to the CO's list of `UnitModel`?

> Anyway, let me know when it's safe to push up more changes without ruining your groove.

I think I'm fine with it, since I'm just going through the commits individually anyway.

ThatIsAUsername:
> You can just... clone ... weapons

Yes. It was a pain. Maybe I should have made multiple versions, one in each unit scheme file? You tell me, since this is your ball of wax.

> if you added BattleMech to the CO's list of UnitModel

I hadn't. Now I did. Now he doesn't crash.

> I think I'm fine with it

Puuuuushed. New cool unit and drop animation ahoy.

src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,514 @@
...
+    BattleMech.healableHabs = new HashSet<TerrainType>(); // BattleMechs have specialized parts, not easy to repair.
+
+    UnitModel antiAir = UnitModelScheme.getModelFromString("Anti-Air", grModels.unitModels);
+    WeaponModel ratatat = antiAir.weapons.get(0).clone();
+    ArrayList<WeaponModel> weapons = new ArrayList<WeaponModel>();
+    weapons.add(BattleMech.weapons.get(0));
+    weapons.add(ratatat);
+    BattleMech.weapons = weapons;
+    return BattleMech;
+  }
```
ThislsAUsername, line 513: A few notes:
I would rather clone the CO's own models instead of pulling a whole extra `GameReadyModels` out.
Name matching is a poor choice, IMO. I would do a `getAllModels()` for `ASSAULT`, then pull out index 1 (or fall back to 0 if impossible). This fits with the documentation in `UnitModel`.
Likewise for the AA, I'd grab the first `SURFACE_TO_AIR` model.

It's pretty much otherwise how I'd do it.

As for improving the setup... I guess we could provide some sort of factory method on `UnitModelScheme`... but we'd run into the same sorts of difficulties setting up the "armor" and weapons.
This is definitely a drawback of abstracting the unit set out :P

Also, I see you have the `moneyCost` setter commented. It'd make sense to let you set it here, but that kind of messes with my idea of keeping the default cost "sacred" so as to avoid order-of-operations issues. You could accomplish this by setting the `moneyCostAdjustment` to the difference between your desired cost and the current cost, or we could go ahead and direct public access to `moneyCost`, I guess.

src/Engine/GameEvents/CommanderAbilityEvent.java
```diff
@@ -18,13 +19,23 @@ public CommanderAbilityEvent(CommanderAbility ability)
...
-    return null;
+    // TODO: CO Ability intro splash
+    GameAnimation anim = null;
+    if( !gameEvents.isEmpty() )
+    {
+      // Just grab the first one for now; TODO create compound animations.
+      anim = gameEvents.peek().getEventAnimation(mapView);
+    }
+    return anim;
   }
```
ThislsAUsername, line 30: Super random idea for setting up ability buff animations:
The animation callsite takes in a `UnitModel` with zero'd stats + the "delta", and maybe a "is this a permanent buff" flag. (TBD whether it's a set of units or a single unit per animation)
That would still leave out stuff like healing or refresh, but it would make for a "reasonable" interface that automatically stays current with all available knobs to turn and allows for multiple changes at once.

src/Engine/GameEvents/TeleportEvent.java
```diff
@@ -76,39 +81,39 @@ public void performEvent(MapMaster gameMap)
...
-    if( null != oldResident )
+    obstacle = gameMap.getLocation(unitDestination).getResident();
+    obstacleDies = false;
+    if( null != obstacle )
     {
       switch(collisionOutcome)
       {
         case KILL:
-          killOldResident = true;
+          obstacleDies = true;
```
ThislsAUsername, line 95: Just a note: this should end up killing a unit that teleports onto its current tile. IDK if we care about that case, but it exists.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -15,12 +15,12 @@
...
-  public final int ACTION_MOVEEAST = 2;
-  public final int ACTION_MOVESOUTH = 3;
-  public final int ACTION_MOVEWEST = 4;
-  public final int ACTION_DIE = 5;
+  public static final int ACTION_IDLE = 0;
+  public static final int ACTION_MOVENORTH = 1;
+  public static final int ACTION_MOVEEAST = 2;
+  public static final int ACTION_MOVESOUTH = 3;
+  public static final int ACTION_MOVEWEST = 4;
+  public static final int ACTION_DIE = 5;
```
ThislsAUsername, line 23: ...This'll be a little ugly to merge. Oops.
D'you want to muddle through that, or should I?

src/Engine/GameEvents/TeleportEvent.java
```diff
@@ -0,0 +1,165 @@
...
+  public enum CollisionOutcome
+  {
+    KILL,
+    SWAP
+  }
+
+  public TeleportEvent(GameMap map, Unit u, XYCoord dest, AnimationStyle animStyle, CollisionOutcome crashResult)
+  {
+    unit = u;
+    unitStart = new XYCoord(unit.x, unit.y);
```
ThislsAUsername, line 45: I might default to the CO's HQ location in the case that the unit isn't on the map, for more consistent facing.

src/AI/AIUtils.java
```diff
@@ -458,4 +458,28 @@ public int compare(Unit o1, Unit o2)
       return diff;
     }
   }
+
+  /** Return the set of locations with enemies or terrain that `unit` could attack in one turn from `start` */
+  public static Set<XYCoord> findPossibleTargets(GameMap gameMap, Unit unit, XYCoord start)
+  {
+    Set<XYCoord> targetLocs = new HashSet<XYCoord>();
+    boolean allowEndingOnUnits = false; // We can't attack from on top of another unit.
```
ThislsAUsername, line 466: I might reverse this, since the units you ordinarily can end on are units that could get out of the way.
If you really wanted to go crazy with it, you could check if units are ready to act and thus get out of the way... though that would introduce a dependency on it being the turn of the CO owning the unit in question.

src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,514 @@
...
+        else
+        {
+          System.out.println("[TechdropAbility.getEvents] WARNING! Could not find drop location.");
+        }
+      }
+
+      return abilityEvents;
+    }
+
+    private TeleportEvent generateDropEvent(MapMaster gameMap, Set<XYCoord> priorDrops, boolean log)
```
ThislsAUsername, line 254: I think this logic could be simplified a lot using the `MassStrikeUtils` thing I set up in the test branch.
It probably would be a bit less efficient, but I think the gains in readability would be significant.
I can make a PR on this PR for that, if you'd like? (Probably wouldn't make it until I'm done reviewing, though.)

src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,514 @@
...
+    {
+      // Create a new Unit to drop onto the battlefield.
+      Unit techMech = new Unit(myCommander, unitModelToDrop);
+      techMech.isTurnOver = false; // Hit the ground ready to rumble.
+      if( myCommander.HQLocation.xCoord >= gameMap.mapWidth / 2 )
+        techMech.x = gameMap.mapWidth; // Make the unit face left as it falls to match the rest of the units.
```
ThislsAUsername, line 260: ...ah, I see you've accounted for my other comment here.

src/CommandingOfficers/Tech.java
```diff
@@ -107,11 +170,11 @@ public int getRepairPower()
 
     GameEventQueue abilityEvents;
 
-    TechdropAbility(Commander myCO, String abilityName, int abilityCost, int buff, int num, int abilityRange)
+    TechdropAbility(Commander myCO, String abilityName, double abilityCost, int buff, int num, int abilityRange)
     {
       super(myCO, abilityName, abilityCost);
 
-      // Create a COModifier that we can apply to Patch when needed.
```
ThislsAUsername: This patch patches Patch out from a patch of not-Patch

src/Units/Unit.java
```diff
@@ -253,8 +253,9 @@ private static int healthFromHP(double input)
    */
   public int damageHP(double damage)
   {
+    if( damage < 0 ) throw new ArithmeticException("Cannot inflict negative damage!");
```
ThislsAUsername, line 256: Hah. Guess I'm glad we threw out the idea of Brenner healing 1.5 HP...

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -162,12 +175,22 @@ public void drawUnitIcons(Graphics g, Commander[] COs, Unit u, int animIndex, in
...
+      {
+        num = SpriteLibrary.getMapUnitHPSprites().getFrame(1); // Tens place.
+
+        // Ones place shares space with the activity icons below if HP > 10.
+        unitIcons.add( SpriteLibrary.getMapUnitHPSprites().getFrame((u.getHP()-u.model.maxHP)) );
+      }
+      else
+        num = SpriteLibrary.getMapUnitHPSprites().getFrame(u.getHP());
+      g.drawImage(num, drawX, drawY + ((unitHeight) / 2), num.getWidth(), num.getHeight(), null);
     }
```
ThislsAUsername, line 194: I think I'll reserve final judgement until I see both versions of this... but I feel like this is kind of a jank way to show it.

src/CommandingOfficers/Tech.java
```diff
@@ -129,6 +138,12 @@ public OverchargeAbility(Commander commander, String abilityName, double ability
       damageBuff = new CODamageModifier(buff);
       defenseBuff = new CODefenseModifier(buff);
       healAmount = healAmt;
+
+      // Only mechanical units get the firepower boost.
+      long exclude = UnitModel.TROOP;
+      ArrayList<UnitModel> models = commander.getAllModels(Tech.mechanicalUnits, true, exclude);
+      for( UnitModel m : models )
+        damageBuff.addApplicableUnitModel(m);
```
ThislsAUsername, line 157: Poor Bikes. What did they ever do to you?

src/CommandingOfficers/Tech.java
```diff
@@ -143,6 +158,8 @@ protected void perform(MapMaster gameMap)
       // Overcharge
       for(Unit u: myCommander.units)
       {
+        // Only mechanical units get overcharged.
+        if( u.model.isNone(UnitModel.TROOP) )
         u.alterHP(healAmount, true);
```
ThislsAUsername, line 176: Technically different rules here :P

src/CommandingOfficers/Tech.java
```diff
@@ -0,0 +1,514 @@
...
+      defenseBuff = new CODefenseModifier(buff);
+      numDrops = num;
+      dropRange = abilityRange;
+      abilityEvents = new GameEventQueue();
+
+      // Only mechanical units get the firepower boost.
+      long exclude = UnitModel.TROOP;
+      ArrayList<UnitModel> models = commander.getAllModels(Tech.mechanicalUnits, true, exclude);
+      for( UnitModel m : models )
+        damageBuff.addApplicableUnitModel(m);
```
ThislsAUsername, line 211: Why not just have Tech do the filtering once in her constructor, and pass in a strongly-typed `Tech` instead of `Commander`?

src/CommandingOfficers/Tech.java
```diff
@@ -214,29 +212,41 @@ protected void perform(MapMaster gameMap)
...
         {
-          dropLocs.add(techDrop.getEndPoint());
+          // Store the new event.
           abilityEvents.add(techDrop);
+
+          // Place it on the map temporarily so we account for it in subsequent drops.
+          dropLocs.add(techDrop.getEndPoint());
+          Unit u = techDrop.getUnit();
+          XYCoord loc = techDrop.getEndPoint();
+          gameMap.moveUnit(u, loc.xCoord, loc.yCoord);
```
ThislsAUsername: This... this feels like a hack.

src/CommandingOfficers/Tech.java
```diff
@@ -349,8 +361,23 @@ else if( !invalidDropCoords.contains(nmexy) )
         }
       }
 
+      // We only want to drop in near enemies; remove any destinations with no "enemy score".
+      Set<XYCoord> friendCoords = new HashSet<XYCoord>(friendScores.keySet());
+      for( XYCoord fc : friendCoords )
+        if( !enemyScores.keySet().contains(fc) )
+          friendScores.remove(fc);
+
+      // If that leaves us with no drop zones, then just reinforce the HQ.
```
ThislsAUsername, line 402: I suppose it's fair to note that this would require a second pass with `MassStrikeUtils`, but oh well.

src/CommandingOfficers/Tech.java
```diff
@@ -219,38 +219,28 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
         if( null != techDrop )
         {
           // Store the new event.
-          abilityEvents.add(techDrop);
-
-          // Place it on the map temporarily so we account for it in subsequent drops.
           dropLocs.add(techDrop.getEndPoint());
-          Unit u = techDrop.getUnit();
-          XYCoord loc = techDrop.getEndPoint();
-          gameMap.moveUnit(u, loc.xCoord, loc.yCoord);
```
ThislsAUsername: It seems you have agreed with my assessment. Works for me.

ThislsAUsername:
The drop animation is pretty cool, on the scale of things.
Speaking of the animation, you currently only animate one drop out of three (though there's a comment to that general effect)
HP over 10 gives a multiplicative boost to attack, which makes her attacks *very* strong during Overcharge.

Having seen it myself now, I also prefer using the marker to display HP over 10. Sure, that'll break if we go up to 20... but I think we may have bigger problems in that case lol

src/CommandingOfficers/Tech.java
ThatIsAUsername: > I would rather clone the CO's own models 

Sensible

> I would do a getAllModels() for ASSAULT, then pull out index 1

That depends on ordering to work right. If we added a unit type in between this would silently return the wrong thing. Using the name look up seems better precisely because it is more fragile. When it breaks, we'll know.

> moneyCost

Yeah, I thought about setting `moneyCostAdjustment`, but then if anyone else got BattleMechs, they'd be cheaper than Tech's and that feels wrong. That said, they aren't buildable or repairable, so it probably doesn't matter for now.

src/Engine/GameEvents/TeleportEvent.java
ThatIsAUsername: Line 85 should prevent that.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: I can merge.

src/Engine/GameEvents/TeleportEvent.java
ThatIsAUsername: That would make the facing wrong whenever it teleports to the wrong side of the HQ.
Maybe the "right" approach is to find the unit's Commander->HQ, figure out which side of the map it's on, and set the start (for purposes of facing, not animation) to be off the map in that direction.

src/AI/AIUtils.java
ThatIsAUsername: Nah, this is a helper utility, not an AI Submodule. It specifically says it's finding what the unit can do _right now_. We could push `allowEndingOnUnits` into the parameter list to provide an affordance for contemplating multi-move strategies, but since this returns only the targets and not the move locations, that seems dubiously useful. I'd prefer to keep it simple and focused.

src/CommandingOfficers/Tech.java
ThatIsAUsername: I'm hesitant to agree to that given that a) I'm unfamiliar with how `MassStrikeUtils` works, and b) this already does.
It would be worth bringing that over for the future though, I'm sure. Had it been part of `master` already, I would have at least tried to use it, if it is flexible enough.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Only works if you drop on the correct side of the HQ. More responses on the other comment.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Indeed.

src/Units/Unit.java
ThatIsAUsername: Hm? `AlterHP` can still be used to add HP. Or was he going to do that by attacking? That would be weird.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: By "both" versions, do you mean:
Screenshot from 2020-06-16 19-37-19

Screenshot from 2020-06-16 19-36-03

src/CommandingOfficers/Tech.java
ThatIsAUsername: We don't have a `Mechanical` flag, or maybe they'd qualify (though for balance they probably shouldn't, not that "balance" is really part of the picture here). Sorry pals.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Different than what?

src/CommandingOfficers/Tech.java
ThatIsAUsername: Guess I could. Make it static even, so no need to cast. It didn't occur to me at the time, and it's not an expensive or frequent operation anyway.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: Yes. I mentioned my conclusion in the summary comment.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Total hack, and it was breaking things. Fixed.

src/Units/Unit.java
ThislsAUsername: Right. It didn't really fit that well, which is why it was scrapped.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Hah. Yes.

src/CommandingOfficers/Tech.java
ThislsAUsername: b) it does... what?
Mostly, it would focus the valuation function on scoring a *single* space, instead of being integrated with a bunch of loops evaluating many spaces.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: Per your comment below, you like the second version better than the first? Once I figured how how to make the base-10 display work, I switched over since it looks far more intuitive.

src/Engine/GameEvents/TeleportEvent.java
ThislsAUsername: That does appear to be what you've done in Tech's case.
I don't mind that approach, I suppose.
Since the HQ facing calculation is `static` on `master`, I would like to see it reuse that function though :eyes:

src/CommandingOfficers/Tech.java
ThatIsAUsername: b) ... works.
Next time I build a globally-targetable ability, maybe I'll try to use the utils, though this PR may be evidence that I like to tweak/fine-tune/customize things beyond what is often easily generalizeable.

src/CommandingOfficers/Tech.java
ThislsAUsername: > > I would do a getAllModels() for ASSAULT, then pull out index 1
> 
> That depends on ordering to work right. If we added a unit type in between this would silently return the wrong thing. Using the name look up seems better precisely because it is more fragile. When it breaks, we'll know.

The ordering being "sensible" for this sort of purpose is part of the contract of `UnitModelScheme`, as documented in the comments in `UnitModel`. If that needs to be more clear, we can enhance that documentation.
If I bring in a Wargroove unit set, there *won't* be a "Md Tank", but a Golem would serve just fine (and would be found via the process I outlined).

> Yeah, I thought about setting `moneyCostAdjustment`, but then if anyone else got BattleMechs, they'd be cheaper than Tech's and that feels wrong.

But... nobody else *can* get BattleMechs. They only exist for Tech, because she manufactures the model.
I guess Cinder could get a discount if she managed to buy one, but she'd have to steal the model first... and that's not a thing.
Anybody who doesn't tweak the cost adjustment would end up with the same price, since it'd come out of this very function with the correct adjustment in place.

> That said, they aren't ... repairable, so it probably doesn't matter for now.

Black Boats would like a word with you...
Also, cost matters for COP charge purposes.

src/CommandingOfficers/Tech.java
ThislsAUsername: I'd like to take a stab at porting it.
Do you mind?

src/CommandingOfficers/Tech.java
ThislsAUsername: Your stat boosts also require the unit to be one of `TANK`, `SHIP`, `HOVER`, or `JET` in addition to not being `TROOP`.

src/CommandingOfficers/Tech.java
ThislsAUsername: Can't be static, because each `Tech` has different models.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: Yessir, I do like having it as a CO mark.
Having HP over 10 is non-standard behavior, and should be treated as a CO buff (because it is one).
Likewise, there *shouldn't* (pending possible scope creep, I guess?) be any unit sets that modify the max HP to be above 10.

I would also contest your claim about it being intuitive: IMO having any number in the bottom left of the unit indicates that the unit is injured. A 1 indicates its almost dead. Changing that expectation is unintuitive.

ThatIsAUsername:
> HP over 10 gives a multiplicative boost to attack, which makes her attacks very strong during Overcharge.

Yeah, Tech seems pretty busted. I'm tempted to remove the Overcharge buff, but then units with sub-10 HP aren't anything special (though healing is still nice). Maybe just drop it to the default +10/10.
I'm also tempted to make TechDrop cost like 8-10 and Steel Rain more like 15-20, though 10+ abilities aren't historically a thing, and it would be odd to have no abilities cost less than 6.

src/CommandingOfficers/Tech.java
ThatIsAUsername: > Black Boats

True, forgot about those. I guess the adjustment field is as good as anything for now.

> Also, cost matters for COP charge purposes.

CO charge is separately determined by `abilityPowerValue`, isn't it?

ThislsAUsername:
> Yeah, Tech seems pretty busted. I'm tempted to remove the Overcharge buff, but then units with sub-10 HP aren't anything special (though healing is still nice). Maybe just drop it to the default +10/10.

You could have her cap her own HP to 10 during combat. Then the "extra HP" becomes basically armor that allows you to eat damage without losing HP at all (as opposed to defense, which makes you lose *less* HP)

> I'm also tempted to make TechDrop cost like 8-10 and Steel Rain more like 15-20, though 10+ abilities aren't historically a thing, and it would be odd to have no abilities cost less than 6.

I don't have strong feelings on any of this.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: > IMO having any number in the bottom left of the unit indicates that the unit is injured. A 1 indicates its almost dead. Changing that expectation is unintuitive.

... I think that the presence of a second digit would make things extremely clear to anyone who saw it. The unit markngs on the other are a whole new affordance we've added, in a spot that's only really used for bookkeeping unit metadata. Putting an HP value there in a different color will just create questions when people first see it.
I mean, folks will figure out what it all means either way after seeing it in action, but base-10 numbers seem more immediately clear.
I'd be open to more opinions though; nothing like actual end-user input to shape expectations.

src/CommandingOfficers/Tech.java
ThatIsAUsername: True I suppose. Is there any units who are classified differently in the two places? I guess I should make it consistent anyway, for form's sake.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Right.

ThatIsAUsername:
> You could have her cap her own HP to 10 during combat

Aww, but I like that more HP == more firepower. It's consistent with how AW works in general.

src/CommandingOfficers/Tech.java
ThislsAUsername: > CO charge is separately determined by `abilityPowerValue`, isn't it?

Planned feature. Not current feature.
Currently on `master`, it's AW2 charge (cost/9k for loss, half that for attack) + 0.1 star per HP you eliminate.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: > ... I think that the presence of a second digit would make things extremely clear to anyone who saw it.

I guess? It's justified wrong, and can take the same space as other icons, though.

> The unit markngs on the other are a whole new affordance we've added, in a spot that's only really used for bookkeeping unit metadata. Putting an HP value there in a different color will just create questions when people first see it.

As it should, IMO. HP > 10 is *allowed*, but under no circumstances should it be considered *standard*.

> I mean, folks will figure out what it all means either way after seeing it in action, but base-10 numbers seem more immediately clear.
> I'd be open to more opinions though; nothing like actual end-user input to shape expectations.

How about you post in the test lab and/or AWBW's customization chat and see what peeps think? :stuck_out_tongue:

src/CommandingOfficers/Tech.java
ThislsAUsername: I don't think so, but it's not out of the question.
IMO the real test *should* just be "not `TROOP`", since `TROOP` is how we indicate the "form factor" of the unit is humanish.
...which makes it weird because I kind of assumed that I wouldn't classify Wargroove Knights as `TROOP`. I guess it's relative, lol.

ThislsAUsername:
> Aww, but I like that more HP == more firepower. It's consistent with how AW works in general.

I think you can justify either conclusion, because AW doesn't ever see HP over 10 so it's not accounted for in their design.
Just the base +10/10 should work, I imagine.

src/CommandingOfficers/Tech.java
ThatIsAUsername: I guess we can figure that out if/when we add Wargroove units.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: > It's justified wrong

It's fixed-width.

> can take the same space as other icons

They rotate. That's an issue with putting it in the mark spot too, anyway.

> As it should, IMO ... under no circumstances should it be considered standard

Disagree ... Agree. Just because something is new doesn't mean it can't also make immediate sense.

> How about you post in the test lab and/or AWBW's customization chat

Could. They may just be annoyed that this isn't one of the "real" features though. Maybe I could placate them by promising to do productive work. Might be worth doing; I'll let your older brother weigh in on it in the meantime.

ThislsAUsername: > It's fixed-width.

*shrug*
> > can take the same space as other icons
> 
> They rotate. That's an issue with putting it in the mark spot too, anyway.

Not nearly as much as an issue, though, IMO.
For base-10 HP, you can see an APC or capping inf with "1" HP.
With the marker, you see an *unhurt" APC or capping inf, and you can wait around for other markers to appear if needed.

> Disagree ... Agree. Just because something is new doesn't mean it can't also make immediate sense.

Right, and I'd argue that "making immediate sense" is all about not violating the expectation of "HP number means fighting at less than 100% capacity"

> Could. They may just be annoyed that this isn't one of the "real" features though.

Maybe. I would be inclined to think they're more reasonable than that.

> Maybe I could placate them by promising to do productive work.

Though I can't imagine this would hurt too much, lol.

> I'll let your older brother weigh in on it in the meantime.

This implies he's meaningfully apprised of the situation?
~~and he's not your older brother?~~

ThatIsAUsername: > Maybe. I would be inclined to think they're more reasonable than that

Yeah, I'm sure they would be.

> This implies he's meaningfully apprised of the situation?

Nope. Which is the point. I sent an email without explaining what either means. We'll see which one lets him figure it out.

> For base-10 HP, you can see an APC or capping inf with "1" HP.

On the one hand yes, this is possible (for about half a second or whatever). On the other hand, it's likely that other units will also be buffed (since this only happens via abilities, which are likely to impact many units), and the odds that an inf or APC is the only affected unit in sight is... vanishingly small. The odds that on top of that, the rotating icon will cause a user to make the wrong choice is even more vanishingly smallerer.
Even with the rotating icon (which affects both anyway), the time-to-understanding is almost certainly less with the base-10 numbers (not to mention that we could make some pretty absurd enemy boss units with this, or e.g. Omega Tech with +25 HP that doesn't go away).

src/CommandingOfficers/Tech.java
```diff
@@ -150,10 +143,12 @@ public OverchargeAbility(Commander commander, String abilityName, double ability
       defenseBuff = new CODefenseModifier(buff);
       healAmount = healAmt;
 
-      // Only mechanical units get the firepower boost.
+      // Only mechanical/non-troop units get the firepower boost.
+      long allUnits = ~0;
       long exclude = UnitModel.TROOP;
-      ArrayList<UnitModel> models = commander.getAllModels(Tech.mechanicalUnits, true, exclude);
-      for( UnitModel m : models )
+      unitsToOverCharge = commander.getAllModels(allUnits, true, exclude);
```
ThislsAUsername, line 149: Maybe introduce a new `getAllModelsNot(long exclude)`, since that would read more clearly IMO.

src/CommandingOfficers/Tech.java
```diff
@@ -267,7 +260,7 @@ private TeleportEvent generateDropEvent(MapMaster gameMap, Set<XYCoord> priorDro
       {
         // Record any locations near owned properties; we want to be able to rescue unprotected structures.
         for( XYCoord xyc : Utils.findLocationsInRange(gameMap, propCoord, 0, dropRange) )
-          if( !unitModelToDrop.propulsion.canTraverse(gameMap.getEnvironment(xyc)) )
+          if( unitModelToDrop.propulsion.canTraverse(gameMap.getEnvironment(xyc)) )
```
ThislsAUsername, line 263: Hah. I thought that one was odd, but I can't say I understand exactly what all the bits are doing so I didn't wanna say anything lol

src/Engine/GameAction.java
```diff
@@ -131,11 +131,10 @@ public AbilityAction(CommanderAbility ability)
     @Override
     public GameEventQueue getEvents(MapMaster map)
     {
-      myEvent.generateAbilityEvents(map); // Let the ability figure out what's gonna go down.
-
-      // We still just return this event as the gateway.
+      // Create an event for the ability itself, and then append any additional events.
       GameEventQueue abilityEvents = new GameEventQueue();
       abilityEvents.add(myEvent);
+      abilityEvents.addAll(myEvent.generateAbilityEvents(map));
```
ThislsAUsername, line 137: I think this is a good hybrid approach for now. We can add `BuffEvent`s or whatever later.

src/Engine/GameEvents/CommanderAbilityEvent.java
```diff
@@ -45,9 +40,10 @@ public void performEvent(MapMaster gameMap)
...
 
   /** Called by AbilityAction before `getEventAnimation`, `performEvent`, or `sendToListener`. */
-  public void generateAbilityEvents(MapMaster gameMap)
+  public GameEventQueue generateAbilityEvents(MapMaster gameMap)
   {
     gameEvents = myAbility.getEvents(gameMap);
+    return gameEvents;
   }
```
ThislsAUsername, line 47: Why not just throw out this function in favor of calling `myAbility.getEvents()` directly in `AbilityAction`?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: > Nope. Which is the point. I sent an email without explaining what either means. We'll see which one lets him figure it out.

A cunning strategy.

> On the one hand yes, this is possible (for about half a second or whatever). On the other hand, it's likely that other units will also be buffed (since this only happens via abilities, which are likely to impact many units), and the odds that an inf or APC is the only affected unit in sight is... vanishingly small.

Sure, but now you're requiring looking around at the other units... and even then, it's only 100% clear with more context about how the CO actually functions.
Say I send someone a screenshot and ask for an opinion. Now the person I'm asking either thinks the inf is 1 HP or thinks it's not capturing (maybe it shot one of my dudes last turn)

OTOH, the vast majority of the time, only one marker will be visible on a given unit at a time (since a CO with multiple markers is likely too complex and COs who apply them to others are fairly rare).
The *most* relevant information in most cases is that the unit is no less than 10 HP, and that's indicated unambiguously by leaving that corner of the unit blank.
You could run into, say, a stunned/revenge-marked unit at 11+ HP, and end up missing the more important piece of info in a screenshot, but that's only possible in a specific CO matchup vs happening in *any* match with Tech.

> this only happens via abilities

I would not make this assumption. I would design around this being a possible D2D (as has been proposed a few times)

> The odds that on top of that, the rotating icon will cause a user to make the wrong choice is even more vanishingly smallerer.

Fair, but we should do what we can to make the game state easily readable.

> Even with the rotating icon (which affects both anyway), the time-to-understanding is almost certainly less with the base-10 numbers

Fair, but IMO it's less readable for the player who knows what's going on (as detailed above)

> (not to mention that we could make some pretty absurd enemy boss units with this, or e.g. Omega Tech with +25 HP that doesn't go away).

IDK, I feel like that's a bridge too far from the basic AW contract.

> ~He totally is. I just wanted to distinguish him from your younger brother.~

~~Okay. Glad we got that one cleared up~~

ThislsAUsername:
Issue roundup (I'm tired of scrolling):
Still seems worthwhile to try splitting up `TeleportEvent` to match how other events work, but I guess it's not super critical because the logic determining the outcome is both pretty simple and deterministic.
I move to defer the whole animation shebang we talked about 'cause we don't have death animations yet anyway.
Dropping a mech on someone should give defensive charge. (We could maybe do this via death listener, since other paths to death generally zero out the HP of the unit in question)
`BattleMech` building: search by `role` (and maybe other criteria), clone own model, `moneyCost`.
Might also be worth tweaking the `BattleMech`'s base attack/defense stats to make it work as intended (based on the difference between "optimal stats" and whatever base unit/weapons you end up with [if you really wanted to go wild with it, you could filter by weapon in a combat modifier]).
HP display (pending external review)

Small fry:
Clean up `getEvents()` callchain through `CommanderAbilityEvent`
Consider adding a `getAllModelsNot(long exclude)`

ThatIsAUsername:
> defer the whole animation shebang

Agreed. This PR has too much stuff and chatter already.

> defensive charge... death listener

Yeah, I didn't even think of that. Maybe not a death listener though, since dying doesn't specifically give charge, losing HP does. Maybe just create a `MassDamageEvent` for the unit instead?

> BattleMech building: search by role (and maybe other criteria), clone own model, moneyCost

Already does clone its own model - as for searching by role instead of name, I think I'd rather not change that until it causes a problem, since a different solution may be more apparent once we have a breaking case. If we search by role instead, we may be guaranteed to find a model, but then Tech may be paying 5 stars to summon an SCV with a rifle or whatever.
I went ahead and set the cost adjustment instead of changing `moneyCost`, so that's probably fine for now.

> Might also be worth tweaking the BattleMech

As in: figure out how strong this unit should be relative to the other units in the set, grab any two weapons that can hit the right things, and buff/nerf them to hit that benchmark? Sounds painful.

> HP display

We got one external review already; that seemed to verify the "initial impression" impression I had. I guess we could float it by more people to see what they say.

> getEvents() callchain

Isn't this done? We even get all three drop animations now.

> getAllModelsNot

Probably a good idea. I almost added it, and then just did the `~0` thing to be lazy.

ThislsAUsername:
> Maybe just create a `MassDamageEvent` for the unit instead?

Sounds good. It'll match user expectations as far as being "non-combat" charge, also.

> Already does clone its own model

Ah. Am blind.
> as for searching by role instead of name, I think I'd rather not change that until it causes a problem, since a different solution may be more apparent once we have a breaking case.

Hah. I suppose I can see the logic to that.
I'm not a super fan of the idea of making code intentionally fragile when IMO the more robust/correct solution is right there, but hey.
> If we search by role instead, we may be guaranteed to find a model, but then Tech may be paying 5 stars to summon an SCV with a rifle or whatever.

That wouldn't happen. It'd have to be an `ASSAULT` unit, aka something at least vaguely like a tank. I guess you could end up with some sort of... tank*hunter* vehicle, but it's at least going to be threatening to hard targets.
You could further require, say, less than 50% base damage from the cheapest `ASSAULT` vs it so as to ensure you get something durable.
> As in: figure out how strong this unit should be relative to the other units in the set, grab any two weapons that can hit the right things, and buff/nerf them to hit that benchmark? Sounds painful.

Nah. I wouldn't go that far. If anything, I'd grab the vehicle I plan to base it off, and just adjust attack and defense based on the base damage matchup with the most basic `ASSAULT` unit.
An even better plan that just occurred to me, though: base the star cost on the base unit's cost, in some way. (So it's a cheaper ability while using DoR units, but it's also not as threatening.)

> We got one external review already; that seemed to verify the "initial impression" impression I had. I guess we could float it by more people to see what they say.

That's fair, but my argument isn't about initial impression so much as visual clarity/information precedence for people who *do* know what's going on

> > getEvents() callchain
> 
> Isn't this done? We even get all three drop animations now.

No? `CommanderAbilityEvent.generateAbilityEvents()` just redirects to `myAbility.getEvents()` and is only called in one place.
Seems silly to leave that method in there when `AbilityAction` works directly with the ability already.

> Probably a good idea. I almost added it, and then just did the `~0` thing to be lazy.

Relatable.

ThatIsAUsername:
> the more robust/correct solution is right there

It's more likely to keep functioning without causing a crash, yes, but that's because it's more likely to start doing the wrong thing without telling anyone. It's not more robust, it's just less obvious in how it fails.

> It'd have to be an ASSAULT unit, aka something at least vaguely like a tank.

Well, sure. But I don't know what extra unit set we are likely to create or how we might bend expectations. In all probability, we'll never bother and this code will never break.

> base the star cost on the base unit's cost

Hmm. That would be possible. Might try that.

ThislsAUsername:
> It's more likely to keep functioning without causing a crash, yes, but that's because it's more likely to start doing the wrong thing without telling anyone. It's not more robust, it's just less obvious in how it fails.

My argument is that using the role is correct such that if something *does* break it, the other thing is wrong.
If we change the expectations on `UnitModelScheme`s, we'll have to re-evaluate, but I don't expect that to happen.
The name, by contrast, is something `UnitModelScheme`s have free reign over.

> Well, sure. But I don't know what extra unit set we are likely to create or how we might bend expectations. In all probability, we'll never bother and this code will never break.

I would still rather not make something that I *know* can be incompatible with a valid component on the other side of the interface.
If we change the interface, I'd expect to change it... but this will crash on valid input, and crashing ain't cool.

ThatIsAUsername:
> but this will crash on valid input, and crashing ain't cool.

It will crash when assumptions are invalidated, yes. This may just be a difference of opinion. I'd rather fail fast/crash/make it obvious that *something* is not how it is expected to be, rather than successfully create *some* unit that may be way overpriced for the value it provides, effectively neutering Tech's abilities without telling you in advance. This way, the crash happens when the match starts, and you don't get 7 turns in before discovering you've just spent your energy to summon a peashooter.

ThislsAUsername:
> It will crash when assumptions are invalidated, yes. This may just be a difference of opinion.

I'm mostly on about the assumptions set down in the `UnitModel` documentation.
You're adding the extra assumption that we'll have a "Md Tank", which isn't explicitly in `UnitModelScheme`'s contract. You're depending on an implementation detail.

> I'd rather fail fast/crash/make it obvious that _something_ is not how it is expected to be, rather than successfully create _some_ unit that may be way overpriced for the value it provides, effectively neutering Tech's abilities without telling you in advance. This way, the crash happens when the match starts, and you don't get 7 turns in before discovering you've just spent your energy to summon a peashooter.

We're *already* creating a unit that's "way overpriced for the value it provides". DoR Md Tanks are pretty bad compared to AW1-3's.
A better approach IMO is to adjust the power costs to suit what you're getting, or nudge the unit stats to compensate.

ThatIsAUsername:
> You're adding the extra assumption that we'll have a "Md Tank"

Yes. Because if that doesn't exist, then I'll immediately be notified to figure out how it should work with the new unit set.

> We're already creating a unit that's "way overpriced for the value it provides"

So that's perhaps fair. On the other hand, DoR Md Tanks still sit in about the same place in the power curve, so it's not completely crazy anyway. Maybe in DoR logic BattleMechs are just a bit cheaper and weaker, because.

> adjust the power costs to suit what you're getting

Tempting. That would be doable, but the info page requires static access to the ability costs.

> nudge the unit stats to compensate

Doing this properly would involve a statistical analysis of available unit firepower/cost o pick something that is sufficiently powerful but not too broken. I just... don't wanna.

ThislsAUsername:
> Yes. Because if that doesn't exist, then I'll immediately be notified to figure out how it should work with the new unit set.

I suspect I would be the person to run into that particular problem, so that argument rings a little hollow to me lol.

> So that's perhaps fair. On the other hand, DoR Md Tanks still sit in about the same place in the power curve, so it's not completely crazy anyway. Maybe in DoR logic BattleMechs are just a bit cheaper and weaker, because.

I... guess?

> > adjust the power costs to suit what you're getting
> 
> Tempting. That would be doable, but the info page requires static access to the ability costs.

...which is wrong as soon as you use a power anyway.
And we have dynamic displays of how much until your next power, etc.
I'm not too worried :P

> > nudge the unit stats to compensate
> 
> Doing this properly would involve a statistical analysis of available unit firepower/cost o pick something that is sufficiently powerful but not too broken. I just... don't wanna.

I'd just base it on a critical relationship, probably vs a normal tank. Wouldn't be too hard, I think.
But fair enough.

ThatIsAUsername:
> I suspect I would be the person to run into that particular problem

Perhaps. But then you'd yell at me and it would be my problem again anyway.

> which is wrong as soon as you use a power anyway

Did you just say "It's already wrong, so it doesn't matter if it's just more wrong in this case"?

> I'd just base it on a critical relationship, probably vs a normal tank

That would at least be more doable than trying to dynamically set star power based on unit cost, I think.

ThatIsAUsername:
Is it just me, or is power accumulation out of whack? If I do a Tech ditto, they start trading Steel Rains every turn even on smaller maps. The BattleMechs start to outnumber the normal units.

ThislsAUsername:
> Perhaps. But then you'd yell at me and it would be my problem again anyway.

I'd probably fix it myself. I don't like having broken stuff.

> Did you just say "It's already wrong, so it doesn't matter if it's just more wrong in this case"?

Uhhh, more or less? You could also put the base cost formula instead of an explicit number, I guess.

> That would at least be more doable than trying to dynamically set star power based on unit cost, I think.

Eh? Nah. Just pick a base star value for the boost+possible smush, then add unit cost times some scale factor (and probably a small discount for the 3x ability).
You have the model of the `BattleMech` on hand when you make the abilities already, so you can just call `getCost()`

ThislsAUsername:
> Is it just me, or is power accumulation out of whack? If I do a Tech ditto, they start trading Steel Rains every turn even on smaller maps. The BattleMechs start to outnumber the normal units.

BattleMechs cost like 20k. That's a lot of charge.
We have a higher charge rate than AW2, thanks to the HP-offense-based charge.
Also, you still charge while abilities are active, which matters a *lot*, especially when that also gives your army a funds injection.

ThatIsAUsername:
> You have the model of the BattleMech on hand when you make the abilities already, so you can just call getCost()

Nein. `BattleMech` is created per-Tech. InfoPages are created statically. Calculating a new star value is easy, making it statically available when it could differ between instances is not, unless we want to make Tech's CoInfo non-final and insert the additional info pages in the constructor or something.

ThislsAUsername:
> > You have the model of the BattleMech on hand when you make the abilities already, so you can just call getCost()
> 
> Nein. `BattleMech` is created per-Tech. InfoPages are created statically.

>Uhhh, more or less? You could also put the base cost formula instead of an explicit number, I guess.

> Calculating a new star value is easy, making it statically available when it could differ between instances is not, unless we want to make Tech's CoInfo non-final and insert the additional info pages in the constructor or something.

If we're really all that concerned about accuracy, we should display the _real_ costs dynamically instead of making them just static text.
Making an exception for Tech doesn't really make sense IMO (especially since we have another CO with non-standard power cost scaling in Ave).

ThatIsAUsername:
> Making an exception for Tech doesn't really make sense 

Agree, consistency is good. So how do we want to handle this. Make up numbers (e.g. 5 and 10) or create an issue to live-update the ability power costs?
I recognize that in AW the star values don't change, so it might make more sense (if we want Tech's power cost to differ based on unit stats) to just let Tech be an exception.

If you repeated your "base cost formula" quote again because I didn't respond to it the first time, you're going to have to explain it better.

ThislsAUsername:
> [C]reate an issue to live-update the ability power costs?

Don't have a strong opinion on this. It'd be a little nice, I guess?
OTOH, we could just standardize to making explicit notes of non-standard behavior and not bother (since these pages are used without an instance of the CO in view as well)

> I recognize that in AW the star values don't change, so it might make more sense (if we want Tech's power cost to differ based on unit stats) to just let Tech be an exception.

Right, but there's no *good* time/place right now to insert the new info such that it will be accurate.
I guess we could have the `COInfo` take in the game rules and do it that way, but ehhh...

> If you repeated your "base cost formula" quote again because I didn't respond to it the first time, you're going to have to explain it better.

I quoted it because that's exactly what my response was/would have been.
Let's say you settle on 2+ceil(mech cost/5k) for TechDrop's base cost formula. You can just... put that instead of a single number. Yeah, it's a little harder to read, but it's statically correct for a dynamic value.
Could even have a separate page about how the mechs are designed.

ThatIsAUsername:
> You can just... put that instead of a single number.

Nope. `BattleMechModel` is not statically available. It's created at Tech-creation-time.

> explicit notes of non-standard behavior

As in just put "Cost of this ability depends on the cost of the BattleMech unit, or some such? That feels like a cop out. I'd rather have the cost be fixed and predictable, and the unit be at a predictable place in the power curve. I might look into tweaking unit stats to try and achieve that since you seem dead-set on not calling existing models by name (though really, that would be the simplest option, having been done already).

ThislsAUsername:
> > You can just... put that instead of a single number.
> 
> Nope. `BattleMechModel` is not statically available. It's created at Tech-creation-time.

I meant to literally put in the formula, or some reasonable approximation thereof.

> As in just put "Cost of this ability depends on the cost of the BattleMech unit, or some such? That feels like a cop out. I'd rather have the cost be fixed and predictable, and the unit be at a predictable place in the power curve.

Fair. I suggested it mostly because you mentioned the ability could/should just be a different strength based on the unit set in use.

> I might look into tweaking unit stats to try and achieve that since you seem dead-set on not calling existing models by name (though really, that would be the simplest option, having been done already).

I am quite set against name-based fetching.
That said, tweaking stats is a separate thing and can be done separately.

ThislsAUsername:
I suppose I might be willing to concede on the `BattleMech` thing. It's your CO. Do as you will, I guess.
I still think it's worth getting other opinions on the HP thing, since the checked source is a) atypical in outlook and intuitions and b) not really our "target demographic". (And I still argue that visual clarity for the knowledgeable > initial understanding, since CO/power descriptions exist to aid in the latter)

ThatIsAUsername:
> We have a higher charge rate than AW2, thanks to the HP-offense-based charge.
Also, you still charge while abilities are active, which matters a lot

Isn't that going to throw off all of the test lab's careful testing? (And didn't you knock Walker a bit ago about "not properly copying the source material" or some such?) At any rate, I might want to make those abilities more expensive in light of that. Steel Rain every turn is crazy.

> I meant to literally put in the formula, or some reasonable approximation thereof

It would have to be based off a unit that doesn't exist yet, relative to other units that will vary based on the unit set; there's just not a meaningful way to express this at CO-selection time.
Trying to buff some other unit to a common place is also not necessarily workable; even just tank->md tank has power scaling that varies based on what they are attacking, so it isn't just a flat multiplier to get from A to B.

> I suppose I might be willing to concede

Cool. If/when it breaks, feel free to yell at me, or disable Tech, or remove 2/3 of her abilities, or whatever.

> I still think it's worth getting other opinions on the HP thing

Alright. I am sourcing more opinions.

> the checked source is...

I'd say the same thing about you. *I'm* the only normal one here.

ThislsAUsername:
> Isn't that going to throw off all of the test lab's careful testing?

You keep assuming that "this is how it is on `master`" means "this is how it is on the test branch".
It's been wrong the last few times.
I told you this explicitly.
I think it's fair to assume I know and have accounted for this.
> (And didn't you knock Walker a bit ago about "not properly copying the source material" or some such?)

AWBW sets out to be a 1:1 copy, and disallows custom content entirely (Except for Sturm, apparently. And mixing AW2 charge with the other games' default boosts. And black tiles. And the damage chart...).
The question about COP costs was mostly for *my* information, as I wanted to know if I was accurately emulating *AWBW*, since that mechanic is very significant for COs with Adder bars.
> At any rate, I might want to make those abilities more expensive in light of that. Steel Rain every turn is crazy.

Reasonable. It's free charge... for using your power.
I think, unless we plan to change how the UI shows, we should probably move to canceling charge that happens while a power is on.
It's cool, don't get me wrong, but it's "unexpected" and the UI doesn't message the mechanic at all.

> It would have to be based off a unit that doesn't exist yet, relative to other units that will vary based on the unit set; there's just not a meaningful way to express this at CO-selection time.

You can make a page on "BattleMech mechanics" and then just using "BattleMech cost" would make perfect sense.

> Trying to buff some other unit to a common place is also not necessarily workable; even just tank->md tank has power scaling that varies based on what they are attacking, so it isn't just a flat multiplier to get from A to B.

You mind explaining that more?

> Cool. If/when it breaks, feel free to yell at me, or disable Tech, or remove 2/3 of her abilities, or whatever.

*scribbles notes furiously*

> > I still think it's worth getting other opinions on the HP thing
> 
> Alright. I am sourcing more opinions.

Seems they're not so eager to reply there. I would have suggested technical-stuff or the AWBW customization chat.

> I'd say the same thing about you. _I'm_ the only normal one here.

I'd say we're both closer to normal than he is, based on extensive conversations with him. lol

ThatIsAUsername:
> You keep assuming that "this is how it is on master" means "this is how it is on the test branch".

Touché, I suppose. Maybe we should put master more in line with real AW. I feel like we've more or less dropped the pretense of not being a glorified clone at this point.

> The question about COP costs was mostly for my information, as I wanted to know if I was accurately emulating AWBW, since that mechanic is very significant for COs with Adder bars

Not sure which question you mean. Or what "Adder bars" are, unless that just means "very short"

> I think, unless we plan to change how the UI shows, we should probably move to canceling charge that happens while a power is on.

Probably. I'm not sure why we chose not to do that in the first place, unless it was "it seemed cool at the time".

> and then just using "BattleMech cost" would make perfect sense

I was tentatively going to base if off of `abilityPowerValue`, since in theory that would be easier to tune. Except we apparently don't use that variable anyway.

> You mind explaining that more?

For example, take the matchups of Tank and MdTank vs Inf, Recon, and MdTank.
Tank does 6, 7, 1 damage respectively
MdTank does 9, 9, 4 damage
Meaning the damage values go up by 3, 2, 3, with percent increases of 50, 28, and 300

> Seems they're not so eager to reply there. I would have suggested technical-stuff or the AWBW customization chat.

Well, it's been up for all of like 10 minutes. Feel free to cross-post though if you know a better place; I haven't been in the AWBW customization chat at all.

ThislsAUsername:
> Maybe we should put master more in line with real AW. I feel like we've more or less dropped the pretense of not being a glorified clone at this point.

I'm not even sure about the "glorified" part.
I've been plotting to make some of this stuff configurable or at least variable anyhow.
To wit:
Defense calcs are determined by the CO (and therefore which calc makes sense). This wouldn't be user-configurable since that doesn't make sense. It occurs to me that might clash horribly with my tag system idea (+270 defense sure is a number)
I might make luck damage calc variable if we decide we want to make a "more sensible" luck system (e.g. scaling with base damage but ignoring defense). I'm thinking it could/should work on the same axis as AW luck, so we might be able to make that user-configurable.
Charge rates should be configurable both by CO and by user. I want to support both DoR-style and AW1-3 style COs, so I think I want to get proper COU mechanics in before I start meddling with this (maybe it doesn't matter, dunno).
   My general plan for this has been: zone-charge and global-charge are configured separately by the user. Whether a CO uses zone or global charge is *not* configurable. The user may select "default", which means all COs behave as they did in their original incarnations. If the user selects non-default, that setting is applied to all COs in the match.

> Not sure which question you mean. Or what "Adder bars" are, unless that just means "very short"

I asked if AWBW emulated cart mechanics w.r.t. powers costing more than their activation threshholds. I thought that's what you were referring to.
Will has a 2/5 power cost pair, and Greyfield has a *one star* power, so that sort of thing is gonna be important.

> Probably. I'm not sure why we chose not to do that in the first place, unless it was "it seemed cool at the time".

We just... didn't prevent it.
And yeah, seemed cool at the time.

> I was tentatively going to base if off of `abilityPowerValue`, since in theory that would be easier to tune. Except we apparently don't use that variable anyway.

See above.

> For example, take the matchups of Tank and MdTank vs Inf, Recon, and MdTank.
> Tank does 6, 7, 1 damage respectively
> MdTank does 9, 9, 4 damage
> Meaning the damage values go up by 3, 2, 3, with percent increases of 50, 28, and 300

Right, but I'd argue Tank vs MdTank is the "critical" relationship, so you can tweak the attack/defense to give a similar boost to all matchups. Defense would be a little more tricky, but you could just "underbuff" that so it's about right for 1.5 terrain stars and I feel like it'd be about right (the scaling would work a little differently, but that's probably fine).
Or, if/when we get DoR-style defense, you could use that and then life is happy.

ThatIsAUsername:
> Defense calcs and CO zones

I don't think I followed all of that, but it's also out of scope for this and I'd probably forget by the time it became relevant again. I will say it sounds like it might be getting enough in the weeds to be unfriendly to users.

> emulated cart mechanics w.r.t. powers costing more than their activation threshholds

Oh, and we do it differently than AWBW for that, right? They increase the cost and then subtract, and we do the other way?
I was actually referring to how much damage terrain takes (vs. MdTank in AW, vs Neo in AWBW).

> Charging mechanics.

Maybe I'll leave Tech's powers as is for now, anticipating that they will be more inline with what I expect on the test branch, and possibly that we will align master with that at some point (unless you think we shouldn't).

> See above

If "above" refers to your previous comment about using unit cost; `abilityPowerValue` isn't visible to users and can't be included in the formula. Actually, including cost in the formula wouldn't be helpful either since they can't see how much `BattleMech`s cost since they aren't buildable. If you were referring to something else, you'll have to start being more explicit about things because I'm clearly not intelligent enough to follow the obscure bread crumbs you are so fond of leaving.

> HP display

So far we have two votes in. How long do we want to wait to make the call?

ThislsAUsername:
> I will say it sounds like it might be getting enough in the weeds to be unfriendly to users.

Eh, Maybe.

> Oh, and we do it differently than AWBW for that, right? They increase the cost and then subtract, and we do the other way?

That's cart. We're the same as AWBW.

> I was actually referring to how much damage terrain takes (vs. MdTank in AW, vs Neo in AWBW).

I don't think I complained about that?
There's been too much map design around that mechanic by now they weren't interested in changing it.

> Maybe I'll leave Tech's powers as is for now, anticipating that they will be more inline with what I expect on the test branch, and possibly that we will align master with that at some point (unless you think we shouldn't).

Sure.

> If "above" refers to your previous comment about using unit cost ... If you were referring to something else, you'll have to start being more explicit about things because I'm clearly not intelligent enough to follow the obscure bread crumbs you are so fond of leaving.

I was referring to the first section, with the charge mechanics stuff.
So. AW:DS uses `abilityPowerValue`.
I said "see above", because having the COs *from* AW:DS use their cart mechanics would necessarily require making the `abilityPowerValue` a used quantity.
It would logically follow that that'd be something I was planning to make configurable.

> `abilityPowerValue` isn't visible to users and can't be included in the formula.

Isn't, but could and should be.
You say "can't", but it definitely could be?

> Actually, including cost in the formula wouldn't be helpful either since they can't see how much `BattleMech`s cost since they aren't buildable.

They could technically figure it out via repairing one with a BBoat.
...but I was *just* suggesting including a "BattleMech mechanics" page that describes how their capabilities are set up (including cost/power value, as appropriate)
There's a lot of UI we're missing, still.

> So far we have two votes in. How long do we want to wait to make the call?

Bah. Fine. I still think it's terrible, but apparently I'm the minority here.

ThatIsAUsername:
> Sure

I opened issue #264 for this.

> It would logically follow that that'd be something I was planning to make configurable

Not necessarily, but then I'm not as well-versed as you in the various cart mechanics. You say honoring the DoR cart mechanics would necessitate using `abilityPowerValue`, and say that implies it would be configurable. That sounds contradictory. Either we must use `abilityPowerValue`, or it's configurable and we don't have to (unless I'm failing to connect your pronouns).

> in the formula.

Yes, we could put it in the formula, but `abilityPowerValue` isn't something that's exposed in the UI in any AW game, so for it to be meaningful we'd have to change a bunch of stuff.

> "BattleMech mechanics" page that describes how their capabilities are set up

I'm actually fairly opposed to this. I don't want a page describing how we clone a `UnitModel` or whatever; I'm only doing that because it's exponentially more painful to add each new unit type otherwise.

> Bah. Fine

Is there anything else outstanding in this branch, or is it good to go?

ThislsAUsername:
> > It would logically follow that that'd be something I was planning to make configurable
> 
> Not necessarily

Yes, necessarily. I stated that I plan to make COs behave as on-cart *by default*, but that you as the user could globally enforce one standard for a match.
I didn't explicitly state I was planning to make HP/funds/power value all available to fiddle with, but it would logically follow from stating that I wanted it to *be* configurable since those are the attributes that have been used in AW games.

> You say honoring the DoR cart mechanics would necessitate using `abilityPowerValue`, and say that implies it would be configurable. That sounds contradictory.

AW1/2 use funds. DS uses `abilityPowerValue`. DoR uses HP.
As for it being contradictory, see above.

> Either we must use `abilityPowerValue`, or it's configurable and we don't have to (unless I'm failing to connect your pronouns).

Whether to use `abilityPowerValue` can/should be configurable, so we *have* to use it... but only sometimes.

> Yes, we could put it in the formula, but `abilityPowerValue` isn't something that's exposed in the UI in any AW game, so for it to be meaningful we'd have to change a bunch of stuff.

Not really. If/when we put out a real detailed unit stats page, it can go there.
As for it being "exposed", high level DS play is all about properly (ab)using `abilityPowerValue` to get your power at a more opportune time. For instance, Recons have the highest power/funds ratio of any vehicle and can ram tanks to charge your power.
Funds-based charge is a little obfuscated, but you have the relevant numbers in front of you.
HP-based charge is uh... really easy to figure out, actually.

> I'm actually fairly opposed to this. I don't want a page describing how we clone a `UnitModel` or whatever; I'm only doing that because it's exponentially more painful to add each new unit type otherwise.

You can just say "it fights like a Md Tank + an AA gun, and costs X more funds than the Md (though powers deploy them for free)"
Then again, maybe it would make more sense to just leave that stuff for the detailed unit stats page, and let the users make inferences if they care.

> > Bah. Fine
> 
> Is there anything else outstanding in this branch, or is it good to go?

Uhhh. I think I'm good?

PR #260, opened 2020-06-27
Tech mod

ThislsAUsername:
I thought about the `MassStrikeUtils` thing.
In the end, I think this is just a bit more complex than fits in that paradigm: there's no fixed "footprint", and the "reward" for targets ranks differently at different radii.
I could, in principle, refactor it to use the same general search pattern, but that'd probably make it *more* complex, not less.

### DIFF
--- a/src/CommandingOfficers/Tech.java
+++ b/src/CommandingOfficers/Tech.java
@@ -260,12 +260,8 @@ public class Tech extends Commander
       {
         // Record any locations near owned properties; we want to be able to rescue unprotected structures.
         for( XYCoord xyc : Utils.findLocationsInRange(gameMap, propCoord, 0, dropRange) )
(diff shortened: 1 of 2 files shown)


src/CommandingOfficers/Tech.java
```diff
@@ -326,34 +314,13 @@ private TeleportEvent generateDropEvent(MapMaster gameMap, Set<XYCoord> priorDro
         // Assign base scores for spaces around this enemy.
         for( XYCoord xyc : Utils.findLocationsInRange(gameMap, nmexy, 0, dropRange) )
         {
```
ThatIsAUsername, line 316: Could add an `if(invalidDropCoords.contains(xyc)) continue;` here to skip some processing.

PR #261, opened 2020-06-28
Teleport action

ThatIsAUsername:
Accidentally started refactoring. Ended up with this. New PR for new discussions.

### DIFF
--- a/src/CommandingOfficers/Tech.java
+++ b/src/CommandingOfficers/Tech.java
@@ -246,13 +246,16 @@ public class Tech extends Commander
         Unit r = gameMap.getLocation(xyc).getResident();
         if( null != r )
         {
-          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
-          smashes.put(r.CO, (oldVal == null) ? 1 : (oldVal+1));
+          Integer oldVal = smashes.get(r.CO);
+          if( null == oldVal )
+            smashes.put(r.CO, 1);
+          else
+            smashes.put(r.CO, oldVal+1);
         }
       }
       for( Commander co : smashes.keySet() )
       {
-        if( co.units.size() <= smashes.get(co) )
+        if( co.units.size() > 1 && co.units.size() <= smashes.get(co) )
           abilityEvents.add(new CommanderDefeatEvent(co));
       }
 
--- a/src/CommandingOfficers/Tech.java
+++ b/src/CommandingOfficers/Tech.java
@@ -244,11 +244,8 @@ public class Tech extends Commander
         Unit r = gameMap.getLocation(xyc).getResident();
         if( null != r )
         {
-          Integer oldVal = smashes.get(r.CO);
-          if( null == oldVal )
-            smashes.put(r.CO, 1);
-          else
-            smashes.put(r.CO, oldVal+1);
+          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
+          if( null != oldVal ) smashes.put(r.CO, oldVal+1);
         }
       }
       for( Commander co : smashes.keySet() )
--- a/src/CommandingOfficers/Tech.java
+++ b/src/CommandingOfficers/Tech.java
@@ -11,10 +11,12 @@ import java.util.Set;
 import AI.AIUtils;
 import CommandingOfficers.Modifiers.CODamageModifier;
 import CommandingOfficers.Modifiers.CODefenseModifier;
+import Engine.GameAction.TeleportAction;
 import Engine.GameScenario;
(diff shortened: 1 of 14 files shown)


src/CommandingOfficers/Tech.java
```diff
@@ -223,23 +225,41 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
+      // Figure out if this caused the enemy's defeat. The TeleportActions can't track
+      // this because multiple units may be smashed at once.
+      Map<Commander, Integer> smashes = new HashMap<Commander, Integer>();
+      for( XYCoord xyc : dropLocs )
+      {
+        Unit r = gameMap.getLocation(xyc).getResident();
+        if( null != r )
+        {
+          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
+          smashes.put(r.CO, (oldVal == null) ? 1 : (oldVal+1));
```
ThislsAUsername, line 250: It should be more efficient to just do a normal if statement, here...

src/CommandingOfficers/Tech.java
```diff
@@ -223,23 +225,41 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
+        }
+      }
+      for( Commander co : smashes.keySet() )
+      {
+        if( co.units.size() <= smashes.get(co) )
+          abilityEvents.add(new CommanderDefeatEvent(co));
```
ThislsAUsername, line 256: So if you smush his last existing unit with a power, you get two defeat events?

ThislsAUsername:
Overall, I think I prefer the idea of using a mass damage event to notify for power gain, so as to avoid overloading the unit death event unnecessarily.

ThatIsAUsername:
> mass damage event

Certainly doable. The only downside is that we'd have to send `MassDamageEvent`+`UnitDieEvent` to get both death and AP effects.

src/CommandingOfficers/Tech.java
ThatIsAUsername: Yes. Realized that after I pushed it up. Not that two defeat events should break anything, but I should probs fix that.

src/Engine/GameAction.java
```diff
@@ -268,7 +276,12 @@ public GameEventQueue getEvents(MapMaster gameMap)
       // If our guy can't survive there, end him.
       if( !unit.model.propulsion.canTraverse(gameMap.getEnvironment(unitDestination)) )
       {
+        ArrayList<Unit> ary = new ArrayList<Unit>();
+        ary.add(unit);
+        boolean fatal = false;
```
ThislsAUsername, line 280: Needs to be true, or you will get 1 less HP's worth of charge than you should.

src/CommandingOfficers/Tech.java
```diff
@@ -246,13 +246,16 @@ public GameEventQueue getEvents(MapMaster gameMap)
         Unit r = gameMap.getLocation(xyc).getResident();
         if( null != r )
         {
-          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
-          smashes.put(r.CO, (oldVal == null) ? 1 : (oldVal+1));
+          Integer oldVal = smashes.get(r.CO);
+          if( null == oldVal )
+            smashes.put(r.CO, 1);
+          else
+            smashes.put(r.CO, oldVal+1);
```
ThislsAUsername, line 251:
```suggestion
          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
          if( null != oldVal ) smashes.put(r.CO, oldVal+1);
```

src/CommandingOfficers/Tech.java
```diff
@@ -223,23 +223,44 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
+      for( XYCoord xyc : dropLocs )
+      {
+        Unit r = gameMap.getLocation(xyc).getResident();
+        if( null != r )
+        {
+          Integer oldVal = smashes.get(r.CO);
+          if( null == oldVal )
+            smashes.put(r.CO, 1);
+          else
+            smashes.put(r.CO, oldVal+1);
```
ThislsAUsername, line 251:
```suggestion
          Integer oldVal = smashes.putIfAbsent(r.CO, 1);
          if( null != oldVal ) smashes.put(r.CO, oldVal+1);
```
Ayy, this one worked, I think

src/Engine/GameAction.java
ThatIsAUsername: Gah. I had it that way, and changed it for testing, and forgot to change it back. Good catch.

ThislsAUsername:
Looks good to me.

PR #267, opened 2020-07-05
236 quick info panel

ThatIsAUsername:


ThislsAUsername:
Wow, you've been busy.

ThislsAUsername:
You don't currently display materials when relevant.
I can't say I'm a super fan of having leading zeros on the numbers.
HQs clip into the coordinates display.
Coordinates are black against a black tint over "something", which can be pretty unreadable.
Terrain health doesn't pop up immediately on first demolition attempt.
It'd be cool (though not necessarily in scope) to see what's currently in a transport.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
```diff
@@ -101,7 +101,7 @@ private static void generateOverlay(GameMap map, XYCoord coord)
       if( unit.ammo >= 0 ) 
         unitAttrs.add(new AttributeArtist(SpriteLibrary.getAmmoIcon(), unit.ammo));
       if( unit.getCaptureProgress() > 0)
-        unitAttrs.add(new AttributeArtist(SpriteLibrary.getCaptureIcon(unit.CO.myColor), unit.getCaptureProgress()));
+        unitAttrs.add(new AttributeArtist(SpriteLibrary.getCaptureIcon(unit.CO.myColor), 20-unit.getCaptureProgress()));
```
ThislsAUsername, line 104: This smells magical to me

ThatIsAUsername:
> materials

Not part of the issue! But I'll think about it.

> Leading zeros.

Should be removable

> coords clip HQ

They do. I could move the coordinates above the grey box, or to a corner on the other side of the screen. Need to figure out where to put the turn counter at some point too.

> coordinate colors

I thought about using bold lettering like for HP/etc, but that quickly got too big to fit in the grey box. Left as is til we figure out something better.

> Terrain health updates.

Fixable. Now if only someone had created the listener hooks when the event was first introduced...

> Contents of transports

Make an issue.

> Unit/Terrain names

I wasn't going to bother with this; the names would take a lot of space.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
ThatIsAUsername: It's magical in `Unit` as well. Where do you propose it should live instead?

ThislsAUsername: My "ideal" answer would be `GameRules`, but I feel like that'd be hard to set up access to.
Could be a static field in `TerrainType`?

ThislsAUsername:
> coordinates

Why not use a full UI panel instead of the semitransparent box?

> turn counter

Super Famicom Wars has it next to your funds, if that helps.
I might go for putting it somewhere on the game stats page?

> Now if only someone had created the listener hooks when the event was first introduced...

Quoth the me: YAGNI.
(or more accurately, I figured it wouldn't be hard to set up)
(and we frequently end up adjusting the listeners for new use cases anyhow)

ThatIsAUsername:
> Why not use a full UI panel

Not sure what you mean. Like a new UI element that stretches across the whole screen? Wouldn't want to do that, but maybe you mean something else?

> turn counter

Yeah, stats page is probably a good place to have it, whether it's on the main HUD or not. I guess it may not need to be always-visible.

ThislsAUsername:
> > Why not use a full UI panel
> 
> Not sure what you mean. Like a new UI element that stretches across the whole screen? Wouldn't want to do that, but maybe you mean something else?

I mean using the beige panels we use for the rest of the UI.
Sure, it'll fully block what's behind it, but it flips with the header to stay out of the way and the users do fullscreen anyhow.

ThatIsAUsername:
I don't think I want fully-opaque panels; I don't like to have to scroll across the middle of the screen to see what's under it, and fullscreen only helps with maps that don't fill the screen (and doesn't help for folks who don't fullscreen). I might be OK with an opaque panel for just the coordinates, maybe in the other bottom corner? I dunno.

ThatIsAUsername:
Not sure I'm really a fan of putting the coordinates in the opposite corner like that. Maybe it should go back inside the tile info box where it's less conspicuous. Or maybe drop it in right below the CO bar, along with the turn counter.
We could also have a "Show Coordinates" option for folks who don't want it (or for folks who do, if we default it to off).
Any opinions?

ThislsAUsername:
> Any opinions?

Not really, no. Coordinate display isn't really a high priority in my book.

ThatIsAUsername:
> Not really, no. Coordinate display isn't really a high priority in my book.

Yeah, that was part of why I had it hidden in low-contrast font in the terrain-preview box. Most people won't ever need to see it. I mostly added it because sometimes when debugging AI it will make it easier to find units.

How about I just make drawing of tile coordinates a configurable option. Most people won't ever bother to turn it on.

ThislsAUsername:
We merging this?

PR #269, opened 2020-07-05
229 air unit crashes

ThatIsAUsername:


src/Units/UnitModel.java
```diff
@@ -189,12 +193,40 @@ public boolean canRepairOn(Location locus)
...
+   * Overriders should collect the output of super.getTurnInitEvents() before returning.
    * @param self Assumed to be a Unit of the model's type.
+   * @param map The current true state of the game map.
    */
   public GameEventQueue getTurnInitEvents(Unit self, MapMaster map)
   {
-    // Most Units don't have any; specific UnitModel types can override.
-    return new GameEventQueue();
+    GameEventQueue queue = new GameEventQueue();
+    if( isAirUnit() && 0 == self.fuel )
```
ThislsAUsername, line 203: This should check for fuel drain, since sea units also are a thing.

src/Units/UnitModel.java
```diff
@@ -189,12 +193,40 @@ public boolean canRepairOn(Location locus)
...
+      // First check if we are on an airport.
+      if(loc.getOwner() == self.CO && canRepairOn(loc))
+        ok = true;
+
+      // Next see if there is a friendly APC nearby.
+      List<XYCoord> adjacents = Utils.findLocationsInRange(map, xyc, 1);
+      for( XYCoord adj : adjacents )
+      {
+        Unit res = map.getLocation(adj).getResident();
+        if( null != res && (res.CO == self.CO) && res.model.hasActionType(UnitActionFactory.RESUPPLY) )
```
ThislsAUsername, line 219: May as well add the auto-resupply logic into this function, since you're now depending on units having the `RESUPPLY` action == doing auto-supply on turn init.
I'm not opposed to that notion, mind, it's just not a constraint we currently enforce and you're touching the logic.

src/Units/UnitModel.java
ThatIsAUsername: Do they die when out of fuel? I guess it make sense since there isn't a way to refuel them, but boats float when not under power.

src/Units/UnitModel.java
ThatIsAUsername: It would be a little strange to move that here, though I guess there's no reason we can't. If we ever make units that can resupply one thing but not others, we'll have issues anyway.
I can pull that in if you want. Seems odd since I think APCs are what made us create the `UnitModel` event hook in the first place.

src/Units/UnitModel.java
ThislsAUsername: Yup.
I don't make the rules, lol

src/Units/UnitModel.java
ThislsAUsername: > It would be a little strange to move that here, though I guess there's no reason we can't.

It'd be even weirder to have a supplying unit that *can't* auto-supply be able to keep an empty jet in the air by its mere presence.

> If we ever make units that can resupply one thing but not others, we'll have issues anyway.

For your reference, then, in case we ever go for a SFW unit set:
Fuel Trucks in Super Famicom Wars do have auto-supply, but cannot supply jets.
I don't recall at the moment whether they can resupply ships.
In addition, they can only resupply copters on open ground (IIRC only forests and mountains are banned. If it's relevant, I believe it's 1:1 with the places T-Copters can drop units from.)

> I can pull that in if you want.

I would prefer that.

> Seems odd since I think APCs are what made us create the `UnitModel` event hook in the first place.

They were.
You're depending on that behavior now, though, so IMO it goes here unless and until we have a reason to change/"better" solution.
The first thing to mind would be an "auto-supply" unit role.

ThatIsAUsername: I almost created a SUPPLY unit role, and then didn't because I realized it would be duplicative of "has the RESUPPLY action"

src/Units/UnitModel.java
```diff
@@ -200,32 +202,49 @@ public boolean canRepairOn(Location locus)
...
+          queue.add(new ResupplyEvent(res, self));
+          resupplying = true;
           break;
         }
       }
+    }
 
-      if(!ok)
-        queue.add(new UnitDieEvent(self));
+    if( !resupplying && (0 == self.fuel) & (isAirUnit() || isSeaUnit()) )
```
ThislsAUsername, line 242: I still think it should be based on whether the unit has fuel drain, rather than unit type.

ThatIsAUsername: But then land units would die. I'm pretty sure _that_ would not be cart-compliant.

src/Engine/UnitActionLifecycles/ResupplyLifecycle.java
```diff
@@ -118,37 +104,29 @@ public GameEventQueue getEvents(MapMaster map)
...
         supplyLocation = myLocation();
 
-        // Add a move event if we need to move.
-        // Note that movePath being null is OK for ResupplyAction when it is being re-used.
-        if( movePath != null )
+        // If we get blocked, don't resupply anything.
+        if( Utils.enqueueMoveEvent(map, unitActor, movePath, eventSequence) )
         {
-          // If we should be blocked, don't resupply anything.
-          if( !Utils.enqueueMoveEvent(map, unitActor, movePath, eventSequence) )
```
ThislsAUsername, line 133: That was backwards this whole time, wasn't it?
I cri.

src/Units/UnitModel.java
ThislsAUsername: Cart *has* no land units with fuel drain.
I guess I could ask Xen/hack a ROM to figure out which case is "correct", but the carts frankly *don't address* that question at all.

ThatIsAUsername: Hm. But why use `idleFuelBurn` instead of domain, since that seems like a more directly-related concept (falling out of the sky/sinking vs... randomly combusting once you are out of combustible materials).

ThislsAUsername: It's more a question of "needing the fuel on a continuous basis to function properly".
I don't pretend to know why naval units drain fuel in the way they do, though it makes perfect sense for air (though perhaps the relative quantities might want adjustment)

I'd honestly argue that infantry should have a fuel drain, since the only obvious "fuel" is food and you can't exactly fight on an empty stomach.

src/Engine/UnitActionLifecycles/ResupplyLifecycle.java
ThatIsAUsername: It was negated the whole time, but the logic was structured differently so it worked.

src/Units/UnitModel.java
ThatIsAUsername: "Needing fuel on a continuous basis to function" is a fair cry from "spontaneously explodes when out of fuel".

> naval units

It could be argues that it takes effort to maintain position with waves or whatever. I dunno.

> infantry

I'm not sure I want to add fuel drain to infantry, even if they don't explode when they run out. It would just be so annoying to try and resupply them all (not that most of them would live long enough for it to matter, but still).

ThislsAUsername: > "Needing fuel on a continuous basis to function" is a fair cry from "spontaneously explodes when out of fuel".

In the way that I meant it, it does.
You can turn your car engine off, and it's fine.
You can turn a jet's engine off, but you're probably in the air... and if so, you'll have "problems".
For sea, maybe it's something like a non-renewable factor in their power plants? If nuclear, some kind of stabilizing agent could conceivably be necessary to keep the thing from exploding, lol.

I still hold that passive fuel drain is the determining factor.

> infantry

It would be annoying, sure, but it would make perfect simulationist sense (which obviously isn't the same as perfect sense for the game).

ThatIsAUsername: > I still hold that passive fuel drain is the determining factor.

And I hold that running out of fuel in a treacherous domain is the determining factor (plus it keeps us from having to make up strange justifications about stabilizing nuclear agents). I'll change it to `!isLandUnit()` if you want, but it doesn't make sense to base it on fuel drain.

> Simulationist Infantry

I still don't think I want that.

ThislsAUsername: > And I hold that running out of fuel in a treacherous domain is the determining factor

And what, pray tell, determines whether a domain is treacherous?
I would argue the bit of info that encodes that *is* the fuel drain - the unit has to expend effort to not-die on a continuous basis.
I would find it very curious if an air unit with no drain (say, a blimp or Eagle copters) exploded on using the last unit of fuel... but could chill indefinitely without the last bit of it.

ThatIsAUsername: > I would find it very curious if an air unit with no drain (say, a blimp or Eagle copters) exploded on using the last unit of fuel... but could chill indefinitely with the last bit of it.

That's a fair point, though I'm not sure even blimps would count as having zero drain. Conversely, I'd find it very curious if a land unit with fuel drain (say if we wanted to limit BattleMech effectiveness or introduce a supply-challenged CO) exploded while just sitting there.

Perhaps the right answer is to use both: Units that don't sit on terra firma and do have fuel drain are susceptible.

ThislsAUsername: Bah. I did a quick test on Dual Strike, and Eagle's copters do indeed go boom IFF they're at exactly 0 fuel.
Go ahead and merge it, I guess.

PR #273, opened 2020-07-07
Fixes #262 Join and MassDamage actions now correctly remove units.

### LINKED ISSUE #262: SEEK sometimes tries to move to invalid locations
Prints out "ERROR! GameInstance.setLocation() was given an out-of-bounds location. Ignoring."

ThatIsAUsername:
Turns out `JOIN`ing units would leave both in the CO's unit list. Then SEEK would still try to find it.


ThislsAUsername:
I'm not sure how I feel about mass damage removing units.
I think that should be left as the territory of death events, since as-is you could do lethal mass damage and just never tell anybody the unit died (nor, in fact, animate its death). Leaving the unit around makes the erroneous program state obvious.

PR #274, opened 2020-07-11
Fancy map preview

### LINKED ISSUE #235: Better map previews
If there's enough space, it'd be nice if the map selection screen would show the actual map tiles at the biggest scale that will fit.
The users like maximizing the game, for some reason.

ThislsAUsername:
Fixes #235

### DIFF
--- a/src/UI/Art/SpriteArtist/MiniMapArtist.java
+++ b/src/UI/Art/SpriteArtist/MiniMapArtist.java
@@ -3,16 +3,17 @@ package UI.Art.SpriteArtist;
 import java.awt.Color;
 import java.awt.Graphics;
 import java.awt.image.BufferedImage;
-
 import Engine.XYCoord;
 import Terrain.MapInfo;
 import UI.UIUtils;
+import UI.UIUtils.COSpriteSpec;
+import UI.UIUtils.Faction;
 
 public class MiniMapArtist
 {
   private static MapInfo lastMapDrawn;
   private static Color[] lastTeamColors;
-  private static BufferedImage lastMapImage;
+  private static BufferedImage lastMiniMapImage, lastFullMapImage;
 
   /**
    * Retrieve a BufferedImage with a 1-pixel-per-tile representation of the provided MapInfo.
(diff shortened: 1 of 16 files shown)


ThatIsAUsername:
0/10 doesn't show predeployed units.

That shift from raw pixels to 1:1-scale is pretty satisfying though. Like it just pops right into focus.

src/UI/Art/SpriteArtist/MiniMapArtist.java
```diff
@@ -101,6 +109,66 @@ private static BufferedImage generateMapImage(MapInfo mapInfo, Color[] teamColor
     return image;
   }
 
+  /**
+   * Generates a BufferedImage containing the full-scale map.
+   */
+  private static BufferedImage generateFullMapImage(MapInfo mapInfo, Color[] teamColors)
```
ThatIsAUsername, line 115: I assume you considered it, but were you not able to just have `SpriteMiniMapArtist` draw itself instead of reimplementing this logic? That seems like the obvious thing to do, though admittedly I haven't looked to see how feasible that would be.

ThislsAUsername:
> 0/10 doesn't show predeployed units.

Thought about that. Would have to massage interfaces a bit to get there, though.
And predeployed is mostly a meme.

> That shift from raw pixels to 1:1-scale is pretty satisfying though. Like it just pops right into focus.

Heh.

src/UI/Art/SpriteArtist/MiniMapArtist.java
ThislsAUsername: 1) It's all tied up in the `GameMap` interface proper.
2) It's spread across no less than 2 functions and 2 classes, and would probably require faking a set of COs and `GameRules`.

I'll warrant it's a little ugly to have it duplicated, but I really don't see a way to fix that.

I'll also warrant that mixing the true-scale and 1:1-scale drawing logic isn't great either, but you can't really get what I was looking for without making that jump.

ThatIsAUsername:
> And predeployed is mostly a meme.

Hm? Shadows Chase has predeployed transports. Really plays havoc with the AI since they are on neutral seaports.

ThislsAUsername:
> Hm? Shadows Chase has predeployed transports. Really plays havoc with the AI since they are on neutral seaports.

My point is mostly that having significant predeployed forces pretty much bars a map from competitive play.
There are exceptions, but they're pretty shenanigans anyway.

ThislsAUsername:
We good here, btw?

ThatIsAUsername:
I was still curious about using `support()` instead of overriding `equals()` since it seems to do essentially the same job, but otherwise yes.

ThislsAUsername:
> I was still curious about using `support()` instead of overriding `equals()` since it seems to do essentially the same job, but otherwise yes.

Because I don't know *a priori* whether (or which) argument to the function might be null, and calling `x.equals(y)` will throw if `x` is null.
And one or the other being null is a standard use case that I would like to be defined by the same logic that otherwise defines matching.
I guess I could make a *static* `equals` function, but `support()` is more clear IMO.

PR #279, opened 2020-07-18
Fixed loading of sprite variations.

ThatIsAUsername:
It was loading variation 0 twice, and never loading the last one. Should be fixed now.


ThislsAUsername:
The `while` blocks on 162 and 187 could be one function, and there's probably more stuff that could be done to clean up...
I'm okay with merging as is, though, since it's not really broken as is.

PR #283, opened 2020-07-26
Modular AI

ThislsAUsername:
I wanted to make Wally's stuff a bit more generic, but extricating it from the assumptions/state in the AI is... a little more than I felt like biting off.

There's more coming down the pipe for another PR or two, but I got mired in fiddling with stuff lol
Specifically:
I want to let AIs serialize normally if they like
I want to add a generic "long term unit goal" thing to `ModularAI` to allow better allocation of units. This will also feed into my plans to revamp Wally's purchase logic (whether it can be generalized is TBD) to buy units to go specific places.

I also have mildly concrete plans for how to handle transports... I'm kinda feeling like getting that into the AI is an important bit of polish (though obviously e.g. ISAI doesn't need it) that I want set up before public playtest. Thoughts on that?


src/AI/WallyAI.java
```diff
@@ -692,7 +692,7 @@ private GameAction findTravelAction(
     boolean ignoreResident = true;
     ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, ignoreResident);
     if( ignoreSafety ) // If we *must* travel, make sure we do actually move.
-      destinations.remove(0);
+      destinations.remove(new XYCoord(unit.x, unit.y));
```
ThatIsAUsername, line 725: You're saying [this](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#remove-int-) doesn't work?

ThislsAUsername: It works just fine. It even still works for that use case, I think - I'm pretty sure I misdiagnosed which piece of code produced the "sit in place when you're not supposed to" behavior.
That said, this is more explicit and doesn't depend on the implementation details (i.e. ordering) of `findPossibleDestinations()`, so I see no reason to change it back.

src/AI/AIUtils.java
```diff
@@ -482,4 +506,23 @@ public int compare(Unit o1, Unit o2)
...
+
+  /**
+   * Finds allied production centers in the input set
+   */
+  public static Set<XYCoord> findAlliedIndustries(GameMap gameMap, Commander co, Iterable<XYCoord> coords, boolean ignoreMyOwn)
+  {
+    Set<XYCoord> result = new HashSet<XYCoord>();
+    for( XYCoord coord : coords )
+    {
+      Commander owner = gameMap.getLocation(coord).getOwner();
```
ThatIsAUsername, line 518: I know this works because `isEnemy()` filters out null COs, but I might still prefer to have an explicit check for whether the location is actually owned.

ThislsAUsername: I feel like we'd have to change a lot of stuff if we changed how `isEnemy()` behaves.
Would a comment suffice?

ThatIsAUsername: I wasn't suggesting a change to `isEnemy()` at all; rather an explicit check here instead of relying on `isEnemy()` implementation details.
It's probably not needed, really; I just didn't immediately see how "null CO" gets handled when I went through the mental checklist of edge cases.

ThislsAUsername: > I wasn't suggesting a change to `isEnemy()` at all; rather an explicit check here instead of relying on `isEnemy()` implementation details.

I'd argue that "returns true if the other CO is null" is part of the function's contract rather than an implementation detail.

> It's probably not needed, really; I just didn't immediately see how "null CO" gets handled when I went through the mental checklist of edge cases.

May as well throw in a comment, then.

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+    // Create a list of every property we don't own, but want to.
+    unownedProperties = new ArrayList<XYCoord>();
+    for( int x = 0; x < gameMap.mapWidth; ++x )
+    {
+      for( int y = 0; y < gameMap.mapHeight; ++y )
+      {
+        XYCoord loc = new XYCoord(x, y);
+        if( gameMap.getLocation(loc).isCaptureable() && myCo.isEnemy(gameMap.getLocation(loc).getOwner()) )
+        {
+          unownedProperties.add(loc);
```
ThatIsAUsername, line 53: Why not just call `AIUtils.findNonAlliedProperties()`?

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+    ArrayList<Unit> eligibleUnits = new ArrayList<Unit>();
+    for( Unit unit : myCo.units )
+    {
+      if( unit.isTurnOver || !gameMap.isLocationValid(unit.x, unit.y) )
+        continue; // No actions for units that are stale or out of bounds.
+      eligibleUnits.add(unit);
+    }
+
+    GameAction nextAction = null;
+    PriorityQueue<Unit> unitQueue = new PriorityQueue<Unit>(13, unitOrderSetter);
```
ThatIsAUsername, line 82: Why 13 instead of `myCo.units.size()`?

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+ */
+public abstract class ModularAI implements AIController
+{
+  protected Commander myCo = null;
+
+  // Updated on turn init
+  protected ArrayList<XYCoord> unownedProperties;
+  // List of possible AI modes, in order of precedence
+  protected ArrayList<AIModule> aiPhases;
+  // Used to determine what the default priority for unit actions is
```
ThatIsAUsername, line 28: A bit ambiguous/awkward. I'm assuming `unitOrderSetter` just defines the order in which units should act, but the comment seems to imply that action types are being ordered somehow.

src/AI/ModularAI.java
ThatIsAUsername: Or rather, `eligibleUnits.size()`

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+
+    GameAction nextAction = null;
+    PriorityQueue<Unit> unitQueue = new PriorityQueue<Unit>(13, unitOrderSetter);
+
+    // Each module gets a shot at the whole unit queue, in order.
+    // If no action's found, we're done
+    for( AIModule phase : aiPhases )
+    {
+      unitQueue.clear();
+      unitQueue.addAll(eligibleUnits);
```
ThatIsAUsername, line 89: Is it really necessary to clear and re-add all the units between calling each submodule?

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+
+    public boolean checked = false;
+    @Override
+    public void initTurn(GameMap gameMap) { checked = false; }
+    @Override
+    public GameAction getNextAction(PriorityQueue<Unit> unitQueue, GameMap map)
+    {
+      if( checked )
+        return null;
+      checked = true;
```
ThatIsAUsername, line 125: Won't this make it so this module can only ever activate an ability at the start of a turn?

src/AI/ModularAI.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    @Override
+    public GameAction getUnitAction(Unit unit, GameMap map)
+    {
+      if( unit.getCaptureProgress() > 0 )
+      {
+        XYCoord position = new XYCoord(unit.x, unit.y);
+        ai.unownedProperties.remove(position);
+        return new CaptureLifecycle.CaptureAction(map, unit, Utils.findShortestPath(unit, position, map));
```
ThatIsAUsername, line 149: This is a pretty decent "default" capture-continuation behavior, but it doesn't really accommodate situations where you don't want to; e.g. a high-value target is adjacent and only reachable from the spot being captured, or the capturing unit is about to be stomped by nearby enemies and won't be able to finish anyway, or the unit has already been damaged and a healthy friend is nearby who can complete the capture faster.

One of the main concerns I have with this whole approach is that although it makes it easier to get a new AI running with basic functionality, it hides a lot of the complexity and reduces the problem to trying to find the right building blocks, when a successful strategy will necessarily use a more (oh boy I get to use this in a sentence) holistic approach.

src/AI/WallyAI.java
```diff
@@ -70,29 +51,25 @@ public AIMaker getAIInfo()
...
   private static final double TERRAIN_PENALTY_WEIGHT = 3; // Exponent for how crippling we think high move costs are
   private static final double MIN_SIEGE_RANGE_WEIGHT = 0.8; // Exponent for how much to penalize siege weapon ranges for their min ranges 
-
-  private ArrayList<XYCoord> unownedProperties;
-  private ArrayList<XYCoord> capturingProperties;
   
+  private Map<UnitModel, Map<XYCoord, Double>> threatMap;
+  private ArrayList<Unit> allThreats;
   private HashMap<UnitModel, Double> unitEffectiveMove = null; // How well the unit can move, on average, on this map
   public double getEffectiveMove(UnitModel model)
```
ThatIsAUsername, line 70: So I know this is unrelated to the whole modular AI breakdown, and as such should be addressed in a separate PR if at all, but why oh why are you a) recalculating each model's effective movement at the beginning of every turn, and b) looping over every map location for every model for every CO instead of just caching some numbers?

src/AI/WallyAI.java
```diff
@@ -589,124 +634,184 @@ else if( unit.model.possibleActions.contains(UnitActionFactory.ATTACK) )
...
             && myCo.isEnemy(loc.getOwner()) )
         {
           goals.add(coord);
         }
       }
-      Utils.sortLocationsByDistance(new XYCoord(unit.x, unit.y), goals);
     }
 
+    if( goals.isEmpty() ) // If there's really nothing to do, go to MY HQ
+      goals.add(myCo.HQLocation);
```
ThatIsAUsername, line 651: Bold

src/AI/WallyAI.java
```diff
@@ -786,22 +889,22 @@ private double findAssaultKills(GameMap gameMap, Queue<Unit> unitQueue, Map<XYCo
           neededAttacks.put(xyc, unit);
           double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc.xCoord, xyc.yCoord).defenderHPLoss;
 
-          if (thisDamage > target.getPreciseHP())
+          if( thisDamage > target.getPreciseHP() )
             continue; // OHKOs should be decided using different logic
 
-          log(String.format("  Use %s to deal %sHP?", unit.toStringWithLocation(), thisDamage));
+//          log(String.format("  Use %s to deal %sHP?", unit.toStringWithLocation(), thisDamage));
```
ThatIsAUsername, line 895: Are these log statements supposed to be commented out, or just removed?

src/AI/WallyAI.java
```diff
@@ -815,9 +918,9 @@ private static int valueUnit(Unit unit, Location locale, boolean includeCurrentH
   {
     int value = unit.model.getCost();
 
-    if (includeCurrentHealth)
+    if( includeCurrentHealth )
       value *= unit.getHP();
-    value *= (unit.getCaptureProgress() > 0 && locale.isCaptureable())? 8 : 1; // Strongly value capturing units
+    value *= (unit.model.hasActionType(UnitActionFactory.CAPTURE) && locale.isCaptureable()) ? 8 : 1; // Strongly value capturing units
```
ThatIsAUsername, line 923: Isn't capture progress a more true metric of whether a unit is actually capturing? Checking the action type tells whether it _could_ start capturing, but then I might argue you should also check if there are any capturable tiles in range, not just directly underneath.

src/AI/ModularAI.java
ThislsAUsername: It sets the order that units are fed to the `aiPhases`, which will typically define the default order in which phases pick actions.

So yes, and also yes.

src/AI/ModularAI.java
ThislsAUsername: Submodules are free to pop units off the queue or whatever. They can't change the comparator on it, or I'd reset that too.
I want a clean slate for each module, aside from intentional crosstalk.

src/AI/ModularAI.java
ThislsAUsername: It means you only ever check for the ability once.
If you place it at the end of the list of modules, you're guaranteeing that if you *reach* the module, you're already otherwise going to end your turn anyway.
I only foresee problems with this if we introduce the ability to use multiple CO powers in one turn.

src/AI/ModularAI.java
ThislsAUsername: > This is a pretty decent "default" capture-continuation behavior, but it doesn't really accommodate situations where you don't want to; e.g. a high-value target is adjacent and only reachable from the spot being captured, or the capturing unit is about to be stomped by nearby enemies and won't be able to finish anyway, or the unit has already been damaged and a healthy friend is nearby who can complete the capture faster.

All of this is true, and a great argument for... putting all that cool logic into higher-priority modules.
Or you can roll your own.
It'd be just as valid to put this at the end of your module list, so long as you account for the fact that not all your units will have finished any applicable captures.

> One of the main concerns I have with this whole approach is that although it makes it easier to get a new AI running with basic functionality, it hides a lot of the complexity

Isn't that... the *whole point* of an abstraction?
This AI base gives you:
a) a very concise and idiomatic way of specifying what sorts of actions your AI will take, and in what order they will be considered/executed.
b) free management of a few key things like finding the "highest priority" unit to act.

You don't really lose any control.

> ...and reduces the problem to trying to find the right building blocks

Not really? It just makes it easier to put them together.

> ...when a successful strategy will necessarily use a more (oh boy I get to use this in a sentence) holistic approach.

How is "defining what I'm gonna do and what my priorities are" not a holistic approach?

src/AI/WallyAI.java
ThislsAUsername: re: first init, that's actually part of why I want to introduce persistent AI. If I fight Wally/Lash and save during a power, Wally's gonna be clinically addicted to/afraid of Recons for the rest of the match.
Then again, it might *be* worth re-evaluating partway through the match if e.g. you're fighting Ave on a map full of roads and suddenly those air units can't push as well as tanks.
...AI is hard.

As far as caching some numbers... what else would I cache?
The move cost/reachable tiles are contingent on the `MoveType`, so I can't cache between unit types.
If you mean caching the move cost per unit per terrain and then just checking that for each tile... I did. It's called a `MoveType`.

I guess I could consider only `MoveType`s, but they're 1:1 for `UnitModel`s anyway because cloning.

src/AI/WallyAI.java
ThislsAUsername: I really thought about deleting them, but it'd be a real pain to get them back if I ever wanted to debug that stuff more.
The system's pretty well working as I expect though, so they're mostly log clutter right now.
If we had log levels, they'd be like `AI_TRACE` or something.

src/AI/WallyAI.java
ThislsAUsername: True, but this means you could e.g. kick a Tank off of a city so an inf can capture it, or kick a lower-HP inf off so you can capture it (if it hadn't used its turn capturing already).
It also means Wally's very interested in getting enemy troops off *his* property.

It *does* provide the "interesting" property that Wally values infantry on properties more even on their own turf, so I guess I should fix that.

src/AI/ModularAI.java
```diff
@@ -68,7 +68,7 @@ public GameAction getNextAction(GameMap gameMap)
     }
 
     GameAction nextAction = null;
-    PriorityQueue<Unit> unitQueue = new PriorityQueue<Unit>(13, unitOrderSetter);
+    PriorityQueue<Unit> unitQueue = new PriorityQueue<Unit>(eligibleUnits.size(), unitOrderSetter);
```
ThatIsAUsername, line 71: This works great after the first turn, but you should really enforce a minimum of 1 (because 0 causes a stack trace).

src/AI/WallyAI.java
ThatIsAUsername: At first glance, you could cache `validTiles` and `totalCosts` to avoid needing to loop over the whole map again.
That said, since this is something that only happens at init time, I'm not too worried about saving cycles.

Yeah, I hadn't really considered the whole starting mid-match thing with AI. Muriel would be affected by that as well.
Ah well. Maybe one day when we are actually good at this we'll be able to fix all the gotchas.

src/AI/ModularAI.java
ThatIsAUsername: > putting all that cool logic into higher-priority modules

Yeah, the priority element is pretty interesting, especially if it allows module A to pass, let module B to do something, and then pick an action the next time around.

> Isn't that... the whole point of an abstraction?

Well, yes. And no. But mostly yes. It's good when you don't need that complexity to solve the problem effectively. It's bad when it hides or obscures features that would be useful.
As you say though, an AI builder can simply swap out modules or provide higher-priority ones to handle stuff they care about.

>  How is "defining what I'm gonna do and what my priorities are" not a holistic approach?

Depends on how you decide what you are going to do. Holistic just means to consider the whole problem space when choosing an action. `CaptureFinisher`, for example, currently doesn't look out for threats to itself (though it could if designed that way). It's not very holistic, (and maybe it doesn't need to be, to be effective).

My comment just meant that this modular strategy encourages throwing together a bunch of myopic building blocks that don't know anything about what the others are doing, and might well end up working at cross-purposes if someone isn't careful. That doesn't mean it's a bad approach, it just means care should be taken, which is just... true for any attempt to create a non-trivial AI.

src/AI/ModularAI.java
ThatIsAUsername: Would submodules ever have a reason to remove units from the queue? I guess if they don't like the first unit they could keep popping or whatever. Aight.

src/AI/ModularAI.java
ThatIsAUsername: That sounds more like yes, and basically no (except maybe really indirectly).
I'd rather make the comment unambiguous that we are setting the order in which units are considered for actions within each phase, and preclude the interpretation that we are setting action priorities in any real sense.
Yes, if we prioritize all ASSAULT units, we are implicitly prioritizing attack actions, but we are only explicitly ordering units, so that's what we should call out.

ThatIsAUsername:
So I haven't reviewed Wally's maybe-new logic in his refactored modules in any depth, and I probably won't unless you want me to, but I think I have an understanding of how the ModularAI system works at least.

I'm fine with merging once we resolve the existing conversations.

src/AI/ModularAI.java
ThislsAUsername: Fair enough.

src/AI/ModularAI.java
ThislsAUsername: > Yeah, the priority element is pretty interesting, especially if it allows module A to pass, let module B to do something, and then pick an action the next time around.

It totally does. And it makes your code *prettier* (IMO) when you use it, instead of uglier, now.

> As you say though, an AI builder can simply swap out modules or provide higher-priority ones to handle stuff they care about.

...and this scheme provides a concise way to define your order of operations/priority, so you can move stuff around without huge deltas.
You could do that anyway by factoring everything into methods, but this is easier to look at than that would be IMO.

> Depends on how you decide what you are going to do. Holistic just means to consider the whole problem space when choosing an action.

That's fair; this design is relatively ill-suited for an AI that considers/scores all potential moves before choosing one.
It would still gain a *little* from the inversion-of-control bit, but yeah.
A reasonable extension might be to have a variant that polls an action (and a score for that action) from all modules and picks the highest... but defining the scoring metric sounds pretty hard lol

Such an AI would probably also lag quite a bit more than Wally does, and that's saying something lol

> `CaptureFinisher`, for example, currently doesn't look out for threats to itself (though it could if designed that way). It's not very holistic, (and maybe it doesn't need to be, to be effective).

Yeah. That's fair.
You can get pretty holistic through the combination of fairly myopic parts, though, so long as they're properly designed.

> My comment just meant that this modular strategy encourages throwing together a bunch of myopic building blocks that don't know anything about what the others are doing, and might well end up working at cross-purposes if someone isn't careful. That doesn't mean it's a bad approach, it just means care should be taken, which is just... true for any attempt to create a non-trivial AI.

The way I look at it is it just makes it *easier* to slap your building blocks together, rather than encouraging you to do so haphazardly.
You'll note I also put in a fair amount of pre/post-conditions in comments along the line of "I expect to be called after X has done stuff this round"

This also nudges the door open for possibly letting COs define AI routines for activating/maximizing benefit from their powers, which I'm kind of a fan of - this would probably require more wiring to let the AI tell its modules the sorts of things it values, so e.g. Meridian's module can transform Tanks to Arty in places the AI prefers before COP-ing.

src/AI/WallyAI.java
ThislsAUsername: I can't cache either of those values - they're unique per `MoveType`.
Sea thinks land is invalid, and land thinks sea is.

I realized the gotcha when I was setting up the saving logic, but I didn't want to open the can of worms of saving AI yet.
I think I'm happy to do that at this point since stuff has matured a bit more, though.

src/AI/ModularAI.java
ThislsAUsername: Huh. That's dumb. I figured it was too innocuous to break anything, _sigh_.

ThislsAUsername:
> So I haven't reviewed Wally's maybe-new logic in his refactored modules in any depth, and I probably won't unless you want me to

I did try to split the commits up so that I wasn't moving code around and making semantic changes at the same time, though I'm not convinced I succeeded all that well.
It might be edutaining to look at what I did with evictions. I *wanted* to make it more generic, but ended up with insufficient spoons. ...It may also have partially been that it would make me sad to see Muriel whipping Wally even worse than she already does.

> I'm fine with merging once we resolve the existing conversations.

The one thing I'm kinda iffy on about this whole thing is that I need to use instance members for passing data between modules (and not doing so would require re-calculating stuff).
I thought about trying to set up some sort of generic store in the base class, but I think that would end up being much pain for little gain.

src/AI/ModularAI.java
ThatIsAUsername: > You'll note I also put in a fair amount of pre/post-conditions in comments along the line of "I expect to be called after X has done stuff this round"

I did not see that. I don't see many comments on the AI modules at all. I'm also not sure about having implicit dependencies between modules anyway, but eh, I guess that's up to the AI designer.

> letting COs define AI routines

That sounds like a fun/scary idea. Meridian will finally be relevant!

ThatIsAUsername:
> I wanted to make it more generic, but ended up with insufficient spoons.

Don't think I've heard that spoons metaphor before. Makes me think of the philosopher's table problem, except we aren't doing multithreading here.

> It may also have partially been that it would make me sad to see Muriel whipping Wally even worse than she already does.

Also not sure how a better eviction process would worsen Wally's chances.

> need to use instance members for passing data between modules

You mean the `unownedProperties`, `aiPhases`, etc, variables at the `ModularAI` level? I don't think I see a problem with that.

> generic store in the base class

Remember `GameInputState`? Pepperidge Farm Remembers.

ThislsAUsername:
> Don't think I've heard that spoons metaphor before. Makes me think of the philosopher's table problem, except we aren't doing multithreading here.

https://en.wikipedia.org/wiki/Spoon_theory

> Also not sure how a better eviction process would worsen Wally's chances.

It wouldn't, but Muriel getting that and/or the ModularAI deal would, and making eviction generic would make that easy and probably quite beneficial.

> You mean the `unownedProperties`, `aiPhases`, etc, variables at the `ModularAI` level? I don't think I see a problem with that.

The `threatMap` and such are also instance variables at the `WallyAI` level, which is why those modules require a `WallyAI` rather than a `ModularAI`.

> > generic store in the base class
> 
> Remember `GameInputState`? Pepperidge Farm Remembers.

*distant sobbing*

ThatIsAUsername:
> The threatMap and such are also instance variables at the WallyAI level, which is why those modules require a WallyAI rather than a ModularAI

That just means those modules have to be specific to Wally. Not sure that's necessarily a problem, though it does make it less transferrable. There are ways you could probably fix that too, it would just take more spoons.

src/AI/ModularAI.java
ThislsAUsername: > I did not see that. I don't see many comments on the AI modules at all. I'm also not sure about having implicit dependencies between modules anyway, but eh, I guess that's up to the AI designer.

Hmm. I've definitely had a bit of a shift in philosophy towards comments.
If I can't say something that meaningfully summarizes a fair amount of logic nor something that expresses design intent significantly more clearly than the code... adding another comment to maintain seems like make-work.
Example of this:
```java
  /** Functions as working memory to prevent eviction cycles */
  private transient Set<Unit> evictionStack;
```
I don't repeat the exact "what" of it, since that'd be kind of redundant. Instead, I say why I'm tracking it at all.
I guess I could go into more detail, but I'm not sure that I'd be clearer than just reading the code, and keeping two copies of the algo (one which the compiler doesn't care about) seems silly.

I did at least note `For use after unit building is complete` a few places. I coulda sworn I added in a thing about the threat map; I should do that.

I feel like you're not gonna get away with having no dependency between modules at all, unless you really like duplicating calculations or something.

> > letting COs define AI routines
> 
> That sounds like a fun/scary idea. Meridian will finally be relevant!

Hah. Indeed.
Pretty sure I've even brought it up before a few times, lol

ThatIsAUsername: Yeah, using comments for the "why"--the intent--is probably the way to go. Which your eviction example follows, but not your "use after unit building" example. Why should the module be used only after producing units?

> I feel like you're not gonna get away with having no dependency between modules at all

I mean, you could get clever and accept that data in the interface, and only recalculate it if it isn't given so the module can be generically used whether or not the AI optimizes it properly. But that would be potentially complex to implement, and may or may not be worth it.

ThislsAUsername: > Why should the module be used only after producing units?

Because it doesn't consider whether the unit's moving onto a production building :P 

> I mean, you could get clever and accept that data in the interface, and only recalculate it if it isn't given so the module can be generically used whether or not the AI optimizes it properly. But that would be potentially complex to implement, and may or may not be worth it.

But now the interface is no longer generic - it has to be specialized to the module in question.
...and if it's unspecialized data, you'd have to just assume what it is and cast it to whatever you think your data is and hope it works.

src/AI/WallyAI.java
```diff
@@ -62,7 +62,11 @@ public AIMaker getAIInfo()
   private static final double AGGRO_FUNDS_WEIGHT = 0.9; // Multiplier on damage I need to get before a sacrifice is worth it
   private static final double RANGE_WEIGHT = 1; // Exponent for how powerful range is considered to be
   private static final double TERRAIN_PENALTY_WEIGHT = 3; // Exponent for how crippling we think high move costs are
-  private static final double MIN_SIEGE_RANGE_WEIGHT = 0.8; // Exponent for how much to penalize siege weapon ranges for their min ranges 
+  private static final double MIN_SIEGE_RANGE_WEIGHT = 0.8; // Exponent for how much to penalize siege weapon ranges for their min ranges
+
+  private static final double TERRAIN_FUNDS_WEIGHT = 2.5; // Multiplier for per-city income for adding value to units threatening to cap
+  private static final double TERRAIN_INDUSTRY_WEIGHT = 20000; // Funds amount added to units threatening to cap an industry
```
ThatIsAUsername, line 68: That is... quite the bump up from the normal city weight.

src/AI/WallyAI.java
```diff
@@ -918,14 +922,30 @@ private static int valueUnit(Unit unit, Location locale, boolean includeCurrentH
...
     return value;
   }
 
+  private static int valueTerrain(Commander co, TerrainType terrain)
+  {
+    int value = 0;
+    if( terrain.isProfitable() )
+      value += co.gameRules.incomePerCity * TERRAIN_FUNDS_WEIGHT;
+    if( co.unitProductionByTerrain.containsKey(terrain) )
+      value += TERRAIN_INDUSTRY_WEIGHT;
```
ThatIsAUsername, line 959: Industries generate income too.

src/AI/WallyAI.java
```diff
@@ -918,14 +922,30 @@ private static int valueUnit(Unit unit, Location locale, boolean includeCurrentH
...
+    if( TerrainType.HEADQUARTERS == terrain
+        || TerrainType.LAB == terrain )
+      value += TERRAIN_HQ_WEIGHT;
```
ThatIsAUsername, line 962: In a high-funds game, the HQ might be undervalued. We allow up to 20K, which with the 2.5 multiplier makes a standard city worth more than HQ.

src/AI/WallyAI.java
```diff
@@ -697,6 +699,11 @@ private GameAction evictUnit(
       log(String.format("%s  Eviction cycle! Bailing.", spacing));
       return null;
     }
+    if( evictionStack.size() > EVICTION_STACK_MAX_DEPTH )
+    {
+      log(String.format("%s  Too many units blocking! Bailing.", spacing));
```
ThatIsAUsername, line 708: Does this actually ever become a problem in non-recursive cases (I know you also check recursion separately)?

ThislsAUsername: Pretty sure It's a significant speedup in big unit clogs. Which Wally v Wally *regularly* produces, lol

src/AI/WallyAI.java
ThislsAUsername: It's additive.

src/AI/WallyAI.java
ThislsAUsername: It's additive.

src/AI/WallyAI.java
ThislsAUsername: Production is important, lol

src/AI/WallyAI.java
ThatIsAUsername: Ah. I guess if I actually used those squishy white orbs in my face I would have known that.

ThatIsAUsername:
I guess fire at ~~Wall~~will, unless you have anything else you want to add here.

PR #284, opened 2020-07-26
Fix left sliders

ThislsAUsername:
This was really bugging me lol

### DIFF
--- a/src/UI/Art/Animation/BaseUnitActionAnimation.java
+++ b/src/UI/Art/Animation/BaseUnitActionAnimation.java
@@ -4,9 +4,7 @@ import java.awt.Graphics;
 import java.util.ArrayList;
 
 import Engine.XYCoord;
-import UI.Art.SpriteArtist.SpriteLibrary;
-import UI.Art.SpriteArtist.SpriteMapView;
-import UI.Art.SpriteArtist.UnitSpriteSet;
+import UI.Art.SpriteArtist.*;
 import UI.Art.SpriteArtist.UnitSpriteSet.AnimState;
 import Units.Unit;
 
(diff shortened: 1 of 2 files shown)


src/UI/Art/Animation/BaseUnitActionAnimation.java
```diff
@@ -4,9 +4,7 @@
 import java.util.ArrayList;
 
 import Engine.XYCoord;
-import UI.Art.SpriteArtist.SpriteLibrary;
-import UI.Art.SpriteArtist.SpriteMapView;
-import UI.Art.SpriteArtist.UnitSpriteSet;
+import UI.Art.SpriteArtist.*;
```
ThatIsAUsername, line 7: I'm not sure we approve of this kind of thing around here.

ThislsAUsername: Why not? It means less churn on the imports, and no annoying warnings after you remove the last reference to something... *like that one we have in Unit ;-;*.

ThatIsAUsername: It's generally considered bad practice to import with wildcards. It clutters the namespace with a bunch of stuff you aren't using, makes name collisions more likely, and makes it harder to see what you actually use from the package. Plus I just prefer to be explicit about things.

ThislsAUsername: The clutter is mostly irrelevant... especially in namespaces that *we* define.
Collisions are mostly not a thing we need to worry about, especially if you "be explicit about things" and specify the enclosing scope next to the invocation.

Also, I literally never look at the imports unless the IDE tells me something's wrong with them.
The signal to noise ratio is awful.
Fiddling with that less often is a win in my book.

ThatIsAUsername: Blah fine. Hopefully we don't find reason to regret it.

PR #288, opened 2020-08-05
Buff highlights

### LINKED ISSUE #287: Ability power indication
Right now there's no way to tell if an enemy Commander Ability is active without going to the status screen.
It would be nice to have a visual "buff" indicator so it's clear from the map view that the game is afoot.

ThatIsAUsername:
Fixes #287

### DIFF
--- a/src/UI/Art/SpriteArtist/UnitSpriteSet.java
+++ b/src/UI/Art/SpriteArtist/UnitSpriteSet.java
@@ -2,6 +2,7 @@ package UI.Art.SpriteArtist;
 
 import java.awt.AlphaComposite;
 import java.awt.Color;
+import java.awt.Composite;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
@@ -173,10 +174,31 @@ public class UnitSpriteSet
     return sprites[state.ordinal()].getFrame(imageIndex);
   }
 
+  AlphaComposite buffComposite = null;
+  long lastCompositeCreationTime = 0;
   public void drawUnit(Graphics g, Unit u, AnimState state, int imageIndex, int drawX, int drawY)
   {
+    Graphics2D g2d = (Graphics2D)g;
+    Composite oldComposite = g2d.getComposite();
+
     boolean flipImage = SpriteMapView.shouldFlip(u);
+
+    // Figure out if we need to draw a buff overlay. If so, get some things together.
     boolean drawBuff = AnimState.IDLE == state && !u.CO.getActiveAbilityName().isEmpty();
+    float buffOpacity = 0;
+    if( drawBuff )
+    {
+      // Set opacity as a function of time.
+      long nowTime = System.currentTimeMillis();
+      buffOpacity = (float)(0.3*Math.sin(nowTime/250.) + 0.6);
+
+      // Only regenerate the AlphaComposite object once per timestep.
+      if(lastCompositeCreationTime != nowTime)
+      {
+        lastCompositeCreationTime = nowTime;
+        buffComposite = AlphaComposite.getInstance(AlphaComposite.SRC_OVER, buffOpacity);
+      }
+    }
 
     BufferedImage frame = getUnitImage(state, imageIndex);
     BufferedImage buffFrame = drawBuff ? buffMask.getFrame(imageIndex) : null;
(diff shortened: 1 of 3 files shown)


src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -115,6 +118,9 @@ public UnitSpriteSet(String unitType, Faction faction, ColorPalette coColors)
       }
     }
 
+    // Make a mask that we can draw with varying opacity to indicate buff effects.
+    buffMask = new Sprite(sprites[AnimState.TIRED.ordinal()]);
+    buffMask.convertToMask(new Color(255, 255, 255, 255));
```
ThislsAUsername, line 123: I would argue for `IDLE` over `TIRED`.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -170,19 +176,38 @@ public BufferedImage getUnitImage(AnimState state, int imageIndex)
...
     {
       g.drawImage(frame, drawX - shiftX + (frame.getWidth()), drawY + shiftY, -frame.getWidth(), frame.getHeight(), null);
+      if( drawBuff )
+      {
+        // Set opacity as a function of time.
+        double opacity = 0.3*Math.sin(System.currentTimeMillis()/250.) + 0.6;
+        ((Graphics2D)g).setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, (float)opacity));
+        g.drawImage(buffFrame, drawX - shiftX + (buffFrame.getWidth()),
+            drawY + shiftY, -buffFrame.getWidth(), buffFrame.getHeight(), null);
+        ((Graphics2D)g).setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 1.0f));
```
ThislsAUsername, line 197: What is this casting witchery?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: potato potato. They have the same outline, yeah? I'm just converting it to a mask anyway.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: [This guy](https://stackoverflow.com/questions/19344878/difference-between-graphics-and-graphics2d#19344937) and [this guy](https://jvm-gaming.org/t/graphics-vs-graphics2d/38668/2) summarize it well.

tl:dr; Spooky witchcraft.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: Technically, the user can provide a `TIRED` sprite, because of how we load things.
I don't really... *expect* that, but y'know.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: That's... not really spooky witchcraft.
Do you really *need* to cast it, though?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: Huh. I guess they can. I don't think it used to be that way. Maybe we should just adjust `TIRED`'s `toString` to `''` so it agrees. I guess that would preclude fancy ridiculous tired animations for moved units. Which is definitely a priority feature we are likely to see in our lifetimes.

Sighfinewhatever.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: Yes. `Graphics` doesn't have the composite interface, but `Graphics2D` does. It just so happens that the only other subclass of `Graphics` is `DebugGraphics`, which is used by precisely nobody ever, so we can just cast to the newer `Graphics2D` object and use the cool new stuff.

It just inherits from the old interface for backwards compatibility.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: > Huh. I guess they can. I don't think it used to be that way.

True facts.
> Maybe we should just adjust `TIRED`'s `toString` to `''` so it agrees.

I'm not opposed to this.
> I guess that would preclude fancy ridiculous tired animations for moved units. Which is definitely a priority feature we are likely to see in our lifetimes.

I lol'd

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: I'd lobby perhaps for moving to use `Graphics2D` explicitly in the method signatures, rather than just casting junk and assuming it works.
...then again, I imagine we'd probably still end up casting it at the top level?

ThatIsAUsername: Yes. `getGraphics` doesn't hand out `Graphics2D` objects.

We could still do the cast at the top level and just have it that way everywhere, but that is definitely beyond the scope of this PR.

ThislsAUsername:
Hmm. The effect is a little overwhelming, IMO.
Maybe nudge down the opacity, or restrict the glow to only certain parts of the image? (Say, pixels with a pretty even RGB balance?)

ThislsAUsername:
...and now there are weird flecks anywhere it's not perfect black:
Untitled

ThatIsAUsername:
Hey man, that's what AW does, except they apply it after attaching the HP block, which I don't. Some of the user-created sprites probably don't follow the AW spriting conventions, and you'll also notice that even the proper AW sprites don't always get a full glowy outline (the tops of their heads often get left out).
I'm not opposed to using a more nuanced shading strategy, since that will provide better support for user-created sprites that aren't made to spec, but I don't think that's going to greatly affect clarity either way.

PR #289, opened 2020-08-06
Fixes #285, SEEK behavior.

### LINKED ISSUE #285: SEEK behavior
Currently SEEK will go to the next unit based on creation order. This sometimes results in jumping to units across the map when there are other units right near the cursor where the player's attention presumably is.

It would be far more intuitive for the cursor to jump to the nearest available unit instead. The unit seek-order should not change between back-to-back SEEK inputs so that the player can leave a local unit cluster if desired.

ThatIsAUsername:


src/Engine/MapController.java
```diff
@@ -153,39 +154,39 @@ private void handleFreeTileSelect(InputHandler.InputAction input)
...
+          // First get all active units, sorted.
+          ArrayList<XYCoord> unitLocations = new ArrayList<XYCoord>();
+          unitLocations.addAll(Utils.findLocationsNearUnits(myGame.gameMap, myGame.activeCO.units, 0));
+          unitLocations.removeIf(xy -> myGame.gameMap.getLocation(xy).getResident().isTurnOver);
+          Utils.sortLocationsByDistance(myGame.getCursorCoord(), unitLocations);
+
+          // Find all usable properties, sorted.
+          ArrayList<XYCoord> usableProperties = Utils.findUsableProperties(myGame.activeCO, myGame.gameMap);
+          usableProperties.removeIf(xy -> myGame.gameMap.getLocation(xy).getResident() != null);
+          Utils.sortLocationsByDistance(myGame.getCursorCoord(), usableProperties);
```
ThislsAUsername, line 172: Om nom delicious lambda

ThislsAUsername:
Having played around with this a bit, I think the distance-sorting should take precedence over unit vs industry.

src/Engine/MapController.java
ThatIsAUsername: `Utils.sortLocationsByDistance` is a perfectly standard function, I assure you

ThatIsAUsername:
I... can't say I'm a huge fan of that. I keep getting surprised by a build menu when I'm trying to move something.

src/Engine/MapController.java
ThislsAUsername: I didn't wanna split the code block. I was referring to 171.

ThatIsAUsername: Yes.

ThislsAUsername:
> I... can't say I'm a huge fan of that. I keep getting surprised by a build menu when I'm trying to move something.

IDK, I tried it and it worked nicely IMO. Maybe make it an option?

ThatIsAUsername:
> Maybe make it an option?

Ehhh, I hate to turn everything into an option. I mean, if you really want to we could, but I wouldn't otherwise.

ThislsAUsername:
> Ehhh, I hate to turn everything into an option. I mean, if you really want to we could, but I wouldn't otherwise.

Agreed on that. You strongly prefer production last, I strongly prefer production mixed in... seems like it'd make a decent option.

Issue #290, opened 2020-08-06
The Commander overlay is never visible when spectating AI


ThatIsAUsername:
The Commander overlay is never visible when spectating AI, but it should be, so you can see power level and funds and which commanders are in play.

PR #292, opened 2020-08-07
241 day counter

### LINKED ISSUE #241: Add a day counter
Add some sort of display so players can see how many full rounds of play have passed.

ThatIsAUsername:
Fixes #241

### DIFF
--- a/src/UI/Art/SpriteArtist/SpriteLibrary.java
+++ b/src/UI/Art/SpriteArtist/SpriteLibrary.java
@@ -41,6 +41,7 @@ public class SpriteLibrary
   // Numbers and letters to overlay on map units.
   private static Sprite mapUnitNumberSprites = null;
   private static Sprite mapUnitLetterSprites = null;
+  private static Sprite mapUnitSymbolSprites = null;
   private static Map<Color,Map<Character,BufferedImage>> mapUnitTextSprites = null;
 
   // Icons for various attributes.
@@ -356,6 +357,15 @@ public class SpriteLibrary
       mapUnitLetterSprites = new Sprite(SpriteLibrary.loadSpriteSheetFile("res/unit/icon/alphabet.png"), 8, 8);
     }
     return mapUnitLetterSprites;
+  }
+
+  public static Sprite getMapUnitSymbolSprites()
+  {
+    if( null == mapUnitSymbolSprites )
+    {
+      mapUnitSymbolSprites = new Sprite(SpriteLibrary.createDefaultBlankSprite(8, 8), 8, 8);
+    }
+    return mapUnitSymbolSprites;
   }
 
   public static Map<Character,BufferedImage> getColoredMapTextSprites(Color color)
--- a/src/UI/Art/SpriteArtist/SpriteMapView.java
+++ b/src/UI/Art/SpriteArtist/SpriteMapView.java
@@ -244,12 +244,12 @@ public class SpriteMapView extends MapView
     }
 
     // Keep track of whether we want to draw tile info.
-    boolean hudVisible = false;
+    boolean showTileDetails = false;
 
     // Draw the cursor/our menu if we aren't animating events
     if( currentAnimation == null || !notifyOnAnimEnd )
     {
-      hudVisible = true; // Only draw tile details when not animating.
+      showTileDetails = true; // Only draw tile details when not animating.
       if( getCurrentGameMenu() == null )
       {
         mapArtist.drawCursor(mapGraphics, currentActor, isTargeting, myGame.getCursorX(), myGame.getCursorY());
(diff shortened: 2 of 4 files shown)


ThatIsAUsername:
Probs merge #290 before doing this one, since it builds on that.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -505,12 +505,57 @@ private void drawHUD(Graphics g, boolean includeTileDetails)
...
+
+    // Draw the turn counter.
+    int xDraw = (counterIsRight
+        ? (SpriteOptions.getScreenDimensions().width / SpriteOptions.getDrawScale()) - 2 - turnNumImage.getWidth()
+        : 2);
+    int yDraw = 3;
+
+    // Draw a CO-colored background with the counter.
+    int arcW = turnNumImage.getHeight()+4;
+    g.setColor(UIUtils.getMapUnitColors(myGame.activeCO.myColor).paletteColors[4]); // 0 is darker, 5 is lighter.
```
ThislsAUsername, line 547: Is it intentional to make this a different shade than the CO panel? Pretty sure that one just uses the "signature"/key color

ThislsAUsername:
...uh.
Judging by the state of the branch, maybe we shoulda just thrown out the other branch and merged this one instead.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -358,6 +359,15 @@ public static Sprite getMapUnitLetterSprites()
     return mapUnitLetterSprites;
   }
 
+  public static Sprite getMapUnitSymbolSprites()
+  {
+    if( null == mapUnitSymbolSprites )
+    {
+      mapUnitSymbolSprites = new Sprite(SpriteLibrary.createDefaultBlankSprite(8, 8), 8, 8);
```
ThislsAUsername, line 366: Can't say I'm a super fan of this.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: Yes. The CO panel is a sprite, not procedurally-generated. But also no, since it looks like I ended up selecting the representative color anyway. I can change it back to simplify.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Aww, I was hoping you wouldn't notice.

Whenever we have a need for those sprites, we can add 'em. Until then, they will just show up as blank spaces (just like unsupported characters in the other fonts. This is, in any case, not a regression in functionality.

ThislsAUsername: I say try to load the resource, and fail as normal if it's not there (which it won't be).
At least that way, you get a warning and a black rectangle.

ThislsAUsername:
Hmm. CI and BH still look weird.
Untitled

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: I might rather have an empty space than a black rectangle. I could just put the resource there as a blank sprite I guess.

Not that big a deal either way, since we aren't using any of those characters yet. Whatever you prefer.

ThatIsAUsername:
I don't claim that every color scheme is well-balanced.
I could switch it to use `UIUtils.getMapUnitColors(myGame.activeCO.myColor).paletteColors[5]`, or I could just use the same color for everyone. I kinda like using the team color for the UI though, and I wanted to use a slightly darker color so it wouldn't demand attention as much as the main CO panel. Plus even if it's drawn darker, the number is clear, and folks will know what it says anyway.
I can try pushing the lighter option I guess; figure out what you like.

ThatIsAUsername:
Oh also, apparently the "representative" color for BH is not actually at index 4 in the BH palette. `paletteColors[4]` looks like this:
Image10000

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: I would rather have the loading logic in place, rather than a cleverly-disguised no-op.
Whether you add a resource to load is your choice.

ThislsAUsername:
> Oh also, apparently the "representative" color for BH is not actually at index 4 in the BH palette. `paletteColors[4]` looks like this:

The key colors are not standardized; I picked something that made the UI readableish while being representative.

Either way, I'm not gonna die on this hill. Pick whatever you like.

PR #294, opened 2020-08-09
Removed artifact from image.

ThatIsAUsername:
Could have sworn I had fixed that before.


ThislsAUsername:
I honestly don't see what you've changed here

ThatIsAUsername:
It's a grey line near the left border at about the same height as her right elbow. It's more noticeable in-game.

PR #296, opened 2020-08-09
Buff highlights mod

ThislsAUsername:
Decided to fiddle around with stuff a bit myself.
Based on the fiddling I did, I don't think it's ever gonna look good to use the same buff mask for `IDLE` and `TIRED`.

### DIFF
--- a/src/UI/Art/SpriteArtist/Sprite.java
+++ b/src/UI/Art/SpriteArtist/Sprite.java
@@ -183,29 +183,25 @@ public class Sprite
     }
   }
 
-  public void eraseNonGrey(int rgbDeltaThreshold)
+  public void convertToInverseBrightnessMask(Color maskColor)
   {
     for( BufferedImage bi : spriteImages )
     {
       WritableRaster raster = bi.getRaster();
+      double mask[] = {maskColor.getRed(), maskColor.getGreen(), maskColor.getBlue(), maskColor.getAlpha() };
       for( int x = 0; x < bi.getWidth(); ++x )
       {
         for( int y = 0; y < bi.getHeight(); ++y )
         {
           double[] pixel = new double[4];
           raster.getPixel(x, y, pixel);
-          double R = pixel[0];
-          double G = pixel[1];
-          double B = pixel[2];
-          if(
-            Math.abs(R - G) >= rgbDeltaThreshold ||
-            Math.abs(R - B) >= rgbDeltaThreshold ||
-            Math.abs(G - B) >= rgbDeltaThreshold
-            )
-          {
-            pixel[3] = 0;
-            raster.setPixel(x, y, pixel);
-          }
+          if( pixel[3] == 0 )
+            continue;
+          double scaledBrightness = (pixel[0] + pixel[1] + pixel[2]) / 3;
+          scaledBrightness = Math.min(255, scaledBrightness * 1.5);
+          // Mask alpha == inverse of the pixel brightness
+          mask[3] = 255 - scaledBrightness;
+          raster.setPixel(x, y, mask);
         }
       }
     }
--- a/src/UI/Art/SpriteArtist/Sprite.java
+++ b/src/UI/Art/SpriteArtist/Sprite.java
@@ -182,4 +182,33 @@ public class Sprite
       }
     }
   }
+
+  public void eraseNonGrey(int rgbDeltaThreshold)
+  {
+    for( BufferedImage bi : spriteImages )
+    {
+      WritableRaster raster = bi.getRaster();
+      for( int x = 0; x < bi.getWidth(); ++x )
+      {
+        for( int y = 0; y < bi.getHeight(); ++y )
+        {
+          double[] pixel = new double[4];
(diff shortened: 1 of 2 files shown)


src/UI/Art/SpriteArtist/Sprite.java
```diff
@@ -182,4 +182,29 @@ public void convertToMask(Color maskColor)
...
+      for( int x = 0; x < bi.getWidth(); ++x )
+      {
+        for( int y = 0; y < bi.getHeight(); ++y )
+        {
+          double[] pixel = new double[4];
+          raster.getPixel(x, y, pixel);
+          if( pixel[3] == 0 )
+            continue;
+          double scaledBrightness = (pixel[0] + pixel[1] + pixel[2]) / 3;
+          scaledBrightness = Math.min(255, scaledBrightness * 1.5);
```
ThatIsAUsername, line 201: Why divide by 3 and then multiply by 1.5 instead of just multiplying by 0.5?

ThatIsAUsername:
I'm not sure I see the value of scaling the glowy effect based on the image color, and it also seems completely unnecessary to apply the effect to all of the other unit states; why use extra memory for an effect that will be not be noticed for the 0.5 seconds a unit is in transit to somewhere?
I am also opposed to applying the effect to "tired" units at all. They aren't supposed to draw attention anymore.
Not to mention, you've damped the intensity down so much the poor units look anemic.

src/UI/Art/SpriteArtist/Sprite.java
ThislsAUsername: Clarity.
I'm grabbing an averaged value, then I'm scaling it up - they're semantically different.

ThislsAUsername:
> ~I'm not sure I see the value of scaling the glowy effect based on the image color~, and it also seems completely unnecessary to apply the effect to all of the other unit states; why use extra memory for an effect that will be not be noticed for the 0.5 seconds a unit is in transit to somewhere?

First, the unit animation happens a lot. Small details mean a lot.
Also, the unit runs in place when selected.

> I am also opposed to applying the effect to "tired" units at all. They aren't supposed to draw attention anymore.

That's a fair argument, I suppose. It looks a little weird to me, though.

> Not to mention, you've damped the intensity down so much the poor units look anemic.

It's hard to make a strong visual that isn't too jarring.

> Also, if you are damping the intensity due to it being overwhelming, frankly I think ratcheting up the frequency does more to contribute to overwhelmingness than just making them glow.

I disagree. Having the outline turn pure white is pretty jarring, especially when it stays that way for a reasonably long period.

> At any rate, AW has a stronger glow effect than you have here - I agree it was a bit much to shade the whole unit, but I still wanna see them sparkle.

What do you mean by "sparkle"?

ThatIsAUsername:
> First, the unit animation happens a lot. Small details mean a lot.
Also, the unit runs in place when selected

Fair, but folks don't normally hang out watching the unit run under the cursor. Unit animations happen often, but only briefly. It also looks perfectly fine to not glow when selected (which is what AW does, at any rate - no glowing when running). I don't think it will make a difference either way, and I'd lean towards not eating more memory than needed, but if you are dead set on glowy runners, I'm not going to object too much (because again, nobody's going to notice either way).

> That's a fair argument, I suppose. It looks a little weird to me, though

You must have been super weirded out playing AW for all those years then. Plus, glowy tired units is substantially weird--and distracting--to me.

> I disagree. Having the outline turn pure white is pretty jarring, especially when it stays that way for a reasonably long period.

"Jarring" typically describes an uncomfortable reaction to a sudden change. Maybe you just mean "blinding"? Not sure we are saying the same thing.

> It's hard to make a strong visual that isn't too jarring.

I had the glow varying from 30-90%, you've got it varying from 20-60%. Video evidence from YT plus some tinkering with an image program suggests that AW varies it from 0-75%. Want to meet there? We can keep a higher floor if you want, but a 60% ceiling just isn't cutting it for me.

> What do you mean by "sparkle"?

It just means I want a significant, noticeable effect.

ThislsAUsername:
> I don't think it will make a difference either way, and I'd lean towards not eating more memory than needed, but if you are dead set on glowy runners, I'm not going to object too much (because again, nobody's going to notice either way).

Eh. May as well save memory and be cart accurate.

> You must have been super weirded out playing AW for all those years then. Plus, glowy tired units is substantially weird--and distracting--to me.

/shrug

> > I disagree. Having the outline turn pure white is pretty jarring, especially when it stays that way for a reasonably long period.
> 
> "Jarring" typically describes an uncomfortable reaction to a sudden change. Maybe you just mean "blinding"? Not sure we are saying the same thing.

It just becomes this... mostly-white blob for a while, and it's awkward IMO.

> I had the glow varying from 30-90%, you've got it varying from 20-60%. Video evidence from YT plus some tinkering with an image program suggests that AW varies it from 0-75%. Want to meet there? We can keep a higher floor if you want, but a 60% ceiling just isn't cutting it for me.

0-90%'s just fine; my problem was apparently more with the duration.

> It just means I want a significant, noticeable effect.

I think I've accomplished that.

PR #300, opened 2020-08-14
Victory/Defeat Panels are now correctly positioned even when the wind…

ThatIsAUsername:
…ow is full-screened.

This also lays the groundwork for doing other full-screen animations (for Battles/CO Abilities/Turn-changes, etc).


src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -287,8 +293,25 @@ public void render(Graphics g)
     // Draw the Commander overlay with available funds.
     drawHUD(screenGraphics, showTileDetails);
 
-    // Copy the screen image into the window's graphics buffer.
-    g.drawImage(screenImage, 0, 0, screenImage.getWidth()*drawScale, screenImage.getHeight()*drawScale, null);
+    return screenImage;
```
ThislsAUsername, line 297: Might make sense to just return the map image itself here, and handle putting it in the right place elsewhere.

ThislsAUsername:
So, uh.
This happened when I resized the window after match completion:
uhhh

ThatIsAUsername:
> This happened

Well, yes. It would. I considered that, and decided that "resizing the window after a match ends" is not a very common use case. We could make it dynamically adjust, but I didn't think it was worth the effort.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: It might. I thought about that, and didn't, because by then it was working, and I try to avoid fixing stuff that works.

ThislsAUsername:
I think the bits of polish I mentioned would be nice, but I guess I'm not too bothered if you aren't.

ThatIsAUsername:
I'll fix it if the minions complain.

ThislsAUsername:
Should this not get squash'd since you've got other branches depending on it?

ThatIsAUsername:
Uhhh, yeah, I guess. I mean, it would be easy to fix the "conflicts" that it causes, but not squashing means we don't have to.

No strong preference. I only pushed up the dependent branch to get some early feedback, so it's gonna keep changing either way.

PR #301, opened 2020-08-15
Turn anims

ThatIsAUsername:
Phase 1 of my plan (or Phase 2 if you count the Victory/Defeat panels PR that laid the groundwork for this).

This is theoretically ready to go as is, but some considerations:
- Do we want to have a turn transition for all games, or just e.g. for FOW games so we have a screen blocker?
- Should I require a button-press to remove the screen (or just in FOW games)?
- Would it add anything to have the CO portrait slide across as well?
- Is the effect suitable for its purpose, or would something else be more suited to indicate a turn change?
- Is it disorienting to cover up the map between turns?

I know most of the AW games just have an overlay that floats across (and only use a full-screen thing in FOW games), but I'm not sure how that should look like in DefendPeace. Plus, if we are going to use a full-screen turn transition for FOW games, we may as well re-use it (though whether the BG color is there could be configurable, I suppose).

Thoughts?


ThatIsAUsername:
I might actually like it better like this, at least for the non-FOW case.
Screenshot from 2020-08-15 14-17-45

ThislsAUsername:
For fog specifically (and also games with Subs/Stealths, I guess), I'd want to make it possible (automatic?) to save and exit during the map blank-out. That makes it fair for both players doing save-sharing.

ThatIsAUsername:
> possible (automatic?) to save and exit.

Auto-save/quit would be fairly easy to implement, but it wouldn't protect from accidentally-or-on-purpose re-opening the game. That would also get annoying in hotseat/parsec games (assuming they trusted each other not to look).

We could:
- Auto-save at the end of each turn in MP games
- Have a game option that replaces "end turn" with "save and exit"
- Have an option to allow password-protecting turns
- Just have the splash screen that requires a button-press to get through
- Some other thing I haven't thought of
- Some combination of the above

ThislsAUsername:
> Auto-save/quit would be fairly easy to implement

I even did technically hack that in not too long ago... and lost the code, lol.
It was an awful hack anyway, though.
> but it wouldn't protect from accidentally-or-on-purpose re-opening the game.

This gets into the questions of: Are we actually trying to protect from cheating, or trying to help the user avoid ruining his game for himself?
At the base level, we're just giving them a file that's then completely trusted on the other side. I suppose if (when?) replays happen, we could do some serious verification by making sure we end up back in the same place, but then I'd want to be very careful to make sure we get our floating-point consistency *right*, or risk messing with the normal users' experience. Such a system would be technically breakable, but it'd be a lot harder... assuming there's working turn protection.
> That would also get annoying in hotseat/parsec games (assuming they trusted each other not to look).

Yeah. I made a separate build just for fog PvP when I supported it at all, lol.
I think we'd definitely want to make a game option for it (call it... hotseat mode? play-by-post mode? save sharing mode? save/quit on turn end?) to enable whatever save-sharing protections we want.
>     * Auto-save at the end of each turn in MP games

This could very well work as a separate option, tbh. Being able to say "hmm, weird bug. Dev, have save" without extra friction would be pretty swell.
This could help with AI debugging too, lol.
That said, re-overwriting your one save every turn is kinda a *thing*, so we'd probably want some sort of scheme for that.
>     * Have a game option that replaces "end turn" with "save and exit"

The simplest solution, I think.
>     * Have an option to allow password-protecting turns

I think I might want to avoid this, if possible. I know I don't know how to handle passwords securely, and we'd literally just have people handing their passwords back and forth.
I think maybe the correct thing to do (if we really are serious) is maybe a public/private PGP key thing:
Each player's public key is in the save, and the private key is a separate file on the local machine (ideally in a separate directory, so as to minimize risk of accidental self-pwning).
When it's X's turn, the save's encrypted using X's public key.
As long as the names match up (or maybe we could do better and make an actual game ID lol), we can use the private key to decrypt the save for only the person who has that correct private key.
Private keys could be per game or per user.

- Per game is maximal security, and there's fewer problems if the user loses his private key.
- Per user means the public key could be saved by other people to make games that are fully secured from the start. (This brings up the topic of letting players pick names, which we probably should do eventually anyway)

This neatly evades storing passwords, and is leagues better from a security standpoint.
It *does* add some friction for bug reports, though any protection scheme would (and "I need your private key file for the game" sounds *way* better than "I need your password").
>     * Just have the splash screen that requires a button-press to get through

As noted, we'd want some way to save/quit without closing the splash screen, and probably a little better than "one button press" to get past it, just to prevent accidents.
>     * Some combination of the above

I think "option to replace end turn with s/q" is the MVP, and the PGP key on top is the pie in the sky.

src/UI/Art/SpriteArtist/Backgrounds/HorizontalStreaksBG.java
```diff
@@ -0,0 +1,110 @@
...
+    dist = (dist > dimensions.width)? 0 : dist;
+
+    // Translate the basis point, bringing it back on-screen if needed.
+    for( Streak ss : streaks )
+    {
+      ss.x -= dist;
+      if( ss.x < -ss.w )
+      {
+        ss.x = dimensions.width;
+      }
```
ThislsAUsername: Maybe we need a `Utils.loopIntoRange()`?
We do keep writing snippets like this

ThatIsAUsername:
> re-overwriting your one save every turn is kinda a thing, so we'd probably want some sort of scheme for that

I'd default to just appending the turn number to the save file.

> I know I don't know how to handle passwords securely

Doesn't take anything too crazy: https://www.youtube.com/watch?v=PsIO0gxJF3g

> we'd literally just have people handing their passwords back and forth

What? Why? That would defeat the purpose of having a password at all. If you just mean so that someone can take the first turn for everyone, we could make one of the password settings be "don't set it until the second turn".

> public/private PGP key thing

That is more complex than I want to deal with, especially if it involves keeping a second separate file around. I'd just use a password that can be salted, hashed, and stored directly in the save file.
Granted, if they forget their password, they are SOL, but that's gonna be an issue with the PGP solution also (especially if they are playing from multiple devices).

> letting players pick names, which we probably should do eventually anyway

Do we? I don't think I do.

> This neatly evades storing passwords

I feel like it's more user-friendly for us to just store hashed passwords in the game file than to make them set up a PGP keypair. It might be (arguably?) more secure to go that route, but I don't think we need to reach for the stars, security-wise.

> It does add some friction for bug reports, though any protection scheme would (and "I need your private key file for the game" sounds way better than "I need your password").

I mean, it's kinda the same thing though. Plus, if we really wanted to avoid that friction, it wouldn't be too hard to have a "master password" that only we could use (wouldn't even have to show up in source code) to break saves if folks forgot a password or had an issue to report. Obviously that's a terrible no-good bad idea for any security-sensitive application, but we aren't securing Fort Knox here. Alternatively, we might be able to figure out how to hack the save file directly to remove/alter the password.

> we'd want some way to save/quit without closing the splash screen, and probably a little better than "one button press"

Yeah, that's an issue; "one button press" would only be useful for hiding intelligence in hotseat games, probably. Save/Quit instead of end turn is probably a good approach. I'd do standard passwords with no extra file juggling for that extra layer of security though.

ThislsAUsername:
> I'd default to just appending the turn number to the save file.

Maybe make an "autosaves" directory just to avoid cluttering the save list?
> Doesn't take anything too crazy: https://www.youtube.com/watch?v=PsIO0gxJF3g

Hmm. That's less painful than I'd imagined.
> > we'd literally just have people handing their passwords back and forth
> 
> What? Why? That would defeat the purpose of having a password at all. If you just mean so that someone can take the first turn for everyone, we could make one of the password settings be "don't set it until the second turn".

Well, y'see, the passwords would have to go in the save file (hashed or not, there's representative data). It's a little spoopy to me to be have people distributing that.
> > public/private PGP key thing
> 
> That is more complex than I want to deal with, especially if it involves keeping a second separate file around. I'd just use a password that can be salted, hashed, and stored directly in the save file.

It *would* be a little more seamless for users, though.
Also, since the password doesn't include any kind of encryption of the file itself, a user could technically just change the saved password lol
...which might be a feature, tbh.
> Granted, if they forget their password, they are SOL, but that's gonna be an issue with the PGP solution also (especially if they are playing from multiple devices).

That's fair. That would be a solid argument for (in the PGP thing) a one-per-user thing then, so they can sync it once and stop thinking about it.
> > letting players pick names, which we probably should do eventually anyway
> 
> Do we? I don't think I do.

...It would make it easier for the users to remember/recognize save names, probably, but I suppose it's not critical.
The thing I was thinking was to have it facilitate "friends", so you could have fully-secured games from the start, but I guess that doesn't really make all that much sense either. (I think it'd be easiest, with passwords or PGP keys, to just make it a menu option in MP games)
> I feel like it's more user-friendly for us to just store hashed passwords in the game file than to make them set up a PGP keypair. It might be (arguably?) more secure to go that route, but I don't think we need to reach for the stars, security-wise.

We could do the PGP key thing pretty much entirely automatically.
User1 selects save-shared game. User1 does his turn, hits the "fingerprint army" option, ends turn/saves (unencrypted).
If there's no keypair set yet, the game generates one and slaps the public one into the save, and both into a local keyfile.
User2 opens (unencrypted) game on his turn, hits the "fingerprint army" (or w/e) option, does his turn, and ends turn/saves (encrypted with User1's public key).
If there's no keypair set yet, the game generates one and slaps the public one into the save, and both into a local keyfile.
User1 can now open the save, and User2 can't. But that's okay, because it's User1's turn now.
The relationship flips after the next save, and nobody has to worry about anything but multi-device... and in that case all you need is to copy the private keyfile to all devices and not worry about it again.

It'd be possible to compromise this scheme also by hacking the game directly, but I don't see a way to fix *that* without making a server and stuff.

> I mean, it's kinda the same thing though.

Not quite. The user can't be reusing a password from something else if it's a PGP key that we generated, lol.
> Plus, if we really wanted to avoid that friction, it wouldn't be too hard to have a "master password" that only we could use (wouldn't even have to show up in source code) to break saves if folks forgot a password or had an issue to report.

That sounds... difficult to set up?
Much easier to just run the game and change the password check to... just not check lol.
This is why I was saying not actually having good security might be a "feature".

> Alternatively, we might be able to figure out how to hack the save file directly to remove/alter the password.

Yeah, probs. Just load it, hit breakpoint, set password to null, boom, password reset. Or figure out how to do it with a hex editor, but that sounds like work lol.
> Yeah, that's an issue; "one button press" would only be useful for hiding intelligence in hotseat games, probably. Save/Quit instead of end turn is probably a good approach. I'd do standard passwords with no extra file juggling for that extra layer of security though.

Hmm. I suppose implementing either would be pretty good experience, lol.

ThatIsAUsername:
> It would be a little more seamless for users, though.

Hmm, yeah. Instead of a "use passwords" option, a generic "Use secure turns" option (or however we want to name it) could require zero extra effort on their part (assuming single-device users). Agreed, that would be more seamless, but slightly less flexible.

Ooh, except maybe not. I know they often start a game by having one person do the first turn for everyone; a PGP keyfile approach might scuttle that. I was thinking we could get around that by having an option to not ask for the password until the second turn. I guess it would be possible to do something similar with the PGP approach.

Hah, we could even just add an "auto first turn" option so they don't even have to manually build the first round of infantry.

> The user can't be reusing a password from something else if it's a PGP key that we generated

That's a valid point. I don't really want someone's Facebook password or whatever. That said, we wouldn't know what the passwords are if we salt/hash them before storing (though of course they don't know what we can/can't do or see).

> Much easier to just run the game and change the password check to... just not check lol

True. I guess to keep the save states safe from _us_ we'd have to encrypt the whole game state using their hashed password. I don't think i care to get that serious (plus debugging is nice).

> Hmm. I suppose implementing either would be pretty good experience, lol

ORRR, we could implement both. Password-protected saves to support playing from multiple devices, PGP keyfile for those who want the zero-effort approach.

ThislsAUsername:
> Ooh, except maybe not. I know they often start a game by having one person do the first turn for everyone; a PGP keyfile approach might scuttle that. I was thinking we could get around that by having an option to not ask for the password until the second turn. I guess it would be possible to do something similar with the PGP approach.

My thinking was that we'd just add another option (for remote games) to the meta actions menu to secure your turns.
So game setup -> save sharing mode ON -> end turn becomes save/quit
IFF save sharing mode ON -> add "use secure turns" option to meta actions (and we could have the option to toggle it off too, to give us the save to debug with)
> Hah, we could even just add an "auto first turn" option so they don't even have to manually build the first round of infantry.

Doesn't work IMO because of the FTA counter inf/predeployed transports/etc.
> That's a valid point. I don't really want someone's Facebook password or whatever. That said, we wouldn't know what the passwords are if we salt/hash them before storing (though of course they don't know what we can/can't do or see).

Correct. I was thinking we may as well salt *and* pepper because why not, but asking for a password at all is kinda spoopy IMO.
> True. I guess to keep the save states safe from _us_ we'd have to encrypt the whole game state using their hashed password. I don't think i care to get that serious (plus debugging is nice).

Yeah. The PGP keypair does not have that particular nicety.
> ORRR, we could implement both. Password-protected saves to support playing from multiple devices, PGP keyfile for those who want the zero-effort approach.

Weren't you the one who was just pushing for *fewer* game options?
That said, I don't really disagree because that sounds hilarious.

ThatIsAUsername:
> IFF save sharing mode ON -> add "use secure turns" option to meta actions (and we could have the option to toggle it off too, to give us the save to debug with)

I'd just have "secure turns" be a game-setting option too, and not mess with it in-game. Changing "end turn" to "save/quit" is still valid though.

> Weren't you the one who was just pushing for fewer game options?

Yes, though that is technically one more option for a hypothetically already-existing "security" option. Technically.

> that also means we get to say we've implemented two-factor authentication

I think that's only true if we require both a password _and_ the key file to get into the game. I think we are currently discussing two separate single-factor authentication schemes (that happen to use different factors).

ThislsAUsername:
> I'd just have "secure turns" be a game-setting option too, and not mess with it in-game. Changing "end turn" to "save/quit" is still valid though.

Problem there is: there's no obvious point to input a password/set the key. With this, I can freely take 0-n turns for everyone, then set protection on my turn before sending it off so other people can take their turns and turn on their own protections.
Also, letting you turn it off completely is good for bug reports, especially if we go the PGP key route.
> Yes, though that is technically one more option for a hypothetically already-existing "security" option. Technically.
> I think that's only true if we require both a password _and_ the key file to get into the game. I think we are currently discussing two separate single-factor authentication schemes (that happen to use different factors).

My impression was that they'd be entirely separate.
You can set your encryption key, and (if present) it's used to encrypt the save during your turn.
You can set your password, and you have to enter that to take your turn once the save is loaded.
So if both are true:
Attempt to load save -> it's gibberish.
Attempt to decrypt -> it's now valid savedata.
Notice user has password set -> prompt for password.

> EDIT: I take it back, this flawless plan is [Foolproof](https://mintmobileplus.com/).

???

ThatIsAUsername:
> Problem there is: there's no obvious point to input a password/set the key.

I don't think anybody will want to take more than n=1 turns before enabling turn protection, so I would probably just give the "Set Passwords?"/"Enable Security" setting options of "No", "Turn 1", and "Turn 2", and prompt/create at that point.
PGP could be a separate setting or a sub-option of the "Security" setting - I'd advocate for ignoring the PGP key until the second turn in either case, for fully-transparent (e.g. no extra in-game button presses) operation.

> My impression was that they'd be entirely separate.

Yeah. I guess there's no reason we couldn't. The truly paranoid can just turn on all the things!

> Foolproof

I was just reminded of that by your "flawless" comment. It's the best new streaming site that lets you watch all of your favorite movies, as long as your favorite movie is Foolproof. Because Ryan Reynolds owns a good chunk of Mint Mobile Wireless and decided they needed a streaming service like all the big networks. And because his sense of humor is damaged (or visionary?).

ThislsAUsername:
> I don't think anybody will want to take more than n=1 turns before enabling turn protection

Valid.
> so I would probably just give the "Set Passwords?"/"Enable Security" setting options of "No", "Turn 1", and "Turn 2", and prompt/create at that point.

That is simpler, but it doesn't allow for asymmetric security (unless, I suppose, it's a game-client decision how much to apply?).
> PGP could be a separate setting or a sub-option of the "Security" setting - I'd advocate for ignoring the PGP key until the second turn in either case, for fully-transparent (e.g. no extra in-game button presses) operation.

That does mean it'll be impossible to give us saves for debugging without also providing the private keyfile, though.
> Yeah. I guess there's no reason we couldn't. The truly paranoid can just turn on all the things!

Always!
> I was just reminded of that by your "flawless" comment. It's the best new streaming site that lets you watch all of your favorite movies, as long as your favorite movie is Foolproof. Because Ryan Reynolds owns a good chunk of Mint Mobile Wireless and decided they needed a streaming service like all the big networks. And because his sense of humor is damaged (or visionary?).

That's... Pretty dank.

ThatIsAUsername:
> asymmetric security

As in, one player has a password and the other doesn't? I wasn't planning on making that an option.

> That does mean it'll be impossible to give us saves for debugging without also providing the private keyfile, though

Couldn't we just... comment out the code that validates the keyfile, as you said?

> dank

Yes. That is the word.

ThislsAUsername:
> As in, one player has a password and the other doesn't? I wasn't planning on making that an option.

Bob only plays on his home computer. George has a home PC and a laptop. They both want security.
Bob takes the PGP key, because it's no fuss and strong protection.
George wants the password, because he can't be bothered to figure out transferring the keyfile.
> EDIT: Or did you just mean assymetric encryption? I'm not sure why automating the process would make that infeasible either though. I also don't know that we need to go full PGP to get hands-off security working.

Okay, but this whole thing is "because we can", so...
> Couldn't we just... comment out the code that validates the keyfile, as you said?

No, because I was talking about actually encrypting the save with the PGP key. I guess you could encrypt and re-decrypt it to check if you got the same output as the original file, but...
To be clear:
PGP keys would actually encrypt the save and make it more-or-less state of the art secure.
Passwords would just put a runtime check in because eh.

ThatIsAUsername:
> Asymmetric

Ah, gotcha. That sounds more complicated to implement, especially on the UI side. I don't think I want to offer that option. I think we can take a seamless approach and tell them to drag the keyfile around if they want to. That way we don't have to worry about getting some kid's AOL password, and they don't need extra steps to type one in.

> Okay, but this whole thing is "because we can", so...

Yes. But A cursory look indicates proper PGP encryption would require adding dependencies and be more complicated to set up, so I'm not sure I want to go that route just yet.

ThislsAUsername:
> I think we can take a seamless approach and tell them to drag the keyfile around if they want to. That way we don't have to worry about getting some kid's AOL password, and they don't need extra steps to type one in.

Okay. Fair enough. Would it just be the same key in the game and out of game?

> Yes. But A cursory look indicates proper PGP encryption would require adding dependencies and be more complicated to set up, so I'm not sure I want to go that route just yet.

You know, I hear the Soviets are already a third of the way through their bag of [pinecones](https://what-if.xkcd.com/124/)...

ThatIsAUsername:
> Okay. Fair enough. Would it just be the same key in the save and in the local file?

Nah. Then someone could find the key by inspecting the save. I'd probably do [something more like this](https://github.com/ThislsAUsername/DefendPeace/pull/303)

PR #303, opened 2020-08-19
WIP Security

ThatIsAUsername:
Here's a first stab at things.

Potential improvements: Touch up the outro text to be less blunt
Fix the spacing on the game-setup-options screen
Instead of a bool, have an enum with Security levels: Off, Save/Exit on turn end, pass-protect, PGP-encrypted saves, to suit every taste
Double-check that behavior is consistent/desirable for different configurations

The guts of the thing are working though; at the end of the second turn (turn 1) it creates a passfile, and then only someone with that passfile can take moves for that player.


src/CommandingOfficers/Commander.java
```diff
@@ -463,6 +468,42 @@ private void readObject(ObjectInputStream stream) throws IOException, ClassNotFo
...
+  /** Return true if this Commander is password-protected, false else. */
+  public boolean hasPassword()
+  {
+    return password != null;
+  }
+
+  /** Salt the provided string with the Commander's salt and return the result. */
+  private UUID hashPass(String pass)
+  {
+    int numBytes = 8 + (2*pass.length());
```
ThislsAUsername, line 480:
```suggestion
    int numBytes = Long.BYTES + (Character.BYTES*pass.length());
```

src/CommandingOfficers/Commander.java
```diff
@@ -463,6 +468,42 @@ private void readObject(ObjectInputStream stream) throws IOException, ClassNotFo
...
+    ByteBuffer bb = ByteBuffer.allocate(numBytes);
+    bb.putLong(passSalt).put(pass.getBytes());
+    UUID hashedPass = UUID.nameUUIDFromBytes(bb.array());
+    return hashedPass;
+  }
+
+  /** Return true if this Commander is password-protected, false else.
+   * Throws an exception if a password has already been set. */
+  public void setPassword(long salt, String pass)
```
ThislsAUsername, line 489: Doc mismatch

src/Engine/GameInstance.java
```diff
@@ -39,21 +39,24 @@
   private Weathers defaultWeather;
 
   private GameScenario gameScenario;
+  boolean isSecurityEnabled;
```
ThislsAUsername, line 42: I feel like this and probably a few other things (fog, default weather, etc?) could move to `GameRules` now that that exists.
This has the benefit of not needing to duplicate the fog value for `MapWindow`, which means we can easily update it during the course of the game.

That said, I think I'm fine leaving this as is for now?

src/Engine/PasswordManager.java
```diff
@@ -0,0 +1,102 @@
...
+      // Generate a new salt/passfile and set cmdr's password.
+      long salt = new Random().nextLong();
+      if( storedPass.isEmpty() )
+      {
+        System.out.println("Generating new password");
+        storedPass = UUID.randomUUID().toString();
+      }
+      System.out.println("Setting password.");
+      cmdr.setPassword(salt, storedPass);
+      writePassfile(storedPass);
```
ThislsAUsername, line 43: I see no reason to rewrite the passfile when you're not changing it?

src/Engine/PasswordManager.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+      catch(FileNotFoundException fnfe)
+      {
+        System.out.println("Somehow we failed to find the passfile after checking that it exists!");
+      }
+      catch( InputMismatchException ime )
+      {
+        System.out.println("Encountered an error while parsing passfile!");
+      }
+    } // ~if file exists
```
ThislsAUsername, line 98: Could just bail instead of having a block :eyes:

src/CommandingOfficers/Commander.java
```diff
@@ -463,6 +468,42 @@ private void readObject(ObjectInputStream stream) throws IOException, ClassNotFo
...
+    return password != null;
+  }
+
+  /** Salt the provided string with the Commander's salt and return the result. */
+  private UUID hashPass(String pass)
+  {
+    int numBytes = 8 + (2*pass.length());
+    ByteBuffer bb = ByteBuffer.allocate(numBytes);
+    bb.putLong(passSalt).put(pass.getBytes());
+    UUID hashedPass = UUID.nameUUIDFromBytes(bb.array());
```
ThislsAUsername, line 483: This apparently uses MD5, so I'd complain if these were *user* passwords.
But they're not, so whatever.
(https://stackoverflow.com/questions/29059530/is-there-any-way-to-generate-the-same-uuid-from-a-string)

ThislsAUsername:
> Fix the spacing on the game-setup-options screen
IMO the spacing is fine; may as well just give it the same center-on-selected-thingy logic, since I suspect we're gonna want at least a few more options at some point or other.

src/CommandingOfficers/Commander.java
ThatIsAUsername: pasta

src/Engine/GameInstance.java
ThatIsAUsername: I mean... maybe. I feel like I considered that, but "need for passwords" sounds more like an attribute of the game, rather than a rule of engagement. It could go either place though, I'm sure. Maybe it will make more sense to move it once `GameRules` (or even `GameScenario`) are a little more mature.

src/Engine/PasswordManager.java
ThatIsAUsername: Me neither.

src/Engine/PasswordManager.java
ThatIsAUsername: Hokay.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Hm. That's a cool utility, almost might be worth using just to have it around. Though if we really want to turn up the security, the right answer might be to add the PGP option later and do it proper.

src/Engine/GameInstance.java
ThislsAUsername: Agreed, honestly.
I think fog and weather should move, but not in this PR.

src/Engine/MapController.java
```diff
@@ -56,8 +56,8 @@ public MapController(GameInstance game, MapView view, boolean initGame)
     isGameOver = false;
     nextSeekIndex = 0;
 
-    if( initGame )
-      // Start the first turn.
+    // Start the first turn (or the next one if loading a protected save.
+    if( initGame || myGame.requireInitOnLoad() )
```
ThislsAUsername, line 60: Why not just throw out `initGame` and make `currentTurnEnded` default to false?

src/Engine/MapController.java
```diff
@@ -56,8 +56,8 @@ public MapController(GameInstance game, MapView view, boolean initGame)
     isGameOver = false;
     nextSeekIndex = 0;
 
-    if( initGame )
-      // Start the first turn.
+    // Start the first turn (or the next one if loading a protected save.
```
ThislsAUsername, line 59:
```suggestion
    // Start the first turn (or the next one if loading a protected save).
```

src/Engine/MapController.java
ThatIsAUsername: Because then it wouldn't call `startNextTurn()` to initialize the first turn when starting a new game... if I'm understanding you correctly.

ThislsAUsername: Sure it would. `currentTurnEnded` is false, so `requireInitOnLoad()` returns true.
The only call that uses this constructor is for loading a save from the main menu.

ThatIsAUsername: `currentTurnEnded` defaults to false, yes, but that means `requireInitOnLoad()` returns false also. Plus, this constructor is called with `initGame==true` by the other `MapController` constructor.

ThislsAUsername: > `currentTurnEnded` defaults to false, yes, but that means `requireInitOnLoad()` returns false also. Plus, this constructor is called with `initGame==true` by the other `MapController` constructor.

...right. Derp.
So you should default it to *true*, so it effectively replaces the `initGame==true`.
We then need only one constructor, with the lower number of params.

ThatIsAUsername: Yeah, I think that makes sense. Probably should have made that leap on my own. I blame after midnight and being distracted by making the config options screen scroll.

src/UI/Art/SpriteArtist/GameOptionSetupArtist.java
```diff
@@ -113,7 +116,7 @@ public static void drawGameOptions(Graphics g, GameOptionSetupController control
 
     // Set up some initial parameters.
     int xDraw = (optionsImage.getWidth() / 2) - (graphicsOptionWidth / 2);
-    int yDraw = graphicsOptionHeight;
+    int yDraw = yDrawStart.geti();
     int firstOptionY = yDraw; // Hold onto this to draw the selector arrows.
     int ySpacing = (graphicsOptionHeight + (optionNamePanel.getHeight() / 2));
```
ThislsAUsername, line 121:
```suggestion
    int ySpacing = (int)(graphicsOptionHeight * 1.5);
```

src/UI/Art/SpriteArtist/GameOptionSetupArtist.java
```diff
@@ -131,6 +134,19 @@ public static void drawGameOptions(Graphics g, GameOptionSetupController control
...
+    if( yDrawDest > optionsImage.getHeight() - graphicsOptionHeight )
+    {
+      int offset = -2*graphicsOptionHeight + optionsImage.getHeight() - yDrawDest;
+      yDrawStart.set(yDrawStart.geti() + offset );
+    }
+    if( yDrawDest < graphicsOptionHeight )
+    {
+      int offset = graphicsOptionHeight - yDrawDest;
+      yDrawStart.set(yDrawStart.geti() + offset );
+    }
```
ThislsAUsername, line 148:
```suggestion
    int yDrawDest = yDrawStart.getDestination() + (int) (ySpacing * selectedGameOption.getDestination()) + (3);
    if( yDrawDest > optionsImage.getHeight() - graphicsOptionHeight )
    {
      int offset = -ySpacing + optionsImage.getHeight() - yDrawDest;
      yDrawStart.set(yDrawStart.geti() + offset);
    }
    if( yDrawDest < graphicsOptionHeight )
    {
      int offset = graphicsOptionHeight - yDrawDest;
      yDrawStart.set(yDrawStart.geti() + offset);
    }
```

ThislsAUsername:
We don't have punctuation on the map unit font, which creates black squaaares

ThislsAUsername:
Also, you can get a brief glimpse of your opponent's vision in fog by re-opening the game when it's not your turn

ThislsAUsername:
I'm getting a null ref exception when I try to load a saved fog game (the turn init animation doesn't have a location)
`Thread [AWT-EventQueue-0] (Suspended (breakpoint at line 114 in GameInstance))	
	GameInstance.setCursorLocation(XYCoord) line: 114	
	SpriteMapView.loadNextEventAnimation() line: 162	
	SpriteMapView.animate(GameEventQueue) line: 131	
	MapController.startNextTurn() line: 598	
	MapController.<init>(GameInstance, MapView) line: 57	
	MainUIController.handleSaveSelectMenuInput(InputHandler$InputAction) line: 228	
	MainUIController.handleInput(InputHandler$InputAction) line: 73	
	Driver.keyPressed(KeyEvent) line: 115	
	JFrame(Component).processKeyEvent(KeyEvent) line: 6589	
	JFrame(Component).processEvent(AWTEvent) line: 6408	
	JFrame(Container).processEvent(AWTEvent) line: 2263	
	JFrame(Window).processEvent(AWTEvent) line: 2049	
	JFrame(Component).dispatchEventImpl(AWTEvent) line: 5007	
	JFrame(Container).dispatchEventImpl(AWTEvent) line: 2321	
	JFrame(Window).dispatchEventImpl(AWTEvent) line: 2772	
	JFrame(Component).dispatchEvent(AWTEvent) line: 4839	
	DefaultKeyboardFocusManager(KeyboardFocusManager).redispatchEvent(Component, AWTEvent) line: 1950	
	DefaultKeyboardFocusManager.dispatchKeyEvent(KeyEvent) line: 870	
	DefaultKeyboardFocusManager.preDispatchKeyEvent(KeyEvent) line: 1139	
	DefaultKeyboardFocusManager.typeAheadAssertions(Component, AWTEvent) line: 1009	
	DefaultKeyboardFocusManager.dispatchEvent(AWTEvent) line: 835	
	JFrame(Component).dispatchEventImpl(AWTEvent) line: 4888	
	JFrame(Container).dispatchEventImpl(AWTEvent) line: 2321	
	JFrame(Window).dispatchEventImpl(AWTEvent) line: 2772	
	JFrame(Component).dispatchEvent(AWTEvent) line: 4839	
	EventQueue.dispatchEventImpl(AWTEvent, Object) line: 772	
	EventQueue$4.run() line: 721	
	EventQueue$4.run() line: 715	
	AccessController.doPrivileged(PrivilegedAction<T>, AccessControlContext) line: not available [native method]	
	ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(PrivilegedAction<T>, AccessControlContext, AccessControlContext) line: 85	
	ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(PrivilegedAction<T>, AccessControlContext) line: 95	
	EventQueue$5.run() line: 745	
	EventQueue$5.run() line: 743	
	AccessController.doPrivileged(PrivilegedAction<T>, AccessControlContext) line: not available [native method]	
	ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(PrivilegedAction<T>, AccessControlContext, AccessControlContext) line: 85	
	EventQueue.dispatchEvent(AWTEvent) line: 742	
	EventDispatchThread.pumpOneEventForFilters(int) line: 203	
	EventDispatchThread.pumpEventsForFilter(int, Conditional, EventFilter) line: 124	
	EventDispatchThread.pumpEventsForHierarchy(int, Conditional, Component) line: 113	
	EventDispatchThread.pumpEvents(int, Conditional) line: 109	
	EventDispatchThread.pumpEvents(Conditional) line: 101	
	EventDispatchThread.run() line: 90	
`

ThatIsAUsername:
Yeah, my silly AI observation change is breaking on the turn init animation. Should be fixed now, along with the fog-peeking exploit.

ThislsAUsername:
We done here?

ThatIsAUsername:
I'd like to format the turn-end text so it doesn't run off-screen so eagerly.
The underscores in the filename are also not being drawn - I assume we should probably fix that too.

ThislsAUsername:
> I'd like to format the turn-end text so it doesn't run off-screen so eagerly.

Fair.
> The underscores in the filename are also not being drawn - I assume we should probably fix that too.

I am very explicitly not worried about that. We (or at least I) depend on that behavior in other places.

ThatIsAUsername:
> We (or at least I) depend on that behavior in other places.

For the small caps font we do (action names in menus for example), but not for the bold font anyway. I'd rather be as correct as possible in dealing with users.
Alternatively, we could change the save format to use a 'v' instead of a '_' I guess.

ThislsAUsername:
> For the small caps font we do (action names in menus for example), but not for the bold font anyway.

I would rather avoid desynching our symbols files :P
> Alternatively, we could change the save format to use a 'v' instead of a '_' I guess.

I am a-ok with this.

ThatIsAUsername:
Wouldn't have to desynch. Could just add another transparent block on the ones that "don't support" the underscore.
Downside of using a 'v' is that it is all caps, so it ends up being e.g. STRONGVPATCH and is annoying to read.

ThatIsAUsername:
We could also, theoretically, fix the string representations of the enum values to not have an underscore, come to think of it.

ThislsAUsername:
> Wouldn't have to desynch. Could just add another transparent block on the ones that "don't support" the underscore.

I... suppose.
> Downside of using a 'v' is that it is all caps, so it ends up being e.g. STRONGVPATCH and is annoying to read.

Can fix that too, lol.
Other punctuation (dot/dash?) would work too.

> We could also, theoretically, fix the string representations of the enum values to not have an underscore, come to think of it.

I... that's true. I don't want to track down everywhere I use underscores in the UI right now, though

src/Engine/GameEvents/TurnInitEvent.java
```diff
@@ -14,14 +17,21 @@
...
   int turn;
   boolean opaque;
-  String msg;
+  Collection<String> msg;
 
   public TurnInitEvent(Commander co, int turnNum, boolean hideMap)
   {
-    this(co, turnNum, hideMap, "Turn "+turnNum);
+    this(co, turnNum, hideMap, new ArrayList<String>());
+    msg.add("Turn "+turnNum);
```
ThislsAUsername: This didn't need to change, though?

ThatIsAUsername: Hah, yeah. I changed this before I re-added the "string" constructor. Oh well.

src/Test/TestSaveLoad.java
```diff
@@ -62,7 +62,7 @@ private boolean testSaveLoad()
     String path = game.writeSave(false);
     File file = new File(path);
     testPassed &= validate(file.exists(),  "    The file we just saved to doesn't exist");
-    testPassed &= validate(GameInstance.isSaveCompatible(path).length() == 0,  "    We are incompatible with the save we just made");
+    testPassed &= validate(GameInstance.getSaveWarnings(path).length() == 0,  "    We are incompatible with the save we just made");
```
ThislsAUsername, line 65: Idiomatic

ThatIsAUsername: The length of the warnings is zero. Perfectly idiomatic.

PR #304, opened 2020-08-24
Seek option

ThislsAUsername:
I broke the previous graphics options selection; I could avoid breaking it at the cost of more code, but I think I like minimal lines.
We also have a consistent framework for saving options super easily, so yay on that score also.

The way I hacked together the display of the new options (hooking the two artists together) is probably not optimal; I really don't have a solid idea of how these classes should be set up.

But hey, it works.


src/UI/InputHandler.java
```diff
@@ -330,25 +325,20 @@ public boolean handleInput(InputAction action)
...
-        }
-        catch( IOException ioe )
-        {
-          System.out.println("Error! Failed to save key bindings file!.\n  " + ioe.toString());
+          // substring to remove the first space
+          configItems.add(Pair.from(key, val.substring(1)));
         }
+        ConfigUtils.writeConfigItems(KEYS_FILENAME, configItems);
+        if( !ConfigUtils.writeConfigs(OTHER_INPUT_FILENAME, Arrays.asList(extraOptions)) )
+          System.out.println("Unable to write extra game input options to file.");
```
ThatIsAUsername, line 341: Why check the return value for one of these but not the other?

ThislsAUsername: No reason, just lazy.

src/UI/Art/SpriteArtist/ControlOptionsSetupArtist.java
```diff
@@ -23,22 +24,33 @@
...
   public static void draw(Graphics g, InputHandler control)
   {
     // Draw a fancy background.
     DiagonalBlindsBG.draw(g);
 
     myControl = control;
+    if( null == context )
+    {
+      context = new OptionSelectionContext();
+      SpriteOptions.initialize(InputHandler.extraOptions, context);
```
ThatIsAUsername, line 38: Feels like `OptionSelectionContext` should be broken out into its own class/file, and probably have its own `initialize()` call instead of having this artist call that artist to initialize this artist's stuff.
I'd also change the name of the new class to like, `OptionPanelTemplate`/`Configuration` or something, I dunno.

ThatIsAUsername:
So it works, and the design is sensible enough, but the aesthetics are moderately horrifying (not that we score high marks there anyway, but still. Since the SEEK option is a different "kind" of thing than the keybindings, maybe we could put it on a separate tab within the control-options page, the same way other games separate families of configurable options.

This would require implementing a mechanism for changing tabs, but should fix the panel mismatch thing and create a home for future options.

src/UI/InputHandler.java
```diff
@@ -1,22 +1,26 @@
...
 import Engine.IController;
 import Engine.OptionSelector;
+import Engine.Pair;
 
 public class InputHandler implements IController
 {
   private static final String KEYS_FILENAME = "res/keys.txt";
+  private static final String OTHER_INPUT_FILENAME = "res/other_input.txt";
+
+  public static GameOptionBool seekBuildingsLastOption = new GameOptionBool("Seek Buildings Last", true);
```
ThatIsAUsername, line 19: Still not sure this is the best phrasing. Maybe "Seek Units First", or "SEEK Prioritizes Units", but that's a little verbose.

Or maybe have the item name as "SEEK", and the options be "Units then Buildings" or "Units with Buildings", and maybe even have a "Units Only" option for the real old-skool folks (though I don't really see a need for that last one).

ThislsAUsername:
> So it works, and the design is sensible enough, but the aesthetics are moderately horrifying (not that we score high marks there anyway, but still.

Heh.
> Since the SEEK option is a different "kind" of thing than the keybindings, maybe we could put it on a separate tab within the control-options page, the same way other games separate families of configurable options.

Honestly, I'd just add a new menu entirely rather than complicating what we have already.
> This would require implementing a mechanism for changing tabs, but should fix the panel mismatch thing and create a home for future options.

New menu does this too.

PR #307, opened 2020-09-05
Inf tweaks

ThatIsAUsername:


ThatIsAUsername:
It always bothered me that the cursor would cling to the edges once you scroll past the center of the group, since you couldn't see more than a row in the direction you were trying to travel.
This should make it easier to see what's around the cursor while panning.
Also removed the warning spam, because yeesh.

PR #308, opened 2020-09-05
Overlays

ThislsAUsername:
This isn't quite done yet - I need to add the actual threat overlays. But the UI plumbing is there.

### DIFF
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -280,13 +280,14 @@ public class AIUtils
     ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeOccupiedDestinations);
     for( WeaponModel wep : unit.model.weapons )
     {
-      if( wep.getDamage(target) > 0 )
+      double damage = (null == target)? 1 : wep.getDamage(target);
+      if( damage > 0 )
       {
         if( !wep.canFireAfterMoving )
         {
           for (XYCoord xyc : Utils.findLocationsInRange(gameMap, origin, wep.minRange, wep.maxRange))
           {
-            double val = wep.getDamage(target) * unit.getHPFactor();
+            double val = damage * unit.getHPFactor();
             if (shootableTiles.containsKey(xyc))
               val = Math.max(val, shootableTiles.get(xyc));
             shootableTiles.put(xyc, val);
@@ -298,7 +299,7 @@ public class AIUtils
           {
             for (XYCoord xyc : Utils.findLocationsInRange(gameMap, dest, wep.minRange, wep.maxRange))
             {
-              double val = wep.getDamage(target) * unit.getHPFactor();
+              double val = damage * unit.getHPFactor();
               if (shootableTiles.containsKey(xyc))
                 val = Math.max(val, shootableTiles.get(xyc));
               shootableTiles.put(xyc, val);
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -280,14 +280,14 @@ public class AIUtils
     ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeOccupiedDestinations);
     for( WeaponModel wep : unit.model.weapons )
     {
-      double damage = (null == target)? 1 : wep.getDamage(target);
+      double damage = (null == target)? 1 : wep.getDamage(target) * unit.getHPFactor();
       if( damage > 0 )
       {
         if( !wep.canFireAfterMoving )
         {
           for (XYCoord xyc : Utils.findLocationsInRange(gameMap, origin, wep.minRange, wep.maxRange))
           {
-            double val = damage * unit.getHPFactor();
+            double val = damage;
             if (shootableTiles.containsKey(xyc))
               val = Math.max(val, shootableTiles.get(xyc));
             shootableTiles.put(xyc, val);
@@ -299,7 +299,7 @@ public class AIUtils
           {
             for (XYCoord xyc : Utils.findLocationsInRange(gameMap, dest, wep.minRange, wep.maxRange))
             {
-              double val = damage * unit.getHPFactor();
+              double val = damage;
               if (shootableTiles.containsKey(xyc))
                 val = Math.max(val, shootableTiles.get(xyc));
               shootableTiles.put(xyc, val);
--- a/src/Engine/GameVersion.java
+++ b/src/Engine/GameVersion.java
@@ -12,7 +12,7 @@ public class GameVersion implements Serializable
   int majorRev, minorRev, hotfix;
   public GameVersion()
   {
-    majorRev = 4; // For substantial, save-breaking changes to the game logic.
+    majorRev = 5; // For substantial, save-breaking changes to the game logic.
     minorRev = 0; // For added content or UI updates
     hotfix   = 0; // For bugfixes (hopefully to never see increment)
   }
--- a/src/UI/Art/SpriteArtist/OverlayArtist.java
+++ b/src/UI/Art/SpriteArtist/OverlayArtist.java
@@ -6,6 +6,7 @@ import java.awt.Composite;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
 
@@ -26,6 +27,9 @@ public class OverlayArtist
   static AlphaComposite buffComposite = null;
   static long lastCompositeCreationTime = 0;
 
+  private static ByteBuffer oldParams;
+  private static BufferedImage overlayImage;
+
   /**
    * Draw any overlays in the input set, as well as the move/target highlight overlays.
    * @param bigG The map image graphics, ignoring viewport size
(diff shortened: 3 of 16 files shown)


ThislsAUsername:
There's a lot of possible visual polish here, but it has the features I wanted in there.

src/AI/AIUtils.java
```diff
@@ -280,13 +280,14 @@ public static double scoreAttackAction(Unit unit, GameAction action, GameMap map
     ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeOccupiedDestinations);
     for( WeaponModel wep : unit.model.weapons )
     {
-      if( wep.getDamage(target) > 0 )
+      double damage = (null == target)? 1 : wep.getDamage(target);
```
ThatIsAUsername, line 283: Why move `wep.getDamage(target)` here and not `unit.getHPFactor()` also? The resulting damage term is only used with `getHPFactor()` to calculate `val`.

Also why multiply `damage` with hp factor in the first place? A weapon that can do 5 damage to a 10HP unit is even more dangerous to a 6HP unit, but this would make it have a lower `val` in the second case.

src/Engine/GameVersion.java
```diff
@@ -12,7 +12,7 @@
   int majorRev, minorRev, hotfix;
   public GameVersion()
   {
-    majorRev = 4; // For substantial, save-breaking changes to the game logic.
+    majorRev = 5; // For substantial, save-breaking changes to the game logic.
```
ThatIsAUsername, line 15: Not save protection, or control modifications, or AI restructuring, or enforced air unit crashes, or any of the other numerous recent (tm) changes. Nope. We draw the line at colorful overlays.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+import Engine.XYCoord;
+import Terrain.GameMap;
+import UI.GameOverlay;
+
+public class OverlayArtist
+{
+  private static final Color HIGHLIGHT_COLOR = new Color(255, 255, 255, 160); // white
+
+  public static final Color MOBILE_FIRE_EDGE = new Color(240,   0,   0, 255); // red
+  public static final Color SIEGE_FIRE_EDGE  = new Color(  0,   0, 160, 255); // dark blue
```
ThatIsAUsername, line 21: Why make this blue, instead of also red? I assume it's because they can't move and shoot, so it's the "potential" range, but if I click a stationary arty without moving, it's still blue, which seems odd.

Plus if I also have its overlay showing, the blue overlay and the team-color overlay kinda step on each other.
Maybe prioritize the "active" overlay and draw it last? E.g. the red/blue one should always be visible over the passive/team-color threat areas.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final Color FIRE_FILL        = new Color(  0,   0,   0,   0);
```
ThatIsAUsername, line 22: Huh.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+                                    GameMap gameMap,
+                                    Collection<GameOverlay> inputOverlays,
+                                    int drawX, int drawY,
+                                    int viewWidth, int viewHeight,
+                                    int tileSize)
+  {
+    if( viewWidth < 1 || viewHeight < 1 )
+      return;
+
+    BufferedImage overlayImage = SpriteLibrary.createTransparentSprite(viewWidth, viewHeight);
```
ThatIsAUsername, line 39: Instead of regenerating this every frame, consider caching it and only regenerating when something changes.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+    Graphics og = overlayImage.getGraphics();
+
+    // Only show stuff we're allowed to
+    Collection<GameOverlay> overlays = new ArrayList<GameOverlay>();
+    for( GameOverlay ov : inputOverlays )
+      if( null == ov.origin || !gameMap.isLocationFogged(ov.origin) )
```
ThatIsAUsername, line 45: This check would be more useful if the `GameOverlay` was ever constructed with `origin != null`.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final Color MOBILE_FIRE_EDGE = new Color(240,   0,   0, 255); // red
+  public static final Color SIEGE_FIRE_EDGE  = new Color(  0,   0, 160, 255); // dark blue
+  public static final Color FIRE_FILL        = new Color(  0,   0,   0,   0);
+
+  private static final int OVERLAY_EDGE_THICKNESS = 2;
+
+  static AlphaComposite buffComposite = null;
+  static long lastCompositeCreationTime = 0;
+
+  public static void drawHighlights(Graphics bigG,
```
ThatIsAUsername, line 29: Some javadocs to describe the input params and their units would be nice.

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+      {
+        if( (h+1)*tileSize < drawY || drawY + viewHeight < (h-2)*tileSize )
+          continue;
+
+        XYCoord coord = new XYCoord(w,h);
+        for( GameOverlay ov : overlays )
+          if( ov.area.contains(coord) )
+            drawOverlayTile(og, ov, coord, w * tileSize - drawX, h * tileSize - drawY, tileSize);
+        
+        if( gameMap.isLocationValid(w, h) )
```
ThatIsAUsername, line 62: You call `drawOverlayTile` before `isLocationValid()`?

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -0,0 +1,110 @@
...
+            og.fillRect(w * tileSize - drawX, h * tileSize - drawY, tileSize, tileSize);
+          }
+        }
+      }
+    }
+
+    // Set opacity as a function of time.
+    long nowTime = System.currentTimeMillis();
+//    float overlayOpacity = (float)(0.5*Math.max(0, Math.sin(nowTime/420.)));
+    float overlayOpacity = (float)(0.15*Math.sin(nowTime/420.) + 0.7);
```
ThatIsAUsername, line 77: Hey this looks familiar.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -244,8 +248,25 @@ private BufferedImage renderMap()
...
+        if( w.canFireAfterMoving )
+          edgeColor = OverlayArtist.MOBILE_FIRE_EDGE;
+        overlays.add(new GameOverlay(null,
+                     Utils.findLocationsInRange(gameMap, myGame.getCursorCoord(), w.minRange, w.maxRange),
+                     OverlayArtist.FIRE_FILL, edgeColor));
+      }
+    }
+    for( Commander co : myGame.commanders )
+    {
+      overlays.addAll(co.getMyOverlays(gameMap, co == gameMap.viewer));
```
ThatIsAUsername, line 267: Given that `co.getMyOverlays()` just returns an empty collection if `co != gameMap.viewer`, why bother with a loop here?

src/UI/GameOverlay.java
```diff
@@ -0,0 +1,30 @@
...
+import Engine.XYCoord;
+
+/** Holds the data to define an overlay to display on the map */
+public class GameOverlay implements Serializable
+{
+  private static final long serialVersionUID = 1L;
+
+  /** Determines whether to draw in FoW; null for "always draw" */
+  public XYCoord origin;
+  public ArrayList<XYCoord> area = new ArrayList<XYCoord>();
```
ThatIsAUsername, line 17: A `Set` might be a more natural choice here. Either works.

src/AI/AIUtils.java
ThislsAUsername: > Why move `wep.getDamage(target)` here and not `unit.getHPFactor()` also? The resulting damage term is only used with `getHPFactor()` to calculate `val`.

Fair. Mostly, I just wanted to enable using it with a `null` target to get all weapons.
> Also why multiply `damage` with hp factor in the first place? A weapon that can do 5 damage to a 10HP unit is even more dangerous to a 6HP unit, but this would make it have a lower `val` in the second case.

It's the *attacker's* HP factor. A 1 HP tank always has the same result from `WeaponModel.getDamage()`, but deals substantially less *real* damage because of HP.

src/Engine/GameVersion.java
ThislsAUsername: I was originally going to forget and then probably do it on the test branch.
I don't really care about the `GameVersion` when it's not in the hands of users, lol.

Save protection starts out null, so it might not technically break the save, but that's a valid major rev.
AI restructuring doesn't actually affect the save file, so it'd be a minor rev (until we start saving AI state).
Air unit crashes doesn't change the save file either. Might even be "hotfix" tier.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: Yeah. I thought about doing that, but sleep sure is a thing.

Good idea on the priority.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: There wasn't a "transparent" option that I could find.

...which isn't to say there *isn't* a transparent option...

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: blehhhhhh

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: It sure *should* be.
...I'm an idiot. I didn't. I blame sleep, and the lack of it.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: \*gasp* I've been caught
Having variable transparency is pretty ideal for visibility to both things, but I didn't want it changing drastically, so...

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: Hmm. I mean, it's not like we're gonna get overlays on invalid tiles.
And if we do, we're just drawing not-on-the-map. Not a huge deal.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: I have *plaaaaaaaaaaaaans*.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: More specifically, I can cache screen dimensions and whatever, but knowing how and when an overlay changes isn't necessarily trivial.
I think drawing only on the viewport rather than caching is the better optimization of valor here.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: Priority done w.r.t. other normal overlays.
There's no particularly *nice* way to make it show up on top of the movement/targeting highlight, though.
I could try implementing that as another overlay, but that involves either double calculation or overhauling a bunch of stuff 'cause the highlight is actually used a *lot* as a side-channel by `MapController` - might be good to do, but sounds like it might be out of scope.

src/AI/AIUtils.java
ThatIsAUsername: > attacker's HP factor

Oh. Right.

src/Engine/GameVersion.java
ThatIsAUsername: Hah, fair enough. I was being facetious. Certainly we needed to roll it before the next real release, so it's the accumulation of all those things that matters.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThatIsAUsername: Fair enough.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThatIsAUsername: I mean. You found it. But why have it if it's transparent? Future plans?

src/UI/Art/SpriteArtist/OverlayArtist.java
ThatIsAUsername: Even if it's just on each cursor movement, that saves a lot of regen cycles.
Drawing only on the viewport? As in without double-buffering?
Also I don't think I've heard the term "crunchy" in this context before. I assume it just means "textured"?

src/UI/Art/SpriteArtist/OverlayArtist.java
ThatIsAUsername: I mean, it should work... sure does look strange though.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: Right, but AFAIK there's no easy way of determining that something has happened to cause the overlays to be different from last time.
I suppose I could just check how many overlays there are, and listen for any and all GameEvents?

"Textured" texturing, yes.
It was an attempt to convey that the texturing is supposed to be more satisfying to look at than the stuff we have now.
Then again, simplicity on a passive visual is good.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: It's required to have a "fill" color in an overlay.
I guess I could allow it to be null, and it would save processor cycles in the end?

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: Are you saying you want it changed?

ThatIsAUsername: Not saying you have to change it. I mean, you are looping over the map dimensions anyway, so `isLocationValid` should never ever return false; it would be just as valid to remove the check as it would be to move it.

Just saying it makes me question how your head works.

PR #309, opened 2020-09-06
Flare fix

ThislsAUsername:
I intended to make this a small PR with one fix, but you know what happens when you playtest, lol

### DIFF
--- a/src/Engine/GameInput/SelectActionTarget.java
+++ b/src/Engine/GameInput/SelectActionTarget.java
@@ -23,8 +23,21 @@ class SelectActionTarget extends GameInputState<XYCoord>
     ArrayList<XYCoord> targets = myStateData.actionSet.getTargetedLocations();
 
     consider(targets.get(0));
+
+    boolean useFreeSelect = myStateData.actionSet.useFreeSelect;
+    // For constrained select, switch to free-tile select in target-rich environments
+    if( !useFreeSelect && targets.size() > 3 )
+    {
+      int maxDist = 0;
+      for( XYCoord a : targets )
+        for( XYCoord b : targets )
+          maxDist = Math.max(maxDist, a.getDistance(b));
+
+      useFreeSelect = maxDist < targets.size();
+    }
+
     return new OptionSet(
-          myStateData.actionSet.useFreeSelect?
+          useFreeSelect?
           InputType.FREE_TILE_SELECT : InputType.CONSTRAINED_TILE_SELECT,
           targets);
   }
@@ -32,6 +45,7 @@ class SelectActionTarget extends GameInputState<XYCoord>
   @Override
   public void consider(XYCoord coord)
   {
+    myStateData.damagePopups = new ArrayList<DamagePopup>();
     for( GameAction action : myStateData.actionSet.getGameActions() )
       if( coord.equals(action.getTargetLocation()) )
       {
--- a/src/Engine/GameInput/SelectActionTarget.java
+++ b/src/Engine/GameInput/SelectActionTarget.java
@@ -24,20 +24,8 @@ class SelectActionTarget extends GameInputState<XYCoord>
 
     consider(targets.get(0));
 
-    boolean useFreeSelect = myStateData.actionSet.useFreeSelect;
-    // For constrained select, switch to free-tile select in target-rich environments
-    if( !useFreeSelect && targets.size() > 3 )
-    {
-      int maxDist = 0;
-      for( XYCoord a : targets )
-        for( XYCoord b : targets )
-          maxDist = Math.max(maxDist, a.getDistance(b));
-
-      useFreeSelect = maxDist < targets.size();
-    }
-
     return new OptionSet(
-          useFreeSelect?
+          myStateData.actionSet.useFreeSelect?
           InputType.FREE_TILE_SELECT : InputType.CONSTRAINED_TILE_SELECT,
           targets);
   }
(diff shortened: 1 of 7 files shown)


src/Engine/MapController.java
```diff
@@ -204,20 +204,22 @@ private void handleFreeTileSelect(InputHandler.InputAction input)
...
+          if( null != resident )
+          {
+            ArrayList<Unit> threats = myGame.activeCO.threatsToOverlay;
+            if( threats.contains(resident) )
+              threats.remove(resident);
+            else
+              threats.add(resident);
+          }
         }
+        myGameInputHandler.back();
```
ThatIsAUsername, line 221: This used to only happen `if( myGameInputHandler.isTargeting() )`. Why the change?

src/Engine/MapController.java
```diff
@@ -383,9 +385,12 @@ public void onStateChange()
     switch (inputType)
     {
       case CONSTRAINED_TILE_SELECT:
-        // Create an option selector to keep track of where we are.
-        myGame.setCursorLocation(myGameInputHandler.getCoordinateOptions().get(myGameInputOptionSelector.getSelectionNormalized()));
+      case FREE_TILE_SELECT:
+        if( null != options && options.size() > 0 )
+          myGame.setCursorLocation(options.get(0));
```
ThatIsAUsername, line 390: I'm not a fan of jumping the cursor in free tile select. Also why change to getting `0` instead of `myGameInputOptionSelector.getSelectionNormalized()`?

src/Engine/MapController.java
ThislsAUsername: I reversed that because it brought back the "unit is now very excited" after you click on it thing.
I might be able to remove it entirely in the `transaction` PR, though.

src/Engine/MapController.java
ThislsAUsername: So, this is due to a thing I thought would be fancy: if you're in a target-dense environment (like a Rocket with a block of 3x3 units in range), you get `FREE_TILE_SELECT` instead of constrained, since the way we iterate those is a little unintuitive (and it's always going to be unintuitive for someone)

By the same token, having to move your cursor over to the enemies manually is annoying, so I fixed that too.

As for the selector: it can be null (presumably only the new situations I introduced), and it's always initialized to 0 AFAIK anyway.

src/Engine/GameInput/SelectActionTarget.java
```diff
@@ -23,15 +23,29 @@ protected OptionSet initOptions()
...
+    boolean useFreeSelect = myStateData.actionSet.useFreeSelect;
+    // For constrained select, switch to free-tile select in target-rich environments
+    if( !useFreeSelect && targets.size() > 3 )
+    {
+      int maxDist = 0;
+      for( XYCoord a : targets )
+        for( XYCoord b : targets )
+          maxDist = Math.max(maxDist, a.getDistance(b));
+
+      useFreeSelect = maxDist < targets.size();
```
ThatIsAUsername, line 36: This is a bit weird. But OK.

src/Engine/MapController.java
ThatIsAUsername: I get the desire to make target-dense situations a little easier to navigate, but this is no bueno for e.g. Flare and Missile Silo actions? It jumps to the NorthWest-most option, so have fun scrolling back to where you want to be.

src/Engine/GameInput/SelectActionTarget.java
ThislsAUsername: It means you have to be denser than a straight line for it to "force" free tile select.
Could also make this system an option, now that we have a special page for that.

src/Engine/MapController.java
ThislsAUsername: Bah. Good point.
Maybe this logic should go in `MapController`, then, so it can special-case constrained input as it comes in?

ThatIsAUsername: This _is_ MapController.

I guess you could use some kind of special-case logic though (only set the cursor for `FREE_TILE_SELECT` if there are fewer than some number of options, maybe?).

ThislsAUsername: To clarify:
I was meaning that I would move the "use free select if options are dense" into `MapController`, so that the rocket with friends would say to use constrained, but `MapController` would just interpret input as free select.

Actually actually: Why not (revert this change and) just make Q toggle to free select while a constrained selection is active? Best of both worlds, I think?

ThatIsAUsername: Maybe, except nobody would know to do that. You'd have to throw up a tooltip somewhere to say how to change modes.

src/Engine/MapController.java
ThatIsAUsername: Does it? I don't see that.

ThislsAUsername: I apparently forgot to push it up. Whoops.
Bah, I need to fiddle with stuff a bit.

ThislsAUsername: Okay, after having doublechecked...
I think I can move it back if necessary, but we can't get rid of it entirely or you can't back out of launching flares.

src/Engine/MapController.java
ThislsAUsername: Found a better way, I think.
It's almost like `MapController` is the place for control machinations.

PR #315, opened 2020-09-07
Power popups

ThislsAUsername:


ThislsAUsername:
FYI, I'm currently looking at trying to do it for Tech

src/Engine/GameInput/SelectUnitAction.java
```diff
@@ -80,9 +80,15 @@ else if( actionType.shouldConfirm )
     return next;
   }
 
+  @Override
+  public void consider(GameActionSet actionSet)
+  {
+    myStateData.damagePopups = actionSet.getSelected().getDamagePopups(myStateData.gameMap);
```
ThatIsAUsername, line 86: Not a fan of this; it makes it look like I'm about to commit an action before I actually am. Plus there's no (apparent) rhyme or reason for which option gets shown (and I don't think there's a good way to provide that).

src/CommandingOfficers/Meridian.java
```diff
@@ -148,6 +151,20 @@ protected void perform(MapMaster gameMap)
...
+    @Override
+    public Collection<DamagePopup> getDamagePopups(GameMap gameMap)
+    {
+      ArrayList<DamagePopup> output = new ArrayList<DamagePopup>();
+
+      for( Unit unit : COcast.justTransformed )
+        output.add(new DamagePopup(
+                       new XYCoord(unit.x, unit.y),
+                       myCommander.myColor,
+                       "Flow"));
```
ThatIsAUsername, line 164: Don't think popups are really needed for this one since the application rules are simple.
Plus "Flow" may not be the most descriptive thing anyway. Maybe "Act", if we can't go over 4 chars? I dunno, I'd just remove it.

src/CommandingOfficers/Tech.java
```diff
@@ -257,7 +260,48 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
+    @Override
+    public Collection<DamagePopup> getDamagePopups(GameMap gameMap)
+    {
+      ArrayList<DamagePopup> output = new ArrayList<DamagePopup>();
+
+      boolean log = true;
+      Set<XYCoord> dropLocs = new HashSet<XYCoord>();
+      for( int i = 0; i < numDrops; ++i )
+      {
+        XYCoord techDrop = findDropLocation(gameMap, dropLocs, log);
```
ThatIsAUsername, line 272: The projected drops don't always match the actual drops even with full visibility, I think because `findDropLocation` isn't deterministic in how it chooses spaces with equal scores.
Also, the "Drop" popups don't really look like they are in the right spot when there's nobody there to squash. I might add like, a little down arrow when there's no resident.

src/Engine/GameInstance.java
```diff
@@ -58,7 +58,7 @@ public GameInstance(MapMaster map, boolean fogOfWarOn, Weathers weather, GameSce
     gameScenario = scenario;
     isSecurityEnabled = useSecurity;
 
-    currentTurn = 0;
+    currentTurn = 1;
```
ThatIsAUsername, line 61: This will break `isSecurityEnforced()`

ThatIsAUsername:
Did you ever ping the crew to see what they think about ability previews/FoW/etc., and if we should actually do them?

src/Engine/GameInput/SelectUnitAction.java
ThislsAUsername: The one that gets shown is the one that would be selected first if you picked that option.
IMO it's nice because it indicates immediately which unit I'll attack if I just mash through. Also reduces button presses for checking out e.g. BBomb explosion places.

src/CommandingOfficers/Meridian.java
ThislsAUsername: There's no other indicator, though, and I don't want to make the user have to remember (especially as you can save in the middle of a turn).
We can go over 4 characters, I just thought "Flow" went well since it's in the power name :V

src/CommandingOfficers/Tech.java
ThislsAUsername: What kind of down arrow?

ThislsAUsername:
> Did you ever ping the crew to see what they think about ability previews/FoW/etc., and if we should actually do them?

I asked in 3 dedicated AW servers, and got 3 A's and 4 B's.
Seems reasonable to make an option and sit it next to the SEEK one.

src/Engine/GameInput/SelectUnitAction.java
ThatIsAUsername: Sure, I get that it gives an advance preview of the first option. Still don't like it. I'm used to seeing the attack stats or whatever just before committing, so having them show up earlier is confusing, and frankly the early preview really doesn't add much.

Re: Mashing: You either a) press to move and bring up the attack option, pause a moment to review it, then mash twice to commit, or you b) press twice to move and bring up the attack options, evaluate, then press once more.
You still have the same work, with the same essential pause to evaluate the option anyway. I'd rather keep a consistent interface language of showing action consequences just before commitment instead of muddying the water this way.

src/CommandingOfficers/Tech.java
ThatIsAUsername: As a quick mock-up, maybe something like this:
Image20000

ThatIsAUsername:
Not sure which is A and B, but that's fairly close either way.
Not sure this make sense to have as an option though, since that is material information that could inform player decisions.

ThislsAUsername:
> Not sure which is A and B, but that's fairly close either way.
> Not sure this make sense to have as an option though, since that is material information that could inform player decisions.

A was preview disabled, B was wrong preview.
I guess this is probably also complicated by the issue of cloaking units, and I didn't consider that in my poll (though I could poll again).
It's totally an interface option, though: "If you can't give me reliable data, don't give me any at all"

src/CommandingOfficers/Tech.java
ThislsAUsername: Drop arrows added. You want to make a sub-PR or something to make Tech drops deterministic?
I will say that Sturm actually has an RNG component on SCOP... but he only drops one thing, so I could just give all 3 options if I implemented that.

ThatIsAUsername: I could probably do that. I had an idea I never implemented that would effectively do that, I think.

How does Sturm's RNG work?

ThatIsAUsername:
> It's totally an interface option, though: "If you can't give me reliable data, don't give me any at all"

I'll buy that. Be tricky to find a good name for the option though. "Show Power previews even in circumstances when they may not be accurate" doesn't quite fit the bill.
Also, with stealth units in the mix, I think you'd have to always show previews even if they were wrong, or it would tip off the player that stealths were present.

src/CommandingOfficers/Ave.java
```diff
@@ -620,38 +644,76 @@ protected void perform(MapMaster gameMap)
...
+                         coord,
+                         myCommander.myColor,
+                         Math.min(victim.getHP()-1, OBLIDO_DAMAGE)*10 + "%"));
+      }
+      for( XYCoord coord : affectedTiles )
+        if( gameMap.getEnvironment(coord).terrainType == TerrainType.FOREST )
+          output.add(new DamagePopup(
+                         coord,
+                         myCommander.myColor,
+                         "BAMF"));
```
ThatIsAUsername, line 692: BAMF? You choose this acronym for a particular reason?

ThislsAUsername: Is it an acronym?

ThatIsAUsername: [*cough*](https://www.urbandictionary.com/define.php?term=BAMF)

src/CommandingOfficers/Tech.java
ThislsAUsername: A-ha! I think I now know why I accidentally ended up editing a post of yours. It doesn't actually *show* the quote option until you've initiated a comment.

> I could probably do that. I had an idea I never implemented that would effectively do that, I think.

Hokay. I have another WIP branch that touches the same general area though. Should maybe chat on how to deal with that.
> How does Sturm's RNG work?

Kinda like Rachel, he has like 3 heuristics for picking his drop location that he randomizes between. It's otherwise deterministic.

src/CommandingOfficers/Ave.java
ThislsAUsername: I was living in the world of comic books, where "BAMF" is what happens when Batman punches you.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -534,6 +533,12 @@ public void drawDamagePreview(Graphics g, DamagePopup data)
     int estimateX = (data.coords.xCoord * tileSize) + (tileSize / 2);
     int estimateY = Math.max((data.coords.yCoord * tileSize) - dmgImage.getHeight() / 2, dmgImage.getHeight() / 2); // Don't want it floating off-screen
     SpriteUIUtils.drawImageCenteredOnPoint(g, dmgImage, estimateX, estimateY);
+
+    if( spaceEmpty && 0 != data.coords.yCoord )
+    {
+      int arrowY = (data.coords.yCoord * tileSize) + (tileSize / 2);
```
ThatIsAUsername, line 539:
```suggestion
      int arrowY = (data.coords.yCoord * tileSize) + (tileSize / 2) - 1;
```

src/CommandingOfficers/Ave.java
ThatIsAUsername: Haha, I kinda figured. I move we find a different onomotopeia, or pick something descriptive like "RAZE" or whatever.

src/CommandingOfficers/Tech.java
ThatIsAUsername: I'm playing with fonts right now. I guess just do your thing, and I probably won't get to it until you're done anyway.

ThislsAUsername:
> I'll buy that. Be tricky to find a good name for the option though. "Show Power previews even in circumstances when they may not be accurate" doesn't quite fit the bill.

Could do "preview powers in fog" or "fog power estimation" or "fog power previews"
Would make sense to disable *all* power previewing at that point, but it makes sense to make that decision at a higher level than the ability logic itself anyway.
> Also, with stealth units in the mix, I think you'd have to always show previews even if they were wrong, or it would tip off the player that stealths were present.

You're going to not preview in fog anyway, and stealths are built uncloaked.
Only time that matters is if you have fog show up for 2 turns while the thing gets built and cloaks (or, I suppose, 1 turn of fog for COs like Cinder/Eagle)
DoR Subs are built dived, buuuuut you'd see the opponent's funds drop a *lot* with no visible unit anyway, so that's mostly just an anti-frustration feature IMO.

ThatIsAUsername:
> Would make sense to disable all power previewing at that point, but it makes sense to make that decision at a higher level than the ability logic itself anyway.

I'm not sure what the option would do except disable all power previews in fog.

> stealths

So you think we should just show the predictions based on knowledge, and it's up to them to know there may be confounding factors? I guess I'm probably fine with that.

ThislsAUsername:
> > Would make sense to disable all power previewing at that point, but it makes sense to make that decision at a higher level than the ability logic itself anyway.
> 
> I'm not sure what the option would do except disable all power previews in fog.

Which *mostly* satisfies the requirement of "don't ever lie to me". And doesn't require the ability logic itself to worry about that setting.
Problem is, it doesn't address the problem of...
> > stealths
> 
> So you think we should just show the predictions based on knowledge, and it's up to them to know there may be confounding factors? I guess I'm probably fine with that.

That's not what I was saying. My point was: you're basically always either...
a) in fog already, so it doesn't matter
b) in a position to know already if there's a stealth unit in the match

So it makes perfect sense to just disable previews whenever a confounding factor appears (fog/cloakers), and not worry about "leaking" info that way (since in a non-fog game, you know when a cloaker is built)

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: I'm assuming this is to break the bottom border of the preview like in your mockup?

src/CommandingOfficers/Tech.java
ThislsAUsername: Made an issue for it.

ThatIsAUsername:
> So it makes perfect sense to just disable previews whenever a confounding factor appears (fog/cloakers), and not worry about "leaking" info that way (since in a non-fog game, you know when a cloaker is built)

So anytime a stealth/sub is built, previews just get disabled (until they die, presumably). What about powers that don't depend on enemy force disposition (e.g. Meridian's)? What if there was a CO who could generate fog temporarily and used it just to create a stealth all sneaky-like, but then this mechanic blew his cover?

ThislsAUsername:
> > So it makes perfect sense to just disable previews whenever a confounding factor appears (fog/cloakers), and not worry about "leaking" info that way (since in a non-fog game, you know when a cloaker is built)
> 
> So anytime a stealth/sub is built, previews just get disabled (until they die, presumably).

The implementation I ran with was just doing so if they *are* cloaked, since that's easier to check for (though an attribute or something would make sense).

> What about powers that don't depend on enemy force disposition (e.g. Meridian's)?

An unfortunate loss, but I'm not too worried about it. If it's something people really care about, they'll complain and maybe we'll change it (it'd be super possible to e.g. put in a flag or whatever that says "I don't depend on enemy disposition")
This is also simpler to set up the option reasonably clearly.
> What if there was a CO who could generate fog temporarily and used it just to create a stealth all sneaky-like, but then this mechanic blew his cover?

(FWIW, AW:DS Drake and Penny can both create fog on cart, so that's valid without customs)
This would require either refreshing or 2 consecutive turns of fog from one power.
I suppose this could be a thing in [fog CO] vs Cinder, since she could build/cloak the Stealth during the power?

I suppose the problem here is that there's *no* good solution. Either you preview the powers (and lie to the player who specifically requested you not), or you hide them (thus providing free information).
The third option that comes to mind is to change the settings I have in the fog preview thing to: never hide, hide in fog, always hide (to replace hide when things are cloaked)

I suppose I could make the argument that you *could* figure out the stealth build by tracking income/funds-vs-unit value, but that's starting to stretch feasibility a little.
Then again, we have a stats page for that very thing, and you could just screenshot it each turn if you were worried (with, I suppose, map screenshots to track repair costs).

ThislsAUsername:
> I suppose the problem here is that there's no _good_ solution. Either you preview the powers (and lie to the player who specifically requested you not), or you hide them (thus providing free information).
> The third option that comes to mind is to change the settings I have in the fog preview thing to: never hide, hide in fog, always hide (to replace hide when things are cloaked)
> 
> I suppose I could make the argument that you _could_ figure out the stealth build by tracking income/funds-vs-unit value, but that's starting to stretch feasibility a little.
> Then again, we have a stats page for that very thing, and you could just screenshot it each turn if you were worried (with, I suppose, map screenshots to track repair costs).

Seeing as you merged the option PR, I'm assuming you're running with the last argument I made there?

ThatIsAUsername:
> Seeing as you merged the option PR, I'm assuming you're running with the last argument I made there?

It's more of a "I think the likelihood of this ever actually being meaningful, combined with its likely impact, is below my threshold for worrying too much about it." If you and the Discord chaps are fine with this, I don't care too much.

ThislsAUsername:
Hah. Works for me.

PR #316, opened 2020-09-07
Bear with me

ThislsAUsername:


src/CommandingOfficers/Bear_Bull.java
```diff
@@ -99,6 +104,12 @@ public void swapD2Ds(boolean setIncome)
     }
   }
 
+  @Override
+  public double getRepairCostFactor()
+  {
+    return 0;
```
ThatIsAUsername, line 110: This is for both? That's awful nice.

ThislsAUsername: Yeah. I figure they need the boost to keep up with the likes of Ave and Tech.

PR #320, opened 2020-09-14
Listen closer

ThislsAUsername:
I wanted this to be a separate PR, but it was gonna cause conflicts and I don't like conflicts.

This is primarily groundwork to allow running multiple games on multiple threads.
I have an MVP for AI battles, but it'd be annoying to de-conflict it with everything so I wanna get stuff settled before pushing it up.


src/CommandingOfficers/Ave.java
```diff
@@ -536,6 +554,14 @@ protected void perform(MapMaster gameMap)
       // Normal CO-power boost.
       myCommander.addCOModifier(damageMod);
 
+      // Freeze enemies around each of Ave's units or buildings.
+      for( Unit victim : findVictims(gameMap) )
+        victim.isStunned = true;
```
ThatIsAUsername, line 559: Do we need a `StunUnitEvent` or similar so we can move this into `getEvents()`? Maybe out of scope for this PR.

src/CommandingOfficers/Patch.java
```diff
@@ -53,32 +54,51 @@ public Commander create(GameScenario.GameRules rules)
...
     addCommanderAbility(new PatchAbility(this, PLUNDER_NAME, PLUNDER_COST, PLUNDER_INCOME, PLUNDER_ATTACK_BUFF));
     addCommanderAbility(new PatchAbility(this, PILLAGE_NAME, PILLAGE_COST, PILLAGE_INCOME, PILLAGE_ATTACK_BUFF));
+  }
 
+  @Override
+  public void registerForEvents(GameInstance game)
+  {
+    super.registerForEvents(game);
     // Passive - Loot
     myLootAbility = new LootAbility(this);
```
ThatIsAUsername, line 72: Interesting choice to create this inside `registerForEvents()`, but, I mean, I _guess_.

src/CommandingOfficers/Patch.java
```diff
@@ -137,48 +179,8 @@ public void revertChanges(Commander commander)
...
     }
   }
 
-  /**
-   * This GameEventListener will inspect published BattleEvents. If the owning Commander took
-   * part in the battle, then any damage done to the opponent will give income to the owning
-   * Commander. The amount of income is a fraction of value of the damage that was done, thus
-   * damaging more expensive units will grant more income.
-   */
-  private static class DamageDealtToIncomeConverter extends GameEventListener
```
ThatIsAUsername, line 150: Not sure I'm a fan of moving this inside `LootAbility`. His passive is unrelated to the active abilities, and it seems to break separation of concerns to toss that in there.
Admittedly, `LootAbility` is already registered for events so I can see why that seems convenient, but I'd still keep it separate.

src/CommandingOfficers/Ave.java
ThislsAUsername: Yes, and yes.
Same for refreshing units.

src/CommandingOfficers/Patch.java
ThislsAUsername: Hmm.
The issue I ran into was that we now require a `GameInstance` reference to un/register, and you only get that at registration time (without fiddling with a whole lot of function APIs)
I suppose I could just make Patch have two listeners he subscribes, one to hold each of these functions?

ThatIsAUsername: That's probably what I would have done.

src/Engine/GameEvents/LambdaEvent.java
```diff
@@ -0,0 +1,49 @@
+package Engine.GameEvents;
+
+import java.util.function.Consumer;
+
+import Engine.XYCoord;
+import Terrain.MapMaster;
+import UI.MapView;
+import UI.Art.Animation.GameAnimation;
+
+public class LambdaEvent implements GameEvent
```
ThatIsAUsername, line 10: Obviously you know I have to object to this on principle.
I can understand there are reasons you would want to have a catch-all event type, but those are the precise reasons I would use to argue against having a catch-all event type.

src/CommandingOfficers/Bear_Bull.java
```diff
@@ -149,13 +159,18 @@ public String toString()
...
+    public GameEventQueue getEvents(MapMaster gameMap)
+    {
       // Damage is dealt after swapping D2Ds so it's actually useful to Bear
-      COcast.liquidateMassDamage = true; // Collect money instead of ability energy
-      GameEvent damage = new MassDamageEvent(findVictims(gameMap), DOWNUPTURN_LIQUIDATION, false);
-      damage.performEvent(gameMap);
-      GameEventListener.publishEvent(damage);
-      COcast.liquidateMassDamage = false;
+      GameEventQueue powerEvents = new GameEventQueue();
+      powerEvents.add( (map) -> COcast.liquidateMassDamage = true ); // Collect money instead of ability energy
```
ThatIsAUsername, line 169: This seems... silly. Using a reified lambda to hand a function manipulating internal state to another entity just breaks... everything about OO best practices.
This also isn't a `GameEvent` in any real sense - it could live quite happily (and more sensibly) in `perform()`.

src/Engine/GameEvents/LambdaEvent.java
ThislsAUsername: Check out how it's used in Bear&Bull.
Without making a whole new event type for *each CO*, there's no way to accomplish setting up event sequences like that.
You could argue, maybe, that I should have implemented differently, but that was the simplest option I could see.

src/CommandingOfficers/Bear_Bull.java
ThatIsAUsername: OK yeah, you can throw `COcast.liquidateMassDamage = true;` in `perform()`, and then toss `COcast.liquidateMassDamage = false;` in `endTurn()` to prevent gaining funds from mass damage events during other CO's turns.
I guess that leaves open the possibility of _another_ mass damage event during your turn that triggers more income accrual, but I mean, that's kind of a non-thing.

ThislsAUsername: It also spreads around the code more broadly, and ties the sequencing of it to other systems.
Putting all three actions next to each other in the event queue makes a lot of semantic (minimizing side-effects) and idiomatic (putting things that happen together and are related together) sense.

ThislsAUsername:
I just noticed there's a fun problem breaking saving in this PR - `GameInstance` is now storing a reference to a SpriteArtist.
May have to fiddle with this a bit.

src/UI/Art/SpriteArtist/MapArtist.java
```diff
@@ -204,6 +204,9 @@ private void redrawBaseTile(XYCoord coord)
       myArtist = artist;
     }
 
+    @Override
+    public boolean shouldSerialize() { return false; }
```
ThatIsAUsername, line 208: Should this be a new method in `GameEventListener`?

src/CommandingOfficers/Bear_Bull.java
ThatIsAUsername: Yeah, fair. Actually, there's no reason you can't just calculate the income up front and inject a `CanHazMoneyPlz` (or maybe just like, `ModifyFundsEvent`, even) event instead of waiting for the damage events to be published. That seems cleaner than either previous idea.

ThislsAUsername: Uh. Hm. That's pretty reasonable, and theoretically simpler than what I had before.

src/CommandingOfficers/Bear_Bull.java
```diff
@@ -164,11 +149,19 @@ protected void perform(MapMaster gameMap)
...
       GameEventQueue powerEvents = new GameEventQueue();
-      powerEvents.add( (map) -> COcast.liquidateMassDamage = true ); // Collect money instead of ability energy
-      powerEvents.add(new MassDamageEvent(findVictims(gameMap), DOWNUPTURN_LIQUIDATION, false));
-      powerEvents.add( (map) -> COcast.liquidateMassDamage = false );
+      powerEvents.add(new MassDamageEvent(COcast, victims, DOWNUPTURN_LIQUIDATION, false));
+
+      int valueDrained = 0;
+      for( Unit victim : victims )
+      {
+        valueDrained += (Math.min(DOWNUPTURN_LIQUIDATION, victim.getHP()) * victim.model.getCost()) / victim.model.maxHP;
```
ThatIsAUsername, line 161:
```suggestion
        valueDrained += (Math.min(DOWNUPTURN_LIQUIDATION, victim.getHP()-1) * victim.model.getCost()) / victim.model.maxHP;
```

src/Engine/GameEvents/ModifyFundsEvent.java
```diff
@@ -1,38 +1,39 @@
...
-    this.lambda = lambda;
+    this.beneficiary = beneficiary;
+    this.value = value;
   }
 
   @Override
   public GameAnimation getEventAnimation(MapView mapView)
   {
-    // TODO?
+    // TODO
```
ThatIsAUsername, line 23: I'm not sure this needs an animation at all; probably don't need a TODO here.

src/Engine/GameEvents/ModifyFundsEvent.java
```diff
@@ -1,38 +1,39 @@
...
     return null;
   }
 
   @Override
   public void sendToListener(GameEventListener listener)
   {
-    // TODO?
+    // TODO
```
ThatIsAUsername, line 30: Might be good to do this instead of leaving it as tech debt.

src/CommandingOfficers/Bear_Bull.java
ThislsAUsername: I thought about that, but you can actually drain a unit to 0.1 HP, and that's pretty niche anyway.
May as well give you the small benefit.

src/Engine/GameEvents/ModifyFundsEvent.java
ThislsAUsername: The carts have an animation for it, actually.

src/CommandingOfficers/Bear_Bull.java
ThatIsAUsername: Really? Map damage is normally in whole increments. I guess why not though if the power actually drains to 0.1.

src/Engine/GameEvents/ModifyFundsEvent.java
ThatIsAUsername: Just the number-scrolling animation? I mean, sure. That's probably low enough of a priority that I'm not sure we need a TODO anyway.

src/CommandingOfficers/Bear_Bull.java
ThislsAUsername: Map damage goes down to 0.1 HP on cart and I believe I've mirrored that here.

ThislsAUsername:
We missing anything here for the merge of this? That goes for the base PR, too.

PR #321, opened 2020-09-27
Add basic FightClub

ThislsAUsername:
Remember rule #1 of fight club


ThislsAUsername:
There's a bunch of commented lines lying around since I didn't figure there would be much reason to build a GUI for this, so having stuff you can turn on/off at will make sense.

Might also make sense to break this code up into multiple files. Dunno.

ThatIsAUsername:
It would be good to have more stats at game-end. Along with winning team, consider:
  Turn count
  End condition (since it resolves stalemates), etc.
  Possibly descriptive stats (units created/killed, income, efficiency, etc).
  Maybe processor/user time for each AI as well, for grins.
  Anything else that seems useful for evaluating performance.

Maybe also have some sort of time-based periodic output so you can tell it hasn't deadlocked or stalled... because Wally. This could even list a match % complete based on elapsed/max turns for that map size.

I also don't think it's super useful to have multiple sets unless they are on different maps, which may be what you had in mind anyway. In terms of AI testing, I would want to set up N matches on M selected maps with alternating start positions.

PR #322, opened 2020-10-04
Implemented font support.

ThatIsAUsername:
Seems to work. Probably some cleanup to do, but this is more than enough for proof-of-concept and to get feedback.

We can probably delete the old font images (unless we want to keep them for posterity; but it would be easy to re-gen them).
The game stats page is no longer aligned correctly because the font is no longer (guaranteed to be) monospaced.
Certain glyphs may not display precisely as intended due to the lack of transparency support. Using an image-parsing approach would let us support that, but might come with other tradeoffs.
It would be easy enough to make this work with the smallcaps font as well; the bold letters would be more challenging; maybe Java has a way to print text with fills/outlines, but I haven't looked.


src/UI/Art/SpriteArtist/HorizontalSelectorTemplate.java
```diff
@@ -98,16 +99,26 @@ private BufferedImage generateOptionPanel(int length, Color fgColor)
...
+//    g.setColor(Color.RED);
+//    g.drawRect(x+drawBuffer, y+drawBuffer, pf.getWidth('H'), pf.getAscent());
+//    g.setColor(Color.YELLOW);
+//    g.drawRect(x+drawBuffer, y+drawBuffer+pf.getAscent(), pf.getWidth('H')*2, pf.getDescent());
+//    g.setColor(Color.BLACK);
+
+    SpriteUIUtils.drawText(g, opt.optionName, x + drawBuffer, y+drawBuffer);
     // Draw the setting panel and the setting value.
-    x = x + (optionNamePanel.getWidth() + optionSpacingPx);
+    x = x + (optionNamePanel.getWidth() + (3 * pf.emSizePx));
```
ThislsAUsername, line 119: I realize now that my `fontsMod` changes just made this line do a round trip back to its previous definition (but with a less useful variable name).
Why'd you delete `optionSpacingPx`? I added `horizontalSpacing` because it was useful for my arrow-positioning logic.

ThatIsAUsername: I think because `optionSpacingPx` was defined relative to `letterWidth`, which I was replacing with Em size. I definitely tweaked the code here multiple times before getting something that worked. I'm definitely fine with putting a proper variable name here again and moving the magic 3 back to where it makes more sense.

The real question is probably: Why does every place that uses `HorizontalSelectorTemplate` have slightly different logic for iterating and drawing all of the options?

ThislsAUsername: That's a great question, and why I was proposing moving more of that logic into `HorizontalSelectorTemplate`.
I don't recall at the moment why we decided that wasn't a great idea, but maybe I should throw some annotation in on my new spacing logic in `fontsMod`?

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -485,65 +483,48 @@ public static BufferedImage getHideIcon(Color color)
...
+          for( final File fileEntry : folder.listFiles() )
+          {
+            String filepath = fileEntry.getAbsolutePath();
+            // Look for files with our extension
+            if( !fileEntry.isDirectory() && filepath.endsWith(".ttf") )
+            {
+              // The name should be formatted as fontname_wxh.ttf
+              String[] nameparts = fileEntry.getName().split("[_x.]");
+
+              if( nameparts.length != 4 ) continue;  // Not formatted correctly.
```
ThislsAUsername, line 504: This does ban using underscores or dots in the rest of the file name. That's slightly awkward, but probably fine.

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -485,65 +483,48 @@ public static BufferedImage getHideIcon(Color color)
...
+            }
+          }
+        }
+      }
+      catch(Exception ex)
+      {
+        System.out.println("WARNING! Exception loading font file. Using fallback");
+        ex.printStackTrace();
+        Font temp = new Font("Consolas", 0, 10);
+        fontStandard = new PixelFont(temp);
```
ThislsAUsername, line 522: This fallback doesn't trigger if there's just no valid font available.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -599,15 +599,16 @@ private void drawTurnCounter(Graphics g, boolean counterIsRight)
...
       BufferedImage dayNum = SpriteUIUtils.getBoldTextAsImage(Integer.toString(turnNum));
       int width = day.getWidth() + dayNum.getWidth();
-      int height = dayNum.getHeight();
+      int height = Math.max(pf.getAscent(), dayNum.getHeight());
 
       turnNumImage = SpriteLibrary.createTransparentSprite(width, height);
       Graphics dcg = turnNumImage.getGraphics();
       dcg.drawImage(day, 0, 0, null);
-      dcg.drawImage(dayNum, turnNumImage.getWidth()-dayNum.getWidth(), 0, null);
+      dcg.drawImage(dayNum, turnNumImage.getWidth()-dayNum.getWidth(), pf.getAscent()-dayNum.getHeight(), null);
```
ThislsAUsername, line 611: I'm not a huge fan of how this looks, but I feel like moving the number away from bottom-aligned makes it look even weirder.
Turn 2 fancy
Might consider just using normal numbers instead of the fancy number?
Turn 2 bland

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Yup. We can either specify the font name format, or come up with a more contrived mechanism for specifying the render size. I'm pretty good with this approach.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThatIsAUsername: Huh. I somehow end up with a working font on Windows. But yeah, that logic is wrong. Fixed.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: I like the bold numbers because they stand out more, but I'm not too tied to it either way.

src/UI/Art/SpriteArtist/HorizontalSelectorTemplate.java
ThatIsAUsername: Sure

PR #323, opened 2020-10-11
Fixes #318 Tech Drop predictions and events now always match.

### LINKED ISSUE #318: Make Tech drops deterministic
Tech currently randomizes when picking mech locations. That's not super ideal for the purposes of previewing where they land accurately.

_Originally posted by @ThislsAUsername in https://github.com/ThislsAUsername/DefendPeace/pull/315#discussion_r487461299_

ThatIsAUsername:


src/Test/TestCommanderTech.java
```diff
@@ -0,0 +1,122 @@
...
+
+    return testPassed;
+  }
+
+  private boolean testTechDropRandomization()
+  {
+    boolean testPassed = true;
+
+    // Let's make sure the predictions and generated events always match.
+    for( int i = 0; i < 10; i++ )
```
ThislsAUsername, line 75: Seems a little silly to run the calc 10 times with the same initial conditions, but whatever boats your float?

ThatIsAUsername: Well, yeah. I initially built the test case under the assumption that the results were non-deterministic, and figured ten identical triple-drops was enough to pretty well guarantee it truly was deterministic.

Turns out it was deterministic the whole time, and just didn't agree with the damage previews. Guess I could simplify now, since the test always failed on the first pass before anyway.

PR #328, opened 2020-10-13
Counterattacks now expend ammo correctly.

### LINKED ISSUE #327: Ammo is only consumed when attacking
... not when counterattacking.

ThatIsAUsername:
Fixes #327

### DIFF
--- a/src/Units/Unit.java
+++ b/src/Units/Unit.java
@@ -205,7 +205,7 @@ public class Unit implements Serializable
       if( ammo > 0 )
         ammo--;
       else
-        System.out.println("WARNING: " + toString() + " fired with no available ammo!");
+        System.out.println("WARNING: " + toStringWithLocation() + " fired with no available ammo!");
     }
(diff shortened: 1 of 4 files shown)


src/Units/Unit.java
```diff
@@ -205,7 +205,7 @@ public void fire(WeaponModel weapon)
       if( ammo > 0 )
         ammo--;
       else
-        System.out.println("WARNING: fired with no available ammo!");
+        System.out.println("WARNING: " + toString() + " fired with no available ammo!");
```
ThislsAUsername, line 208:
```suggestion
        System.out.println("WARNING: " + toStringWithLocation() + " fired with no available ammo!");
```

ThislsAUsername:
Looks good otherwise, I think, though I'm curious at the extra mobile attack testcase.

ThatIsAUsername:
The new test was failing before I made the change.

PR #330, opened 2020-10-13
Fixes #325 Added transport join/death/delete test cases with fixes.

### LINKED ISSUE #325: Ensure transports behave correctly on removal
Specifically, join and death events should make sure to handle cargo units correctly.
Should probably test on the cart to verify.

ThatIsAUsername:


ThatIsAUsername:
Note I still think we don't get Ability power for units that die in transports. Should we add handling for AP from dying units with nonzero health in `receiveUnitDieEvent()` similar to `receiveBattleEvent`, or would you recommend a different approach?

ThislsAUsername:
Hmm. I don't know how worried I am about giving ability power for units dying in transit, since I really doubt the carts do.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
```diff
@@ -169,6 +175,11 @@ public GameEventQueue getEvents(MapMaster gameMap)
           if( event.defenderDies() )
           {
             attackEvents.add(new UnitDieEvent(event.getDefender()));
+            if( event.getDefender().model.holdingCapacity > 0 )
+              for( Unit u : event.getDefender().heldUnits )
+              {
+                attackEvents.add(new UnitDieEvent(u));
+              }
```
ThislsAUsername, line 182: Would it be feasible to put this check into the `UnitDieEvent`, or maybe make a helper function to create the set of them?
I don't like the idea of having to track down units floating in the void after their transports get wrecked.

src/Engine/UnitActionLifecycles/JoinLifecycle.java
```diff
@@ -29,7 +29,8 @@ public GameActionSet getPossibleActions(GameMap map, Path movePath, Unit actor,
       if( !ignoreResident && resident != null )
       {
         // TODO: Consider if and how off-CO joins should be allowed if tags ever happens
-        if( resident.model.equals(actor.model) && resident != actor && resident.isHurt() )
+        int cargoSize = resident.heldUnits.size() + actor.heldUnits.size();
+        if( resident.model.equals(actor.model) && resident != actor && resident.isHurt() && cargoSize <= resident.model.holdingCapacity )
```
ThislsAUsername, line 33: Checked the cart on this. They actually just don't let carrying transports participate in joining at all.
That said, I'm fine with this as is because eh... why not?

src/Engine/UnitActionLifecycles/JoinLifecycle.java
```diff
@@ -98,6 +99,8 @@ public GameEventQueue getEvents(MapMaster gameMap)
           // Find the unit we want to join.
           recipient = gameMap.getLocation(pathEnd).getResident();
           isValid &= (null != recipient) && recipient.isHurt();
+          int cargoSize = recipient.model.holdingCapacity > 0 ? recipient.heldUnits.size() + donor.heldUnits.size() : 0;
+          isValid &= cargoSize <= recipient.model.holdingCapacity;
```
ThislsAUsername, line 103: Why have the conditional assignment?
0 <= 0+0

ThislsAUsername:
I think I'd like to see the death helper function or such happen, but otherwise looks good.

src/Engine/UnitActionLifecycles/JoinLifecycle.java
ThatIsAUsername: Yeah, I know. But I wanted to.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThatIsAUsername: We could certainly remove any transported units in the `UnitDieEvent`, but then they couldn't have their own death event, which I'd rather have.
A utility function would have to be somewhere accessible to both `BattleLifecycle` and `DeleteLifecycle`, and wouldn't save much syntax here anyway. This is a simple enough operation that I'm not sure I care to break it out.

ThislsAUsername: It's not about saving syntax; it's about ensuring program correctness through a consistent approach.
If I make a helper function that queues a death event for a unit and all the unit's cargo and all the *cargo's* cargo (which, I suppose I'll note, you haven't accounted for here), then I shouldn't ever have to call the `UnitDieEvent` constructor explicitly anywhere else, and I don't have to remember to check for ~~traps~~ edge cases.

ThatIsAUsername: Hmm, yes. Okay. Done.

ThatIsAUsername:
Is this good to go?

PR #331, opened 2020-10-13
Fixes #324 FREE_TILE_SELECT override now only applies to ATTACK actions.

### LINKED ISSUE #324: Transports use free tile select when dropping
But only if they are holding multiple units. This is weird.
Transports should always use constrained tile select when dropping off units, rather than changing input modes unexpectedly.

ThatIsAUsername:
Alright. So this works, but I'd still be more inclined to just take over the `FREE_TILE_SELECT` override altogether. It introduces unexpected behavior in edge cases (though admittedly it's unlikely to elicit input mistakes), but it also fails to engage at times when you would think it should, e.g. here:
Screenshot from 2020-10-13 13-59-40

It's not really a huge issue, I'm just not a fan of special cases, especially when the special handling affects non-special cases.

### DIFF
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -255,7 +255,8 @@ public class MapController implements IController, GameInputHandler.StateChanged
 
         boolean useFreeSelect = false;
         // Switch to free-tile select in target-rich environments
-        if( !useFreeSelect && targetLocations.size() > 3 )
+        if( !useFreeSelect && targetLocations.size() > 3
+            && (myGameInputHandler.myStateData.actionSet.getGameActions().get(0).getType() == UnitActionFactory.ATTACK))
         {
           int maxDist = 0;
           for( XYCoord a : targetLocations )
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -253,18 +253,10 @@ public class MapController implements IController, GameInputHandler.StateChanged
           System.out.println("WARNING! Attempting to choose a target for a non-targetable action.");
         }
 
-        boolean useFreeSelect = false;
         // Switch to free-tile select in target-rich environments
-        if( !useFreeSelect && targetLocations.size() > 3
-            && (myGameInputHandler.myStateData.actionSet.getGameActions().get(0).getType() == UnitActionFactory.ATTACK))
-        {
-          int maxDist = 0;
-          for( XYCoord a : targetLocations )
-            for( XYCoord b : targetLocations )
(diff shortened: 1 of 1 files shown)


ThislsAUsername:
I believe this change should break the handling for the flare launch action, which isn't ideal.
I think it'd be better to just enhance the criteria in the if statement rather than limiting what types it can apply to.

Also, your "fail" case isn't surprising to me. Maybe what you're looking for is "activate if there are a bunch of contiguous targeted squares"?

ThatIsAUsername:
Nah, flare still works (except that it still disappears after one turn, of course).

Sure it isn't surprising because you created the capability so you know when it will activate. Someone who just knows that it's supposed to activate in "cluttered or target-rich environments" and who knows it activates without the meteor present would probably be surprised that it didn't activate with the meteor there.

ThislsAUsername:
> Alright. So this works, but I'd still be more inclined to just take over the `FREE_TILE_SELECT` override altogether.

Did you intend to write "take _out_ the `FREE_TILE_SELECT` override"?
If so, I'm slightly down with that. It is a special case for a pretty marginal set of use cases - specifically, in cases where it's easier and less confusing to navigate by the standard meanings of up/down/left/right.

Another salient option would be to improve our ordering, presumably something like "rows top to bottom, items left to right".

ThatIsAUsername:
Yes, I meant remove the override so that behavior is always consistent.
I don't think I'd be opposed to a well-defined ordering, but I'm not sure it adds much; when units are clustered any sane user is going to double-check that they are on the right spot anyway before committing.

ThislsAUsername:
> Yes, I meant remove the override so that behavior is always consistent.

I'm fine with that, I guess.
> I don't think I'd be opposed to a well-defined ordering, but I'm not sure it adds much; when units are clustered any sane user is going to double-check that they are on the right spot anyway before committing.

Sure, but when you iterate diagonally, it's not always intuitive what the ordering actually is. And holding the button to cycle through n/2 targets is a drag (as opposed to just navigating freely)

ThatIsAUsername:
Cycling through N/2 targets isn't typically a big deal, as N isn't typically a big number. Though if we did want to have a free-select switch condition, maybe just switch if N is greater than say, 5 or 6, regardless of location?

Or just keep it consistent. That works too.

ThislsAUsername:
> Though if we did want to have a free-select switch condition, maybe just switch if N is greater than say, 5 or 6, regardless of location?

I'm down with that. Might be worth playing with the threshold a bit and seeing how it feels, though it should definitely not activate on 4.

PR #336, opened 2020-12-27
Muriel dev

ThatIsAUsername:
Quality of life improvements so Muriel doesn't get in her way as much.

### DIFF
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -509,6 +509,31 @@ public class AIUtils
     }
   }
 
+  /**
+   * Compares GameActions based on their `getMoveLocation()`'s distance from the provided `destination`.
+   * GameActions closer to `destination` will be sorted to be first.
+   */
+  public static class DistanceFromLocationComparator implements Comparator<GameAction>
+  {
+    XYCoord target;
+
+    public DistanceFromLocationComparator(XYCoord destination)
+    {
+      target = destination;
+    }
+
+    @Override
+    public int compare(GameAction o1, GameAction o2)
+    {
+      XYCoord o1c = o1.getMoveLocation();
+      XYCoord o2c = o2.getMoveLocation();
+      int o1Dist = (null == o1c) ? Integer.MAX_VALUE : o1c.getDistance(target);
+      int o2Dist = (null == o2c) ? Integer.MAX_VALUE : o2c.getDistance(target);
+      int diff = o1Dist - o2Dist;
+      return diff;
+    }
+  }
+
   /** Return the set of locations with enemies or terrain that `unit` could attack in one turn from `start` */
   public static Set<XYCoord> findPossibleTargets(GameMap gameMap, Unit unit, XYCoord start, boolean includeTerrain)
   {
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -90,9 +90,9 @@ public class AIUtils
   }
   /**
    * Finds all actions available to unit, and organizes them by type instead of by location.
-   * Assumes caller isn't interested in moving into units' current spaces
    * @param unit The unit under consideration.
    * @param gameMap The world in which the Unit lives.
+   * @param includeOccupiedDestinations Determines whether to consider actions that would require moving to a space already occupied by a friendly unit.
    * @return a Map of ActionType to ArrayList<GameAction>.
    */
   public static Map<UnitActionFactory, ArrayList<GameAction> > getAvailableUnitActionsByType(Unit unit, GameMap gameMap, boolean includeOccupiedDestinations)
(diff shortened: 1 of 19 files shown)


src/AI/AIUtils.java
```diff
@@ -70,6 +84,18 @@
...
   public static Map<UnitActionFactory, ArrayList<GameAction> > getAvailableUnitActionsByType(Unit unit, GameMap gameMap)
+  {
+    boolean includeOccupiedDestinations = false;
+    return getAvailableUnitActionsByType(unit, gameMap, includeOccupiedDestinations);
+  }
+  /**
+   * Finds all actions available to unit, and organizes them by type instead of by location.
+   * Assumes caller isn't interested in moving into units' current spaces
+   * @param unit The unit under consideration.
+   * @param gameMap The world in which the Unit lives.
```
ThislsAUsername, line 95: *cough*

src/Engine/GameAction.java
```diff
@@ -30,6 +30,7 @@
    * Returns a GameEventQueue with the events that make up this action. If the action
    * was constructed incorrectly, this should return an empty GameEventQueue.
    */
+  public abstract Unit getActor();
```
ThislsAUsername, line 33: Why didn't you just add a default implementation?

src/AI/Muriel.java
```diff
@@ -219,269 +239,441 @@ public GameAction getNextAction(GameMap gameMap)
...
+    // Flag the actor as an obstacle so the unit it's pushing can't push back.
+    if( !unitSelector.isObstacle(actor) )
+      unitSelector.flagObstacle(actor);
+    obstructedActions.put(actor, desiredAction, actionValue);
+
+    // Give the obstacle priority so it has to move first.
+    unitSelector.flagObstacle(obstacle);
+  }
+  /** Compare ga against any currently-obstructed actions. Return true if it is better
+   * than any of those, and false if an existing known action should take priority. */
```
ThislsAUsername, line 315:
```suggestion
  /** Compare input action against known blocked actions doable from that space. Return true if it is better
   * than any of those, and false if an existing known action should take priority. */
```

src/Test/TestAIMuriel.java
```diff
@@ -168,4 +172,134 @@ private boolean testClearAttackRoute()
...
   }
 
+  /** Two units want to go the same way. One is where the other should end up. Make sure they coordinate. */
+  /** START: [ ] [ ] INF [ ] [ ] INF [ ] [ ] [ ] HQ  */
+  /** GOAL : [ ] [ ] [ ] [ ] [ ] INF [ ] [ ] INF HQ */
+  private boolean testWalkInLine()
+  {
+    // Generate a custom tiny map to keep things simple.
+    TerrainType[] HQCol = {TerrainType.HEADQUARTERS};
+    TerrainType[] GRCol = {TerrainType.GRASS};
```
ThislsAUsername, line 182: Pretty sure we have 2-character abbreviations already defined...

src/Test/TestAIMuriel.java
```diff
@@ -0,0 +1,305 @@
...
+
+    testPassed &= validate(lt.x == 5, "    Muriel failed to move left infantry as intended!");
+    testPassed &= validate(rt.x == 8, "    Muriel failed to move right infantry as intended!");
+
+    // Clean up
+    cleanupTest();
+    return testPassed;
+  }
+
+  /** Two units want to go the same way. One is where the other should end up. Make sure they coordinate. */
```
ThislsAUsername, line 215:
```suggestion
  /** This inf needs to wade through a sea of Md Tanks to get to the HQ. */
```

ThislsAUsername:
Looks pretty good. If you're about done, it might be simpler to get this merged before I get too deep into fixing Wally.

src/Engine/GameAction.java
ThatIsAUsername: I dum

src/Test/TestAIMuriel.java
ThatIsAUsername: We do, but for some reason they are protected members of `IMapBuilder`, so I just did this instead. We could just make those public unless you see a reason not to; they are `static final` anyway, so I'm not really sure why we wouldn't.

ThislsAUsername: I don't see any reason to make them non-public either, lol

src/Engine/GameAction.java
ThislsAUsername: Not necessarily. Having to provide an implementation means you have to think about what it means for each subclass, which might help avoid surprises down the line.
I'm generally in favor of terseness, but letting the compiler help you keep track of stuff can also help.

ThatIsAUsername:
> If you're about done, it might be simpler to get this merged before I get too deep into fixing Wally.

I don't think I'm making any more significant changes at this point. At some point I may move the AI tests into a separate file, run them on all AIs, and set per-AI thresholds on what constitutes "passing" to prevent regressions, and I definitely have more improvements to eventually make, but that's no hurry.

PR #337, opened 2020-12-28
Fonttweaks

ThatIsAUsername:
The font should be sized correctly on both Windows and Linux now.
Most of the glyphs are narrower, and the font is a pixel shorter, resulting in substantially improved space efficiency.
There are instructions in `res/dpfont_glyphs.txt` for how to regenerate the font file if needed.
I added another space of indent to the info pages as well, since it looks cleaner that way.
I also reworked Strong's InfoPage, and would suggest we modify others to match. In short:
- Put numbers first (I also made them bold in the font so they stand out more clearly. AW did this too).
- More brevity (and less unnecessary verbiage to explain things that can easily be inferred from the implicit context on the page).
- No leading dashes - we currently have a mix of with/without dashes, and with paragraph indents we don't need 'em.


src/CommandingOfficers/Strong.java
```diff
@@ -30,29 +30,27 @@ public instantiator()
...
-          "Strong can build infantry on air/sea ports\n"));
+          "Can build infantry on air/sea ports\n"));
       infoPages.add(new InfoPage(
           "Mobilize ("+MobilizeAbility.MOBILIZE_COST+"):\n" + 
           "+"+MobilizeAbility.MOBILIZE_BUFF+"% attack for all units\n" +
           "+"+MobilizeAbility.MOBILIZE_DEFENSE_BUFF+"% defense for all units\n" +
           "+2 movement for all units\n" +
-          "Refreshes footsoldiers\n" +
-          "Strong can build footsoldiers on cities, industries, and the HQ\n"));
+          "All footsoldiers can move again\n" +
```
ThislsAUsername, line 52: I feel like "refresh" is maybe a better term:
"Move again" doesn't necessarily imply *act* again (and vice versa)
"Refresh" sounds like a thing that happens instantly on activation, while "can move again" could easily mean "get refreshed upon acting the first time"

src/CommandingOfficers/Ave.java
```diff
@@ -65,7 +65,7 @@
   public static final int SNOW_THRESHOLD = 100; // Big numbers for integer math.
   public static final int SNOW_PER_TURN = 400;
   public static final int SNOW_MELT_RATE = 100;
-  private int MAX_SNOW_SPREAD_RANGE = 2;
+  private int MAX_SNOW_SPREAD_RANGE = 1;
```
ThislsAUsername, line 68: Any particular reason for reducing her radii by 1?

src/CommandingOfficers/Ave.java
```diff
@@ -78,28 +78,28 @@ public instantiator()
...
           "Commander Ave (AH-vey) supports her skiing habit by slowly growing a mountain of fresh " +
-          "snow around each of her buildings. This also allows her to gradually but inexorably " +
+          "snow around each of her buildings. This allows her to gradually but inexorably " +
           "grind her opponents down beneath a wall of ever-encroaching ice."));
       infoPages.add(new InfoPage(
           "Passive:\r\n" +
-          "- Ave generates snow around all owned properties, which spreads over time.\n" +
-          "- The radius of effect is small at first, but can be expanded by her abilities.\n" +
-          "- Her units move normally in snow, but take a movement and defense penalty in forests."));
+          "Ave Generates snow around all owned properties, which spreads over time.\n" +
```
ThislsAUsername, line 85: Why capitalize "Generates"?

src/CommandingOfficers/Patch.java
```diff
@@ -24,16 +24,16 @@ public instantiator()
       infoPages.add(new InfoPage(
           "Commander Patch is a pirate, who does piratey things like lootin' and plunderin'\n"));
       infoPages.add(new InfoPage(
-          "Passive:\n" + 
-          "- Patch gets a turn's worth of income from any property he captures\n"));
+          "Passive:\n" +
+          "1 extra turn of income when capturing any property\n"));
```
ThislsAUsername, line 28:
```suggestion
          "Instantly gain an extra turn of income on capturing a property\n"));
```

src/CommandingOfficers/Tech.java
```diff
@@ -49,21 +49,21 @@ public instantiator()
...
+              "Deploys " + TECHDROP_NUM + " BattleMech to the front lines\n" +
+              "+" + TECHDROP_BUFF + "% attack for all mechanical units\n" +
+              "+" + TECHDROP_BUFF + "% defense for all units\n" +
+              "\nNOTE: BattleMechs require special parts and cannot be repaired on buildings\n"));
       infoPages.add(new InfoPage(
           OVERCHARGE_NAME + " (" + OVERCHARGE_COST + "):\n" +
-              "All mechanical units are repaired by " + OVERCHARGE_HEAL + " HP, allowing more than 10 HP for this turn.\n" +
-              "All mechanical units gain " + OVERCHARGE_BUFF + "% attack.\n" +
-              "All Units gain " + OVERCHARGE_BUFF + "% defense"));
+              "Repairs all mechanical units by " + OVERCHARGE_HEAL + " HP, allowing more than 10 HP for this turn\n" +
```
ThislsAUsername, line 58:
```suggestion
              "+" + OVERCHARGE_HEAL + " HP to mechanical units, allowing more than 10 HP for this turn\n" +
```
Also, I feel like it's worth making it standard that "repair" means "liable for costs" and "heal" means non-liable?

src/CommandingOfficers/Strong.java
ThatIsAUsername: "Move again" is the verbiage used on the carts, so I'm not worried about it being understood. Plus moving without being able to act just... isn't a thing.
"Refresh" can work too, but it could also just mean to pass around the lemonade.
Potato potato, really.

src/CommandingOfficers/Ave.java
ThatIsAUsername: She snowballs too hard otherwise?

src/CommandingOfficers/Ave.java
ThatIsAUsername: Because oops.

src/CommandingOfficers/Ave.java
ThatIsAUsername: I guess this doesn't need to be a balance patch, and I should probably acquire some actual data before nerfing. Besides, someone needs to hold Tech in check.

src/CommandingOfficers/Strong.java
ThislsAUsername: > "Move again" is the verbiage used on the carts, so I'm not worried about it being understood. Plus moving without being able to act just... isn't a thing.

It's not a thing as the game is now, but it's a reasonably common proposed custom unit/CO mechanic... though in general, I think finding other ways to implement it than "you can move again, but not do anything other than wait" would be preferable for clarity/clean design purposes.

> "Refresh" can work too, but it could also just mean to pass around the lemonade.

Hah, but "resupply" (or even rearm/refuel) and "heal" are their own separate and unambiguous terms. If it wasn't meaningful re: game mechanics, it wouldn't be in the description.

src/CommandingOfficers/Ave.java
ThislsAUsername: You have also received a user request specifically to not nerf her, if I recall correctly, lol.

ThislsAUsername:
Looking at the numbers some more, I don't mind the bolding as much as I thought.
What I did mind was the zero being so... full.
dpfont_glyphs

src/CommandingOfficers/Ave.java
ThatIsAUsername: Well, better not disappoint the fans then.

src/CommandingOfficers/Ave.java
```diff
@@ -46,8 +46,8 @@
  *    Ave's units gain a 10 percent increase in firepower.
  *
  * Glacio:
- *    Increases the snow-aura around her buildings by 2 spaces for the next turn.
- *    Snows on every tile in a 2-space radius around each of her units.
+ *    Increases the snow-aura around her buildings by 3 spaces for the next turn.
+ *    Snows on every tile in a 3-space radius around each of her units.
```
ThislsAUsername: You appear to have missed updating the code for this.

src/CommandingOfficers/Strong.java
ThatIsAUsername: I think if we ever add the ability to move-shoot-move, we would need to clarify that in the ability description anyway. Anyone who's played AW will expect "unit can move again" to entail also acting, unless you specify "Can move once more after acting"
So I'm super not worried about this being misinterpreted; this seems like another instance of you trying to impose narrow distinctions on similar terms that aren't defined clearly anywhere (and in this case, your new meaning contradicts the commonly-understood one). That said, if you prefer to use "Refresh", I'm not gonna get too bothered over the precise verbiage.

> Hah, but "resupply" (or even rearm/refuel) and "heal" are their own separate and unambiguous terms
"Resupply" and "heal" can already be done independently on the carts; being able to move and not act cannot.

src/CommandingOfficers/Tech.java
ThatIsAUsername: I think I'm fine with that convention.

src/CommandingOfficers/Ave.java
ThatIsAUsername: yarr

ThatIsAUsername:
> Looking at the numbers some more, I don't mind the bolding as much as I thought.
> What I did mind was the zero being so... full.
> Suggested tweak:

Hmm. I'm not strictly opposed, though I might point out that the AW font has bolded zeros too. It looks a little strange seeing non-bolded zeros as part of large numbers with bolded digits, but the non-bolded zero does look better on its own.

Actually, what if we just remove the slash?
dpfont_glyphs0
Screenshot from 2021-01-08 14-49-49

ThislsAUsername:
> Actually, what if we just remove the slash?

I'm down with that. Otherwise ready to merge?

PR #339, opened 2021-01-25
Wally

ThislsAUsername:
Wally can actually sometimes win VS Muriel now. Huzzah.

re: the build clogging - it's occurring because Wally isn't always able to evict everything he built last turn.
The new OHKO dealie helps with that, and so does "if bad thing, don't do the bad thing".
I could fix that "properly" by testing for successful eviction and modifying build choices on that, but I feel like that's an oddball-enough occurrence to not be worth leaning into.

Probably a WIP, since I don't think I actually got to the things we nominally wanted to fix before release.
One day, I'll figure that out. Probably.

### DIFF
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -150,18 +150,21 @@ public class AICombatUtils
    * <p>Prunes any unnecessary attacks and empty spaces from the return value.
    * <p>Does not consider launchable units
    * @param target Your victim
-   * @param co Whose turn it is now
-   * @param attackCandidates The units CO is willing and able to commit
+   * @param attackCandidates The units you are willing and able to commit
    * @param excludedSpaces Any tiles you don't want to consider attacks from
    * @return The lethal combination of units organized by strike location, or null on failure.
    */
   public static HashMap<XYCoord, Unit> findMultiHitKill(
                                    GameMap gameMap, Unit target,
-                                   Commander co, Collection<Unit> attackCandidates,
+                                   Collection<Unit> attackCandidates,
                                    Collection<XYCoord> excludedSpaces)
   {
     if( target.getHP() < 1 ) // Try not to pick fights with zombies
       return null;
+    if( attackCandidates.size() < 1 )
+      return null;
+
+    Commander co = attackCandidates.iterator().next().CO;
 
     final int minRange = 1;
     final XYCoord targetCoord = new XYCoord(target);
@@ -237,7 +240,7 @@ public class AICombatUtils
       neededAttacks.put(xyc, null);
     }
 
-    double damage = findMultiHitKill(gameMap, attackers, neededAttacks, target, 0);
+    double damage = findMultiHitKill(gameMap, target, attackers, neededAttacks, 0);
     if( damage >= target.getHP() )
     {
       // Prune excess attacks and empty attacking spaces
@@ -269,12 +272,16 @@ public class AICombatUtils
    * Attempts to find a combination of attacks that will create a kill.
    * Considers units in the order provided.
    * Recursive.
-   * @param unitQueue The set of potential attackers
+   * @param attackCandidates The set of potential attackers
    * @param neededAttacks The set of locations to consider, pre-populated with any mandatory attacks, to be populated
    * @param pDamage The cumulative base damage done by those mandatory attacks
-   * @return The cumulative base damage of all attacks in the neededAttacks
+   * @return The cumulative base damage of all attacks already in the neededAttacks
    */
-  public static double findMultiHitKill(GameMap gameMap, Collection<Unit> unitQueue, Map<XYCoord, Unit> neededAttacks, Unit target, double pDamage)
+  public static double findMultiHitKill(
+                                GameMap gameMap, Unit target,
+                                Collection<Unit> attackCandidates,
+                                Map<XYCoord, Unit> neededAttacks,
+                                double pDamage)
   {
     // Base case; we found a kill
     if( pDamage >= target.getPreciseHP() )
@@ -290,7 +297,7 @@ public class AICombatUtils
       if( null != neededAttacks.get(xyc) )
         continue;
 
-      Queue<Unit> assaultQueue = new ArrayDeque<Unit>(unitQueue);
+      Queue<Unit> assaultQueue = new ArrayDeque<Unit>(attackCandidates);
       while (!assaultQueue.isEmpty())
       {
         Unit unit = assaultQueue.poll();
@@ -309,7 +316,7 @@ public class AICombatUtils
           neededAttacks.put(xyc, unit);
           double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc.xCoord, xyc.yCoord).defenderHPLoss;
 
-          thisDamage = findMultiHitKill(gameMap, unitQueue, neededAttacks, target, damage + thisDamage);
+          thisDamage = findMultiHitKill(gameMap, target, attackCandidates, neededAttacks, damage + thisDamage);
 
           // If we've found a kill, we're done
           if( thisDamage >= target.getPreciseHP() )
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -1,0 +1,142 @@
+package AI;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import CommandingOfficers.Commander;
+import Engine.GameAction;
+import Engine.Utils;
+import Engine.XYCoord;
+import Engine.Combat.BattleSummary;
+import Engine.Combat.CombatEngine;
+import Engine.Combat.StrikeParams;
+import Terrain.GameMap;
+import Terrain.Location;
+import Terrain.TerrainType;
+import Units.Unit;
+import Units.UnitModel;
+import Units.WeaponModel;
+
+public class AICombatUtils
+{
+  /**
+   * Evaluates an attack action based on caller-provided logic
+   * @param unit The attacking unit
+   * @param action The action to evaluate
+   * @param map The user's current game knowledge
+   * @param combatScorer Evaluates combat with a unit
+   * @param demolishScorer Evaluates targeting terrain
+   * @return
+   */
+  public static double scoreAttackAction(Unit unit, GameAction action, GameMap map,
+                                         Function<BattleSummary, Double> combatScorer,
+                                         BiFunction<TerrainType, StrikeParams, Double> demolishScorer)
+  {
+    double score = 0;
+    Location targetLoc = map.getLocation(action.getTargetLocation());
+    Unit targetUnit = targetLoc.getResident();
+    if( null != targetUnit )
+    {
+      BattleSummary results = CombatEngine.simulateBattleResults(unit, targetUnit, map, action.getMoveLocation());
+      score = combatScorer.apply(results);
+    }
+    else
+    {
+      StrikeParams params = CombatEngine.calculateTerrainDamage(unit,
+          Utils.findShortestPath(unit, action.getMoveLocation(), map), targetLoc, map);
+      score = demolishScorer.apply(targetLoc.getEnvironment().terrainType, params);
+    }
+
+    return score;
+  }
+
+  /**
+   * @return The area and severity of threat from the unit, against the specified target type
+   */
+  public static Map<XYCoord, Double> findThreatPower(GameMap gameMap, Unit unit, UnitModel target)
+  {
+    XYCoord origin = new XYCoord(unit.x, unit.y);
+    Map<XYCoord, Double> shootableTiles = new HashMap<XYCoord, Double>();
+    boolean includeOccupiedDestinations = true; // We assume the enemy knows how to manage positioning within his turn
+    ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeOccupiedDestinations);
+    for( WeaponModel wep : unit.model.weapons )
+    {
+      double damage = (null == target)? 1 : wep.getDamage(target) * unit.getHPFactor();
+      if( damage > 0 )
+      {
+        if( !wep.canFireAfterMoving )
+        {
+          for (XYCoord xyc : Utils.findLocationsInRange(gameMap, origin, wep.minRange, wep.maxRange))
+          {
+            double val = damage;
+            if (shootableTiles.containsKey(xyc))
+              val = Math.max(val, shootableTiles.get(xyc));
+            shootableTiles.put(xyc, val);
+          }
+        }
+        else
+        {
+          for( XYCoord dest : destinations )
+          {
+            for (XYCoord xyc : Utils.findLocationsInRange(gameMap, dest, wep.minRange, wep.maxRange))
+            {
+              double val = damage;
+              if (shootableTiles.containsKey(xyc))
+                val = Math.max(val, shootableTiles.get(xyc));
+              shootableTiles.put(xyc, val);
+            }
+          }
+        }
+      }
+    }
+    return shootableTiles;
+  }
+
+  /**
+   * @return The range at which the CO in question might be able to attack after moving.
+   */
+  public static int findMaxStrikeWeaponRange(Commander co)
+  {
+    int range = 0;
+    for( UnitModel um : co.unitModels )
+    {
+      for( WeaponModel wm : um.weapons )
+      {
+        if( wm.canFireAfterMoving )
+          range = Math.max(range, wm.maxRange);
+      }
+    }
+    return range;
+  }
+
+  /** Return the set of locations with enemies or terrain that `unit` could attack in one turn from `start` */
+  public static Set<XYCoord> findPossibleTargets(GameMap gameMap, Unit unit, XYCoord start, boolean includeTerrain)
+  {
+    Set<XYCoord> targetLocs = new HashSet<XYCoord>();
+    boolean allowEndingOnUnits = false; // We can't attack from on top of another unit.
+    ArrayList<XYCoord> moves = Utils.findPossibleDestinations(start, unit, gameMap, allowEndingOnUnits);
+    for( XYCoord move : moves )
+    {
+      boolean moved = !move.equals(start);
+
+      for( WeaponModel wpn : unit.model.weapons )
+      {
+        // Evaluate this weapon for targets if it has ammo, and if either the weapon
+        // is mobile or we don't care if it's mobile (because we aren't moving).
+        if( wpn.loaded(unit) && (!moved || wpn.canFireAfterMoving) )
+        {
+          ArrayList<XYCoord> locations = Utils.findTargetsInRange(gameMap, unit.CO, move, wpn, includeTerrain);
+          targetLocs.addAll(locations);
+        }
+      } // ~Weapon loop
+    }
+    targetLocs.remove(start); // No attacking your own position.
+    return targetLocs;
+  }
+
+}
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -1,5 +1,6 @@
 package AI;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -104,7 +105,7 @@ public class AICombatUtils
   /**
    * @return The range at which the CO in question might be able to attack after moving.
    */
-  public static int findMaxStrikeWeaponRange(Commander co)
+  public static int findMaxMobileWeaponRange(Commander co)
   {
     int range = 0;
     for( UnitModel um : co.unitModels )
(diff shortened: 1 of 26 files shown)


src/Engine/GameInstance.java
```diff
@@ -287,81 +284,43 @@ private String getSaveName()
...
-    catch (IOException ex)
-    {
-      System.out.println(ex.toString());
-    }
-
-    return filename;
+    // Method for serialization of object
+    out.writeObject(new GameVersion());
+    out.writeObject(this);
+    currentTurnEnded = temp;
```
ThatIsAUsername, line 323: Why restore this value? Is there any situation where we would want to save the game at turn-end without ending the turn?

src/Engine/SerializationUtils.java
```diff
@@ -0,0 +1,64 @@
+package Engine;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+
+public class SerializationUtils
```
ThatIsAUsername, line 10: Why are we making changes to game load/save functionality as part of a Wally QoL PR?

Also why create this at all? It looks like it just creates an extra layer of boilerplate before passing through to the GameInstance functions anyway.

src/Test/TestSaveLoad.java
```diff
@@ -59,14 +62,46 @@ private boolean testSaveLoad()
...
+         ObjectInputStream in = new ObjectInputStream(bais);)
+    {
+      in.readObject(); // Pull out and discard our version info
+      loaded = (GameInstance) in.readObject();
+    }
+    catch (Exception ex)
+    {
+      System.out.println(ex.toString());
+      testPassed = false;
+    }
```
ThatIsAUsername, line 102: What was the goal of the save/load changes? Whatever it was, "ease of use" was apparently not on the radar, judging by the amount of extra boilerplate here to do the same thing.
This doesn't even use the new class you added either (which would admittedly explain why this got more complicated, but doesn't explain why the new class was needed in the first place, or why it isn't being tested).

src/Test/TestAIMuriel.java
```diff
@@ -12,29 +17,30 @@
...
 import Terrain.TerrainType;
+import Terrain.Maps.MapReader;
 import Units.Unit;
 
 public class TestAIMuriel extends TestCase
 {
   private static Commander testCo1;
   private static Commander testCo2;
-  private static Muriel testAI;
+  private static AIMaker[] ais = { Muriel.info, WallyAI.info };
```
ThatIsAUsername, line 27: No love for ISAI/Spender?

src/Test/TestAIMuriel.java
```diff
@@ -55,37 +60,65 @@ private void cleanupTest()
...
-    testPassed &= validate(testClearAttackRoute(), "  Route clearing test failed.");
-    testPassed &= validate(testWalkInLine(), "  Line walking test failed.");
-    testPassed &= validate(testInfWadeThroughTanks(), "  Infantry move priority test failed.");
-    testPassed &= validate(testTankWadeThroughInfs(), "  Tank move priority test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testBuildMegatank(ai)), "  Build Megatank test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testHuntStall(ai)), "  Hunting stall-test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testClearAttackRoute(ai)), "  Route clearing test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testWalkInLine(ai)), "  Line walking test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testInfWadeThroughTanks(ai)), "  Infantry move priority test failed.");
+    testPassed &= validate(testTankWadeThroughInfs(Muriel.info), "  Tank move priority test failed.");
```
ThatIsAUsername, line 68: Why not apply all AIs to all challenges?

src/Test/TestAIMuriel.java
```diff
@@ -55,37 +60,65 @@ private void cleanupTest()
   public boolean runTest()
   {
     boolean testPassed = true;
-    testPassed &= validate(testBuildMegatank(), "  Build Megatank test failed.");
-    testPassed &= validate(testHuntStall(), "  Hunting stall-test failed.");
-    testPassed &= validate(testClearAttackRoute(), "  Route clearing test failed.");
-    testPassed &= validate(testWalkInLine(), "  Line walking test failed.");
-    testPassed &= validate(testInfWadeThroughTanks(), "  Infantry move priority test failed.");
-    testPassed &= validate(testTankWadeThroughInfs(), "  Tank move priority test failed.");
+    testPassed &= validate(testForAllAIs((ai) -> testBuildMegatank(ai)), "  Build Megatank test failed.");
```
ThatIsAUsername, line 63: Really?
Look, I get that lambdas are like, the most excellent thing ever invented in the history of programming, but here is yet another case where it probably isn't warranted.

Granted, it's super cool to be able to just say "run this test on all the AIs", but on the other hand, that ties their success together. Which, if I recall the initial vision for this capability, is exactly not what we want. We want to set per-AI thresholds (Muriel must pass 100% of existing tests, Wally must pass 85%, ISAI must pass 2%, whatever) so that we can catch regressions when we go tweaking things.

And I know you are now going to go and simply reverse your lambda, but really - a simple loop is all we need here.

src/Test/TestAIMuriel.java
```diff
@@ -55,37 +60,65 @@ private void cleanupTest()
...
+  {
+    setupTest(ai);
 
     // Add an enemy to counter.
     addUnit(testMap, testCo2, "Megatank", 1, 1);
 
-    // Put Muriel in control
-    testCo1.setAIController(testAI);
-
     // Give Muriel resources.
```
ThatIsAUsername, line 97: "Muriel" still referenced all over.

src/Test/TestAIMuriel.java
```diff
@@ -55,37 +60,65 @@ private void cleanupTest()
...
     testCo1.initTurn(testMap);
-    GameAction act = testCo1.getNextAIAction(testMap);
-    performGameAction(act, testGame);
+    GameAction act = null;
+    boolean testPassed = true;
+    do
+    {
+      act = testCo1.getNextAIAction(testMap);
+      if( null != act )
+        testPassed &= validate(performGameAction(act, testGame), "    "+ai.getName()+" generated a bad action!");
```
ThatIsAUsername, line 108: if `null == act` seems like it should also be a test failure.
Also why turn this into a loop? There's like, nothing that can be done except unit production anyway. One action should be all that is required.

ThatIsAUsername: Unless Wally is simply building other things first instead - which may be forgivable. This might be a case for 1) reducing the available funds to force the issue, and b) this being more of a Muriel-specific tests - who's to say that Artillery isn't the correct response, generally speaking? If we are creating an AI obstacle course, this test probably doesn't belong there.

src/Test/TestAIMuriel.java
```diff
@@ -94,9 +127,9 @@ private boolean testBuildMegatank()
   }
 
   /** Confirm that Muriel will build the correct counter, even when there is only one possible counter. */
```
ThatIsAUsername, line 129: Wrong Javadoc. My bad.

src/Test/TestAIMuriel.java
```diff
@@ -112,24 +145,30 @@ private boolean testHuntStall()
...
     // Ask Muriel what to do.
     testCo1.initTurn(testMap);
-    GameAction act = testCo1.getNextAIAction(testMap);
-    performGameAction(act, testGame);
+    GameAction act = null;
+    do
+    {
+      act = testCo1.getNextAIAction(testMap);
+      if( null != act )
+        testPassed &= validate(performGameAction(act, testGame), "    "+ai.getName()+" generated a bad action!");
```
ThatIsAUsername, line 153: Same set of objections to this being a) a loop when it probably isn't needed, and 2) not being "universally true" in that an AI can be working correctly even when not "passing" this test.

src/Test/TestAIMuriel.java
```diff
@@ -144,14 +183,14 @@ private boolean testClearAttackRoute()
     // Where are things?
     XYCoord facPos = new XYCoord(7, 8);
 
-    // Add an enemy tank on a neutral fac, flanked by friendly infs, with a friendly tank nearby.
-    Unit nmeTank = addUnit(testMap, testCo2, "Tank", facPos);
-    Unit myTank = addUnit(testMap, testCo1, "Md Tank", facPos.up().left());
+    // Add an enemy copter on a neutral fac, flanked by friendly infs, with a friendly AA nearby.
+    Unit nmeCopter = addUnit(testMap, testCo2, "B-Copter", facPos);
```
ThatIsAUsername, line 187: Why change the units around?

src/AI/AICombatUtils.java
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * Attempts to find a combination of attacks that will create a kill.
+   * Recursive.
+   * @param unitQueue The set of potential attackers
+   * @param neededAttacks The set of locations to consider, pre-populated with any mandatory attacks, to be populated
+   * @param pDamage The cumulative base damage done by those mandatory attacks
+   * @return The cumulative base damage of all attacks in the neededAttacks
+   */
+  public static double findAssaultKill(GameMap gameMap, Collection<Unit> unitQueue, Map<XYCoord, Unit> neededAttacks, Unit target, double pDamage)
```
ThatIsAUsername, line 216: I can see merit in breaking some of these functions into a separate file, but I'm not sure `findAssaultKill` really fits into the "utility" category. It contains logic specific to Wally's operation - there isn't necessarily one "right" answer, so it's not the kind of question I would normally put in a utility class.

src/Test/TestAIMuriel.java
```diff
@@ -289,17 +329,52 @@ private boolean testTankWadeThroughInfs()
...
 
     // Clean up
     cleanupTest();
     return testPassed;
   }
 
+  /** Confirm that the AI will clear its factory to build a counter, when there is only one possible counter. */
+  private boolean testProductionClearing(AIMaker ai)
+  {
+    setupTest(MapReader.readSingleMap("src/Test/TestProductionClearing.map"), ai);
```
ThatIsAUsername, line 348: This map does not exist. Not sure we want to create a bunch of map files just for simple test cases, but it might be amusing to have them as actual files so people can try them too.

src/Engine/GameInstance.java
ThislsAUsername: So it's a pure function. It generates a valid game save, and doesn't affect the object you called it on.

src/Engine/SerializationUtils.java
ThislsAUsername: To abstract the direct file I/O out of GameInstance, because writing an unnecessary file to test game save logic drives me batty.

src/Test/TestSaveLoad.java
ThislsAUsername: See above.

src/Test/TestAIMuriel.java
ThislsAUsername: They'd just fail all the tests anyway, so...

src/Test/TestAIMuriel.java
ThislsAUsername: I suppose it would be roughly as concise (and probably more clear, really) to just wrap the tests in a loop.

src/Test/TestAIMuriel.java
ThislsAUsername: And in the file name. I did mention this is a WIP, yes?

src/Test/TestAIMuriel.java
ThislsAUsername: Yeah. Wally built an inf first, if I recall correctly.
Artillery doesn't answer a Megatank. It just doesn't do enough damage, and is too vulnerable (and has too little threat area).
Bombers could work, but there's no airport.

If you did want to force the issue, I'd suggest providing only one factory.

src/Test/TestAIMuriel.java
ThislsAUsername: Not super worried about documentation on tests, to be perfectly honest.

src/Test/TestAIMuriel.java
ThislsAUsername: Added the loop 'cause Wally builds stuff before allocating travel, and order isn't really important there.
Blocking your own factory is *almost always* real dumb, and isn't something I'd ever expect an AI to do intelligently... and to counter a *Recon* slightly better? Nah.

src/Test/TestAIMuriel.java
ThislsAUsername: To change the value proposition, since "judge this a good attack to make" doesn't seem to be a core part of the test (if I thought it was, I'd have thrown it out of scope for Wally).
Wally only wants to go aggro if he can deal about as much damage as he'd take by losing the unit... and this new unit pair satisfies that.

He should, at some point, figure out that you shouldn't count the unit you're about to first-strike as a significant threat to counterattack a play, but that's more AI juice than I think I want to delay the next build for.

src/AI/AICombatUtils.java
ThislsAUsername: Why not? It finds the cheapest combination of units to clear a particular tile. There is a huge quantity of applications for that.

src/Test/TestAIMuriel.java
ThislsAUsername: ...I coulda sworn I added it. Derp.
IMO it's a lot less effort (and a lot clearer) to use our map format than to hand-code the toy maps.
Encodes terrain/ownership/units in a concise and readable format.

src/AI/AICombatUtils.java
ThatIsAUsername: Specifically the cheapest combination of units that will do the job? This should be clarified in the docstring (along with any other assumptions/requirements/context that is needed to use this properly).
Also not sure that `findAssaultKill` is really the best name for this, but eh.

src/Terrain/MapMaster.java
```diff
@@ -66,7 +66,7 @@ public MapMaster(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
             // If the CO has no HQ yet, assign this one.
             if( COs[co].HQLocation == null )
             {
-              System.out.println("Assigning HQ at " + x + ", " + y + " to " + COs[co]);
+//              System.out.println("Assigning HQ at " + x + ", " + y + " to " + COs[co]);
```
ThatIsAUsername, line 69: Hmm.

src/Test/TestAIMuriel.java
ThatIsAUsername: I guess. Not sure if we really want a bunch of tiny maps hanging around--but then that might make it easier to let a human try out the AI tests. That could be fun, especially once we make some more complex ones.

Also not sure why you created this test case at all since it's basically a slightly more complicated version of the other megatank test, but eh.

src/Test/TestAIMuriel.java
ThatIsAUsername: sigh

src/Test/TestAIMuriel.java
ThatIsAUsername: If we really want to test "Building Megatank as the correct counter" then I would either reduce budget or factory count to ensure the AI performs correctly under the most constrained conditions.

src/Test/TestAIMuriel.java
ThatIsAUsername: They would. But it would be funny.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -296,7 +296,7 @@ public void drawUnitIcons(Graphics g, Commander[] COs, Unit u, int animIndex, in
     if( !statusIcons.isEmpty() )
     {
       int iconIndex = (animIndex%(statusIcons.size()*ANIM_FRAMES_PER_MARK))/ANIM_FRAMES_PER_MARK;
-      BufferedImage statusIcon = statusIcons.get(iconIndex);
+      BufferedImage statusIcon = statusIcons.get(Math.abs(iconIndex));
```
ThatIsAUsername, line 299: Wasn't this issue fixed elsewhere already?

src/Test/TestAIMuriel.java
ThislsAUsername: Eh... fiiiiiine

src/AI/AICombatUtils.java
ThislsAUsername: I was trying to make "Assault" and "Siege" terms for "units that can attack after moving" and "can't", since (in)direct is specifically about attack range.
You're right, though; it now happily uses indirects.

src/Terrain/MapMaster.java
ThislsAUsername: I figure we had it in there for some reason. I just got annoyed by seeing it spammed, lol

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThislsAUsername: It was... but not on this branch, and I ran into it during testing. Guess I accidentally checked it in.

src/AI/AICombatUtils.java
ThatIsAUsername: Even aside from the fact that the function can use indirects... You can't just assign precise, domain-specific meaning to terms and then expect everyone to understand, especially when you don't record your definitions anywhere. You seem to do this often though, so I'm going to keep pointing it out.
It's one thing if you do it inside an AI, and that's the only place the term matters (though I'd still say you should treat it as public code as far as style goes); it's quite another if you are putting it in a utility for the world to misunderstand.

ThislsAUsername: > You seem to do this often though, so I'm going to keep pointing it out.

Fair enough.
I suppose "(im)mobile weapon" works for most use cases, and is more clear.

Not having a concise term for "move and fire indirect" bugs me, though.
"Mobile indirect" doesn't work, for obvious reasons.
"Assault indirect" seems like a reasonable alternative to me, and "siege unit/weapon" is pretty obvious on its own.

> it's quite another if you are putting it in a utility for the world to misunderstand.

I suppose that's why we have code review, isn't it?

src/Test/TestAIMuriel.java
ThislsAUsername: > I guess. Not sure if we really want a bunch of tiny maps hanging around

The test maps will be in the test directory, not the general maps directory (unless we really want that).
The map *will* be defined somewhere, one way or the other. The question is basically how readable/reusable it is.

src/Test/TestAIMuriel.java
ThislsAUsername: It might be funny, if it didn't cause a test failure and prevent starting the game, lol

src/AI/AICombatUtils.java
```diff
@@ -207,13 +267,14 @@ public static int findMaxStrikeWeaponRange(Commander co)
...
    * Attempts to find a combination of attacks that will create a kill.
+   * Considers units in the order provided.
    * Recursive.
    * @param unitQueue The set of potential attackers
    * @param neededAttacks The set of locations to consider, pre-populated with any mandatory attacks, to be populated
    * @param pDamage The cumulative base damage done by those mandatory attacks
    * @return The cumulative base damage of all attacks in the neededAttacks
    */
-  public static double findAssaultKill(GameMap gameMap, Collection<Unit> unitQueue, Map<XYCoord, Unit> neededAttacks, Unit target, double pDamage)
+  public static double findMultiHitKill(GameMap gameMap, Collection<Unit> unitQueue, Map<XYCoord, Unit> neededAttacks, Unit target, double pDamage)
```
ThatIsAUsername, line 277: The two functions are both public, but one considers units in cost order and the other in prior order? They are also a bit cavalier in terms of parameter ordering or naming. One is
GameMap, target, Commander, attackCandidates, excludedSpaces; the other is
GameMap, unitQueue, neededAttacks, target, pDamage

src/AI/AICombatUtils.java
```diff
@@ -144,40 +145,99 @@ public static int findMaxStrikeWeaponRange(Commander co)
...
-                                   minRange, findMaxStrikeWeaponRange(co));
+                                   targetCoord,
+                                   minRange, findMaxMobileWeaponRange(co))
+        );
 
-    HashMap<XYCoord, Unit> neededAttacks = new HashMap<XYCoord, Unit>();
-    // Figure out where we can attack from, and include attackers already in range by default.
-    for( XYCoord xyc : coordsToCheck )
+    // Consider the cheapest units first.
+    PriorityQueue<Unit> attackers = new PriorityQueue<Unit>(11, new AIUtils.UnitCostComparator(true));
```
ThatIsAUsername, line 175: Seems like you just changed quite a bit more than simply adding docs. Which is always a bit nervous right before merging in a thing.
I'm sure the unit tests for this function still pass though, so I'm sure there's no real cause for alarm.

src/Engine/XYCoord.java
```diff
@@ -13,6 +15,10 @@ public XYCoord(int x, int y)
     xCoord = x;
     yCoord = y;
   }
+  public XYCoord(Unit u)
```
ThatIsAUsername, line 18: Interesting. Why this instead of `unit.getCoord()`?

src/AI/AICombatUtils.java
ThislsAUsername: > The two functions are both public, but one considers units in cost order and the other in prior order?

Right. The first one is basically a more user-friendly interface to the other function.
I'm not opposed to changing assumptions and such (let the caller pick the ordering or whatever).
The second function is public because it allows you to do stuff like partial calcs: "I know I want to use this unit to hit that one, but can I finish it off this turn?"

Your point about naming/order is fair, though.

src/Engine/XYCoord.java
ThislsAUsername: No reason, really.

src/AI/AICombatUtils.java
ThislsAUsername: Yeah, I did. Made Wally take under 40 seconds for all the current defaults in Fight Club, too.
And yes, the AI tests all still pass.

ThislsAUsername: ...I suppose there aren't any real *unit* tests, but it'd probably be more work than it's worth to set up a proper unit test.

ThatIsAUsername: Cool. And yeah, probably.

ThatIsAUsername:
I guess this is probably ready to go, unless you feel like double-checking any unresolved conversations.

PR #341, opened 2021-03-01
Status QoL

ThislsAUsername:
For reference: https://discord.com/channels/541366642127536130/541366642739773441/810965053339074610

Screenshot from 2021-03-01 00-45-16

Screenshot from 2021-03-01 00-45-31

Untitled


src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -51,36 +82,47 @@ public static void drawCommanderOverlay(Graphics g, Commander commander, boolean
...
       else
       {
         // We are drawing an ability name, and we don't want to flip it.
-        g.drawImage( powerBarImage, pbXPos, (overlayImage.getHeight()) - (POWERBAR_BUFFER), powerBarImage.getWidth(), powerBarImage.getHeight(), null );
+        g.drawImage( powerBarImage, pbXPos, (overlayImage.getHeight()+verticalOffset) - (POWERBAR_BUFFER), powerBarImage.getWidth(), powerBarImage.getHeight(), null );
       }
     }
   }
 
+  private static final int ICON_VALUE_SPACING = SpriteLibrary.baseIconSize + 2;
```
ThatIsAUsername, line 118: This kinda looks like it's declared here to be used in the function below, but it's only used in the function above. Maybe decrease the scope on this guy?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -7,35 +7,66 @@
...
   private static int animIndex = 0;
   private static long animIndexUpdateTime = 0;
   private static final int animIndexUpdateInterval = 8;
   
   private static Map<String, Sprite> activeAbilityTextSprites = new HashMap<String, Sprite>();
 
+  /**
+   * Draws an overlay for all COs passed in, left-justified, with status text and optionally highlights one CO
+   */
+  public static BufferedImage drawAllCommanderOverlays(Commander[] coList, GameMap drawableMap, int drawingWidth, int drawingHeight, Commander coToHighlight)
```
ThatIsAUsername, line 27: When it switches to drawing all overlays, and an ability is primed, the pip pulses much more slowly than it does when only a single CO overlay is being shown.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -15,9 +15,7 @@
   public static final int OVERLAY_WIDTH = 97;
   public static final int OVERLAY_HEIGHT = 20;
 
-  private static int animIndex = 0;
-  private static long animIndexUpdateTime = 0;
-  private static final int animIndexUpdateInterval = 8;
+  private static final int animIndexUpdateInterval = 13;
```
ThatIsAUsername, line 18: Why change from 8 to 13?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -143,7 +132,7 @@ private static void updateAnimIndex()
   public static BufferedImage buildCoPowerBar(Commander co, double[] abilityPoints, double currentPower, boolean leftSide)
   {
     final double pixelsPerPowerUnit = 3.0;
-    int slowAnimIndex = (animIndex/32) % 2;
+    int slowAnimIndex = (getAnimIndex()/32) % 2;
```
ThatIsAUsername, line 135: `getAnimIndex()` is called multiple times in this function - maybe just call it once and store it instead of recalculating each time?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: Because even at 13, it's faster than it was before. Not sure why.

ThatIsAUsername:
I don't think I see anything else to whine about. If you've got what you want, go ahead and merge.

PR #342, opened 2021-03-21
Tech debt

ThislsAUsername:
Am I missing anything?

I did originally want event listeners to *only* affect game state by pushing out events, but I think I've changed on that.
Practically speaking, if we want to use a series of events that includes "return events", we have a very simple way of doing that: just don't apply any return events. The logic will be the same in both cases.
No operations will be duplicated or missed, since the direct changes the listeners do will be done independently by their listeners in the current game instances.
Sound reasonable?

### DIFF
--- a/src/Engine/GameEvents/MapChangeEvent.java
+++ b/src/Engine/GameEvents/MapChangeEvent.java
@@ -4,7 +4,7 @@ import java.util.ArrayList;
 
 import Engine.XYCoord;
 import Terrain.Environment;
-import Terrain.Location;
+import Terrain.MapLocation;
 import Terrain.MapMaster;
 import UI.MapView;
 import UI.Art.Animation.GameAnimation;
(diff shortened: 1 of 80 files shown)


src/Engine/GameEvents/MapChangeEvent.java
```diff
@@ -61,13 +61,17 @@ public void performEvent(MapMaster gameMap)
   @Override
   public XYCoord getStartPoint()
   {
-    return changes.get(0).where;
+    if( changes.size() > 0 )
+      return changes.get(0).where;
+    return null;
```
ThatIsAUsername, line 66: What was making this cause a crash?

src/Terrain/MapFog.java
```diff
@@ -9,7 +9,7 @@
 import Engine.XYCoord;
 import Units.Unit;
 
-public class MapWindow extends GameMap
+public class MapFog extends GameMap
```
ThatIsAUsername, line 12: Not super stoked on `MapFog` as a name, since this object contains much more than just fog tracking. Are there any alternatives that would be clearer on the purpose of this class?

src/Engine/GameEvents/MapChangeEvent.java
ThatIsAUsername: I wouldn't count on being able to see the lack of weather changing. There can be a lot going on, and the user won't necessarily know what "correct" is.
That said, this probably isn't going to be a critical thing either way. I'd prefer to at least have a warning though.

PR #343, opened 2021-03-21
Keep the turn number inside the renderable area

ThislsAUsername:


ThatIsAUsername:
This is good, but now it's misaligned with "Turn". They should probably share a baseline like they did before.
Screenshot from 2021-03-26 16-51-25

PR #344, opened 2021-10-02
Buff rework

ThislsAUsername:
I was gonna go for adding sandstorm and rain's effects as UnitMods, but then I realized those would be decent-size PRs of their own.
I think I know what I want to do here, but I want to get the deeper engine things sorted first.

I plan to set up another PR pointing at this one to implement removing `COModifiers` entirely, since that's a bit of a drastic pivot.

Some general design questions:
What bits in `UnitContext` are we fine with being null, as a rule? Ideally, I guess anything can and should be allowed to be null at any point, but that seems... annoying to work with.
Should we define a non-null default value for e.g. the Environment?

I think my next major goal is creating `Army`, to support tags.


src/CommandingOfficers/Strong.java
```diff
@@ -198,15 +193,15 @@ protected void perform(MapMaster gameMap)
...
       // Grant a global +2 movement buff.
       COMovementModifier moveMod = new COMovementModifier(2);
       for( UnitModel model : myCommander.unitModels )
       {
         moveMod.addApplicableUnitModel(model);
       }
-      myCommander.addCOModifier(moveMod);
+      modList.add(moveMod);
 
       // Lastly, all troops are refreshed and able to move again.
```
ThislsAUsername, line 206: This ending up outside `perform()` was an oversight. I am fixing it in the next branch (and can pull the fix forward if we decide against that branch, I guess)

ThislsAUsername:
A curious asterisk on these changes that I didn't consider (and I'm not sure I'm upset with):
`UnitModifiers` on the level of the individual unit cannot affect cost.

src/CommandingOfficers/Commander.java
```diff
@@ -567,4 +538,30 @@ public int getIncomePerTurn()
     }
     return count * (gameRules.incomePerCity + incomeAdjustment);
   }
+
+  @Override
+  public void applyToUMLImpl(UnitModList uml)
```
ThatIsAUsername: Didn't realize we had a Unified Markup Language implementation built into our game. CoOl.

src/CommandingOfficers/Commander.java
```diff
@@ -567,4 +530,23 @@ public int getIncomePerTurn()
...
+    return new ArrayList<UnitModifier>(unitMods);
+  }
+
+  @Override
+  public void add(UnitModifier unitModifier)
+  {
+    unitMods.add(unitModifier);
+  }
+  @Override
+  public void remove(UnitModifier unitModifier)
```
ThatIsAUsername, line 548: `add` and `remove` are odd function names for a `Commander` object.

src/CommandingOfficers/Ave.java
```diff
@@ -542,33 +546,42 @@ public GameEventQueue getEvents(MapMaster gameMap)
...
 
       // Add snow in an expanded range around Ave's areas.
-      int maxSnowRange = coCast.MAX_SNOW_SPREAD_RANGE + GLACIO_SNOW_SPREAD;
-      Set<XYCoord> tilesInRange = Utils.findLocationsNearProperties(gameMap, coCast, maxSnowRange);
-      tilesInRange.addAll(Utils.findLocationsNearUnits(gameMap, coCast, GLACIO_SNOW_SPREAD));
+      int maxSnowRange = GLACIO_SNOW_SPREAD;
+      if (co instanceof Ave)
+      {
+        maxSnowRange += ((Ave)co).MAX_SNOW_SPREAD_RANGE;
+      }
```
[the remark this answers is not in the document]
ThislsAUsername: Because the general future plan is to apply powers to all COs in a given `Army`.
It probably should be checking for equality with `coCast`, though, so you can't use two `Ave`s to expand your snow radius faster than normal.
(Or you could have `Nix`'s cost scale with the current radius explicitly, so the two `Ave`s would have their cost/range increase in sync but still benefit from the extra charge of a tag)

src/CommandingOfficers/Strong.java
```diff
@@ -117,45 +121,40 @@ public void modifyUnitAttackOnUnit(BattleParams params)
...
-    protected void perform(MapMaster gameMap)
+    protected void enqueueCOMods(Commander myCommander, MapMaster gameMap, ArrayList<COModifier> modList)
     {
-      // Grant the base firepower/defense bonus.
-      myCommander.addCOModifier(damageMod);
-      myCommander.addCOModifier(defenseMod);
-      myCommander.addCOModifier(damageModTroop);
+      GenericCOModifier damageMod = new CODamageModifier(STRONGARM_BUFF);
+      GenericCOModifier defenseMod = new CODefenseModifier(STRONGARM_BUFF);
+      GenericCOModifier damageModTroop = new CODamageModifier(STRONGARM_FOOT_BUFF);
```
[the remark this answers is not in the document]
ThislsAUsername: Because the filtered damage modifier originally stored a list of `UnitModel`s it wanted to modify, and I wanted that to be ephemeral (so that you could apply it to several COs, then remove it from all of them at once)

src/Units/UnitContext.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class UnitContext extends UnitState
+{
+  private static final long serialVersionUID = 1L;
+
+  // Groups are set together
+  public Unit unit;
+
+  public GamePath path;
+  public XYCoord coord;
```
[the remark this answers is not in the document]
ThislsAUsername: The intent was to have an `XYCoord` so that you can specify whether you *intend* to specify a coordinate or not.

src/CommandingOfficers/Commander.java
ThislsAUsername: Some of that will get pulled up to `Army`, but yeah, `Commander` is the primary extension point in the engine so it needs its fingers in a lot of pies.

src/Engine/Combat/CombatContext.java
```diff
@@ -1,50 +1,140 @@
...
+      Unit pDefender, WeaponModel defenderWep, List<UnitModifier> defenderMods,
+      int pBattleRange, int attackerX, int attackerY)
+  {
+    CombatContext c = new CombatContext(map,
+        new UnitContext(map, pAttacker, attackerWep, attackerX, attackerY),
+        new UnitContext(map, pDefender, defenderWep, pDefender.x, pDefender.y),
+        pBattleRange);
+    c.attacker.mods.clear();
+    c.attacker.mods.addAll(attackerMods);
+    c.defender.mods.clear();
```
ThatIsAUsername, line 34: Why is it necessary to clear the mods immediately after creating the objects and before adding the mods?

ThislsAUsername: I had it in my head that I was gonna support passing in a `UnitContext` straight from the top (thus pushing the `UnitContext` creation itself into an overload), and that it shouldn't assume the current mods were the right mods for combat consideration.
I don't think I planned out the API design well enough to know what assumptions I should be making.

src/Units/UnitContext.java
```diff
@@ -0,0 +1,105 @@
...
+    env = input;
+    // Air units shouldn't get terrain defense
+    if( null != env && !model.isAirUnit() )
+      terrainStars = env.terrainType.getDefLevel();
+  }
+
+  @Override
+  public String toString()
+  {
+    return model.name;
```
ThatIsAUsername, line 103: Seems a little wonky to have the `toString` here match the `UnitModel` `toString`. Not a huge deal, but I'd probably find some way to differentiate them or provide more relevant info.

src/Units/UnitContext.java
```diff
@@ -0,0 +1,105 @@
...
+    path = other.path;
+    coord = other.coord;
+    maxHP = other.maxHP;
+    attackPower = other.attackPower;
+    defensePower = other.defensePower;
+    env = other.env;
+    terrainStars = other.terrainStars;
+    weapon = other.weapon;
+    mods.addAll(other.mods);
+  }
```
ThatIsAUsername, line 75: If possible, I always try to consolidate overloads into a single actual implementation, and have the others simply forward to the real/comprehensive one.

src/Engine/Combat/CombatContext.java
```diff
@@ -1,50 +1,140 @@
...
+    }
+
+    if( null == attacker.weapon )
+    {
+      attacker.weapon = attacker.unit.chooseWeapon(defender.model, battleRange, attackerMoved);
+    }
+    if( null == defender.weapon )
+    {
+      defender.weapon = defender.unit.chooseWeapon(attacker.model, battleRange, false);
+    }
```
ThatIsAUsername, line 85: We probably don't need to choose a defender weapon until after we determine it can counterattack.

src/Engine/Combat/CombatContext.java
```diff
@@ -46,12 +59,45 @@ public CombatContext(GameMap map,
...
     // Only attacks at point-blank range can be countered
     if( (1 == battleRange) && (null != defender.weapon) )
     {
       canCounter = true;
     }
 
+    if( attacker.mods.isEmpty() )
+      attacker.mods.addAll(attacker.unit.getModifiers());
+    if( defender.mods.isEmpty() )
+      defender.mods.addAll(defender.unit.getModifiers());
```
ThatIsAUsername, line 96: Should this be an unconditional clear/addAll? It seems like preexisting mods would be unexpected in any case.

src/AI/AICombatUtils.java
```diff
@@ -254,7 +254,7 @@ public static int findMaxMobileWeaponRange(Commander co)
         }
         double thisShot =
             CombatEngine.simulateBattleResults(attacker, target, gameMap,
-                                               space.xCoord, space.yCoord).defenderHPLoss;
+                                               space.xCoord, space.yCoord).defender.deltaPreciseHP * -1;
```
ThatIsAUsername, line 257: The intent might be more clear if you use `Math.abs` instead of just multiplying an expected negative result by -1.

src/AI/AICombatUtils.java
```diff
@@ -314,7 +314,7 @@ public static double findMultiHitKill(
         if( movePath.getPathLength() > 0 )
         {
           neededAttacks.put(xyc, unit);
-          double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc.xCoord, xyc.yCoord).defenderHPLoss;
+          double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc.xCoord, xyc.yCoord).defender.deltaPreciseHP * -1;
```
ThatIsAUsername, line 317: Same as 257.

src/AI/Muriel.java
```diff
@@ -453,9 +453,9 @@ private boolean queueUnitAction(GameMap gameMap, Unit unit, boolean allowDeferri
         // Sift through all attack actions we can perform.
         double damageValue = AICombatUtils.scoreAttackAction(unit, action, gameMap,
             (results) -> {
-              double hpDamage = Math.min(results.defenderHPLoss, results.defender.getPreciseHP());
+              double hpDamage = Math.min(results.defender.deltaPreciseHP * -1, results.defender.unit.getPreciseHP());
```
ThatIsAUsername, line 456: I would use `abs` here too.

src/CommandingOfficers/Commander.java
```diff
@@ -357,17 +357,17 @@ public GameEventQueue receiveBattleEvent(final BattleSummary summary)
...
     if( this == summary.attacker.CO )
     {
-      minion = summary.attacker;
-      myHPLoss = summary.attackerHPLoss;
-      enemy = summary.defender;
-      myHPDealt = summary.defenderHPLoss;
+      minion = summary.attacker.unit;
+      myHPLoss = summary.attacker.deltaHP * -1;
+      enemy = summary.defender.unit;
+      myHPDealt = summary.defender.deltaHP * -1;
```
ThatIsAUsername, line 362: Oh. This `*-1` business is everywhere. Maybe have a way to ask for HP lost? Magic numbers are messy.

src/Engine/Combat/BattleSummary.java
```diff
@@ -1,33 +1,21 @@
...
   {
-    attacker = atk;
-    defender = def;
-    attackerWeapon = aw;
-    defenderWeapon = dw;
-    attackerTerrain = atkTerrain;
-    defenderTerrain = defTerrain;
-    attackerHPLoss = atkHPLoss;
-    defenderHPLoss = defHPLoss;
+    super();
```
ThatIsAUsername, line 17: Unnecessary `super()`?

src/Units/UnitState.java
```diff
@@ -0,0 +1,190 @@
...
+  public int damageHP(double damage, boolean allowNegative)
+  {
+    if( damage < 0 )
+      throw new ArithmeticException("Cannot inflict negative damage!");
+    int before = getHP();
+    health = health - healthFromHP(damage);
+    if( !allowNegative )
+      health = Math.max(0, health);
+    return getHP() - before;
+  }
```
ThatIsAUsername, line 127: So let me make sure I understand this. We have a flag to determine if negative damage (healing?) is allowed.
Then we make sure damage is nonnegative, throwing an exception if it is.
Then we force the resulting health to a min of zero if the flag says to (this part makes sense).
It also looks like everywhere this is called, the callers populate `allowNegative` with `isSim`.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
```diff
@@ -191,12 +192,14 @@ public GameEventQueue getEvents(MapMaster gameMap)
       BattleSummary summary = CombatEngine.simulateBattleResults(attacker, defender, map,
                               moveCoord.xCoord, moveCoord.yCoord);
 
+      int attackerHPLoss = (int) (-10 * summary.attacker.deltaPreciseHP);
+      int defenderHPLoss = (int) (-10 * summary.defender.deltaPreciseHP);
```
ThatIsAUsername, line 196: Is this always accurate? If an attack does 1.7 damage, reducing a unit from 5.3 to 3.4 health, then the precise delta of 1.7 would be an HP change of 2 (6->4).

ThatIsAUsername: I guess it's the same calculation we did before, so it's at least not more broken than it was.

src/Units/UnitState.java
```diff
@@ -0,0 +1,190 @@
...
+import CommandingOfficers.Commander;
+import Engine.XYCoord;
+import Terrain.MapLocation;
+import Units.Unit.CargoList;
+
+public abstract class UnitState implements Serializable
+{
+  private static final long serialVersionUID = 1L;
+
+  public final CargoList heldUnits;
```
ThatIsAUsername, line 13: CargoList is final? And part of `UnitState`?

src/Units/UnitState.java
```diff
@@ -50,17 +50,34 @@ public UnitState(UnitState other)
...
     isTurnOver = other.isTurnOver;
-    health = other.health;
+    setResourceState(other);
     captureProgress = other.captureProgress;
     captureTarget = other.captureTarget;
 
     heldUnits = new CargoList(model);
     heldUnits.addAll(other.heldUnits);
   }
+  public void setResourceState(UnitState other)
```
ThatIsAUsername, line 61: This feels like an overly general name. Maybe `copyUnitState` would be clearer?

src/CommandingOfficers/Commander.java
```diff
@@ -63,7 +63,7 @@
   private double myAbilityPower = 0;
 
   private ArrayList<CommanderAbility> myAbilities = null;
-  private String myActiveAbilityName = "";
+  private CommanderAbility myActiveAbility = null; // This should move up to Army later on.
```
ThatIsAUsername, line 66: Should it? What if an army has multiple commander with currently-active abilities? I guess we'll figure that out when someone gets around to implementing armies.

src/CommandingOfficers/CommanderAbility.java
```diff
@@ -38,45 +39,72 @@ public String toString()
...
+   * Public hook to handle cleanup.
+   * <p>One call per use of the power should be sufficient.
+   */
+  public final void deactivate(MapMaster gameMap)
   {
-    if( myCommander.getAbilityPower() < myPowerCost )
+    for( Commander co : new ArrayList<Commander>(coModsApplied.keySet()) )
     {
-      System.out.println("WARNING!: Performing ability with insufficient ability power!");
+      if( coModsApplied.containsKey(co) )
```
ThatIsAUsername, line 77: Unnecessary conditional to check if `co` is part of the map it is part of (because we are iterating over it).

src/CommandingOfficers/Strong.java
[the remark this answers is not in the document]
ThatIsAUsername: Big future plans, eh? I'd wait until then to change how things work, just in case they need to work in a different way than you envision right now. At this point it's just unnecessary code churn.

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: I believe I've cleaned this up already.

src/CommandingOfficers/Commander.java
ThislsAUsername: Removed the comment in PR2

src/Units/UnitState.java
ThislsAUsername: Done in PR2

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThislsAUsername: Changed to *HealthLoss in PR2

src/Units/UnitState.java
ThislsAUsername: Changed param name to `allowOverkill` in PR2

src/CommandingOfficers/Ave.java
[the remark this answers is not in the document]
ThatIsAUsername: As I understand it, we are no longer planning to apply powers to all COs in a given army?

src/Engine/UnitMods/CountTracker.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+  public void incrementCount(RootKeyType co, InstanceKeyType coord)
+  {
+    HashMap<InstanceKeyType, Integer> coCounts = getCountFor(co);
+    int prevCount = getCountFor(co, coord);
+    coCounts.put(coord, prevCount + 1);
+  }
+  public void resetCountFor(RootKeyType co)
+  {
+    buildCounts.remove(co);
```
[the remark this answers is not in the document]
ThislsAUsername: It's a list of counts. I guess you could clear the list instead, or zero every count individually, but this is simpler and saves a little memory/disk space.
Should I rename it to `removeCountsFor()`?

src/CommandingOfficers/Commander.java
```diff
@@ -74,63 +74,38 @@
...
 
     myAbilities = new ArrayList<CommanderAbility>();
   }
 
-  protected void addCommanderAbility(CommanderAbility ca)
+  @Override
+  public void unregister(GameInstance game)
   {
-    myAbilities.add(ca);
+    GameEventListener.super.unregister(game);
```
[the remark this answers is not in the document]
ThislsAUsername: `GameEventListener` is an interface, so that's the syntax you have to use to invoke the method's default implementation.
Java do be how Java do.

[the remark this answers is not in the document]
ThatIsAUsername: Fair enough. Could also do `GameEventListener.registerEventListener(this, game);`, but I guess I don't have an objection to directly invoking the parent version either.

src/Engine/Combat/CombatContext.java
ThislsAUsername: I don't think it matters one way or the other?

src/Engine/Combat/CombatContext.java
ThislsAUsername: My general thought here was that the caller *may* (but is not required to) specify the actual modifiers to be used.
The obvious signal to that effect is "are there mods specified already?".
This should be relevant later on for the detailed combat calc UI page.

src/Units/UnitContext.java
ThislsAUsername: Done.

src/CommandingOfficers/Strong.java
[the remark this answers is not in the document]
ThislsAUsername: I did actually have it set up that way at one point in development.
That said, the new modifiers are even less stateful than the old ones, so I can definitely move 'em back.

src/Units/UnitContext.java
ThislsAUsername: There are three semantic variants of this constructor: unit, model, and copy
Model and copy have only one each, and don't really make sense to merge in that way.
I think I can make some progress on the unit ones, though.

src/Engine/Combat/CombatContext.java
ThatIsAUsername: So then; do we need this, or can we remove the superfluous `clear`s?

src/Units/UnitContext.java
ThatIsAUsername: If you have three different use cases for this class (different enough that the constructors are irreconcilable), I'd consider whether to try and cram it all into one class.

src/Engine/Combat/CombatContext.java
ThatIsAUsername: Not really, no. Would save a few CPU cycles.

src/Engine/UnitMods/CountTracker.java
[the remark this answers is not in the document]
ThatIsAUsername: I could just as easily argue that if we `reset` and then `get` the count, it has to re-create the entry which takes more time. That said, I don't think we need to worry about this amount of either memory or CPU time, and the exposed behavior seems to be consistent enough. I'm not gonna quibble, do what makes sense.

src/Engine/Combat/CombatContext.java
ThatIsAUsername: I'm not sure I'm comfortable using implicit signals like that. It might be more clear to add a boolean in the constructor to say whether to pull in CO unit modifiers?
Alternatively: when we go to implement it, there may be a more obvious way to do that job (e.g. just use the normal process of applying modifiers to the CO instead of doing funny implicit business).

src/Engine/Combat/CombatContext.java
ThislsAUsername: We don't. I deleted it entirely here: https://github.com/ThislsAUsername/DefendPeace/pull/346/commits/4c5377a89365d749c86db252c4487e5d259ecab6

src/Units/UnitContext.java
ThislsAUsername: The copy constructor isn't a separate use case. Or at least, I sure hope it isn't.

That leaves unit and model-only cases, and they're actually used for the same things - this allows us to collect modifications and preview stuff without having a specific `Unit` in question.

src/Engine/Combat/CombatContext.java
ThislsAUsername: I take it you are lobbying to unconditionally clear-and-add the modifiers, then?

ThatIsAUsername: Yeah, probably, until and unless it becomes necessary. Maybe even warn if there are modifiers present already.

ThislsAUsername: Okay, one problem: I'm already unconditionally adding the modifiers in `UnitContext`'s constructor - as far as I can tell, this code shouldn't ever actually affect anything.
Thoughts?

ThatIsAUsername: Sounds like we can just nuke this altogether then. Should we also apply them in the constructor, so we don't have to worry about conditionally/double applying them?

ThislsAUsername: Uh, what?
I'm already applying modifiers in the `UnitContext` constructor, so I don't think anywhere else is necessary?

ThatIsAUsername: You are adding them in the constructor, not applying them. It may not make sense to apply them right away anyway though.

ThislsAUsername: Oh. Derp.

That's what I was proposing in some of the other discussions we were having.
The trouble with that IMO is that how and when combat modifiers get applied is currently an implementation detail of the combat engine. Changing that would require e.g. telling a `UnitContext` whether it's the defender or attacker *on construction*, which I don't think is compatible with Counter Break.

src/Engine/GameEvents/CommanderAbilityEvent.java
```diff
@@ -33,7 +36,14 @@ public GameEventQueue sendToListener(GameEventListener listener)
   @Override
   public void performEvent(MapMaster gameMap)
   {
-    myAbility.activate(gameMap);
+    if( myCommander.getAbilityPower() < myAbility.getCost() )
+    {
+      System.out.println("WARNING!: Performing ability with insufficient ability power!");
+    }
+    myCommander.activateAbility(myAbility, gameMap);
```
ThatIsAUsername, line 43: `CommanderAbility.activate()` also calls `Commander.activateAbility()` and does the cost adjustment. This function in general is doing a lot more than it did before, and now it's duplicative.

Should we revert this event and the corresponding action to how they were before? This appears to be residual change from when you pulled Commanders out of `CommanderAbility`.

PR #346, opened 2021-10-15
Kill CO mods

ThislsAUsername:
I think I'm happy with the current state of the code.

The next "big feature" I want to do is `Army`, to support tags (and probably tags along with)

More roadmap stuff:
I definitely want to get a unit set together with my naval rework in, at some point.
I also kinda want to have a more detailed unit info page (with descriptions in words, probably) before that happens, since we already have two unit sets and that's a lotta tech debt to pile on.

From there, the main feature between us and porting lots of COs is some CO organization scheme so it's not just a mile-long list.


src/Test/TestCombatMods.java
```diff
@@ -89,16 +88,16 @@ private boolean testIronWill()
...
     infB.isTurnOver = false;
-    
+
     // Check our damage for each first strike pre-power...
     BattleSummary normalAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 5);
 
     venge.modifyAbilityPower(42); // juice up
-    performGameAction(new GameAction.AbilityAction(venge, venge.getReadyAbilities().get(0)), testGame); // activate Iron Will
-    
+    venge.getReadyAbilities().get(0).activate(testMap); // activate Iron WIll
```
ThatIsAUsername, line 96: aaaaaahhhhhhh the unnecessary capital I returned.

src/CommandingOfficers/Modifiers/CODefenseModifier.java
```diff
@@ -1,13 +1,9 @@
...
+import Engine.Combat.StrikeParams.BattleParams;
+import Engine.UnitMods.UnitModifierWithDefaults;
 
-import java.util.ArrayList;
-
-import CommandingOfficers.Commander;
-import CommandingOfficers.Modifiers.COModifier.GenericCOModifier;
-
-public class CODefenseModifier extends GenericCOModifier
+public class CODefenseModifier implements UnitModifierWithDefaults
```
ThatIsAUsername, line 6: Should this (and others) be renamed to e.g. `UnitDefenseModifier` for consistency?

src/CommandingOfficers/Modifiers/COMovementModifier.java
```diff
@@ -1,12 +1,11 @@
...
 
+import java.io.Serializable;
 import java.util.ArrayList;
-
 import CommandingOfficers.Commander;
-import CommandingOfficers.Modifiers.COModifier.GenericCOModifier;
 import Units.UnitModel;
 
-public class COMovementModifier extends GenericCOModifier
+public class COMovementModifier implements Serializable
```
ThatIsAUsername: Strikes me as a bit odd that this doesn't inherit from `UnitModifier`.

src/Test/TestCombatMods.java
```diff
@@ -128,16 +127,14 @@ private boolean testRetribution()
...
+    venge.getReadyAbilities().get(1).activate(testMap); // activate Retribution
 
     // ...and after power
     BattleSummary retribAB = CombatEngine.simulateBattleResults(infA, infB, testMap, 7, 4);
     BattleSummary retribBA = CombatEngine.simulateBattleResults(infB, infA, testMap, 7, 4);
 
     // Check that Venge's Retribution works properly without breaking things (other than balance)
     boolean testPassed = true;
-    testPassed &= validate(infB.model.getDamageRatio() > 110, "    Retribution didn't buff offense.");
-    testPassed &= validate(infB.model.getDefenseRatio() < 100, "    Retribution didn't reduce defense.");
```
ThatIsAUsername, line 140: Why delete? Aren't these effects still part of Retribution?

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
```diff
@@ -0,0 +1,27 @@
+package CommandingOfficers.Modifiers;
```
ThatIsAUsername, line 1: It seems strange to have this and other filters in the `CommandingOfficers` tree when everything it's related to is in the `UnitMods` tree.

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
```diff
@@ -0,0 +1,27 @@
...
+import Engine.UnitMods.UnitModFilter;
+import Engine.UnitMods.UnitModifier;
+import Units.Unit;
+import Units.UnitContext;
+
+public class UnitInstanceFilter extends UnitModFilter
+{
+  private static final long serialVersionUID = 1L;
+
+  public final ArrayList<Unit> instances = new ArrayList<>();
```
ThatIsAUsername, line 14: It would probably make for a cleaner design to have a method to add more units instead of just directly accessing the public collection.

src/CommandingOfficers/Modifiers/COMovementModifier.java
ThislsAUsername: Yeah. That was a temporary thing because I hadn't set up movement in the `UnitModifier` interface yet.

src/Test/TestCombatMods.java
ThislsAUsername: Yes, they are in Retribution... but as `UnitModifier`s, so the stat change won't show up until combat calculation happens.
I could manually invoke the combat internals to inspect the stat changes there, but it's just as good to observe the damage output does what I expect (and that test's still in there).

It might be a good idea to get rid of these functions and just use raw instance variables at this point, though, for two reasons:
My original intent for them is OBE - hiding the actual value behind g/setters was to enforce relative changes only, so that ability stat changes would be more likely to be cleanly reversible... but we're not touching the stat on `UnitModel` anymore, so the base stat should represent the D2D and nothing else. (It might be a good idea to move those functions to `UnitContext` and/or `StrikeParams` to enforce addition/subtraction there?)
Having a getter kinda implies (now) that the `UnitModifier`s will be consulted to produce a reasonably-close-to-correct value, but that's not what happens here. I don't really want to change that, either, as I don't want to add more spaghetti to `CombatEngine` than I need to for the actual functionality I want to support.

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
ThislsAUsername: Sorta, yeah. I'm not convinced I did a very good job of putting files in the right places in this PR, lol.

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
ThislsAUsername: Would it?
There's no reason to expect UIF to be modifying the input in any way; the instantiating class should have pretty complete control of the content.
I guess I should probably use a `HashSet` instead of an `ArrayList`, but those have a common external interface for adding stuff anyway.
Once it goes out of the scope of the creating function and is handled only by its interface, then it doesn't matter if the collection's public or not.
And, finally, it adds more boilerplate code and error checks to write.

src/Engine/UnitMods/UnitModifierWithDefaults.java
```diff
@@ -0,0 +1,53 @@
...
+
+import Engine.Combat.CombatContext;
+import Engine.Combat.StrikeParams;
+import Engine.Combat.StrikeParams.BattleParams;
+import Units.UnitContext;
+
+/**
+ * Defaults all the interface items to no-ops for ease of implementation
+ */
+public interface UnitModifierWithDefaults extends UnitModifier
```
ThatIsAUsername, line 11: Why introduce a new class to implement `UnitModifier` instead of just giving `UnitModifier` the default implementations?

src/CommandingOfficers/Meridian.java
```diff
@@ -75,150 +73,139 @@ public Meridian(GameScenario.GameRules rules)
     arty.possibleActions.add(new TransformLifecycle.TransformFactory(tank, "~TANK"));
 
     addCommanderAbility(new ChangeAndFlow(this));
-    addCommanderAbility(new VehicularCharge(this));
+    addCommanderAbility(myVehicularCharge);
```
ThatIsAUsername, line 76: Why does Vehicular Charge need to be created in advance?

src/Engine/UnitMods/UnitModifierWithDefaults.java
ThislsAUsername: I'd like to have the option of getting the compiler to tell me whether I've implemented all the methods in this interface or not.
This is relevant for the filter classes, for example.

src/CommandingOfficers/Meridian.java
ThislsAUsername: It doesn't, but it's fewer lines of code and equivalent meaning this way.
Meridian keeps a reference to it around so she can mark all of her own units she's currently debuffing

I considered pushing her markings into some sort of generic StateTracker that just handles arbitrary markings, but that isn't something the current implementation of unit markings/state trackers lends itself to (deliberately; whether that's good in practice is a valid consideration)

src/AI/AICombatUtils.java
```diff
@@ -187,7 +198,7 @@ public static int findMaxMobileWeaponRange(Commander co)
       final XYCoord attackerCoord = new XYCoord(u);
       boolean requiresMoving = false;
       int dist = targetCoord.getDistance(attackerCoord);
-      if( !u.canAttack(target.model, dist, requiresMoving) )
+      if( !u.canAttack(gameMap, target.model, dist, requiresMoving) )
```
ThatIsAUsername, line 201: We didn't need a `gameMap` here before. I don't think we really do now, either, except that `UnitContext` requires it. I don't really like how `UnitContext` is trying to be all things to all people, or the fact that we now have functions like `getRangeContext`. That smells to me, probably because it means we don't really know what we have if we have a `UnitContext` object, unless we just purpose-built it; I'd be vaguely curious how often we create them.

At any rate, it definitely feels like there should be a simpler and cleaner way to handle this.

src/Units/UnitContext.java
```diff
@@ -54,6 +68,8 @@ public UnitContext(Unit u)
     mods.addAll(u.getModifiers());
     initModel();
   }
+  // Note: the Commander argument is here as a placeholder for Army
+  // The plan is for Unit to have an Army reference, and UnitModel to have the Commander reference
```
ThatIsAUsername, line 72: Why not give Commander the Army reference?

src/Units/UnitContext.java
```diff
@@ -87,6 +106,8 @@ public void setPath(GamePath pPath)
   {
     path = pPath;
     coord = path.getEndCoord();
+    if( null != map && null != coord )
+      setEnvironment(map.getEnvironment(coord));
```
ThatIsAUsername, line 110: Are these expected to ever be null? Should this not just be an error condition?

src/Units/UnitContext.java
```diff
@@ -97,9 +118,84 @@ public void setEnvironment(Environment input)
...
+    if( null != w )
+    {
+      rangeMin = w.rangeMin;
+      rangeMax = w.rangeMax;
+      // Build a scratch UnitContext because we don't trust UnitMods with our state
+      UnitContext uc = new UnitContext(this);
+      for( UnitModifier mod : mods )
+        mod.modifyAttackRange(uc);
+      rangeMin = uc.rangeMin;
+      rangeMax = uc.rangeMax;
```
ThatIsAUsername, line 133: Mmmm, Pungent.

src/Engine/GameInstance.java
```diff
@@ -70,6 +71,9 @@ public GameInstance(MapMaster map, Weathers weather, GameScenario scenario, bool
     playerCursors = new HashMap<Integer, XYCoord>();
     for( int i = 0; i < commanders.length; ++i )
     {
+      // This is hacky, but hey
+      commanders[i].addUnitModifier(new SandstormModifier());
```
ThatIsAUsername, line 75: sigh

src/AI/AICombatUtils.java
ThislsAUsername: > We didn't need a `gameMap` here before. I don't think we really do now, either, except that `UnitContext` requires it.

`UnitContext` does not require it; however, I prefer to provide maximum information to make the results as accurate as possible. For example, not having the map means we won't have knowledge of the weather, which means we won't apply -1 range if there's a sandstorm going.
> I don't really like how `UnitContext` is trying to be all things to all people

I had the same complaint about the `GameInputState` state data, though that's a much smaller code surface area, to be fair.
The core issue here is that `UnitContext` is basically the information-carrier to enable all the extension points that `Commander` wants. We *could* make a bunch of classes with subsets of `UnitContext`'s current fields for calculating different things, but we'd be likely to end up with even more duplication than we already have.
>, or the fact that we now have functions like `getRangeContext`.

`getRangeContext()` exists because I didn't want to calculate min and max range separately (I don't feel strongly about this; maybe we should?), and didn't feel like making up an entirely new object for the purpose.
> That smells to me, probably because it means we don't really know what we have if we have a `UnitContext` object, unless we just purpose-built it;

You are indeed correct. A `UnitContext` is modified based on... context. I don't need to touch the range if all I'm calculating is movement, for example. ...At least until someone makes a CO that changes stuff based on other unit stats.
We could just institute a policy of "apply all modifications to every facet on instantiation", but that would require some rewickering of stuff.
> I'd be vaguely curious how often we create them.

Depends on use case, but "a lot".
For example, 4 minimum for combat calculations, usually more.
> At any rate, it definitely feels like there should be a simpler and cleaner way to handle this.

I'm trying to design this to support a lot of features we don't have yet, so new COs and such can just slot in without changing anything outside the new code and code to enable invoking the new code.
That necessarily involves pulling in a lot of data.
If there is a cleaner solution, I'm not sure what it would look like. Maybe that immutable state thing you were talking about before?

src/Units/UnitContext.java
ThislsAUsername: `Commander` will also have the `Army`, but I think `Commander` may not have a list of units anymore (since that becomes a nigh-duplicate source of truth).
That means that `Army` owns units, `Commander` owns `UnitModel`s, and `Unit`s get references to them.

It just seems like the "right way" to structure the data, to me, but I'm certainly willing to hear counterarguments.

src/Units/UnitContext.java
ThislsAUsername: Yes.
Them not being present is not an error.
If I want to contemplate a unit type outside of the context of a specific unit, then it feels pretty meaningless to include the map unless we care about the terrain for some reason.

src/Units/UnitContext.java
ThislsAUsername: The alternatives (while also being reasonably defensive) are:
Split `UnitContext` into a bunch of separate classes for different contexts (messy, probably lots of duplication, and could end up looking really dumb depending on CO design)
Access protection (requires a bunch of interfaces and functions I don't want to write)
Make state immutable, so we'd be making a copy anyway

src/Engine/GameInstance.java
ThislsAUsername: The list of `UnitModel` initializations lived in `Commander`'s constructor for the longest time, lol

I'm not sure where I want this sort of thing to go.
Should `GameInstance` (or perhaps `GameRules`?) have a list of modifiers that pertain to all units in the game?
Maybe `GameRules` should have a static function to provide all global modifiers? I imagine we want something like that somewhere, but I'm not sure where it goes.
I also want to allow `UnitModelScheme`s to provide modifiers, though I'm not sure whether I just want to say "put it on the `UnitModel` if it applies to it" and not worry about handling the global-modifier case separately.

src/CommandingOfficers/Modifiers/UnitIndirectRangeModifier.java
```diff
@@ -0,0 +1,28 @@
...
+  public UnitIndirectRangeModifier(int range)
+  {
+    rangeChange = range;
+  }
+
+  @Override
+  public void modifyAttackRange(UnitContext uc)
+  {
+    if( uc.rangeMax > 1 )
+      uc.rangeMax += rangeChange;
```
ThatIsAUsername, line 25: If this only acts on units that already have a range that is greater than 1, is it impossible to buff Max artillery?

src/Units/UnitContext.java
```diff
@@ -35,15 +38,26 @@
...
-  public UnitContext(GameMap map, Unit u, WeaponModel w, int x, int y)
+  public UnitContext(GameMap map, Unit u)
+  {
+    this(map, u, null);
+  }
+  public UnitContext(GameMap map, Unit u, WeaponModel w)
+  {
+    this(map, u, w, null, new XYCoord(u));
+  }
+  public UnitContext(GameMap map, Unit u, WeaponModel w, GamePath path, XYCoord coord)
```
ThatIsAUsername, line 53: Why have a constructor with both a path and a coord instead of just using the path's end?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -256,22 +260,35 @@ private BufferedImage renderMap()
...
+        if( threatMode == SelectedUnitThreatAreaMode.All
+            || threatMode == SelectedUnitThreatAreaMode.Current )
+          if( w.canFireAfterMoving || currentPath.getPathLength() == 1 )
+          {
+            UnitContext uc = new UnitContext(gameMap, currentActor, w, currentPath, cursorCoord);
+            overlays.add(new GameOverlay(null,
+                         Utils.findLocationsInRange(gameMap, cursorCoord,
+                                                    (1 == uc.rangeMin)? 0 : uc.rangeMin, uc.rangeMax),
+                         OverlayArtist.FIRE_FILL, OverlayArtist.NOW_FIRE_EDGE));
+          }
```
ThatIsAUsername, line 283: This is pretty cool.

src/CommandingOfficers/Commander.java
```diff
@@ -95,12 +96,22 @@ public Commander(CommanderInfo info, GameScenario.GameRules rules)
     myAbilities = new ArrayList<CommanderAbility>();
   }
 
+  @Override
+  public void registerForEvents(GameInstance game)
```
ThatIsAUsername, line 100: I'd probably support changing this to `initForGame` as well since it's not just registering for events anymore.

src/CommandingOfficers/Modifiers/UnitIndirectRangeModifier.java
ThislsAUsername: Range 1 is direct combat. Max artillery have range 2-2.

src/Units/UnitContext.java
ThislsAUsername: The `path` argument can be null.
This is used, for instance, while you have your Rockets move area up, and pan to somewhere else to consider what area you could cover.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Thank you.
It shooooould work with my naval rework Bship plans, too.

src/CommandingOfficers/Commander.java
ThislsAUsername: It's from the `GameEventListener` interface; `CommanderAbility` isn't (necessarily) a listener, while `Commander` always is.
That said, the registration is always handled separately anyway, so sure.

src/Test/TestCountTracker.java
```diff
@@ -0,0 +1,72 @@
...
+      for( int i = 0; i < 3; ++i )
+      {
+        ct.hasCountFor(null);
+        ct.getCountFor(null);
+        ct.getCountFor(null, null);
+        ct.getCountFor(0, null);
+        ct.getCountFor(null, 0.0);
+        ct.incrementCount(null, null);
+        ct.incrementCount(0, null);
+        ct.incrementCount(null, 0.0);
```
ThatIsAUsername, line 35: What is this testing? Just that you can call each of these functions without it crashing? No verifying that return values are correct, or that the 0/null counts actually increment?

src/Engine/UnitMods/CountTracker.java
```diff
@@ -7,33 +7,35 @@
...
+    return countMap.containsKey(root);
   }
-  public HashMap<InstanceKeyType, Integer> getCountFor(RootKeyType co)
+  public HashMap<InstanceKeyType, Integer> getCountFor(RootKeyType root)
   {
-    if( !buildCounts.containsKey(co) )
-      buildCounts.put(co, new HashMap<InstanceKeyType, Integer>());
-    return buildCounts.get(co);
+    if( !countMap.containsKey(root) )
+      countMap.put(root, new HashMap<InstanceKeyType, Integer>());
```
ThatIsAUsername, line 19: Seems a bit odd that calling an accessor mutates the collection here, especially since the sibling `getCountFor` does not.

src/Test/TestCombatMods.java
```diff
@@ -158,4 +163,84 @@ private boolean testRetribution()
...
+      performGameAction(new BattleLifecycle.BattleAction(testMap, aa, Utils.findShortestPath(aa, 7, 3, testMap), 7, 4), testGame);
+
+      testPassed &= validate(cinder.money == currentFundsReturn * 1000, "    Expected to make "+currentFundsReturn*1000+", but got "+cinder.money+" instead.");
+    }
+    // Remove those amounts, and make sure that works
+    for( int i = 0; i < 55; ++i )
+    {
+      currentFundsReturn -= i;
+      ddtic.stopTracking(cinder, i);
+      ddtic.stopTracking(cinder, i); // Remove twice, just to be sure
```
ThatIsAUsername, line 202: What? Why? This is literally inside a test case. If you want to be sure, put in a test that will fail if it isn't doing what it should.

src/Test/TestCombatMods.java
```diff
@@ -158,4 +163,84 @@ private boolean testRetribution()
...
+      // Add the victim, and yeet him
+      addUnit(testMap, venge, UnitModel.TROOP, 7, 4);
+      performGameAction(new BattleLifecycle.BattleAction(testMap, aa, Utils.findShortestPath(aa, 7, 3, testMap), 7, 4), testGame);
+
+      testPassed &= validate(cinder.money == currentFundsReturn * 1000, "    Expected to make "+currentFundsReturn*1000+", but got "+cinder.money+" instead.");
+    }
+    // Remove those amounts, and make sure that works
+    for( int i = 0; i < 55; ++i )
+    {
+      currentFundsReturn -= i;
```
ThatIsAUsername, line 200: Isn't this going to check the same cases as the previous loop, just in reverse as `currentFundsReturn` goes back to `0`?

src/Test/TestCombatMods.java
ThatIsAUsername: Also, if you tell it to stop tracking, shouldn't it fail to increase Cinder's warchest?

src/Test/TestCombatMods.java
```diff
@@ -158,4 +163,84 @@ private boolean testRetribution()
...
+      GameEventListener.publishEvent(event, testGame);
+
+      // Add the victim, and yeet him
+      addUnit(testMap, venge, UnitModel.TROOP, 7, 4);
+      performGameAction(new BattleLifecycle.BattleAction(testMap, aa, Utils.findShortestPath(aa, 7, 3, testMap), 7, 4), testGame);
+
+      testPassed &= validate(cinder.money == currentFundsReturn * 1000, "    Expected to make "+currentFundsReturn*1000+", but got "+cinder.money+" instead.");
+    }
+    // Do the second half backwards, for giggles
+    for( int i = 99; i > 42; --i )
```
ThatIsAUsername, line 218: Is this supposed to be a continuation of the previous loop, but reversing the direction? The cutoff isn't even in the same place (though I guess that doesn't matter with how the tests are set up).
This does make the test more complex than it needs to be though. I'm not sure why we even need the second loop (see comment above), but if so you should at least not go out of your way to make the test more difficult to understand.

src/Test/TestStateTracker.java
```diff
@@ -0,0 +1,58 @@
...
+    testPassed &= validate(testInitialization(), "  Initialization test failed.");
+
+    return testPassed;
+  }
+
+  private boolean testInitialization()
+  {
+    boolean testPassed = true; // Assume nothin's busted
+    int currentTrackers = testGame.stateTrackers.keySet().size();
+    testPassed &= validate(currentTrackers == 2, "    Expected 2 trackers, but the game has "+currentTrackers+" instead.");
```
ThatIsAUsername, line 46: I assume there are 2 trackers expected because Venge and Cinder each create one? That's a level of indirection above what is typical for a test case, so it would be good to at least note in a comment what is driving this test.

src/Test/TestStateTracker.java
```diff
@@ -0,0 +1,58 @@
...
+
+    for( int i = 0; i < 10; ++i )
+    {
+      DamageDealtToIncomeConverter.instance(testGame, DamageDealtToIncomeConverter.class);
```
ThatIsAUsername: This seems like odd syntax. Isn't this equivalent to `StateTracker.instance(testGame, DamageDealtToIncomeConverter.class)`? Would it be possible to allow a subclass-specific override so we can just call `DamageDealtToIncomeConverter.instance(testGame)`, since the sublcass already knows its type?

src/Units/UnitContext.java
```diff
@@ -196,6 +196,6 @@ public void chooseWeapon(ITargetable targetType, int range, boolean afterMoving)
   @Override
   public String toString()
   {
-    return model.name;
+    return "Context for "+model.name+" at "+coord+", using weapon "+weapon;
```
ThatIsAUsername, line 199: Better, but can't weapon (and coord?) be null? Or the context may have been created just to calculate a price difference, which isn't reflected here at all?

src/Engine/Combat/CombatEngine.java
```diff
@@ -109,11 +111,12 @@ public static BattleSummary calculateBattleResults(UnitContext attacker, UnitCon
 
   public static double calculateOneStrikeDamage( Unit attacker, int battleRange, Unit defender, GameMap map, int terrainStars, boolean attackerMoved )
   {
-    UnitContext attackerContext = new UnitContext(map, attacker, null, attacker.x, attacker.y);
+    XYCoord dest = new XYCoord(attacker);
+    UnitContext attackerContext = new UnitContext(map, attacker, null, Utils.findShortestPath(attacker, dest, map), dest);
```
ThatIsAUsername, line 115: This didn't need a path before... seems strange to require one now.

src/Test/TestCombatMods.java
ThatIsAUsername: If the model value is now only ever going to be the D2D, then it makes sense to remove the associated calls and use it directly. I'd probably also rename the fields to clarify that they are the D2D value, not necessarily the current value.

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
ThatIsAUsername: It would avoid requiring knowledge of the object internals (and uncertainty about whether direct modification is kosher). Probs not a huge deal here though I suppose.

src/Engine/UnitMods/UnitModifierWithDefaults.java
ThatIsAUsername: Do you ever (or ever plan to) implement every interface method in the same subclass (excepting `UnitModifierWithDefaults` and `UnitModFilter`, of course)? Those should be stable once we stabilize the set of modifiable things, so I'm not too worried. I guess it doesn't hurt anything either.

src/AI/AICombatUtils.java
ThatIsAUsername: Not super thrilled with creating a million objects all the time - seems like it would create a lot of extra runtime work that we didn't use to have. I also don't know that I want to enforce "full" population of `UnitContext` every time one is created, since that just creates more busywork for no apparent gain.
The immutable state thing would be fun to try, but would also require quite a lot of rewickering.

src/Units/UnitContext.java
ThatIsAUsername: I'd probably leave the unit list with the commander. An Army can iterate through all units via the commanders, and each commander still needs to iterate through their own units (since those are the ones they can move). Unless I'm missing something in your design. Army owns the COs, who own the units.

src/Engine/GameInstance.java
ThatIsAUsername: I mean yeah. It's fine there, at least for now. Just... sigh.

Though frankly, due to weather being tile specific, this is probably as good an approach as any (until a better one appears). Otherwise this restriction would probably just have to be baked into the engine wherever weapon ranges are determined, but fortunately we can filter modifiers on a per-unit basis according to location.

src/Units/UnitContext.java
ThatIsAUsername: Hm. Then I'd probably have a constructor with a coord and no path, just to avoid needing dummy parameters.

Can we pull in named parameters from Python? That's a cool feature I'd like to just have everywhere.

src/CommandingOfficers/Modifiers/UnitIndirectRangeModifier.java
ThatIsAUsername: Ah, touché. And if it gets set less than `minRange`, `UnitContext.setWeapon` will sort that.

src/AI/AICombatUtils.java
ThislsAUsername: > Not super thrilled with creating a million objects all the time - seems like it would create a lot of extra runtime work that we didn't use to have.

Technically correct, but the only time the actual game engine takes up most of our processing is when the AI is active.
If it ends up eating lots of our CPU budget, then I'd expect it to be a case of e.g. running a calculation every frame when we don't need to.
As for creating lots of objects... that's kinda the nature of the beast. You need scratch space to calculate stuff.
> I also don't know that I want to enforce "full" population of `UnitContext` every time one is created, since that just creates more busywork for no apparent gain.

Being able to not populate all of its fields is a core feature. I wouldn't want to drop that.
What I was proposing was to apply all *modifications* to the `UnitContext` on instantiation, so you don't ever have to wonder whether e.g. the context's range has been modified yet.
It would certainly be more CPU work, and would require rewickering the combat stuff a bit (since that operates on the assumption that it gets a "clean" context that it can then clone and modify in various ways)

src/Units/UnitContext.java
ThislsAUsername: > An Army can iterate through all units via the commanders

Fair point, though a nested for loop (for each commander, for each unit) seems a bit more complex than a for with a nested if (for each unit, if CO == x)
> each commander still needs to iterate through their own units (since those are the ones they can move)

`Army` determines what is movable during a given turn. For example, Wally wouldn't necessarily care which CO a given unit belonged to; he just wants to move every unit in his `Army`.
We could still end up having COs iterating through their own units for `turnInit`, D2D, or power reasons... but I was thinking most everything would be handled generically through `Army`.
> Unless I'm missing something in your design. Army owns the COs, who own the units.

Either object graph is defensible; they contain basically the same amount of info and duplication thereof:
image
(the lines are supposed to be grabby stick-figure hands)

src/Units/UnitContext.java
ThislsAUsername: It's mostly pointless to do that - the only instance where `null` is passed in explicitly is in `AICombatUtils.findThreatPower()` for the case where the unit can't move-and-fire.
That case could easily pass in a path as well. I was just lazy.

> named parameters

Scala has them. It runs on the JVM and is interoperable with Java: https://www.geeksforgeeks.org/scala-named-arguments/

src/Test/TestCountTracker.java
ThislsAUsername: It's testing that `CountTracker` is null-safe.
As for correctness, I don't know and don't really care what's correct.
That may change in the future, but for now there's no defined "correct" behavior for these cases - is `null` a valid key? I don't know.

src/Engine/UnitMods/CountTracker.java
ThislsAUsername: This is specifically getting a count *collection* - if I returned `null` here, the caller would have to check for null and possibly instantiate one needlessly.
It's intended to be read/write access, too, so that's also relevant.

src/Test/TestCombatMods.java
ThislsAUsername: That's the intent *of* the test case - I'm making sure double-removal doesn't break my assumptions.
This is the test that would fail.

`stopTracking()` works on one particular value at a time - there's probably a better name for it, but the semantics are "Cancel giving me money at this particular exchange rate, assuming you were giving me money in the first place"

src/Test/TestCombatMods.java
ThislsAUsername: Yes, but it's removing them in a different order than they were added.

src/Test/TestCombatMods.java
ThislsAUsername: Yes. The intent is to make sure the tracker works regardless of add/remove order, including when there are more removes than adds.
If there was no reason to make the test more complex, I don't think I would have.

src/Test/TestStateTracker.java
ThislsAUsername: Yeah, probably... but boiiiilerplate.
This would be a great use case for overriding a static method, if that was a thing defined in the language, lol.

src/Units/UnitContext.java
ThislsAUsername: They can be, but `null`'s string concatenation is the empty string rather than a crash.

That's fair, but I don't really feel like adding a read-back of literally every field.
If you really want it, then I suggest we start using Lombok https://projectlombok.org/features/all

src/Engine/Combat/CombatEngine.java
ThislsAUsername: I think I want to start putting paths into combat stuff in general, since path-dependence is a pretty common thing among custom CO concepts.

src/AI/AICombatUtils.java
ThatIsAUsername: > As for creating lots of objects... that's kinda the nature of the beast. You need scratch space to calculate stuff.

We didn't do that before, and I hope you aren't arguing that this is the only valid solution. Things like this are probably why OO has a bit of a reputation for inefficiency and bloat. Once you start thinking in objects, everything becomes one.

> What I was proposing was to apply all modifications

Would that even make sense? E.g. if we only partly populate the context, and then apply a modifier that affects something unpopulated? At any rate, I'm not sure that's the right answer either.

src/Units/UnitContext.java
ThatIsAUsername: > It's mostly pointless to do that

It would help with clarity - it's confusing to have a constructor that demands two apparently-conflicting parameters with no guidance on when one should be provided over the other. It would be cleaner--or at least clearer at point of use-- to just have two constructors to accept either a path or the endpoint.

> That case could easily pass in a path as well. I was just lazy.

It's not lazy to not do useless work. It's pointless at best, and worse if it creates confusion and more unnecessary work.

src/Engine/UnitMods/CountTracker.java
ThatIsAUsername: > It's intended to be read/write access

As in, someone could add to the returned collection and have it reflected in the next call to `getCountFor`? Is it used this way anywhere? Is that tested?

I'll grant that it's probably better to create the empty collection due to user expectations, and since it shouldn't affect any future calls.

src/Test/TestCombatMods.java
ThatIsAUsername: Ah. Well, the comment seems to say "We'd better remove this value twice, since once may not be enough to do the job". Which implies the test case is working around known deficiencies in the implementation.

Sometimes your comments leave me more confused than if there wasn't one at all.

src/Test/TestCombatMods.java
ThatIsAUsername: > If there was no reason to make the test more complex, I don't think I would have.

And yet you overlap the third loop with the second, meaning you execute it 55-43=12 extra times after `currentFundsReturn` is already zero.
I mean, I guess it's worth adding in one order and removing in a different order, but using one loop to add everything and two loops to remove it all in different orders is probably more than needed to prove it does what it says (and is certainly more intricate than is probably needed or expected).

src/Test/TestStateTracker.java
ThatIsAUsername: Patch?

src/Units/UnitContext.java
ThatIsAUsername: It's not that I really want a more complex toString; it's more that I want a less complex `UnitContext`.

If wishes were fishes, I know.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: Like Wargroove's knights with their "Do more damage if you move x spaces" mechanic? That's... uh. I mean. I guess.

I generally prefer to add new things at time of need though, otherwise YAGNI.

src/Units/UnitContext.java
ThatIsAUsername: Also what do you mean that we don't trust UnitMods with the state? Isn't their whole point that they modify state?

src/Units/UnitModel.java
```diff
@@ -80,8 +84,8 @@ public static String standardizeID(String input)
   public long carryableExclusionMask;
   private int COstr = 100;
   private int COdef = 100;
-  public double COcost = 1.0;
-  public int moneyCostAdjustment = 0;
+  public double costMultiplier = 1.0;
+  public int costShift = 0;
```
ThatIsAUsername, line 88: These don't appear to be used anymore. Should COstr/def (and maybe other things) also be eliminated in favor of using UnitMods so that the model can be properly immutable?

src/Engine/Combat/CombatEngine.java
ThislsAUsername: YAGNI doesn't apply here IMO, because this is a data coherency issue and also the sort of thing that could introduce subtle inconsistencies.

Let's say, for example, that I add a CO that gains +50 base damage based on pathing.
The AI is contemplating an infantry that's in range of an enemy tank.
Muriel uses `calculateOneStrikeDamage()`, which leaves out pathing info, and determines that attacking the tank is a bad idea.
Wally uses the full combat logic, including the path, and discovers a very cost-effective trade.
Where's the bug? (assuming, for the sake of argument, that +50 base damage isn't a terrible idea)

src/AI/AICombatUtils.java
ThislsAUsername: > We didn't do that before, and I hope you aren't arguing that this is the only valid solution.

There are certainly more possible solutions, but I don't see any with better tradeoffs than this one.
> Things like this are probably why OO has a bit of a reputation for inefficiency and bloat.

Absolutely correct, but I don't feel like it matters that much in this case.
> Would that even make sense? E.g. if we only partly populate the context, and then apply a modifier that affects something unpopulated?

Sure. I want to know about movement, so I don't populate the weapon. I apply range modifications, and those range modifications see that there's no weapon, so they no-op (or change a meaningless range number)

src/Units/UnitContext.java
ThislsAUsername: Their whole point is modifying state, yes... but they _can_ modify more fields than just the ones they're supposed to.
Perhaps what I should be doing is enforcing "If a modifier changes something unrelated to what it's supposed to be changing, that's a bug".

src/Units/UnitContext.java
ThislsAUsername: > It would help with clarity - it's confusing to have a constructor that demands two apparently-conflicting parameters with no guidance on when one should be provided over the other.

Either can be `null`, and that is how you indicate how much information you're interested in providing to whatever calculation you're doing.
> It would be cleaner--or at least clearer at point of use-- to just have two constructors to accept either a path or the endpoint.

I dunno about that. The callsites currently (mostly) just pass in the endpoint and `Utils.findShortestPath()` for that endpoint - the caller doesn't necessarily know or care whether `path` is null.
...actually, that's a solid point. If I'm just calling `Utils.findShortestPath()` with arguments I already have, that's overload material.

> It's pointless at best, and worse if it creates confusion and more unnecessary work.

It probably should be there, per my comments in the conversation where you said YAGNI.

src/Engine/UnitMods/CountTracker.java
ThislsAUsername: Yes, no, no.
I don't really feel the need to write tests for `HashMap`. If that's broken, we have bigger fish to fry.

src/Test/TestCombatMods.java
ThislsAUsername: The overlap is intentional, though it probably doesn't need to be run 12 times.
It probably also doesn't need to be run every time the game starts, but here we are.

You're probably right about it being more than is needed.
I don't think I'd have bothered writing the test at all if you hadn't suggested it, though - we're already beyond my personal expectations of what's needed/expected, so I covered all my bases.

src/Units/UnitModel.java
ThislsAUsername: They are used, specifically as the basis for `UnitContext`'s comparable fields, and as implementation hooks for CO D2Ds.
If we wanted immutable `UnitModel`s and also D2Ds, we *could* throw in Lombok to get copy-on-write semantics running, I think.

I don't hate the idea of doing away with these fields entirely and running with `UnitModifier` only, though.

src/Engine/UnitMods/CountTracker.java
ThatIsAUsername: Fair enough. I guess you are returning a pointer to a hashmap, so any changes will be reflected back at the mothership.
I'm not sure I'd encourage that kind of use, but that's neither here nor there.

src/Test/TestCombatMods.java
ThatIsAUsername: I mean, a confusing-but-correct test is better than no test, but a clear test is yet better. On the other hand, YGWYPF.

src/Engine/Combat/CombatEngine.java
ThatIsAUsername: > the sort of thing that could introduce subtle inconsistencies

Yes. This is. Because we didn't pass a path into `calculateOneStrikeDamage`, we invented one, which may not be the same as what the user entered. Want subtle bugs? This is how you get them.

> Where's the bug?

No bug in the scenario you gave. Muriel just isn't making full use of available tactical information.

Actually, not true; there is a bug, since Wally could find an interesting path that grants extra damage, and then we `findShortestPath` instead of using the carefully crafted non-shortest path, and Wally does an attack expecting a buff and then doesn't get it.

src/Units/UnitModel.java
ThatIsAUsername: Are we directly comparing `UnitContext`s anywhere? That seems like a bad idea, given how willy-nilly they are created.
I'm not gonna advocate for copy-on-write, but managing D2D/all buffs with `UnitModifiers` seems fine, and is where I thought you were going (and you already have a unit modifier to change cost, so this seems redundant now).

src/Units/UnitContext.java
ThatIsAUsername: If we can't trust the unit modifiers to only change what they should, we have bigger problems. There should ideally be a unit test per modifier to ensure it always and only does what it should, and then make sure we aren't double-applying. This kind of defensive construct feels... bad.

> If a modifier changes something unrelated to what it's supposed to be changing, that's a bug

Yes by definition?

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
ThislsAUsername: It would have been a bigger concern for the old `COModifiers`, because they had statefulness in that they modified the models' state while active.
With this? If it's not on the list, it's not getting modified.

src/Units/UnitContext.java
ThislsAUsername: > There should ideally be a unit test per modifier to ensure it always and only does what it should

Most of them fall into the level of "trivial", in my book. Adding to or subtracting from a single number is about all they do.
> Yes by definition?

Okay. Sounds good.

src/Engine/Combat/CombatEngine.java
ThislsAUsername: > Yes. This is. Because we didn't pass a path into `calculateOneStrikeDamage`, we invented one, which may not be the same as what the user entered. Want subtle bugs? This is how you get them.

That's a very good point, and why I was vaguely intending to have callers pass in `UnitContext`s to `CombatEngine`, for the most part. I'll go fix that.

src/Units/UnitModel.java
ThislsAUsername: > Are we directly comparing `UnitContext`s anywhere? That seems like a bad idea, given how willy-nilly they are created.

I don't think so. I meant that changes to these fields in the `UnitModel` would propagate to any `UnitContext`s that were created from it.
> managing D2D/all buffs with `UnitModifier`s seems fine, and is where I thought you were going

Huh. Okay. I don't mind. We wanna do that in this PR?

src/Units/UnitContext.java
```diff
@@ -125,12 +125,8 @@ public void setWeapon(WeaponModel w)
     {
       rangeMin = w.rangeMin;
       rangeMax = w.rangeMax;
-      // Build a scratch UnitContext because we don't trust UnitMods with our state
-      UnitContext uc = new UnitContext(this);
       for( UnitModifier mod : mods )
-        mod.modifyAttackRange(uc);
-      rangeMin = uc.rangeMin;
-      rangeMax = uc.rangeMax;
+        mod.modifyAttackRange(this);
```
ThatIsAUsername: Is there ever a case where we might call `setWeapon` more than once on the same `UnitContext`? Should we explicitly make that an error case?

ThislsAUsername: I don't think so, but it's not really an error case.
We set the range to our new weapon's range before modification, and the modifiers shouldn't be changing anything but min/max range.

src/CommandingOfficers/CommanderAbility.java
```diff
@@ -60,21 +68,59 @@ public final void activate(MapMaster gameMap)
...
+  // Defaults to nothing since most powers don't need special revert logic.
+  protected void revert(MapMaster gameMap)
+  {}
 
   /** Allows a CommanderAbility to generate events that will be animated and published. */
   public GameEventQueue getEvents(MapMaster gameMap)
   {
     return new GameEventQueue();
   }
+  public GameEventQueue getRevertEvents(Commander co, MapMaster gameMap)
```
ThatIsAUsername, line 117: Is this needed? It's never overridden or called anywhere

src/CommandingOfficers/Cinder.java
```diff
@@ -69,93 +69,72 @@ public Cinder(GameScenario.GameRules rules)
...
    * Cinder builds units at 8HP ready to act.
-   * To compensate for the ability to continue producing units in a single turn,
-   * the cost of units increases exponentially for repeated purchases from a single property.
    */
   @Override
   public GameEventQueue receiveCreateUnitEvent(Unit unit)
   {
     XYCoord buildCoords = new XYCoord(unit.x, unit.y);
-    if( this == unit.CO && buildCounts.containsKey(buildCoords) )
+    if( this == unit.CO && myView.isLocationValid(buildCoords) )
```
ThatIsAUsername, line 90: Is there any reason to check `isLocationValid` here? We didn't before, and we don't report anything if the check fails.

src/Units/UnitContext.java
ThatIsAUsername: Are you still making a change for this? Wouldn't hurt to beef up the docs for the six constructors while you're at it.

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: Good catch. I do want to support the possibility for future things, though I guess it's not strictly necessary.
The usecase that springs to mind at the moment is "Iron Will's new HP-loss-on-revert effect could be done via events and actually animate that way"

src/CommandingOfficers/Cinder.java
ThislsAUsername: Yes. Carriers can build Seaplanes, and those Seaplanes aren't on a valid map tile (and thus don't interact with the D2D since IDK what to do with it)

src/Units/UnitContext.java
ThislsAUsername: I was originally planning to have the coord-only overload call `Utils.findShortestPath()`, but you pointed out in the other thread that that's a terrible idea.
I should add docs on the constructors, though.

src/AI/AIUtils.java
```diff
@@ -206,7 +206,7 @@ public static GameAction moveTowardLocation(Unit unit, XYCoord destination, Game
 
     if( path.getPathLength() > 0 && validMoves.size() > 0 ) // Check that the destination is reachable at least in theory.
     {
-      path.snip(unit.model.movePower+1); // Trim the path so we go the right immediate direction.
+      path.snip(unit.getMovePower(gameMap)+1); // Trim the path so we go the right immediate direction.
```
ThatIsAUsername, line 209: Should we change the name of `model.movePower` to `baseMovePower` so it's clear that it's no longer the final word?

I guess we could also make it inaccessible, but I'm less a fan of that.

ThislsAUsername: I suppose.
Doing the code change brought to mind our current design of building the BattleMech unit type with Tech. I can see three options here, for when we move to having immutable `UnitModel`s:
1) Don't actually make their fields `final`, so she can just make a copy, modify it, and put it in the global unit type list.
2) Do make their fields `final`, and provide a function to make a modified copy so she can do the above.
3) Force all `UnitModelScheme`s to define a BattleMech unit type.

I think we were planning to do 3 before, but we do have the option of 1 and 2. Thoughts?

src/Units/UnitModel.java
ThatIsAUsername: I don't think I'm opposed. Like I said, I thought that was part of your intent, though to be fair I'm not really sure what the scope of this PR is even supposed to be.

src/CommandingOfficers/CommanderAbility.java
ThatIsAUsername: If we want to have the hook in place, we should be calling it wherever it's supposed to be invoked.

Though this is the sort of thing I probably wouldn't add at all until I had at least one active client.

src/CommandingOfficers/Cinder.java
ThatIsAUsername: Isn't that done via the `UnitProduceEvent`, and thus would never hit this code anyway?

Though it's maybe odd that we have two different events for adding units to the game.

src/AI/AIUtils.java
ThatIsAUsername: I'm fine with sticking to 3, but I'm not necessarily opposed to the others either.

src/Units/UnitModel.java
ThislsAUsername: The scope was supposed to be "Introduce `UnitModifier` so we don't have to directly change `UnitModel`s when activating most abilities".
It didn't occur to me that that same concept would work just as well to implement most D2Ds.
I feel like this PR is big enough at this point that we want to make a new one for new projects like "fully-immutable, shared `UnitModel`s for all COs in a match".

src/CommandingOfficers/Cinder.java
ThislsAUsername: Probably yes, and yes.
I suppose it does make some sense - one is adding directly to the board, while the other is adding into another unit's inventory.
Maybe we should just name them better?

src/AI/AIUtils.java
ThislsAUsername: I think I'll stick to 1 for the immediate future, in the spirit of KISS.

src/CommandingOfficers/Cinder.java
```diff
@@ -192,76 +171,96 @@ protected void perform(MapMaster gameMap)
...
-        }
-        else
-        {
-          // Guess he's not gonna make it.
-          // TODO: Maybe add a debuff event/animation here as well.
-          Utils.enqueueDeathEvent(minion, results);
-        }
+        minion.alterHP(-refreshCost);
+        minion.isTurnOver = false;
+        attackCounts.incrementCount(co, minion);
```
ThatIsAUsername, line 235: Isn't `attackCounts` a `StateTracker` that has been registered to receive events? It seems odd to be manually managing the counts from here.
Or alternatively: it seems weird to have a `StateTracker` to hold the counts.

ThatIsAUsername: Actually, `CountTracker` doesn't even extend `StateTracker`, it's just in that package.

src/Engine/StateTrackers/CountTracker.java
```diff
@@ -0,0 +1,41 @@
+package Engine.StateTrackers;
+
+import java.io.Serializable;
+import java.util.HashMap;
+
+public class CountTracker<RootKeyType, InstanceKeyType> implements Serializable
```
ThatIsAUsername, line 6: The name and location makes it sound like this is a) a `StateTracker` in its own right, and b) that it is doing any kind of counting on its own, when it isn't.
Not sure what a better name would be though: maybe `KeyedSpreadsheet` or something. Maybe change the template type names to `XKey` and `YKey` too since that's basically what they are, and the current names "Root" and "Instance" are--well, they don't help me to understand how this thing is structured, at any rate.

Adding some javadocs wouldn't hurt either.

src/CommandingOfficers/Strong.java
```diff
@@ -116,45 +119,58 @@ public void modifyUnitAttackOnUnit(BattleParams params)
...
+      modList.add(defenseMod);
+      modList.add(damageModTroop);
+      modList.add(moveMod);
+    }
 
-      damageMod = new CODamageModifier(STRONGARM_BUFF);
-      defenseMod = new CODefenseModifier(STRONGARM_BUFF);
-      damageModTroop = new CODamageModifier(STRONGARM_FOOT_BUFF);
-      for( UnitModel model : commander.getAllModels(UnitModel.TROOP) )
+    private HashMap<Commander, UnitProductionModifier> upmsApplied = new HashMap<>();
```
ThatIsAUsername, line 150: Is there any reason to have this be a map instead of just the modifier?

src/CommandingOfficers/Modifiers/COModifier.java
```diff
@@ -1,70 +0,0 @@
-package CommandingOfficers.Modifiers;
-
-import java.io.Serializable;
-import java.util.ArrayList;
-
-import CommandingOfficers.Commander;
-import Units.UnitModel;
-
-public interface COModifier extends Serializable
```
ThatIsAUsername, line 9: At least two implementing classes have now been orphaned now that this is gone. I probably would have kept this around (along with the handling code in `Commander`) so long as valid subclasses exist. Thoughts?

src/CommandingOfficers/Strong.java
```diff
@@ -169,44 +185,50 @@ protected void perform(MapMaster gameMap)
...
+    }
 
-      // Grant a global +2 movement buff.
-      COMovementModifier moveMod = new COMovementModifier(2);
-      for( UnitModel model : myCommander.unitModels )
+    private HashMap<Commander, UnitProductionModifier> upmsApplied = new HashMap<>();
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      // Changing production capabilities probably doesn't make sense to apply to other COs, right?
```
ThatIsAUsername, line 216: Probs not.

src/CommandingOfficers/Tech.java
```diff
@@ -116,63 +119,78 @@ public int getRepairPower()
...
+        if( typesToOverCharge.contains(u.model) )
+        {
+          // Track units that aren't already overhealed
+          if( u.getHP() <= u.model.maxHP
+              && u.getHP() + healAmount > u.model.maxHP )
+            overCharged.add(u);
           u.alterHP(healAmount, true);
+        }
+      }
+      unitsOverCharged.put(myCommander, overCharged);
```
ThatIsAUsername, line 179: Why bother storing the list of overcharged units instead of just trimming any >10 units when the power ends? Is it just to avoid iterating over the whole list again?

And why bother tracking them by commander? a) It is only ever applied to `myCommander` anyway, and b) who each unit belongs to has no impact on the revert logic.

I assume it has something to do with `Army` plans, but we don't have that yet; I'm pretty sure I've stated my opinion before on implementing logic to support things that don't exist yet.

src/CommandingOfficers/Tech.java
```diff
@@ -227,7 +242,7 @@ public GameEventQueue getEvents(MapMaster gameMap)
       Set<XYCoord> dropLocs = new HashSet<XYCoord>();
       for( int i = 0; i < numDrops; ++i )
       {
-        UnitSpawnAction techDrop = generateDropEvent(gameMap, dropLocs, log);
+        UnitSpawnAction techDrop = generateDropEvent(myCommander, gameMap, dropLocs, log);
```
ThatIsAUsername, line 245: Leftover unnecessary `myCommander` parameter, here and several places below.

src/CommandingOfficers/Cinder.java
ThislsAUsername: Yes. We don't have a general `Util` namespace, so that's where I put it.

src/CommandingOfficers/Strong.java
ThislsAUsername: There was until we had those architecture discussions in person.

src/CommandingOfficers/Modifiers/COModifier.java
ThislsAUsername: I would much rather `CommanderAbility`s handle their side-effects on permanent state explicitly - the way it was before created too much room to call the wrong thing and end up with unwanted permanent changes (see: Ave's stacking attack boost bug).
The two remaining cases (production modifier and temporary transformation) are honestly super dissimilar to the point where I'm not sure having a common interface for them makes a lot of sense.

src/CommandingOfficers/Tech.java
ThislsAUsername: My intent is to cleanly interact with units that have >10 HP before the power starts, to handle other COs' units properly.

Tracking by commander was to account for the possibility of other COs activating this power for themselves, independent of the `Tech` instance they're stealing it from. Yeah, yeah, YAGNI and there are other design problems with it.
I'm not super attached to that idea.

I don't feel like YAGNI works quite so well in this instance, since this is possibly subtle bug territory. I'd rather change things now when I'm thinking about them than later when I'm maybe not thinking about them.

src/Engine/StateTrackers/CountTracker.java
ThislsAUsername: Hmm. Would you prefer `CountManager`? `CountTrackerImpl`? Maybe `CountList` or even `CountListList`?
Agreed that the name isn't the best.

`KeyedSpreadsheet` doesn't indicate anything at all, to me.
`XKey` and `YKey` are similarly unintuitive - I didn't conceive of this as mappable to Cartesian space until I read those words just now.
"Root" key is the key for a particular list of counts; I guess it's comparable to a customer at a restaurant.
"Instance" key is the key to a particular number *in* a list of counts; I guess it's comparable to the individual dishes the restaurant customer has ordered at least one of.

Yeah, docs wouldn't hurt.

src/CommandingOfficers/Modifiers/COModifier.java
ThatIsAUsername: OK, fair enough.

src/CommandingOfficers/Tech.java
ThatIsAUsername: > to handle other COs' units properly

This... doesn't handle other CO's units

> other COs activating this power for themselves

This is also not a thing that happens. Even in your alternate coop paradigms, nobody's activating somebody else's powers, right?

> subtle bug territory

How so? Seems more complex to manage unnecessary CO associations. Plus Tech just doesn't have another way to have units with >10 HP, so simply shaving anything higher than that is guaranteed to be correct. (unless you try real hard to contrive a scenario).

src/Engine/StateTrackers/CountTracker.java
ThatIsAUsername: I mean, it's effectively a 2D array, just with objects instead of numeric indices. But `2DWhatever` isn't a valid variable name. `CountManager` or whatever is probably fine.

src/CommandingOfficers/Tech.java
ThislsAUsername: > This... doesn't handle other CO's units

Overcharge seems like a perfectly reasonable thing to apply to all COs in an Army.
Sure, it doesn't do that *today*, but I'm really not interested in ripping out code I wrote to handle a case that I'm about to implement, just because I haven't yet implemented it.
> This is also not a thing that happens. Even in your alternate coop paradigms, nobody's activating somebody else's powers, right?

Yes, agreed. There is *some* vague possibility of that being in-scope eventually, but I imagine we can burn that bridge if and when we decide we want to cross it. (It occurs to me that probably the simple solution is to just clone the power for that weird case and not make everything more complex than it needs to be)
> How so? Seems more complex to manage unnecessary CO associations. Plus Tech just doesn't have another way to have units with >10 HP, so simply shaving anything higher than that is guaranteed to be correct. (unless you try real hard to contrive a scenario).

That it's more complex is strictly correct. That is the cost for handling more complex situations correctly.
You're also correct (today) that Tech can't get >10 HP units any other way, and that that behavior is correct today.
Technically speaking, we also don't have any COs *other* than Tech that can produce >10 HP units, and she always reverts on powers... so tags won't technically cause a problem with the COs we have either.
If we ever add a CO with >10 HP units outside of power turns, it'll suddenly become a problem.

Maybe the solution here is to formalize how >10 HP works - you aren't allowed to go above your max *ever* (so Tech adjusts the max accordingly), and damage is always calculated as if 10 HP is the max.
That's kinda how I expected it to work in the first place, but didn't object at the time because your implementation works and doesn't break anything today.

ThatIsAUsername: > If we ever add a CO with >10 HP units outside of power turns, it'll suddenly become a problem

Only if Tech is attempting to revert her power on someone else's units, which shouldn't be a thing.

> Maybe the solution here is to formalize how >10 HP works - you aren't allowed to go above your max ever

There is a certain appeal to having a max that is actually a proper max. I probably didn't do it that way to start with because then you're into either mutating UnitModels or doing remodeling, and either way is extra complexity. Plus doing it this way offers an HP-based firepower buff which is also fun (but ofc we could just use a normal buff).

ThislsAUsername: > Only if Tech is attempting to revert her power on someone else's units, which shouldn't be a thing.

Which *should* be a thing, in like two PRs.
> There is a certain appeal to having a max that is actually a proper max.

Agreed.
> I probably didn't do it that way to start with because then you're into either mutating UnitModels or doing remodeling, and either way is extra complexity.

You had to change how `Unit` worked to accommodate that, though. Ideally, exceptions to rules that COs create should be located in the CO's file.
> Plus doing it this way offers an HP-based firepower buff which is also fun (but ofc we could just use a normal buff).

Again, we can define this however we like. We could have a "nominal max HP" that gets used for damage calculations, vs the "true max HP", which enforces how much HP you can have.
You're already using a normal buff - you explicitly wanted the multiplicative effect of overcharged HP alongside an attack buff. I'm not sure the numbers come out all that differently, but hey.

src/CommandingOfficers/Modifiers/UnitInstanceFilter.java
ThatIsAUsername: It's arguably better practice not to have an internal object as part of the public API, but we do that elsewhere anyway.

I guess part of my objection is just that after creating a filter I'd naturally look for an "add..." method and then have to figure out to use `instances` directly instead. Discoverability is lower with this approach.

Not a hill I'm gonna die on though. If I care that much, I might just add the function later.

src/CommandingOfficers/Cinder.java
ThatIsAUsername: Yeah, better names might be a good idea.

In the meantime, if we are going to add a check here that could cause the operation to fail, we need to provide some kind of feedback if the operation fails. I'm fine with either printing a warning or removing the unnecessary check.

src/CommandingOfficers/Tech.java
ThatIsAUsername: > Which should be a thing, in like two PRs.

Right. The "one team, two armies" approach. Having like three different ways to handle allied COs is going to add a lot of subtle bug potential - I do hope you have a robust testing approach in the works to ensure it works as it should.

> not interested in ripping out code I wrote to handle a case that I'm about to implement, just because I haven't yet implemented it

This is understandable. On the other hand, I don't like to merge code changes into the release branch that aren't needed for anything, and may need to be changed again once we implement the proper use case. I won't make you change it back this time, but you really really need to get better at keeping your PRs focused.

src/Test/TestCombatMods.java
ThislsAUsername: I assume this is fair to kick out to the next PR, since it directly pertains?

src/CommandingOfficers/Cinder.java
ThislsAUsername: I just looked at the code - both units-producing-units and buildings-producing-units call `receiveCreateUnitEvent()`.
This is a perfectly valid check, and not an error case.

src/CommandingOfficers/Tech.java
ThislsAUsername: > Right. The "one team, two armies" approach. Having like three different ways to handle allied COs is going to add a lot of subtle bug potential - I do hope you have a robust testing approach in the works to ensure it works as it should.

*Team'd COs. Allies (as of this writing, anyway) aren't planned to interact meaningfully.
I imagine it would be sufficient to have a single set of operations on a game state defined with different tag behaviors, and observe whether it results in valid behavior in each individual case.
> This is understandable. On the other hand, I don't like to merge code changes into the release branch that aren't needed for anything

If there's a release branch, it's not `master` at the moment. But I get what you're saying.
> you really really need to get better at keeping your PRs focused.

Agreed.

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: Done.

src/Test/TestCombatMods.java
ThatIsAUsername: Yeah, that's fair.

src/CommandingOfficers/Cinder.java
ThatIsAUsername: Huh. Interesting that `UnitProduceEvent` calls `receiveCreateUnitEvent` instead of having its own listener. Explains why I didn't catch that connection.

Fair enough I guess, but I'm not sure how I feel about having a `UnitProduceEvent` with no listener/re-using an existing listener.

src/CommandingOfficers/Venge.java
```diff
@@ -237,6 +238,14 @@ protected void revert(MapMaster gameMap)
       }
       boostedUnits.clear();
     }
+
+    @Override
+    public GameEventQueue getRevertEvents(MapMaster gameMap)
```
ThatIsAUsername, line 243: Did this part of Ironwill just not work before?

src/CommandingOfficers/Tech.java
ThatIsAUsername: > If there's a release branch, it's not master

Why not? We shouldn't be merging stuff into master until it is known to be good.

src/CommandingOfficers/Strong.java
ThatIsAUsername: So should we simplify this then?

src/CommandingOfficers/Venge.java
ThislsAUsername: It worked just fine. It was in `revert()` rather than actually pushing out an event.

src/CommandingOfficers/Cinder.java
ThislsAUsername: IDK, it seems sensible to have an all-in-one place to listen for new units appearing.
Perhaps it also makes sense to provide separate listener hooks for those specific events, though.

src/CommandingOfficers/Tech.java
ThislsAUsername: Because we've made releases, and none of them were from `master`?
To be fair, this is kind of irregular and ideally this shouldn't work this way.

src/CommandingOfficers/Venge.java
ThatIsAUsername: Well, it looks like that is still there. Except it heals them because `IRONWILL_WOUND` is now positive. An 8 HP unit which gets iron willed will still be at 8 afterwards.

src/CommandingOfficers/Cinder.java
ThatIsAUsername: That's probably what I would have done. That or just make production use the `UnitCreateEvent`. Changing that is certainly out of scope for this PR though.

src/CommandingOfficers/Tech.java
ThatIsAUsername: True enough; master should certainly be stable though.

src/CommandingOfficers/Strong.java
ThislsAUsername: Done.

ThatIsAUsername:
I think we've pretty much resolved all the discussions at this point. Anything left to iron out? If not, we can probably merge.

PR #347, opened 2021-10-18
Switch the InfoView/Controller to scroll vertically smoothly

ThislsAUsername:
Should probably still add an indicator when you can swap pages horizontally, but I'm not gonna do that at this particular moment.


src/UI/Art/SpriteArtist/DamageChartView.java
```diff
@@ -0,0 +1,220 @@
...
+  private SlidingValue viewY = new SlidingValue(0);
+  private SpriteCursor spriteCursor = new SpriteCursor(0, 0, SpriteLibrary.baseSpriteSize, SpriteLibrary.baseSpriteSize, UIUtils.getCOColors()[0]);
+
+  public DamageChartView(DamageChartController control)
+  {
+    this.control = control;
+    unitArray = new BufferedImage[control.units.length];
+
+    // Just pull one sprite to start with. Loading everything at once might take a while.
+    unitArray[0] = SpriteLibrary.getMapUnitSpriteSet(control.units[0].name, control.shooters, control.shooterColor).sprites[0].getFrame(0);
```
ThatIsAUsername, line 38: Did you actually see lag trying to do all at once? Presumably the unit images should already be loaded by the time we get here anyway.

src/UI/Art/SpriteArtist/DamageChartView.java
```diff
@@ -0,0 +1,220 @@
...
+    startY = findViewDestination(selY, gridHeight, unitSizePx, unitBuffer, startY, myHeight - unitSpacingV, marginY);
+
+    viewX.set(startX, false);
+    viewY.set(startY, false);
+
+    // Render it in true size, and then we'll scale it to window size as we draw it.
+    BufferedImage image = SpriteLibrary.createTransparentSprite(myWidth, myHeight);
+    Graphics myG = image.getGraphics();
+
+    // Draw a highlight on the current cursor row/column
```
ThatIsAUsername, line 82: It's more of a lowlight at present. But I mean, it works.

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -53,81 +57,111 @@ public void render(Graphics g)
...
-        break;
-      case BASIC:
-        BufferedImage infoText = SpriteUIUtils.drawProseToWidth(page.info, drawingWidth);
-        myG.drawImage(infoText,3*paneOuterBuffer, 3*paneOuterBuffer, null);
-        break;
-    }
-    // Destroy stale state so we don't get into weird situations
-    if( page.pageType != PageType.GAME_STATUS )
-      statsCO = null;
+    myG.fillRect(2*PANE_OUTER_BUFFER, 2*PANE_OUTER_BUFFER, paneHSize                      , paneVSize                      );
```
ThatIsAUsername, line 88: Cool spacing, bro.

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -53,81 +57,111 @@ public void render(Graphics g)
...
+    int drawingWidth = paneHSize - PANE_OUTER_BUFFER;
+    if( pages != prevPages )
+      pageImage = renderPage(pages, drawingWidth);
+    prevPages = pages;
+
+    final int shiftDown = myControl.getShiftDown() * 2;
+    final int possibleShift = Math.max(0, pageImage.getHeight() - paneVSize);
+    int pixelShift = 0;
+    if( possibleShift > 0 )
+      pixelShift = (possibleShift + shiftDown % possibleShift) % possibleShift;
```
ThatIsAUsername, line 99: Comments are great.

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -53,81 +57,111 @@ public void render(Graphics g)
...
+      {
+        case CO_HEADERS:
+        case GAME_STATUS: // Fall-through
+          Commander[] coList = myControl.getGame().commanders;
+          MapPerspective drawableMap = getDrawableMap(myControl.getGame());
+
+          for( Commander thisCO : coList )
+          {
+            BufferedImage statusHeader =
+                SpriteLibrary.createTransparentSprite(drawingWidth, CommanderOverlayArtist.OVERLAY_HEIGHT + 5);
```
ThatIsAUsername, line 125: Is it +5 to make room for the power bar?

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -111,6 +111,8 @@ public static BufferedImage drawProseToWidth(String prose, int reqWidth)
       String line = lines.get(i);
       if( font.getWidth(line) <= reqWidth ) // if the line's short enough already, don't split it further
         continue;
+      if( i > 100 ) // bail; the text is too long or the space too small
+        return SpriteLibrary.createDefaultBlankSprite(1, 1);
```
ThatIsAUsername, line 115: This is probably fine, but it wouldn't hurt to document more clearly why it's needed.

src/UI/CO_InfoController.java
```diff
@@ -57,13 +58,16 @@ public boolean handleInput( InputAction action )
     switch( action )
     {
       case DOWN:
+        shiftDown++;
```
ThatIsAUsername, line 61: Might want to up the scroll speed. It gets cumbersome when there is a lot of text. Or maybe scroll for about a second at the slow speed and then increase it (double it or whatever).
For extra cool, it might be nice to remove the pause between the initial keypress and the OS repeat kicking in.

src/UI/GameStatsController.java
```diff
@@ -13,23 +13,26 @@
...
-    coOptionSelector = new OptionSelector(myGame.commanders.length);
+    infoPages = new ArrayList<ArrayList<InfoPage>>();
+
+    ArrayList<InfoPage> headers = new ArrayList<InfoPage>();
+    headers.add(new InfoPage(InfoPage.PageType.CO_HEADERS));
+    infoPages.add(headers);
+
+    ArrayList<InfoPage> status = new ArrayList<InfoPage>();
+    status.add(new InfoPage(InfoPage.PageType.GAME_STATUS));
+    infoPages.add(status);
```
ThatIsAUsername, line 33: I think we decided it would make sense to break the detailed info into separate pages per CO.

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -53,81 +57,111 @@ public void render(Graphics g)
...
+    final int contentHeight = Math.min(Math.abs(paneVSize - PANE_OUTER_BUFFER), pageImage.getHeight());
+    // Draw the cropped page into our pane
+    myG.drawImage(pageImage.getSubimage(0, pixelShift, contentWidth, contentHeight), 3 * PANE_OUTER_BUFFER, 2 * PANE_OUTER_BUFFER, null);
 
     // Finally, draw our rendered image onto the window.
     g.drawImage(image, 0, 0, imageWidth*drawScale, imageHeight*drawScale, null);
   }
 
-  
+  private BufferedImage renderPage(ArrayList<InfoPage> pages, int drawingWidth)
```
ThatIsAUsername, line 110: It would be nice to add arrows in the top- and bottom-right corners of the text pane to show that it can scroll up and down. Maybe disappear them if it all fits in one page.

src/UI/Art/SpriteArtist/DamageChartView.java
```diff
@@ -0,0 +1,220 @@
...
+    for(int i = 0; i < control.units.length; ++i)
+    {
+      int yOff = 0;
+      int xOff = unitSpacingH * (i+1) - viewX.geti();
+      if( (xOff > myWidth ) || (xOff < unitSpacingH) )
+        continue;
+      drawTargetAt(myG, i, xOff, yOff);
+    }
+
+    // t = target, s = shooter
```
ThatIsAUsername, line 97: Would be good to delineate that left is shooter and top is target. Maybe just by using the current player's color on the left and some other color on the top.
This creates the potential issue that if you choose a different player color along the top, it could give the impression that this table takes current CO choices into account, when the values are bland. Maybe a header above the table to specify that these are default values? But that's less useful - players will want values relevant to their current CO by default.

src/UI/Art/SpriteArtist/DamageChartView.java
ThislsAUsername: Copy/pasted comment, to be fair.
Also, this view is guaranteed to show all unit types (including sea/air), which aren't guaranteed to ever be loaded otherwise.

src/UI/DamageChartController.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public boolean handleInput(InputAction action)
+  {
+    boolean done = false;
+    switch(action)
+    {
+      case SEEK:
+        outOfAmmo = !outOfAmmo;
```
ThatIsAUsername, line 52: Would be good to have a visual indicator that this option is available.

src/UI/Art/SpriteArtist/InfoView.java
ThislsAUsername: Yes; that's copied from the CO overlay artist, and probably should be a constant?

src/UI/CO_InfoController.java
ThislsAUsername: Putting this off until we start routing input through the view side rather than the controller side.

src/UI/DamageChartController.java
ThislsAUsername: There's no real estate for it, at present, unless you expand the screen.
Even something non-obvious like "press Q" wouldn't really fit even in the top corner.

I guess it'd be possible to just sacrifice the real estate for the prompt *sometimes*, and have a hidden config setting (that gets automatically toggled when you use Q here) to hide the prompt and use all the space?

src/UI/Art/SpriteArtist/DamageChartView.java
ThatIsAUsername: Every unit type, yes, but in only two colors.

PR #351, opened 2021-12-27
Immutable models

ThislsAUsername:
I think this one's a bit more focused than the last one.
I even found bugs and didn't fix them.

### DIFF
--- a/src/AI/Muriel.java
+++ b/src/AI/Muriel.java
@@ -551,7 +551,7 @@ public class Muriel implements AIController
     //////////////////////////////////////////////////////////////////
     // We didn't find an immediate ATTACK or CAPTURE action we can do.
     // Things that can capture; go find something to capture, if you are moderately healthy.
-    if( unit.model.hasActionType(UnitActionFactory.CAPTURE) && (unit.getHP() >= 7) )
+    if( unit.hasActionType(UnitActionFactory.CAPTURE) && (unit.getHP() >= 7) )
     {
       log(String.format("Seeking capture target for %s", unit.toStringWithLocation()));
       XYCoord unitCoords = new XYCoord(unit.x, unit.y);
@@ -589,7 +589,7 @@ public class Muriel implements AIController
 
     //////////////////////////////////////////////////////////////////
     // Everyone else, go hunting.
-    if( queuedActions.isEmpty() && unit.model.hasActionType(UnitActionFactory.ATTACK) )
+    if( queuedActions.isEmpty() && unit.hasActionType(UnitActionFactory.ATTACK) )
     {
       log(String.format("Seeking attack target for %s", unit.toStringWithLocation()));
       ArrayList<XYCoord> enemyLocations = AIUtils.findEnemyUnits(myCo, gameMap); // Get enemy locations.
--- a/src/AI/Muriel.java
+++ b/src/AI/Muriel.java
@@ -174,8 +174,8 @@ public class Muriel implements AIController
     if( myDamage == 0 && otherDamage != 0 ) invRatio = 10000;
     UnitModel myModel = myUnit.model;
     UnitModel otherModel = otherUnit.model;
-    double costRatio = damageRatio * ((double)otherModel.getCost() / myModel.getCost());
-    double otherCostRatio = invRatio * ((double)myModel.getCost() / otherModel.getCost());
+    double costRatio = damageRatio * ((double)otherUnit.getCost() / myUnit.getCost());
+    double otherCostRatio = invRatio * ((double)myUnit.getCost() / otherUnit.getCost());
     myUnitEffectMap.put(new UnitModelPair(myModel, otherModel), new UnitMatchupAndMetaInfo(damageRatio, costRatio));
     myUnitEffectMap.put(new UnitModelPair(otherModel, myModel), new UnitMatchupAndMetaInfo(invRatio, otherCostRatio));
 
@@ -196,7 +196,7 @@ public class Muriel implements AIController
 
     // Order the units by cost.
     Collections.sort(myCo.units, (Unit u1, Unit u2) ->
-      (int)(u2.model.getCost()*u2.getHP() - u1.model.getCost()*u1.getHP()));
+      (int)(u2.getCost()*u2.getHP() - u1.getCost()*u1.getHP()));
 
     // If we are already capturing any of these properties, remove them from the list.
     for( Unit unit : myCo.units )
@@ -456,7 +456,7 @@ public class Muriel implements AIController
               double hpDamage = Math.min(results.defender.getPreciseHPDamage(), results.defender.unit.getPreciseHP());
 
               if( shouldAttack(unit, results.defender.unit, gameMap) )
-                return (results.defender.model.getCost() / 10) * hpDamage;
+                return (results.defender.unit.getCost() / 10) * hpDamage;
 
               return 0.;
             }, (terrain, params) -> 0.); // Don't mess with terrain
@@ -812,7 +812,7 @@ public class Muriel implements AIController
         int gruntsWanted = (int)Math.ceil(myCo.units.size() * INFANTRY_PROPORTION);
         int gruntFacilities = CPI.getNumFacilitiesFor(infModel)-1; // The -1 assumes we are about to build from a factory. Possibly untrue.
         if( gruntFacilities < 0 ) gruntFacilities = 0;
-        costBuffer = (int)Math.min(gruntFacilities, gruntsWanted) * infModel.getCost();
+        costBuffer = (int)Math.min(gruntFacilities, gruntsWanted) * myCo.getCost(infModel);
         log(String.format("  Low on Infantry: witholding %s for possible extra grunts", costBuffer));
       }
 
@@ -821,7 +821,7 @@ public class Muriel implements AIController
       HashSet<UnitModel> counters = new HashSet<UnitModel>();
       for( UnitModel counter : CPI.availableUnitModels )
       {
-        if( myCo.money < counter.getCost() ) continue; // If we can't afford it, don't bother.
+        if( myCo.money < myCo.getCost(counter) ) continue; // If we can't afford it, don't bother.
         UnitMatchupAndMetaInfo umami = getUnitMatchupInfo(counter, enemyToCounter);
         if( umami.costEffectivenessRatio >= COST_EFFECTIVENESS_MIN )
         {
@@ -929,14 +929,14 @@ public class Muriel implements AIController
         // Figure out if we can afford the desired unit type.
         int maxBuildable = CPI.getNumFacilitiesFor(idealCounter);
         log(String.format("    Facilities available: %s", maxBuildable));
-        int cost = idealCounter.getCost();
+        int cost = myCo.getCost(idealCounter);
         if( cost <= (budget - costBuffer))
         {
           // Go place orders.
           log(String.format("    I can build a %s for a cost of %s", idealCounter, cost));
           MapLocation loc = CPI.getLocationToBuild(idealCounter);
-          shoppingCart.add(new PurchaseOrder(loc, idealCounter));
-          budget -= idealCounter.getCost();
+          shoppingCart.add(new PurchaseOrder(loc, myCo, idealCounter));
+          budget -= cost;
           CPI.removeBuildLocation(loc);
           orderedSomething = true;
 
@@ -950,17 +950,18 @@ public class Muriel implements AIController
           }
           break; // Loop around, re-sort the enemies by strength, and figure out what to build next.
         }
-        else {log(String.format("    %s cost %s, I have %s (witholding %s).", idealCounter, idealCounter.getCost(), budget, costBuffer));}
+        else {log(String.format("    %s cost %s, I have %s (witholding %s).", idealCounter, cost, budget, costBuffer));}
       } // ~while( !availableUnitModels.isEmpty() )
     } // ~while( still choosing units to build )
 
     // Build infantry from any remaining facilities.
     UnitModel infModel = myCo.getUnitModel(UnitModel.TROOP);
-    while( (budget >= infModel.getCost()) && (CPI.availableUnitModels.contains(infModel)) )
+    final int infCost = myCo.getCost(infModel);
+    while( (budget >= infCost) && (CPI.availableUnitModels.contains(infModel)) )
     {
       MapLocation loc = CPI.getLocationToBuild(infModel);
-      shoppingCart.add(new PurchaseOrder(loc, infModel));
-      budget -= infModel.getCost();
+      shoppingCart.add(new PurchaseOrder(loc, myCo, infModel));
+      budget -= infCost;
       CPI.removeBuildLocation(loc);
     }
 
@@ -1040,18 +1041,20 @@ public class Muriel implements AIController
   private static class PurchaseOrder implements Comparable<PurchaseOrder>
   {
     MapLocation location;
+    Commander co;
     UnitModel model;
 
-    public PurchaseOrder(MapLocation loc, UnitModel um)
+    public PurchaseOrder(MapLocation loc, Commander co, UnitModel um)
     {
       location = loc;
+      this.co = co;
       model = um;
     }
 
     @Override
     public int compareTo(PurchaseOrder other)
     {
-      return model.getBuyCost(location.getCoordinates()) - other.model.getBuyCost(location.getCoordinates());
+      return co.getBuyCost(model, location.getCoordinates()) - other.co.getBuyCost(other.model, location.getCoordinates());
     }
   }
 
--- a/src/AI/SpenderAI.java
+++ b/src/AI/SpenderAI.java
@@ -20,6 +20,7 @@ import Terrain.GameMap;
 import Terrain.MapLocation;
 import Terrain.TerrainType;
 import Units.Unit;
+import Units.UnitContext;
 import Units.UnitModel;
 
 /**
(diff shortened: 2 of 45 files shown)


src/Units/UnitContext.java
```diff
@@ -158,6 +169,54 @@ public void setWeapon(WeaponModel w)
     }
   }
 
+  public int getCostTotal()
+  {
+    return (int) (costBase*costMultiplier)+costShift;
+  }
```
ThatIsAUsername, line 175: Before we had `getCost` and `getBuyCost` - is this going to apply a temporary price change to e.g. power accumulation calculations?

src/Units/Unit.java
```diff
@@ -218,9 +218,19 @@ public boolean isFullySupplied()
...
+  {
+    return this.CO.getCost(model);
+  }
+
   public int getRepairCost()
   {
-    return model.getRepairCost(new UnitContext(this));
+    UnitContext uc = new UnitContext(this);
+    for( UnitModifier mod : getModifiers() )
+      mod.modifyCost(uc);
```
ThatIsAUsername, line 229: Does this mean temporary stat/cost changes also affect repair costs? Is that intended?

src/Units/UnitContext.java
ThislsAUsername: These values should get modified by any relevant modifiers before this function is called.
`getCost()` and `getBuyCost()` still exist, they just moved out of `UnitModel` since it no longer has all the information to calculate the result.

Temporary price changes still apply, and still should change power accumulation.

src/Units/Unit.java
ThislsAUsername: There are no non-dynamic stat/cost changes anymore, so yes and yes.

src/UI/Art/Animation/MoveAnimation.java
```diff
@@ -14,12 +15,11 @@
...
   private final double tilesPerMs;
 
-  public MoveAnimation(int tileSize, Unit actor, GamePath path)
+  public MoveAnimation(int tileSize, GameMap map, Unit actor, GamePath path)
   {
     super(tileSize, actor, null);
     this.path = path;
-    // We don't need an *exact* speed, so this should be fine
-    tilesPerMs = actor.model.getMovePower() / (double) maxTime;
+    tilesPerMs = actor.getMovePower(map) / (double) maxTime;
```
ThatIsAUsername, line 22: This seems a bit unnecessary. Also I still don't know why we need the map to calculate move power.

ThatIsAUsername: Actually, `UnitContext.calculateMovePower()` does not require the map. Why does this?

ThislsAUsername: Gotta nyoom at accurate velocity :v
I have vague plans for a CO that gets bonus movement based on the terrain the unit starts on. I suppose it would be possible to do that via some hacky stuff in `initTurn()`, but it'd be kind of a pain.

src/CommandingOfficers/Tech.java
```diff
@@ -552,7 +552,7 @@ private UnitModel createBattleMechModel()
     UnitModel BattleMech = mdTank.clone();
     BattleMech.name = "BattleMech";
     BattleMech.role = BattleMech.role | UnitModel.SURFACE_TO_AIR;
-    BattleMech.costShift = mdTank.getCost() + (antiAir.getCost()/2);
+    BattleMech.costBase = mdTank.costBase*2 + (antiAir.costBase/2);
```
ThatIsAUsername, line 563: x2?

ThislsAUsername: Before it was `costShift`, which added onto the original Md Tank cost.
This is setting `costBase`, so I included the cost of the original Md Tank.

If you don't want that, okay. I was just keeping the semantics the same.

src/CommandingOfficers/Meridian.java
```diff
@@ -59,17 +60,17 @@ public Commander create(GameScenario.GameRules rules)
...
 
     // Meridian's basic tanks and arty cost the same
-    UnitModel tank = getUnitModel(UnitModel.ASSAULT);
-    UnitModel arty = getUnitModel(UnitModel.SIEGE);
-    int costShift = (tank.getCost() - arty.getCost())/2;
-    tank.costShift -= costShift;
-    arty.costShift += costShift;
+    tank = getUnitModel(UnitModel.ASSAULT);
+    arty = getUnitModel(UnitModel.SIEGE);
+    costShift = (getCost(tank) - getCost(arty))/2;
```
ThatIsAUsername, line 77: Would it be better to handle this as a filtered unit modifier?

ThislsAUsername: Thought about that. I even wrote the class to filter by a single `UnitModel`.
I feel like it's cleaner to just have the CO do it, in this case - either I'm making two filters for two separate operations, or I'm making a new filter type *and* a new custom modifier type that both changes cost and adds an action.

Also, I'm slightly disposed towards avoiding more modifier instances - they're not free since they end up no-oping on basically any calculation you want to do with a unit.

src/UI/Art/Animation/MoveAnimation.java
ThislsAUsername: > Actually, `UnitContext.calculateMovePower()` does not require the map. Why does this?

`UnitContext` has a pointer to the map in it, which will be accessible by the modifiers.

src/AI/Muriel.java
```diff
@@ -551,7 +551,7 @@ else if( obst.isTurnOver || unitSelector.isObstacle(obst) )
     //////////////////////////////////////////////////////////////////
     // We didn't find an immediate ATTACK or CAPTURE action we can do.
     // Things that can capture; go find something to capture, if you are moderately healthy.
-    if( unit.model.hasActionType(UnitActionFactory.CAPTURE) && (unit.getHP() >= 7) )
+    if( unit.hasActionType(UnitActionFactory.CAPTURE) && (unit.getHP() >= 7) )
```
ThatIsAUsername, line 554: Do you not `calculatePossibleActions` here because we are just checking for a default action type?

ThislsAUsername: I do actually `calculatePossibleActions` - the new `Unit` copy of `hasActionType` invokes it.

src/CommandingOfficers/Tech.java
```diff
@@ -495,7 +496,7 @@ private XYCoord findDropLocation(GameMap gameMap, Set<XYCoord> priorDrops, boole
       for( XYCoord pdc : options )
       {
         // No unfortunate accidents
-        if( !unitModelToDrop.propulsion.canTraverse(gameMap.getEnvironment(pdc)) )
+        if( !new UnitContext(myCommander, unitModelToDrop).calculateMoveType().canTraverse(gameMap.getEnvironment(pdc)) )
```
ThatIsAUsername, line 506: This is like... so ugly.

ThislsAUsername: I could hide it in a function call. Or we could just assume that nobody's ever going to change Tech's movetype such that some terrain is no longer traversable.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: The current approach of using a single immutable shared `UnitModel` does lend itself to more modifiers though - thoughts on using an immutable model per CO? Not sure it would actually be better, but there might be some benefits.

ThislsAUsername: It does, though if we're super worried about it we can just make sure that the D2D ends up in the `Commander` directly. The reason I didn't do so for Strong/Venge/etc is basically laziness - it's just as clear and fewer lines to do it this way.

As for "immutable" models per CO, that's... basically what we have already, as of merging the last PR?
If we modify the model directly for the D2D, then it's not really immutable.
If we *don't* modify the model directly for the D2D, then why not make the models common to all COs?

The benefit of this design is that we don't need an enum to map unit types across COs - everyone just uses the same models, so you can check equality/etc safely across CO lines.

src/Units/UnitModelScheme.java
```diff
@@ -28,11 +28,20 @@ public String toString()
...
-  public abstract GameReadyModels getGameReadyModels();
+  private GameReadyModels grms = null;
+  public final GameReadyModels getGameReadyModels()
+  {
+    if(grms != null)
+      return grms;
+    grms = buildGameReadyModels();
+    return grms;
+  }
+  public abstract GameReadyModels buildGameReadyModels();
```
ThatIsAUsername: Why create this extra layer of indirection?

src/Units/UnitModel.java
```diff
@@ -60,32 +59,29 @@ public static String standardizeID(String input)
     return input.toLowerCase().replaceAll(" ", "_").replaceAll("-", "_");
   }
 
+  public int costBase;
+  public int baseMovePower;
+  public MoveType baseMoveType;
+  public ArrayList<UnitActionFactory> baseActions = new ArrayList<UnitActionFactory>();
+  public int baseCargoCapacity = 0;
+
+  // If you want to modify anything below this line, you're gonna have to implement it into UnitContext
```
ThatIsAUsername, line 68: `healableHabs` is below the line, but it's modified for BattleMech (and not in UnitContext).

src/CommandingOfficers/CommanderInfo.java
```diff
@@ -19,9 +20,14 @@ public CommanderInfo(String name)
 
   public abstract Commander create(GameScenario.GameRules rules);
 
+  public void injectUnits(GameReadyModels grms)
+  {
+  }
+
```
ThatIsAUsername, line 26: Why make this part of `CommanderInfo` instead of `Commander`?

src/Units/UnitModelScheme.java
ThislsAUsername: It was an easy way to make everyone share the same `UnitModel`s without forcing every `UnitModelScheme` to do the caching independently or figure out somewhere else to put it?
The UMS instance will sit in the rules object for the lifetime of the game, so this seems like a reasonable solution to me.

src/Units/UnitModel.java
ThislsAUsername: BattleMech doesn't count because it's a *new* unit type, not really a CO changing an existing unit type.
I suppose I should be more precise with the comment, though.

src/Units/UnitContext.java
ThatIsAUsername: > still should change power accumulation

Do the carts work that way? Less power accumulation when price discounts are in effect? I don't think we did that before, at any rate.

src/CommandingOfficers/CommanderInfo.java
ThislsAUsername: I wanted to inject the unit type before any `Commander` was instantiated, so all `Commander` constructors see it; otherwise, you could potentially get different behavior depending on CO ordering.
I guess you technically could anyway, but "inject units" is a much more narrow scope than the `Commander` constructor.

src/Units/Unit.java
ThatIsAUsername: Are repair costs discounted on the carts when units are on sale? Is this "intended" because it's how we prefer things to work, or because they naturally work that way with this new architecture approach? I don't think repairs were discounted before.

src/Units/UnitContext.java
ThislsAUsername: Exactly one cart works that way (AW2), but yes. Hachi (S)COP actually drastically lowers how much charge you can get from killing his units in AW2.
We didn't have anything in place to prevent it before, and I'm pretty sure I tested to verify that we *did* take discounts into account.

src/Units/Unit.java
ThislsAUsername: I can't imagine any way they wouldn't be. It affects COP charge accumulation in AW2. In any case, the only situation where it'd possibly be relevant is BBoat repairs - you can't activate your power before repairs happen, and it doesn't stay active for the following repair phase.
It's intended because I can't imagine it works differently on cart, though I haven't tested this specific case.
As far as our implementation, I don't think there's been any time where repairs weren't discounted any time units were discounted (Bear would be really bad otherwise, until the repair cost change).

src/CommandingOfficers/Meridian.java
ThatIsAUsername: > It does, though if we're super worried about it we can just make sure that the D2D ends up in the Commander directly.

We could. Though the Commander is a UnitModifier itself anyway, so... potato potato.

> As for "immutable" models per CO, that's... basically what we have already, as of merging the last PR?

Maybe. It was the initial vision for unit models, at any rate, and it would avoid some of the ugliness of e.g. not knowing how far your units can move without doing some work. On the other hand, it's nice to have common models across the board, and that wasn't possible in the initial incarnation.

src/CommandingOfficers/Tech.java
ThatIsAUsername: That is probably a safe assumption, but I also try to avoid thinking any assumption is safe.

src/AI/SpenderAI.java
```diff
@@ -221,7 +223,7 @@ else if( unitQueue.isEmpty() )
             boolean validTarget = false;
             ArrayList<XYCoord> validTargets = new ArrayList<>();
 
-            if( unit.model.possibleActions.contains(UnitActionFactory.CAPTURE) )
+            if( uc.calculatePossibleActions().contains(UnitActionFactory.CAPTURE) )
```
ThatIsAUsername, line 226: Would it be more idiomatic to call `unit.hasActionType(UnitActionFactory.CAPTURE)`?

ThatIsAUsername: I guess it could incur extra overhead if you check for the presence of more than one action type (as Wally does), but it also feels annoying to have to create a `UnitContext` to learn about the `Unit` I already have.

src/CommandingOfficers/Meridian.java
ThislsAUsername: The Commander is a UnitModifier, yes, but it's always going to be there, therefore it's always going to be less processing overall to put functions into it rather than adding them to `Commander`'s list.
> Maybe. It was the initial vision for unit models, at any rate

I can believe that, but me-of-yesteryear obviously didn't really get that memo, hah
> it would avoid some of the ugliness of e.g. not knowing how far your units can move without doing some work

I don't think that's true, actually. If the models are immutable, then Strong's inf movement boost during powers is going to have to go into a `UnitModifier` and you're gonna have to check for that anyway (and do it everywhere, because inconsistency is bad).

src/AI/SpenderAI.java
ThislsAUsername: Probably more idiomatic, yes. I probably just wrote this before I wrote that.

It's definitely annoying and I don't think there's a solution under the current paradigm - the `Unit` itself is no help and the `UnitModel` is generic/sacrosanct.

I suppose another way of making this work would be to have modifiers return just the stat they're supposed to change... but then you run into the question of how you give them all the relevant information.
`UnitContext` also does allow for more complex interactions. Let's say I have a CO that lets his tanks capture, and another CO whose power grants a defense boost to units that can capture. You could handle that with the original setup, but immutable/shared models means that you have to provide the modifications at the time the other CO asks for them.

src/CommandingOfficers/Meridian.java
ThatIsAUsername: > then Strong's inf movement boost during powers is going to have to go into a UnitModifier

Nah, it could just be a remodel. That's not necessarily a cleaner/better approach in all respects, but it would certainly work.

src/Units/UnitModel.java
ThatIsAUsername: I could say the same for max HP, except Tech doesn't actually change the field, she just kinda ignores it.

src/CommandingOfficers/Meridian.java
ThislsAUsername: That's a fair point, I suppose. It wouldn't support the tags system I want without some cleverness, though.

src/Units/UnitModel.java
ThislsAUsername: I would kinda prefer she actually modify the max HP so we can iron out those semantics, but what we have today works and we don't have any other planned COs that use the mechanic... so it's not a super high priority.

src/Units/UnitModel.java
```diff
@@ -60,32 +59,29 @@ public static String standardizeID(String input)
...
+  public static final int MAXIMUM_HP = 10;
+  public static final int DEFAULT_STAT_RATIO = 100;
   public String name;
   public long role;
-  public Commander CO;
-  public int costBase;
   public double abilityPowerValue;
```
ThatIsAUsername, line 73: Isn't this a function of cost? I didn't realize we just set the unit's AP potential explicitly. One of the things that makes Tech's Steel Hail so OP is that killing the first batch pretty much provides enough power for the next one.

ThatIsAUsername:
I think I've probably whined about this PR enough; merge whenever you think it's ready.

src/Units/UnitModel.java
ThislsAUsername: It *is* a function of cost. Dual Strike has different charging rules, though, and I wanted to get the data to support it in sooner rather than later.

ThatIsAUsername: So it's a function of cost, but we've just gone ahead and precalculated it? Might change the BattleMech to be zero in that case.

ThislsAUsername: Nope, we haven't precalculated it.
Our current charge (and AW1-2's) is based on unit cost.
AWDS charge uses an independent value defined per unit; we have that value defined, but we don't currently use it.
AWDoR charge (and ours) uses HP damage and ignores cost.

PR #355, opened 2021-12-29
Labbing

ThislsAUsername:
I decided to add AWBW Lab semantics while I was messing with HQ stuff, and discovered a bunch of fun bugs. So that's neat.


src/Test/TestCommanderAve.java
```diff
@@ -120,6 +121,9 @@ boolean testSnowSpread()
       }
     }
     //System.out.println("Snow map:\n" + Ave.getSnowMapAsString());
+    Unit infantry = addUnit(testMap, Ave, UnitModel.TROOP, snowCity.xCoord, snowCity.yCoord);
+    GamePath path = Utils.findShortestPath(infantry, new XYCoord(10, 5), testMap);
+    testPassed &= validate( 2 == path.getFuelCost(Ave, infantry.model, testMap), "    Ave's units burn too much fuel moving through snow." );
```
ThatIsAUsername, line 126: This seems unrelated to the rest of the test it's in, but I guess some test is better than no test.

src/Engine/Utils.java
```diff
@@ -654,6 +655,51 @@ public static void enqueueDeathEvent(Unit unit, GameEventQueue eventQueue)
     } while( !unitsToDie.isEmpty() );
   }
 
+  /**
+   * @return Whether the owner of this property will lose when it's captured
```
ThatIsAUsername, line 659: Or destroyed?

src/Engine/Utils.java
```diff
@@ -654,6 +655,51 @@ public static void enqueueDeathEvent(Unit unit, GameEventQueue eventQueue)
...
+          break;
+        }
+      }
+      if( numLabs < 2 && numHQs == 0 )
+      {
+        playerHasLost = true;
+      }
+    }
+
+    return playerHasLost;
```
ThatIsAUsername, line 700: The function is named "will" lose, but the variable is "has" lost.

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
```diff
@@ -111,11 +111,12 @@ public GameEventQueue getEvents(MapMaster map)
...
           {
             // Check if capturing this property will cause someone's defeat.
-            if( (propertyType == TerrainType.HEADQUARTERS) && (null != captureLocation.getOwner()) )
+            if( Utils.willLoseFromLossOf(map, captureLocation) )
             {
               // Someone is losing their big, comfy chair.
               CommanderDefeatEvent defeat = new CommanderDefeatEvent(captureLocation.getOwner());
-              defeat.setPropertyBeneficiary(actor.CO);
+              if( (propertyType == TerrainType.HEADQUARTERS) )
+                defeat.setPropertyBeneficiary(actor.CO);
```
ThatIsAUsername, line 119: So capturing labs won't award properties, only HQs?

src/Terrain/MapMaster.java
```diff
@@ -92,15 +99,30 @@ public MapMaster(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
...
+          {
+            COs[co].HQLocations.add(coord);
+          }
+        }
+      // If we own some property, call that good enough for our purposes
+      else if( hasProperty )
+        COs[co].HQLocations.addAll(COs[co].ownedProperties);
+      // If we don't have property, use our first unit's starting location
+      else if( !COs[co].units.isEmpty() )
+        COs[co].HQLocations.add(new XYCoord(COs[co].units.get(0)));
```
ThatIsAUsername, line 119: Why even bother setting the HQ when there isn't an HQ?

src/Terrain/MapMaster.java
```diff
@@ -92,15 +99,30 @@ public MapMaster(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
...
+
+      // If we don't have an HQ and do have Labs, then Labs are our HQs
+      if( hasLab )
+        for( XYCoord coord : COs[co].ownedProperties )
+        {
+          if( getEnvironment(coord).terrainType == TerrainType.LAB )
+          {
+            COs[co].HQLocations.add(coord);
+          }
+        }
```
ThatIsAUsername, line 113: So Labs are HQs if there are no HQs? Can one map have both (and are labs useless in that case)?

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThislsAUsername: That is AWBW lab semantics, so yes.
Don't ask me why they do it that way.

src/Terrain/MapMaster.java
ThislsAUsername: The idea in my head was "to provide AIs with something to fall back on re: objectives e.g. if everything is covered by fog".
I don't have super strong opinions on it, but it seems useful as a "avoid sitting around picking your nose" thing.

src/Terrain/MapMaster.java
ThislsAUsername: Yep, more or less.

One map can definitely have both:
On AWBW, that results in labs only being useful for lab units. If you define a team that has only labs and no HQs, the game doesn't count them as a viable player.
Here, you can mix teams with only labs or only HQs (which is how the cart works, FWIW).

Lab units/unit bans in general are a future kind of thing, but I do think it can create some neat scenarios.

src/Terrain/MapMaster.java
ThatIsAUsername: Hm. Seems like an odd edge case. I guess we can just let it ride and see what happens.

src/Terrain/MapMaster.java
ThatIsAUsername: I didn't know the carts even allowed labs with no HQ (save for a couple campaign missions where it could just be a reskin).
I guess that means a team can also have both one or more HQs plus one or more labs, but then the labs don't matter (except for special units). Cool.

ThatIsAUsername:
If you want to address the verbiage in the comments above go ahead, otherwise I think this (and the other two PRs) are probably good to go.

ThislsAUsername:
...I fixed it, but forgot to push up the commit. I am good at this game.

ThatIsAUsername:
I still don't see that commit.

ThislsAUsername:
I just put it as the first commit on the new PR, rather than bother to rewrite history.

PR #356, opened 2021-12-30
Army

ThislsAUsername:
This got a little out of hand again, sorry.
It turns out that rewriting a fundamental part of your data structure has ripple effects.


ThislsAUsername:
The current state has team-merge tags only, but they're pretty much fully functional.
There are still a bunch of TODOs I need to address, though.

src/Terrain/MapPerspective.java
```diff
@@ -167,7 +167,7 @@ public boolean isLocationFogged(int x, int y)
   }
   public boolean isFogOn()
   {
-    return (null == viewer) || viewer.gameRules.isFogEnabled;
+    return (null == viewer) || viewer.cos[0].gameRules.isFogEnabled;
```
ThatIsAUsername, line 170: Yeah, it definitely feels weird to be consulting a random CO just to see the game rules. Seems like that should live in the `GameInstance`, or maybe in the `MapMaster/Perspective`? Maybe `Army` even, since that's effectively the CO replacement.

src/Test/TestCombat.java
```diff
@@ -29,10 +30,10 @@ private void setupTest()
     GameScenario scn = new GameScenario();
     testCo1 = new Strong(scn.rules);
     testCo2 = new Patch(scn.rules);
-    Commander[] cos = { testCo1, testCo2 };
+    Army[] cos = { new Army(testCo1), new Army(testCo2) };
 
     testMap = new MapMaster(cos, MapLibrary.getByName("Firing Range"));
-    testGame = new GameInstance(testMap);
+    testGame = new GameInstance(cos, testMap);
```
ThatIsAUsername, line 36: Why not keep around the basic `GameInstance(MapMaster)` constructor and just auto-create the armies for simple/standard situations with no tags? Would avoid boilerplate in most places we create games.

src/UI/GameBuilder.java
```diff
@@ -28,27 +34,70 @@
...
+        }
+        else // else, merge with the match we found
+        {
+          // credit: https://www.baeldung.com/java-concatenate-arrays
+          // Clone the final army's commander list into a larger space, then copy the new Commanders into it
+          Commander[] newCoList = Arrays.copyOf(match.cos, match.cos.length + army.cos.length);
+          System.arraycopy(army.cos, 0, newCoList, match.cos.length, army.cos.length);
+          match.cos = newCoList;
+          match.HQLocations.addAll(army.HQLocations);
+        }
```
ThatIsAUsername, line 85: Why create each army with one CO and then merge instead of just creating each army with the full list of COs up front?

src/UI/PlayerSetupInfo.java
```diff
@@ -80,10 +80,6 @@ public Commander makeCommander(GameScenario.GameRules rules)
     co.faction = new Faction(availableFactions[currentFaction].name, availableFactions[currentFaction].basis);
     co.faction.flip = flipUnits;
 
-    co.team = currentTeam;
-
-    co.setAIController(availableAis[currentAi].create(co));
```
ThatIsAUsername, line 85: Why move these assignments up to `GameBuilder`? I guess now you need the actual `Army` to pass to the AI, but the team doesn't seem to need that treatment.

src/AI/Muriel.java
```diff
@@ -960,15 +960,18 @@ private void queueUnitProductionActions(GameMap gameMap)
     // Build infantry from any remaining facilities.
     UnitModel infModel = myCo.cos[0].getUnitModel(UnitModel.TROOP);
     MapLocation loc = CPI.getLocationToBuild(infModel);
-    int infCost = loc.getOwner().getCost(infModel);
-    while( (budget >= infCost) && (CPI.availableUnitModels.contains(infModel)) )
+    int infCost = Integer.MAX_VALUE;
+    if( null != loc && null != loc.getOwner() )
+      infCost = loc.getOwner().getCost(infModel);
```
ThatIsAUsername, line 966: Seems like the unit models should be owned by game scenario/rules/instance now, since this is some convoluted logic for something that should be very simple.

src/Test/TestAIBehavior.java
```diff
@@ -165,15 +167,15 @@ private boolean testClearAttackRoute(AIMaker ai)
     Unit iRight = addUnit(testMap, testCo1, "Infantry", facPos.right());
 
     // The infs all want to cap the fac, but can't because it is occupied. Gotta let the AA through.
-    testCo1.initTurn(testMap);
-    testCo1.money = 0; // No production needed for this test.
+    testCo1.army.initTurn(testMap);
+    testCo1.army.money = 0; // No production needed for this test.
```
ThatIsAUsername, line 171: Why not store the army as part of the test setup instead of calling `testCo1.army` every time?

src/CommandingOfficers/Ave.java
```diff
@@ -726,8 +728,8 @@ public GameEventQueue getEvents(MapMaster gameMap)
 
     public Set<XYCoord> getTilesInRange(GameMap gameMap)
     {
-      Set<XYCoord> tilesInRange = Utils.findLocationsNearUnits(gameMap, Ave, OBLIDO_RANGE);
-      tilesInRange.addAll(Utils.findLocationsNearProperties(gameMap, Ave, OBLIDO_RANGE));
+      Set<XYCoord> tilesInRange = Utils.findLocationsNearUnits(gameMap, Ave.army.getUnits(), OBLIDO_RANGE);
+      tilesInRange.addAll(Utils.findLocationsNearPoints(gameMap, Ave.army.getOwnedProperties(), OBLIDO_RANGE));
```
ThatIsAUsername, line 732: This... is a change in behavior.

src/CommandingOfficers/Commander.java
```diff
@@ -567,12 +357,16 @@ public int getIncomePerTurn()
...
   @Override
   public List<UnitModifier> getModifiers()
   {
-    // TODO Add call to pull modifiers from Army when that becomes a thing?
-    return new ArrayList<UnitModifier>(unitMods);
+    // Intended order of operations: model, D2D, environment, abilities, unit-specific
+    ArrayList<UnitModifier> output = new ArrayList<UnitModifier>();
+    output.add(this);
+    output.addAll(unitMods);
+    output.addAll(this.army.getModifiers());
```
ThatIsAUsername, line 368: I tried starting this game with Team Merge on and got a NPE pointing to this line.
Screenshot from 2022-01-21 16-39-14
Text:
```java
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at CommandingOfficers.Commander.getModifiers(Commander.java:368)
	at Units.UnitContext.<init>(UnitContext.java:103)
	at CommandingOfficers.Commander.getCostContext(Commander.java:204)
	at CommandingOfficers.Commander.getCost(Commander.java:212)
	at CommandingOfficers.Meridian.<init>(Meridian.java:77)
	at CommandingOfficers.Meridian$instantiator.create(Meridian.java:58)
	at UI.PlayerSetupInfo.makeCommander(PlayerSetupInfo.java:77)
	at UI.GameBuilder.createGame(GameBuilder.java:51)
	at UI.PlayerSetupController.handlePlayerSetupInput(PlayerSetupController.java:126)
	at UI.PlayerSetupController.handleInput(PlayerSetupController.java:86)
	at UI.GameOptionSetupController.handleInput(GameOptionSetupController.java:52)
	at Engine.Driver.keyPressed(Driver.java:115)
	at java.desktop/java.awt.Component.processKeyEvent(Component.java:6590)
... off into the bowels of awt
```

ThatIsAUsername: Actually, it worked with all humans as well, which maybe means one of the COs is broken?

src/Test/TestCapture.java
```diff
@@ -190,17 +191,17 @@ private boolean testCaptureHQ()
     captureAction = new CaptureLifecycle.CaptureAction(testMap, mech, Utils.findShortestPath(mech, 13, 1, testMap));
     GameEventQueue events = captureAction.getEvents(testMap);
 
-    // Make sure a CommanderDefeatEvent was generated as a result (the actual event test is in TestGameEvent.java).
+    // Make sure a ArmyDefeatEvent was generated as a result (the actual event test is in TestGameEvent.java).
```
ThatIsAUsername, line 194: *an* - do u ven engish bro

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -139,26 +149,24 @@ private BufferedImage renderPage(ArrayList<InfoPage> pages, int drawingWidth)
...
-          {
-            BufferedImage statusHeader =
-                SpriteLibrary.createTransparentSprite(drawingWidth, CommanderOverlayArtist.OVERLAY_HEIGHT + 5);
-            Graphics headerG = statusHeader.getGraphics();
-            CommanderOverlayArtist.drawCommanderOverlay(headerG, thisCO, 0, true);
+          int ySpaceNeeded = 0;
+          final int overlayHeight = SpriteLibrary.getCoOverlay(armyList[0].cos[0], true).getHeight();
+          for( Army thisArmy : armyList )
+            // Draw a little generously
+            ySpaceNeeded += (thisArmy.cos.length + 1) * overlayHeight;
```
ThatIsAUsername, line 158: Generously? Can we not draw with exactness?

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -7,108 +7,122 @@
...
+      }
+      else
+      { // Draw the overlay on the right side.
+        int xPos = mapViewWidth - overlayImage.getWidth();
+        int valueXPos = mapViewWidth - ICON_VALUE_SPACING - spriteA.getWidth()*energyLength - xTextOffset;
+        g.drawImage(overlayImage, xPos, yCurrentOffset, overlayImage.getWidth(), overlayImage.getHeight(), null);
+        drawIconAndValue(g, SpriteLibrary.MapIcons.ENERGY, energy, energyLength, valueXPos, yTextOffset + yCurrentOffset);
+        int pbXPos = mapViewWidth - xAbilityOffset - powerBarImage.getWidth();
+        g.drawImage( powerBarImage, pbXPos, yPowerBarOffset, powerBarImage.getWidth(), powerBarImage.getHeight(), null );
+      }
```
ThatIsAUsername, line 111: Gotta say, I think I like the power bar/funds better where they used to be. It looks more natural if the bar extends past the end of the overlay, and looks more natural than the brown funds box that's replaced it.

The only downsides of the old scheme are 1) using slightly more vertical space to put a power bar between each CO overlay and 2) Repeating the funds count across multiple COs if they are sharing funds. I don't think either of those is a huge deal though. Plus, isn't there a planned tag mode without shared funds?

src/UI/COStateInfo.java
```diff
@@ -114,10 +122,10 @@ public String getFullStatusLabels()
     sb.append("Unit count:")   .append("\n");
     sb.append("Vehicle count:").append("\n");
     sb.append("Unit funds:")   .append("\n");
-    if( showAbilityInfo )
+    for( Commander co : currentEnergy.keySet() )
     {
-      sb.append("Ability Power:").append("\n");
-      sb.append("Next Ability:") .append("\n");
+      sb.append(co.coInfo.name + " Energy:").append("\n");
+      sb.append(" - Next Energy:") .append("\n");
```
ThatIsAUsername, line 128: Why the change from "Ability" to "Energy"? I think the former sounds more natural.

src/UI/GameStatsController.java
```diff
@@ -73,9 +74,20 @@ public boolean handleInput( InputAction action )
...
+
+  @Override
+  public ArrayList<CommanderInfo> getSelectedCOInfo()
+  {
+    ArrayList<CommanderInfo> output = new ArrayList<>();
+    for( Commander co : getSelectedArmy().cos )
+    {
+      output.add(co.coInfo);
+    }
+    return output;
```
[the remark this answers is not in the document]
ThatIsAUsername: Same for `CO_InfoController.`

src/UI/Art/SpriteArtist/SpriteLibrary.java
```diff
@@ -605,7 +605,8 @@ public static BufferedImage getCoOverlay(Commander co, boolean leftSide)
     if( !coOverlays.containsKey(co) )
     {
       // If we don't already have this overlay, go load and store it.
-      Sprite overlay = new Sprite(SpriteLibrary.loadSpriteSheetFile("res/ui/co_overlay.png"), CommanderOverlayArtist.OVERLAY_WIDTH, CommanderOverlayArtist.OVERLAY_HEIGHT);
+      final BufferedImage sprite = SpriteLibrary.loadSpriteSheetFile("res/ui/co_overlay.png");
+      Sprite overlay = new Sprite(sprite, sprite.getWidth()/2, sprite.getHeight());
```
ThatIsAUsername, line 609: Why width/2?

src/AI/Muriel.java
```diff
@@ -929,13 +933,13 @@ private void queueUnitProductionActions(GameMap gameMap)
         // Figure out if we can afford the desired unit type.
         int maxBuildable = CPI.getNumFacilitiesFor(idealCounter);
         log(String.format("    Facilities available: %s", maxBuildable));
-        int cost = myCo.getCost(idealCounter);
+        MapLocation loc = CPI.getLocationToBuild(idealCounter);
+        int cost = loc.getOwner().getCost(idealCounter);
```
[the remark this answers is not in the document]
ThislsAUsername: No, since `Army` would need to know what location you're looking at.
CPI could probably handle it, though.

src/AI/Muriel.java
```diff
@@ -955,20 +959,26 @@ private void queueUnitProductionActions(GameMap gameMap)
     } // ~while( still choosing units to build )
 
     // Build infantry from any remaining facilities.
-    UnitModel infModel = myCo.getUnitModel(UnitModel.TROOP);
-    final int infCost = myCo.getCost(infModel);
-    while( (budget >= infCost) && (CPI.availableUnitModels.contains(infModel)) )
+    UnitModel infModel = myCo.cos[0].getUnitModel(UnitModel.TROOP);
```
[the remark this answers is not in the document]
ThislsAUsername: I do like `Army.getShoppingList(XYCoord)`, since that gives you the knowledge of what CO to look at.
The awkward bit here is that (per past-me's "modularity" concerns) `GameInstance` isn't very visible (and injecting that dependency could be a little awkward).
It's probably a good idea to feed AIs the rules on construction, and that does seem like a reasonable "master location" for the list of models.

src/AI/WallyAI.java
```diff
@@ -99,13 +100,14 @@ public double getEffectiveMove(UnitModel model)
     //             term for how fast you are   term for map coverage
     double ratio = (validTiles / totalCosts) * (validTiles / totalTiles); // 1.0 is the max expected value
     
-    double effMove = uc.calculateMovePower() * ratio;
+//    double effMove = model.calculateMovePower() * ratio;
+    double effMove = model.baseMovePower * ratio;
```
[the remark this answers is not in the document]
ThislsAUsername: That's one of a few compromises in Wally that I made 'cause AI is hard and the code isn't structured to work nicely with the new object graph.
Specifically, the `UnitModel` no longer is tethered directly to a specific CO, so you actually can't calculate the "true move" with just a model.

src/AI/Muriel.java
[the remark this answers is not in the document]
ThatIsAUsername: Shouldn't it be `getBuyCost` anyway? In which case, that function already requires a location, and rehoming it should work fine.

src/CommandingOfficers/Commander.java
```diff
@@ -58,28 +47,14 @@
   public Faction faction;
```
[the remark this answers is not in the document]
ThislsAUsername: I don't think we need to *move* `GameRules`; having it handy everywhere isn't a problem IMO.
`Faction` could move up, but I don't see the harm in leaving it here (if the users want to confuse themselves, by all means)... and that sounds like a lot of annoying refactoring that's not really necessary.
Fair enough re: `MapPerspective`.

src/UI/GameBuilder.java
ThislsAUsername: Because the list of "people who own units" needs to be in order and the right size (and you can't just assume a jagged iteration of armies->COs will be right, since who's grouped is user-defined).
Then `HQLocations` moved to `Army` and I said "eh".
It's probably possible to have `MapMaster`'s constructor handle only COs and offer some API for pulling the HQs out, but I figure that's probably more work than what we have now.

src/UI/PlayerSetupInfo.java
ThislsAUsername: It seemed like a good idea to co-locate all the relevant initialization.

src/Test/TestAIBehavior.java
ThislsAUsername: Apathy.

src/CommandingOfficers/Commander.java
ThislsAUsername: I ran into this while testing on one of the other later branches.
This is the relevant diff:
```@@ -363,7 +365,10 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
     ArrayList<UnitModifier> output = new ArrayList<UnitModifier>();
     output.add(this);
     output.addAll(unitMods);
-    output.addAll(this.army.getModifiers());
+    if( null != army )
+      output.addAll(army.getModifiers());
+    else
+      System.out.println("WARNING!: Getting Commander modifiers when CO.army == null!");
     return output;
   }
```
I could throw that in here, but I don't want to develop merge conflicts.
Maybe when we're happy with your current comments here and the upstream branches, we just merge everything down onto this for any final adjustments?

src/Test/TestCapture.java
ThislsAUsername: nop

src/UI/Art/SpriteArtist/InfoView.java
ThislsAUsername: As I recall, I was trying to account for the height of the brief summary, which varies by the input font.
Having a little extra space is unlikely to hurt anything.

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -92,7 +102,7 @@ public void render(Graphics g)
 
     final int drawingWidth  = paneHSize - PANE_OUTER_BUFFER*2;
     final int drawingHeight = paneVSize - PANE_OUTER_BUFFER*2;
-    if( pages != prevPages )
+    if( pages != prevPages || pages.get(0).pageType == PageType.CO_HEADERS )
```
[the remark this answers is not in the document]
ThislsAUsername: They don't animate, otherwise.
It makes me sad.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: > It looks more natural if the bar extends past the end of the overlay, and looks more natural than the brown funds box that's replaced it.

Very solid points.
I am considering moving to a "star"-based ability energy storage scheme, since I think I can support all the use-cases we care about with it. That's not really here or there, though.
> 1) using slightly more vertical space to put a power bar between each CO overlay

There's also the question of whether we should have transparent gaps in there or have the bar pips overlap with the next CO header. ...Thinking about it, I guess it wouldn't be a big deal to just overlap, since they're drawn top-to-bottom (so the top of each header would be an unbroken line).
> Repeating the funds count across multiple COs if they are sharing funds.

This is the main reason I moved things around. I don't like duplicate data == wasted space.
I suppose we could pull the power bars *and* gold out, and put the CO name or something else in.
> Plus, isn't there a planned tag mode without shared funds?

No. If there is, then I'm gonna have to refactor 'cause `Army` owns the purse at present.

src/UI/Art/SpriteArtist/SpriteLibrary.java
ThislsAUsername: The sprite has the left and right overlays next to each other, so this tells the `Sprite` to treat them as sub-frames.

src/Test/TestCombat.java
ThislsAUsername: The `MapMaster` requires armies as input anyway.
It's certainly *possible* to do what you're suggesting, but we'd need some sort of interim structure to hold the stuff the map wants to dump into the armies... and that's probably more work.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Yeah, that's a good plan.

src/UI/Art/SpriteArtist/InfoView.java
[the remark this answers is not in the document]
ThatIsAUsername: But... this wasn't necessary before, was it?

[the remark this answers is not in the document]
ThislsAUsername: They didn't animate before, either.
The point of caching the page is so you don't have to redraw it, but we *want* to redraw this, so...

src/AI/ModelForCO.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+
+import CommandingOfficers.Commander;
+import Engine.Army;
+import Units.UnitModel;
+
+public class ModelForCO implements Serializable
```
ThatIsAUsername, line 12: I feel like I'm always having to read code in depth to figure out just what you are trying to do. Comments are great, man. Too bad this class doesn't have any.

src/AI/CommanderProductionInfo.java
```diff
@@ -24,47 +25,50 @@
...
-    availableProperties = new HashSet<MapLocation>();
-    propertyCounts = new HashMap<Terrain.TerrainType, Integer>();
-    modelToTerrainMap = new HashMap<UnitModel, Set<TerrainType>>();
+    myArmy = army;
+    availableUnitModels = new HashSet<>();
+    availableProperties = new HashSet<>();
+    propertyCounts = new HashMap<>();
+    for( Commander co : army.cos )
+      propertyCounts.put(co, new HashMap<>());
+    modelToTerrainMap = new HashMap<>();
```
ThatIsAUsername, line 46: Why remove all the type parameters?

src/AI/Muriel.java
```diff
@@ -78,13 +79,13 @@ public ActionWithValue put(Unit u, GameAction ga, Double value) {
...
 
-  private Commander myCo = null;
+  private Army myArmy = null;
 
   private StringBuffer logger = new StringBuffer();
   private boolean shouldLog = true;
   private int turnNum = 0;
 
-  private ArrayList<Commander> enemyCos = null;
+//  private ArrayList<Commander> enemyCos = null;
```
ThatIsAUsername, line 88: No need for a dead comment if this isn't used anymore.

src/AI/CommanderProductionInfo.java
ThatIsAUsername: I get that Java can infer them based on the declarations above, but is there any reason to prefer not having them there? Or did you just decide you wanted fewer characters to look at?

src/AI/CommanderProductionInfo.java
```diff
@@ -15,56 +16,59 @@
...
-  public Set<UnitModel> availableUnitModels;
+  public Army myArmy;
+  public Set<ModelForCO> availableUnitModels;
   public Set<MapLocation> availableProperties;
-  public Map<Terrain.TerrainType, Integer> propertyCounts;
+  public Map<Commander, Map<Terrain.TerrainType, Integer>> propertyCounts;
   public Map<UnitModel, Set<TerrainType>> modelToTerrainMap;
 
   /**
    * Build a model of the production capabilities for a given Commander.
```
ThatIsAUsername, line 34: Commander -> Army, I suppose.

src/AI/ModelForCO.java
ThislsAUsername: I thought it was clear enough from the name and constructors.
There's not much *to* it other than what amounts to boilerplate and constructors.

src/AI/CommanderProductionInfo.java
ThislsAUsername: I like fewer characters, and it means fewer places to update if I ever want to change the types.

src/AI/CommanderProductionInfo.java
ThislsAUsername: No, actually. Commanders can have different prices for the same unit.

src/AI/Muriel.java
```diff
@@ -794,34 +796,39 @@ private void queueUnitProductionActions(GameMap gameMap)
...
+      boolean useDamageRatio = (myArmy.money > (incomePerTurn*2)); // Rich people can afford to think differently.
       if(useDamageRatio) log("  High funds - sorting units by damage ratio instead of cost effectiveness.");
 
       // If we are low on grunts, make sure we save money to build more.
-      UnitModel infModel = myCo.getUnitModel(UnitModel.TROOP);
+      UnitModel infModel = myArmy.cos[0].getUnitModel(UnitModel.TROOP);
+      int infCount = 0;
+      for( Unit u : myArmy.getUnits() )
+        if( infModel == u.model )
+          ++infCount;
```
ThatIsAUsername, line 812: Why not use `myUnitCounts` to count the number of infantry?

src/AI/Muriel.java
```diff
@@ -841,41 +848,41 @@ private void queueUnitProductionActions(GameMap gameMap)
...
           for( Unit enemyUnit : unitLists.get(enemyCo) )
           {
-            UnitMatchupAndMetaInfo umami = getUnitMatchupInfo(counter, enemyUnit.model);
+            UnitMatchupAndMetaInfo umami = getUnitMatchupInfo(counter, new ModelForCO(enemyUnit.CO, enemyUnit.model));
             if( useDamageRatio )
             {
               if( umami.damageRatio >= 1.0 ) score++; // Plus one if it's worth building.
               if( umami.damageRatio >= 1.5 ) score++; // An extra bump in score if they are very good vs this type.
-              if( umami.damageRatio < 0.8 ) score--; // Discount if my counter is countered.
+              if( umami.damageRatio < 0.8 ) score--; // Discount if myArmy counter is countered.
```
ThatIsAUsername, line 861: Did you do a find/replace for "my" -> "myArmy"? There are a few of these.

src/AI/ModelForCO.java
ThatIsAUsername: But seriously, didn't we just make it so that all Commanders share the same underlying models? Why do we now need a class to keep track of models on a per-CO basis?

ThatIsAUsername: The fact that there's not much to it doesn't really help, because everything needed to understand it is contextual/external. The places I see it used it just seems duplicative of the plain `UnitModel` (and there are often literally duplicate functions, to handle both it and this).
Admittedly, I haven't invested a ton of time to try and figure it out, but I also shouldn't have to reverse-engineer your logic to understand your intent. It just feels like everything you touch gets more complex for no apparent reason.

src/Engine/Army.java
```diff
@@ -0,0 +1,416 @@
...
+      }
+    }
+    return false;
+  }
+  public int getBuyCost(UnitModel um, XYCoord coordinates)
+  {
+    MapLocation loc = myView.getLocation(coordinates);
+    return loc.getOwner().getBuyCost(um, coordinates);
+  }
+
```
ThatIsAUsername, line 177: What will this do if called on a location owned by nobody (or a different army)?

src/AI/InfantrySpamAI.java
```diff
@@ -222,9 +223,11 @@ public GameAction getNextAction(GameMap gameMap)
           if( loc.getEnvironment().terrainType == TerrainType.FACTORY && buyer.army == myArmy && loc.getResident() == null )
           {
             ArrayList<UnitModel> units = buyer.getShoppingList(loc);
-            if( !units.isEmpty() && buyer.getBuyCost(units.get(0), loc.getCoordinates()) <= myArmy.money )
+            final int buyCost = buyer.getBuyCost(units.get(0), loc.getCoordinates());
+            if( !units.isEmpty() && buyCost <= budget )
             {
               GameAction action = new GameAction.UnitProductionAction(buyer, units.get(0), loc.getCoordinates());
+              budget -= buyCost;
```
ThatIsAUsername, line 230: Why is this needed? Is it just that ISAI could potentially go over budget if merged with another CO who used most/all of the budget already, and this would result in an invalid/turn-ending action?

src/UI/CO_InfoController.java
```diff
@@ -106,7 +106,7 @@ public GameInstance getGame()
   }
 
   @Override
-  public ArrayList<CommanderInfo> getSelectedCOInfo()
+  public ArrayList<CommanderInfo> getSelectedCOInfoList()
```
ThatIsAUsername, line 109: Why did this end up being actually necessary? And is the whole list selected (as the function name implies)?

src/AI/CommanderProductionInfo.java
ThatIsAUsername: OK, fair, but the function is taking in an Army, not a Commander, and the prices for the latter can be taken abstractly under the umbrella of the former.
Modeling the production capabilities of an Army inherently requires modeling the underlying Commanders, which (I believe) this does now.

src/AI/ModelForCO.java
ThislsAUsername: > But seriously, didn't we just make it so that all Commanders share the same underlying models? Why do we now need a class to keep track of models on a per-CO basis?

The AIs care about the *modified* characteristics of the unit types in question, not just the *base* values.
I suppose a `UnitContext` would work as well, but it wouldn't be as obvious that we want a "type" of unit.
Wally, for example, wants to track how fast units are re: the current terrain. If he's using Sturm, Recons should look a lot more attractive.
Muriel wants to know how efficiently each COs' units can fight each other - stat modifications will definitely have implications there.

`UnitModel` used to be 100% bound up with the CO (even when it had no direct reference to the CO, its state would be modified re: any D2D/powers), and our AIs are designed around that fact.
We *could* change that but the easier solution was to make a class that represents the abstraction layer the AI wants.
> everything needed to understand it is contextual/external

A fair point, I think.
We're back to that weird place where I don't know where to put the comments to explain it, though.
> The places I see it used it just seems duplicative of the plain `UnitModel` (and there are often literally duplicate functions, to handle both it and this).

I think this is address sufficiently above?
> I also shouldn't have to reverse-engineer your logic to understand your intent

Some minimal amount of that is necessary, lest we start needing to comment... IDK, basic for-loop headers.
You're probably right that some stuff is more obtuse than it should be, though.
> It just feels like everything you touch gets more complex for no apparent reason.

It's getting more complex because I would like to handle more complex scenarios gracefully.
Perhaps I'm optimizing for the edge cases a bit too much, because it's the most "interesting" to me.

src/AI/CommanderProductionInfo.java
ThislsAUsername: Oh, you're talking about the comment that didn't get updated.
Fair enough.

src/AI/Muriel.java
ThislsAUsername: Seems likely.
Wasn't intentional.

src/Engine/Army.java
ThislsAUsername: Barf.
I'm not sure what it should do other than that, TBH.

src/UI/CO_InfoController.java
ThislsAUsername: The `InfoController` can be used to check out CO info during game setup, which means you can't always fetch CO info by proxy of the current CO (since before the game, there's no CO).

I imagine the list-ness of it could start mattering if/when we start building Tag UI, but I don't think it matters at the moment.

src/AI/InfantrySpamAI.java
ThislsAUsername: I was using "zombie apocalypse mode" (Cinder+ISAI) to test things, and noticed that ISAI would never use `Sear` because of invalid buy actions.

src/AI/Muriel.java
ThislsAUsername: There's a relevant semantic difference:
Just counting the inf gives you the number of units.
`myUnitCounts` reckons injured units as fractional units, which may or may not be as relevant for infantry.

src/AI/AIUtils.java
```diff
@@ -116,8 +117,8 @@ public static Map<UnitActionFactory, ArrayList<GameAction> > getAvailableUnitAct
   }
 
   /**
-   * Find locations of every property that is not owned by myCo or an ally.
-   * @param myCo Properties of this Commander and allies will be ignored.
+   * Find locations of every property that is not owned by myArmy or an ally.
+   * @param myArmy Properties of this Commander and allies will be ignored.
```
ThatIsAUsername, line 121: Comment says it takes an army now, but it still takes a Commander. Given that everywhere it's invoked it's given `army.cos[0]`, I think it makes sense to make the parameter list match the docstring.

src/AI/Muriel.java
```diff
@@ -955,20 +960,27 @@ private void queueUnitProductionActions(GameMap gameMap)
     } // ~while( still choosing units to build )
 
     // Build infantry from any remaining facilities.
-    UnitModel infModel = myCo.getUnitModel(UnitModel.TROOP);
-    final int infCost = myCo.getCost(infModel);
-    while( (budget >= infCost) && (CPI.availableUnitModels.contains(infModel)) )
+    UnitModel infModel = myArmy.cos[0].getUnitModel(UnitModel.TROOP);
```
ThatIsAUsername, line 963: Isn't this incorrect when `myArmy` has multiple COs?

Also on line 808.

src/AI/Muriel.java
ThatIsAUsername: This changes Muriel's behavior though, which is def out of scope for this PR. Why did this need to change?

src/Terrain/MapMaster.java
```diff
@@ -65,14 +64,14 @@ public MapMaster(CommandingOfficers.Commander[] COs, MapInfo mapInfo)
...
-            COs[co].HQLocations.add(new XYCoord(x, y));
+            propertyOwners[co].HQLocations.add(new XYCoord(x, y));
             hasHQ = true;
           }
           else if( terrainType == TerrainType.LAB )
           {
             hasLab = true;
           }
-          location.setOwner(COs[co]);
+          location.setOwner(propertyOwners[co].cos[0]);
```
ThatIsAUsername, line 74: No support for assigning properties to multiple COs in the same army, eh?

src/AI/ModelForCO.java
ThatIsAUsername: > The AIs care about the modified characteristics of the unit types in question, not just the base values.

Yeah, sure. But this doesn't allow that. It literally just maps COs to their UnitModels. The models themselves aren't differentiated/modified per the CO's abilities (since we no longer have a way to do that outside of UnitContext).

It's also only used in Muriel, and only to separate enemy unit types into lists by CO. Wally doesn't even use it (It only invokes the ModelForCO type because CPI now returns it, but then it just pulls out the UnitModel itself).

Am I just grossly missing something, or is this class just a more complicated way of doing nothing special?

> EDIT: I've thought a bit more about it, and I think I've hit upon our methodology difference:
My instinct, when adding functionality, is to make a general infrastructure to support permutations of that form.
Yours is to change code outside of your new addition as little as possible

I think both values are important: avoid changing what doesn't need to be changed (because needless code churn is bad), and also make changes of the same type easier in the future (because needless code churn is bad). ModelForCO... doesn't do anything but add complexity, as far as I can tell.

ThislsAUsername: > Yeah, sure. But this doesn't allow that. It literally just maps COs to their UnitModels.

Having CO + UnitModel lets you calculate whatever it is you want. I don't see the problem?
> The models themselves aren't differentiated/modified per the CO's abilities (since we no longer have a way to do that outside of UnitContext).
> 
> It's also only used in Muriel, and only to separate enemy unit types into lists by CO.

I guess another way to do it would be to use `UnitContext`s, but then you can't as easily use them as keys for stuff like the `UnitMatchupAndMetaInfo`.
You can just use `UnitModel` itself, but then you can't store different matchups per CO - and that's something you specifically called out as being broken (I had the code commented out).
I could revert that change so Muriel ends up entirely CO-agnostic as far as unit builds, but that's sort of a bad idea when people like Max and Grit exist.
> Wally doesn't even use it (It only invokes the ModelForCO type because CPI now returns it, but then it just pulls out the UnitModel itself).

Correct. I didn't feel like putting a bunch of effort in to fix his current state when he needs more involved surgery anyway.
Wally is broken by this change, but not badly (given our current CO loadout), and I do intend to fix that later along with more involved changes.
It just seemed kinda rude to break the AI you wrote and not fix it.
> Am I just grossly missing something, or is this class just a more complicated way of doing nothing special?

If you know of a datastructure that can take a CO and a `UnitModel` as a duet unique key and you want to make the AIs start using that, I guess I'm down to try doing it.
Otherwise, all I can think of is a two-tier `HashMap`... and I feel like that will have bigger impacts on the clarity of the code than this does.
> I think both values are important: avoid changing what doesn't need to be changed (because needless code churn is bad), and also make changes of the same type easier in the future (because needless code churn is bad).

Quite fair.
> ModelForCO... doesn't do anything but add complexity, as far as I can tell.

It adds a new object type to be a key into Maps that want to track some value for a unit type for a specific CO.
`UnitContext` does not work since there's no sensible way to set up `.Equals()` such that it would do what you expect in both this and its other use cases (I guess you could make it ignore `UnitState` values if there's no individual unit reference, but that seems like a great way to get subtle bugs...).
`UnitModel` does not work since you entirely lose the CO information if you just use it.

src/AI/Muriel.java
ThislsAUsername: Oh. My brain decided at some point that this is what Muriel did before, and therefore I was doing this to maintain the old function.
Sorry for the confusion.

src/AI/Muriel.java
ThislsAUsername: No. This is just "`Army` doesn't have a copy of the model list directly on hand for querying".
The code on the following lines has to cope with the CO-agnostic nature of `UnitModel`; this instance by just asking for the price directly, 808 via `getAverageCostFor()`.

src/Terrain/MapMaster.java
ThislsAUsername: That's only relevant under Team Merge, which has its own support for that very concept.
The other tag variants always have Lead CO == Owner of All Properties.

src/AI/Muriel.java
ThislsAUsername: On the note of "making the unit type list more available", would this be a case where it makes sense to extend a collection class?
A list of units *is-a* list of things, but it also wants to provide non-standard query APIs that only make sense in this context.

An alternative would be to move the implementations of the filter functions to static methods and have the `Commander` copies just call those.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -7,108 +7,122 @@
...
   {
     BufferedImage output = SpriteLibrary.createTransparentSprite(drawingWidth, drawingHeight);
     Graphics g = output.getGraphics();
     int verticalOffset = 0;
 
-    for( Commander CO : coList )
+    for( Army army : armyList )
     {
+      BufferedImage headerWorkspace = SpriteLibrary.createTransparentSprite(drawingWidth, drawingHeight);
+      int newOffset = CommanderOverlayArtist.drawCommanderOverlay(headerWorkspace.getGraphics(), army, verticalOffset, true);
```
ThatIsAUsername, line 33: This is creating a new sprite the size of the entire canvas for each commander overlay. Then you are drawing to this canvas in the correct absolute offset position, and drawing this image to the original canvas at 0, 0.

Why do we need to double-buffer this? It seems wildly inefficient.

src/UI/Art/SpriteArtist/InfoView.java
ThatIsAUsername: Seems like it wouldn't be too hard to account for that space. As is, it looks like you are adding an extra overlay's worth of height for each team in play. Maybe a bit excessive.

src/AI/ModelForCO.java
ThatIsAUsername: > Having CO + UnitModel lets you calculate whatever it is you want. I don't see the problem?

This is the kind of raison d'être that should probably go in the class javadoc. That said, you are now creating ephemeral units every time you need CO-specific unit info (e.g. to calculate matchups). Seems like it would be cheaper to just create one unit per type per CO up front instead of one unit per type per CO per other type per other CO ephemerally. But that feels almost like going back to CO-specific unit models, except they'd be mutable.

> It just seemed kinda rude to break the AI you wrote and not fix it.

Hah, well I do appreciate that.

src/AI/Muriel.java
ThatIsAUsername: > No. This is just "Army doesn't have a copy of the model list directly on hand for querying".

Ah. Well if `UnitModel`s aren't CO-specific anymore, it might make sense to have some way of obtaining them without going through a CO. Both for convenience and to clarify intent. That can probably be future cleanup though.

Re: extending a collection class - maybe? Not sure what you have in mind here in terms of API.

ThatIsAUsername:
I think I've made it through everything again, and resolved comments that needed it. Any comments still unresolved probably haven't been addressed.

src/AI/FightClub.java
```diff
@@ -179,18 +182,19 @@ public void write(int b) throws IOException{}
         defaultOut.println("  Starting game " + gameIndex + " on map " + mi.mapName + " with combatants:");
         for( int i = 0; i < numCos; ++i )
           defaultOut.println("    Team " + combatants.get(i).team + ": "
-                               + combatants.get(i).getControllerName() + " controlling " + combatants.get(i).coInfo.name);
+                               + combatants.get(i).getControllerName() + " controlling " + combatants.get(i).cos[0].coInfo.name);
```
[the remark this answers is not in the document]
ThislsAUsername: For `FightClub` specifically, or in general?
It's probably possible for `FightClub`, but I feel like trying to do that in general could result in some real jank (and code written against this new API could silently have bugs for multi-CO environments)

src/Engine/MapController.java
```diff
@@ -467,14 +467,14 @@ public void onStateChange()
...
         // Pull out the first enemy available, or ourselves
-        Commander targetCO = myGame.activeCO;
-        for( Commander co : myGame.commanders )
-          if( myGame.activeCO.isEnemy(co) )
+        Commander targetCO = myGame.activeArmy.cos[0];
+        for( Army army : myGame.armies )
+          if( myGame.activeArmy.isEnemy(army) )
           {
-            targetCO = co;
+            targetCO = army.cos[0];
```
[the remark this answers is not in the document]
ThislsAUsername: I'm thinking the currently active one would just be the first one. Just seems convenient.
As for picking opponents to compare to... agreed, if stat changes were accounted for. This is just the base damage chart, though.

src/UI/GameStatsController.java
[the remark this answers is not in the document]
ThislsAUsername: Pretty sure this has been addressed?

src/AI/WallyAI.java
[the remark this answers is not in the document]
ThislsAUsername: Since you left this open - I'm happy leaving Wally borked for now.

src/AI/ModelForCO.java
ThislsAUsername: > That said, you are now creating ephemeral units every time you need CO-specific unit info (e.g. to calculate matchups).

I thought Muriel was already doing that?
`ModelForCO` is just a pair of final pointers, so it's not a full unit.
> Seems like it would be cheaper to just create one unit per type per CO up front instead of one unit per type per CO per other type per other CO ephemerally. But that feels almost like going back to CO-specific unit models, except they'd be mutable.

I mean, caching something like that in your AI isn't the same as CO-specific unit models.
If you really wanted to be penny-pinching about it, I'd say to allocate exactly two `UnitContext`s (for attacker/defender) and swap out properties as you do calcs.
> *Edit: I see you did add a javadoc note that could be construed as providing that insight, but it's pretty oblique; it describes what an AI could use it to do, but not how this class helps it to do that. Or maybe I was just too stupid or distracted to make the necessary connection, who knows.

I thought "track stuff" plus "overrides `equals()` and `hashCode()`" was enough to be reasonably clear.
Should I add "Overrides equals() and hashCode() so it can be used as a Map key."?
I do want the docs to be useful, and part of that is keeping them minimal enough that people will actually read them, lol.

src/AI/Muriel.java
ThislsAUsername: The API would be:
Commander's unit list and anywhere else it's cached would now be our new bespoke class.
That class would own the `getAllModels(long role)` family of functions, with Commander's versions just dispatching to the collection (to avoid code churn).

...I think that's it?

src/UI/Art/SpriteArtist/InfoView.java
ThislsAUsername: This has been addressed on `presentationFixes`, I think?
image

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: `drawAllCommanderOverlays()` wants to draw the highlight underneath the actual header.
It doesn't know how big any given *Army* overlay will be, since I didn't initially have hooks to calculate that "for real" before, and didn't get around to revising.
So it draws the overlay first, gets how big it is, and uses that for the highlight before committing the actual drawn header.

I will take a shot at fixing this on `presentationFixes`.

src/AI/Muriel.java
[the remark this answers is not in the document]
ThislsAUsername: Taking a gander, it looks like there's nowhere we invoke `Army.cos[0].getShoppingList()`, so I don't think adding it to Army's interface makes sense right now?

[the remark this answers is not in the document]
ThatIsAUsername: Fair enough about `getShoppingList`. Still seems... awkward to have to query army for a CO to get a unit model that is defined independently of any CO or army though.

src/Engine/MapController.java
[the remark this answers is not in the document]
ThatIsAUsername: > I'm thinking the currently active one would just be the first one.

Are you saying `Army` will shuffle the `co`s collection each turn so that index 0 is always the active one? Because well, it currently doesn't, and I don't think I want it to.

That said, it doesn't matter whether we actually get the active one if the damage chart doesn't take CO differences into account. Though that calls into question whether we should include the damage chart at all if it's going to be misleading.

src/AI/ModelForCO.java
ThatIsAUsername: It's more the "colocates CO and UnitModel to allow creating a unit to calculate CO-specific unit stats that aren't directly available from just the model", and less the "now you can use this pair as a map key" rationale I'm looking for to explain the class's existence, and that is only indirectly mentioned.

src/AI/Muriel.java
ThatIsAUsername: I'm not sure how that's any better than what we have. But maybe I'm still not picking up what you're putting down.

src/Engine/MapController.java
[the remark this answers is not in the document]
ThislsAUsername: Yes.
You are correct that it currently does not - our current only tag paradigm doesn't have an "active CO".
Why don't you want it to? It seems weird to me that the first CO in your list would not be the one that's currently active (e.g. I'm in a cart-tag with Grit and Max).

IMO it's valuable to know the base percentages, and I am planning another GUI to do detailed damage matchups between COs - it's just not a thing yet, because we have like 5 pending PRs and other priorities.
An option would be to make the attacker/defender a fixed color rather than matching your own, but then labeling the axes becomes more necessary.

src/AI/ModelForCO.java
ThislsAUsername: Oh. Okay.

src/AI/Muriel.java
ThislsAUsername: It's not a better API, it's allowing our same query API to exist in other places than just `Commander`.
So I could go to the `GameRules` and say "Give me the first defined `TROOP`".

src/Engine/MapController.java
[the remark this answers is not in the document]
ThatIsAUsername: I mean, it could, but then you have extra logic, and the added complexity that `army.cos[0]` is not always the same CO. Not to mention that if it's the other army's turn, none of the COs would be active.

It feels like an overloading of responsibilities. `Army.cos` is a list of the COs in the army. If you want to know if or who is the active CO, I'd probably expect a separate accessor for that (which could return null if there is no such commander).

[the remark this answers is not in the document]
ThislsAUsername: > I mean, it could, but then you have extra logic, and the added complexity that `army.cos[0]` is not always the same CO.

We're gonna have to have logic that switches which CO is currently active regardless. It's a little bit more effort to shift the array contents around, but I think having a single source of truth is worth it.
As far as I can tell, `army.cos[0]` being the active CO instead of always the same CO is just a flat benefit, given our current codebase.
> Not to mention that if it's the other army's turn, none of the COs would be active.

Not at all true - the active CO is still who determines the D2D in cart tags and the one who gets full charging regardless.
> It feels like an overloading of responsibilities.

It is. I had exactly the same thoughts. And then I thought more - I believe it doesn't add any realized complexity for our use cases.
If you have a counterexample, I would like to hear it before I run into it myself.
> `Army.cos` is a list of the COs in the army. If you want to know if or who is the active CO, I'd probably expect a separate accessor for that (which could return null if there is no such commander).

That's probably fair, though it'd be a convenience accessor in all currently-planned cases - you can derive whether there's an active CO by the current tag mode, and you'd know what the active CO is (if any) based on CO list ordering.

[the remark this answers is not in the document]
ThatIsAUsername: > Not at all true - the active CO is still who determines the D2D in cart tags and the one who gets full charging regardless

That is certainly a different definition for "active" than I had in mind. Not to mention that with team merges, you'd arguably have multiple "active" COs anyway.

I mean hey, if you feel strongly about it, I don't care too much if you want to implement CO shuffling. I think it's cleaner and simpler to use an accessor instead, but I'm not the one doing the work. I assume that's not something you plan to do in this PR (on account of not being relevant until we have a tag mode)?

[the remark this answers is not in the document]
ThislsAUsername: > That is certainly a different definition for "active" than I had in mind.

What were you talking about, then?

> I assume that's not something you plan to do in this PR (on account of not being relevant until we have a tag mode)?

Quite correct.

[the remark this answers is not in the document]
ThatIsAUsername: Active. As in, currently acting. As in, the one currently able to move units. There can only be one of those at a time. I can see how your definition also makes sense though, in which case there would always be an active CO in each army, which is convenient. Still might have to sort the issue of having multiple active COs for team merges, but that's a problem for the future.

src/CommandingOfficers/Commander.java
ThislsAUsername: Sounds good. And sorry.

PR #357, opened 2021-12-31
Qis

ThislsAUsername:
I got permission to use some art, so here we are.


ThislsAUsername:
The biggest outstanding issue with this PR is that his COP name is too long. Dunno how to fix that, though.

src/CommandingOfficers/Qis.java
```diff
@@ -0,0 +1,234 @@
...
+          "Commander Qis (pronounced 'keese') has spent his adult life fleeing cold weather, and failing to escape it.\n" +
+          "Whether the cause is simply incredible misfortune or some sort of curse is unclear... he is quite curt on the subject.\n" +
+          "What is known is that Qis has become incredibly adept at using the ever-present snow to his advantage."));
+      infoPages.add(new InfoPage(
+          "Passive:\r\n" +
+          "Watery terrain slows your non-naval units, but all units have perfect movement in snow.\n" +
+          "The path each unit takes on your turn is coated with snow until your next turn."));
+      infoPages.add(new InfoPage(
+          GroovinMovin.NAME+" ("+GroovinMovin.COST+"):\n" +
+          "+"+GroovinMovin.BUFF+" move for all units\n" +
```
ThatIsAUsername, line 47: The '+' is not rendered in the CO description screen. How long has that been broken? It's not just a problem for Qis.

src/CommandingOfficers/Qis.java
```diff
@@ -0,0 +1,234 @@
...
+          if( terrain.isWater() )
+            uc.moveType.setMoveCost(weather, terrain, uc.moveType.getMoveCost(weather, terrain) + 1);
+        }
+      }
+    }
+    // Perfect movement on snow
+    for( TerrainType terrain : TerrainType.TerrainTypeList )
+    {
+      if( MoveType.IMPASSABLE > uc.moveType.getMoveCost(Weathers.SNOW, terrain) )
+        uc.moveType.setMoveCost(Weathers.SNOW, terrain, 1);
```
ThatIsAUsername, line 85: So his units can move better on snowy mountains than on clear mountains?

src/Units/MoveTypes/MoveType.java
```diff
@@ -88,6 +88,8 @@ public boolean canTraverse(Environment tile)
   /** Sets the cost to move through terrain during weather. */
   public void setMoveCost(Weathers weather, TerrainType terrain, int cost)
   {
+    if( cost > IMPASSABLE )
+      cost = IMPASSABLE;
```
ThatIsAUsername, line 92: Does this happen?

src/CommandingOfficers/Qis.java
```diff
@@ -0,0 +1,234 @@
...
+          "Watery terrain slows your non-naval units, but all units have perfect movement in snow.\n" +
+          "The path each unit takes on your turn is coated with snow until your next turn."));
+      infoPages.add(new InfoPage(
+          GroovinMovin.NAME+" ("+GroovinMovin.COST+"):\n" +
+          "+"+GroovinMovin.BUFF+" move for all units\n" +
+          "Snow trails left this turn last +1 turn for each space moved so far"));
+      infoPages.add(new InfoPage(
+          TrailOfChill.NAME+" ("+TrailOfChill.COST+"):\n" +
+          "+"+TrailOfChill.BUFF+" move for all units\n" +
+          "All snow on the map will last "+TrailOfChill.DURATION+" turns"));
```
ThatIsAUsername, line 52: On larger maps, this seems like it would just turn into perpetual snow.

src/CommandingOfficers/Qis.java
```diff
@@ -39,7 +39,8 @@ public instantiator()
           "What is known is that Qis has become incredibly adept at using the ever-present snow to his advantage."));
       infoPages.add(new InfoPage(
           "Passive:\r\n" +
-          "Watery terrain slows your non-naval units, but all units have perfect movement in snow.\n" +
+          "Watery terrain slows your non-naval vehicles, and they lose -1 movement when starting in rain\n" +
```
ThatIsAUsername, line 42: "lose -1 movement" is awkward phrasing. They gain movement?

src/Terrain/TerrainType.java
```diff
@@ -90,14 +90,14 @@ private TerrainType(int attributeFlags, int defense, Color mainColor, String nam
   private static final String TEMP_AIRPORT_NAME = "TEMP_AIRPORT";
   public static final TerrainType TEMP_AIRPORT = new TerrainType( TEMP_AIRPORT_FLAGS, TEMP_AIRPORT_DEFENSE, TEMP_AIRPORT_COLOR, TEMP_AIRPORT_NAME, GRASS );
 
-  private static final int BRIDGE_FLAGS = LAND | WATER;
+  private static final int BRIDGE_FLAGS = LAND;
```
ThatIsAUsername, line 93: But... bridge terrains do have water. Does this mean ships can't go there now?

src/Terrain/TerrainType.java
```diff
@@ -90,14 +90,14 @@ private TerrainType(int attributeFlags, int defense, Color mainColor, String nam
...
   private static final int BRIDGE_DEFENSE = 0;
   private static final Color BRIDGE_COLOR = new Color(189, 189, 189);
   private static final String BRIDGE_NAME = "BRIDGE";
   public static final TerrainType BRIDGE = new TerrainType( BRIDGE_FLAGS, BRIDGE_DEFENSE, BRIDGE_COLOR, BRIDGE_NAME, SEA );
 
   private static final int CITY_FLAGS = LAND | CAPTURABLE | PROFITABLE | PROVIDES_COVER | HEALS_LAND;
-  private static final int CITY_DEFENSE = 2;
+  private static final int CITY_DEFENSE = 3;
```
ThatIsAUsername, line 100: I had sort of assumed we were using AW2 values for things - seems out of scope for just adding a CO, but I don't object to using AW2 values.

src/CommandingOfficers/Qis.java
```diff
@@ -0,0 +1,245 @@
...
+      infoPages.add(new InfoPage(
+          TrailOfChill.NAME+" ("+TrailOfChill.COST+"):\n" +
+          "+"+TrailOfChill.BUFF+" move for all units\n" +
+          "All snow on the map will last "+TrailOfChill.DURATION+" turns"));
+      infoPages.add(new InfoPage(
+          "Hit: Deserts, Desserts\n" +
+          "Miss: Showing up first"));
+      infoPages.add(new InfoPage(
+          "Original art credit:\n" +
+          "@Smile ❤#3047 Discord ID 374728627373211651"));
```
ThatIsAUsername, line 58: There doesn't seem to be a way to put comments on non-text files, so I'll put it here.

1) You said you got permission for custom art? This is a recolor. Really not sure how I feel about that.
2) It doesn't line up correctly. When full-screened it probably doesn't matter, but at the default size...
Screenshot from 2022-01-01 09-45-28
If we want full-body portraits, maybe we need a way to anchor them so they are framed properly at smaller sizes? I dunno.

src/CommandingOfficers/Qis.java
ThislsAUsername: We haven't ever had a + symbol, except when using system fonts. I just put the + in there by habit.
...I guess that's in-scope for `presentationFixes`, innit

src/CommandingOfficers/Qis.java
ThislsAUsername: Yes. I even accidentally let his tanks move on mountains accidentally.
The whole gameplay schtick here is that you move a unit through the heavy terrain to snow it for the dudes behind that dude. The effect is *sorta* like Sturm, but not if you're just YOLOing one unit (and it requires more planning)

src/Units/MoveTypes/MoveType.java
ThislsAUsername: I originally added +1 to all movement costs in rain, which was awful.
So yes, that would happen. I didn't find anything wrong, but I don't want to run into a subtle bug down the line.

src/CommandingOfficers/Qis.java
ThislsAUsername: It could. The neat bit is that he can only make snow where his dudes go, though - it just interacts in fun ways with other COs.
The SCOP is intentionally a bit more niche than COP.

src/CommandingOfficers/Qis.java
ThislsAUsername: It is awkward phrasing. I guess I probably picked up using double negatives like that from the AWBW customization chat.

src/Terrain/TerrainType.java
ThislsAUsername: It is out of scope, but he was slowed by bridges and I didn't like that part.

src/CommandingOfficers/Qis.java
ThislsAUsername: It's not *just* a recolor. But yes, I'm aware it's based on Koal.
Yeah, I think I'll tweak the alignment deal in `presentationFixes` and report back if I do.

src/Terrain/TerrainType.java
ThislsAUsername: Correct. That's how AW2 works.
DoR-style bridges are definitely interesting, but most of the map design we have now assumes AW2 bridges anyway.

src/CommandingOfficers/Qis.java
ThatIsAUsername: I could swear we had that working at one point. I wouldn't have written a bunch of CO ability descriptions with pluses if it hadn't been working at the time.

ThislsAUsername: It was probably working before we had fonts? IDK.
Either way, I fixed it in `presentationFixes`.

src/CommandingOfficers/Qis.java
ThatIsAUsername: It's little more than a recolor, and I'd take badly-drawn originals over recolors any day. I know you got permission from this guy to use it, but I'm not sure it's _his_ permission you'd really need, anyway. But I mean, it's your CO I guess.

And changing the face/eyes to focus on the tree? Now you're just trolling.

ThislsAUsername: Hmm. I think I can see your point about recolors. This *is* a parody of Koal, though, so leaning into it a little more isn't bad IMO.
The point about permission is probably fair as well.

As for trolling... Yeah, probably guilty as charged.

src/CommandingOfficers/Qis.java
```diff
@@ -97,7 +97,7 @@ public void modifyMoveType(UnitContext uc)
       // Watery terrain costs +1 move
       for( TerrainType terrain : TerrainType.TerrainTypeList )
       {
-        if( terrain.isWater() )
+        if( terrain.isWater() || terrain == TerrainType.RIVER )
```
ThatIsAUsername, line 100: Hmm. We probably have a good reason for isWater() == false for rivers, since it seems counterintuitive. Do you remember why?

src/CommandingOfficers/Qis.java
ThatIsAUsername: It would also be more consistent to rephrase to put the numbers up front; most/all of the other COs do that, I think.
e.g. "-1 movement for non-naval vehicles when moving through watery terrain" or whatever.

ThatIsAUsername:
As far as the COP name being too long, I think swapping the funds/ability bar back where they were would resolve that...

src/CommandingOfficers/Qis.java
ThislsAUsername: I can't think of a way to phrase it concisely and clearly, and I think those virtues probably trump consistency?

src/CommandingOfficers/Qis.java
ThislsAUsername: Sea units can't go there.
The flag probably should be called... `OCEAN`?
So `SEA` tiles are a kind of `OCEAN`, as well as reefs and shoals.

src/CommandingOfficers/Qis.java
ThatIsAUsername: Ah, fair enough I guess. I'm not sure we are perfectly consistent otherwise anyway.

PR #358, opened 2021-12-31
Presentation fixes

ThislsAUsername:
A bunch of assorted UI-ish jank that I've been putting off.


src/Engine/GameActionSet.java
```diff
@@ -78,7 +78,7 @@ public String toString()
     {
       // NOTE: Ability and Unit Production actions return null for getType(), since they have no associated UnitActionFactory.
       // However, they are also never put into GameActionSets, so this is probably fine.
-      return actionOptions.get(0).getType().name();
+      return actionOptions.get(0).getType().name(actionOptions.get(0).getActor());
```
ThatIsAUsername, line 81: This comment may not have aged well. It says production actions aren't put into `GameActionSets`, but this change is only needed for production actions.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
```diff
@@ -68,11 +68,11 @@ public static void drawTileDetails(Graphics g, GameMap map, XYCoord tileToDetail
     // Draw the tile coordinates.
     if(SpriteOptions.getCoordinatesEnabled())
     {
-      String coordStr = String.format("(%d, %d)", tileToDetail.xCoord, tileToDetail.yCoord);
+      String coordStr = String.format("(%d,%d)", tileToDetail.xCoord, tileToDetail.yCoord);
```
ThatIsAUsername, line 59: Really trying to make the most of screen real estate?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -324,12 +324,12 @@ private BufferedImage renderMap()
     for( DamagePopup popup : mapController.getDamagePopups() )
       drawDamagePreview(mapGraphics, popup, gameMap.isLocationEmpty(popup.coords));
 
-    // When we draw the map, we want to center it if it's smaller than the view dimensions
+    // When we draw the map, we want to right-justify it if it's smaller than the view dimensions
```
ThatIsAUsername, line 335: Why? This looks pretty weird at full screen.

ThatIsAUsername: I mean yeah, it brings up the extra CO overlays sooner, but still. Not sure it's a net improvement.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -623,7 +625,7 @@ private void drawHUD(Graphics g, boolean includeTileDetails)
 
   private int lastTurnNum = -1;
   private BufferedImage turnNumImage;
-  private void drawTurnCounter(Graphics g, boolean counterIsRight)
+  private void drawTurnCounter(Graphics g, boolean overlayIsLeft)
```
ThatIsAUsername, line 639: Was the turn counter so in the way that it had to move over with the CO overlay? The CO overlay looks funny now that it's not top-justified. Not the end of the world though, I guess.

src/Engine/GameEvents/GameEventListener.java
```diff
@@ -90,4 +91,31 @@ default public void unregister(GameInstance gi)
   default public GameEventQueue receiveMassDamageEvent(Commander attacker, Map<Unit, Integer> lostHP){ return null; };
   default public GameEventQueue receiveModifyFundsEvent(Army beneficiary, int fundsDelta){ return null; };
 
+  /**
+   * Want to be lazy, but too lazy to implement laziness? This class is for you.
+   */
+  public static interface CacheInvalidationListener extends GameEventListener
```
ThatIsAUsername, line 97: This... really isn't less work than just doing it in MTDA directly the way it was before. So far as I can tell, it is strictly more work since you still need all the same calls, and it's less intuitive to boot.

src/CommandingOfficers/Commander.java
```diff
@@ -365,7 +365,10 @@ public int getIncomePerTurn()
     ArrayList<UnitModifier> output = new ArrayList<UnitModifier>();
     output.add(this);
     output.addAll(unitMods);
-    output.addAll(this.army.getModifiers());
+    if( null != army )
+      output.addAll(army.getModifiers());
+    else
+      System.out.println("WARNING!: Getting Commander modifiers when CO.army == null!");
```
ThatIsAUsername, line 371: a) Commanders don't have modifiers anymore?
2) Did this branch really need to go on top of the army branch?
z) Is this really a warning case? Is there always (supposed to be) an Army in the picture?

src/Engine/GameActionSet.java
ThislsAUsername: The comment is referring to building units from buildings, but that's probably fair.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
ThislsAUsername: The alignment was bugging me.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Hmm. I suppose I could right-justify until there's "plenty of space" and then go as close to center-aligned as possible?

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: Yes and no. If you're at the top corner of the map and also the window, there's no way to see what's under it (and your cursor can get under it).
I tried moving it under the CO header(s), but that didn't look quite right either.

src/Engine/GameEvents/GameEventListener.java
ThislsAUsername: Okay, but:
It's less likely to get out of date (it was already missing some event hooks).
I can re-use that work elsewhere, making it less work overall.

src/CommandingOfficers/Commander.java
ThislsAUsername: ...nice list labels.
a) It's Commander *`UnitModifiers`*. I guess I could specify that.
2) Yes, because Army added the multi-CO UI (because that's kinda critical for using `Army`)
z) One of the AIs hit that during game init once, which is fine(?) because there won't be any modifiers in `Army` at that point other than the environment hooks.

res/dpfont_settings.json
```diff
@@ -4,7 +4,7 @@
 	"in-glyphs": [
 		"ABCDEFGHIJKLMNOPQRSTUVWXYZ",
 		"abcdefghijklmnopqrstuvwxyz",
-		"0123456789%./-~,;:!?'&()_"
+		"0123456789%./-~,;:!?'&()_+"
```
ThatIsAUsername, line 7: Didn't you have to modify dpfont_flyphs.png as well?

ThislsAUsername: Nope. The + was already there.

src/CommandingOfficers/Commander.java
ThatIsAUsername: If a Commander will always have an army, it seems strange that an AI would be able to act before an army was available. Might be worth an issue to investigate later?

ThatIsAUsername: Also I just tried running and it fails to launch due to a missing army now:
```java
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException: Cannot read field "cos" because "this.myGame.activeArmy" is null
	at UI.Art.SpriteArtist.SpriteMapView.<init>(SpriteMapView.java:96)
	at UI.Art.SpriteArtist.SpriteEngine.createMapView(SpriteEngine.java:31)
	at UI.PlayerSetupController.handlePlayerSetupInput(PlayerSetupController.java:134)
	at UI.PlayerSetupController.handleInput(PlayerSetupController.java:86)
	at UI.GameOptionSetupController.handleInput(GameOptionSetupController.java:52)
	at Engine.Driver.keyPressed(Driver.java:115)
```

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -91,6 +92,7 @@ public SpriteMapView(GameInstance game)
     menuArtist = new MenuArtist(game, this);
 
     myGame = game;
+    COMPREHENSIVE_HUD_H_SIZE = SpriteLibrary.getCoOverlay(myGame.activeArmy.cos[0], true).getWidth() + 42;
```
ThatIsAUsername, line 96: Why+42? I know you love your magic 42s, but you know I hate magic.

ThatIsAUsername: I get it was there before, and it's just a buffer for good measure, so not the end of the world, but descriptive words are nice.

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -586,80 +608,106 @@ public static boolean shouldFlip(Unit u)
...
+      }
+    }
+
+    if( drawAllHeaders )
       overlayIsLeft = true;
 
+    if( drawOppositeHud )
+    {
+      drawTurnCounter(g, overlayIsLeft);
+      drawOppositeDetails(g, overlayIsLeft);
```
ThatIsAUsername, line 647: So now the coords will only be drawn if the window is expanded, even if the user toggles coords on in options?
If the user wants coords, the user should get coords.

I don't think the coords or the turn counter are so in the way that we have to ban them. Even if they completely covered the corner squares (which they don't), those squares are never (I guess maybe "almost never", technically?) game-relevant anyway.
Your comment about not being able to see what's under them also isn't even true, since we have the tile overlay.

I really don't understand why you are so hostile to having them there, to the point where you are overriding user preferences to banish them.

ThislsAUsername: They're also drawn any time a menu is up. I guess we could change the option to "always show opposite UI/details/something"?
I've found them to get in the way at default size, and I quite like playing with the map/window perfectly sized to each other. It is map dependent whether it's a problem, though.

They're not banned. Also, they do cover menu options and damage popups sometimes, which is a real drag since those are smaller and don't show up on the details pane.
As for the details pane working to show you what's there... I guess that's true. It's pretty bad QoL, though.

A user complained about it, and I realized I really didn't like it either.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
ThatIsAUsername: Hah, pretty sure I tried it both ways and kept the space in there because it looked weird without it.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: User complaints are more valid than arbitrary preference changes. That said, there is an option to hide it.

> they do cover menu options and damage popups sometimes

Huh. I would have thought the menus were already being drawn last already. That would be annoying.

I'd probably be fine with grouping them in an "always show details" option or whatever, but then I'm sure someone would want to hide one but not the other. Guess we could wait for the complaint though. I guess another option is to just not have the option, and don't show that stuff unless there is room/when menus are up. Neither is exactly crucial information.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -67,62 +73,69 @@ public static int drawCommanderOverlay(Graphics g, Army army, int yInitialOffset
...
-    int yCurrentOffset = yInitialOffset;
+    final int yShiftPerOverlay = SpriteLibrary.getCoOverlay(army.cos[0], overlayIsLeft).getHeight() - 1;
+
+    final int fundsLength = ("" + army.money).length();
+    final int fundsWidth = ICON_VALUE_SPACING + fundsLength * spriteA.getWidth() + 1;
+    final int drawX = overlayIsLeft ? 0 : mapViewWidth - fundsWidth;
+    final int drawY = yInitialOffset + yShiftPerOverlay*army.cos.length + yTextOffset;
+    SpriteUIUtils.drawMenuFrame(g, SpriteUIUtils.MENUBGCOLOR, SpriteUIUtils.MENUFRAMECOLOR,
+        drawX, drawY, fundsWidth, fundsHeight, moneyBufferPx);
+    drawIconAndValue(g, SpriteLibrary.MapIcons.FUNDS, army.money, fundsLength, drawX, drawY + moneyBufferPx);
```
ThatIsAUsername, line 88: A couple comments:
1: I like putting the power bar back underneath the CO portrait, but in team merge mode you still need to add a bit of buffer between the two overlays so the bar doesn't cover the next CO's forehead.
2: Tossing the funds count in a differently-formatted frame underneath the CO portrait looks odd. Since we already have the faces to say who's who, can we put funds back inside the overlay area instead of the name? Yes, this would mean all COs would report the same amount in merge mode, but I don't think that's really a bad thing. If anything, it just underscores that they have a shared purse.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThislsAUsername: There's no option to hide the turn counter, and of the two that's the guiltier party (damage popups for the top *two* rows hide under it)

Nah, menus are drawn on the map, from back when we didn't draw anything outside the map.
There wasn't a pressing need to change that, so...

Okay. Do you have a preference between "no option" and "always show option"? If you don't, I'll probably do whichever's easier.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: Fair 'nuff.
I don't like putting redundant info on the GUI, but I guess I'll try it.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: No strong preference.

PR #359, opened 2022-01-01
View mode

ThislsAUsername:
This is the fabled second QoL button. The overlay-toggle feature was specifically requested by the users.

### DIFF
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -219,7 +219,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
           // If we hit BACK while over a unit, add it to the threat overlay for this CO
           MapLocation loc = myGame.gameMap.getLocation(myGame.getCursorCoord());
           Unit resident = loc.getResident();
-          if( null != resident )
+          if( getOverlayMode() == OverlayMode.THREATS_MANUAL && null != resident )
           {
             ArrayList<Unit> threats = myGame.activeArmy.threatsToOverlay;
             if( threats.contains(resident) )
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -649,7 +649,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
 
   public enum OverlayMode
   {
-    THREATS, GUNS, VISION, NONE,
+    THREATS_MANUAL, THREATS_ALL, VISION, NONE,
   }
   private int[] armyOverlayModes;
   public OverlayMode getOverlayMode()
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -52,6 +52,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
     myView = view;
     myView.setController(this);
     inputMode = InputMode.INPUT;
+    armyOverlayModes = new int[game.armies.length];
     isGameOver = false;
     nextSeekIndex = 0;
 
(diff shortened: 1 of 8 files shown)


src/Engine/OverlayCache.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    return markCache;
+  }
+
+  private OverlayCache(GameInstance game)
+  {
+    this.game = game;
+    registerForEvents(game);
+    InvalidateCache();
```
ThatIsAUsername, line 37: on construction?

src/Engine/OverlayCache.java
```diff
@@ -0,0 +1,122 @@
...
+import Engine.MapController.OverlayMode;
+import Engine.GameEvents.GameEventListener.CacheInvalidationListener;
+import Terrain.MapPerspective;
+import UI.GameOverlay;
+import Units.Unit;
+
+public class OverlayCache implements CacheInvalidationListener
+{
+  private static final long serialVersionUID = 1L;
+  private static OverlayCache markCache;
```
ThatIsAUsername, line 18: `markeCache` might not be the best name for this variable.

src/Engine/MapController.java
```diff
@@ -641,6 +647,16 @@ public XYCoord getContemplationCoord()
     return myGameInputHandler.getCoordinateOptions();
   }
 
+  public enum OverlayMode
+  {
+    THREATS, GUNS, VISION, NONE,
```
ThatIsAUsername, line 652: Took me a while to figure out that `THREATS` means "only threats which are specifically flagged by the user". Based on the name alone, it seems potentially duplicative of `GUNS`.

src/Engine/MapController.java
```diff
@@ -221,6 +226,7 @@ private void handleFreeTileSelect(InputHandler.InputAction input)
               threats.remove(resident);
             else
               threats.add(resident);
+            OverlayCache.instance(myGame).InvalidateCache();
```
ThatIsAUsername, line 229: When in `THREATS` mode, flagging/unflagging units works immediately. When in any other mode, it only takes effect once the mode changes again. This is really only obvious because of the `T` mark that gets applied. Strange nonetheless.

src/Engine/MapController.java
ThislsAUsername: Yeah. I thought about putting "user" or "enemy" something in some of them, but I couldn't figure a way to word it.
Any ideas?

src/Engine/MapController.java
ThislsAUsername: The flagging goes through immediately; the mark doesn't show up because the overlay didn't change.
FWIW, I kinda wanted to put `OverlayCache` on the UI side of things, but that would prevent `MapController` from directly notifying it when the list changes. (there are ways around that, but I'm not super worried about it at this point)

src/Engine/MapController.java
ThatIsAUsername: I dunno, maybe `TAGGED` or `MANUAL` vs `ALL`.

ThatIsAUsername:
I think my only real complaint now is that there is little/no visual indicator of what view mode you are in. It can be confusing rotating through the four since manual and none look the same if nothing is flagged. This isn't really a huge deal though.

There's also no instructions for how to flag units. Most of our controls are self-explanatory, but the discoverability of this feature is low/nonexistent.

Also: Do we really need both the manual and none options?

ThislsAUsername:
> there is little/no visual indicator of what view mode you are in

I get that, but IDK if I really want to clutter up the UI more than it already is.
(On that note, I realized recently that we could try leaving the coords/turn number on the opposite side, but only displaying them when there's a menu up - this gives the user maximum visibility most of the time, but also easy access to that info. Then we could put the current view mode stacked with the coords)
> There's also no instructions for how to flag units. Most of our controls are self-explanatory, but the discoverability of this feature is low/nonexistent.

It's the same as on the carts, though obviously we don't have a tutorial (and people probably wouldn't do it if we had one).
> Also: Do we really need both the manual and none options?

It was a specific user request. The all threats and vision modes are me being silly.

ThatIsAUsername:
Fair enough. No objections to merging then if you think it's ready.

PR #360, opened 2022-01-02
Tileset

ThislsAUsername:
You asked about the tileset, so this is the changes Jay's made so far.
He also swapped the HQ for the OS one, but IMO OS's HQ should go to *OS*, not everybody.


ThislsAUsername:
IMO, the only pretty-objective improvement is the AW city/factory just because they're visually distinct.
...also, it appears I was wrong and I don't have his mountain edits (or he never made any) - those look like AW mountains to me.

ThatIsAUsername:
Are those other ones not also just the AW sprites? They all look standard to me (though I didn't bust out the original sprites to compare). I thought I recall seeing something more different when you showed me before.

ThislsAUsername:
You were indeed correct.
...also, I think that might just be an AW road, but I ain't sure

ThislsAUsername:
I think I'm cool with merging whenever you're done with this.

ThatIsAUsername:
I mean, I might tweak the forests if I get time, but that's not essential. More important: are the properties all still hydrophilic?

That might be part of why we left the roads connecting to HQs but not other properties before (plus cart precedent).

ThatIsAUsername:
Cool. This is probably good then. I can do more spriting if I ever want to and have time at the same time again.

PR #362, opened 2022-03-12
Map picker view tweaks

ThislsAUsername:


ThatIsAUsername:
Kinda enjoyed the pixely map. This is probs better though.

Do we also want to scale it in between the 1x -> 2x (and beyond) sizes?

ThislsAUsername:
Not really? I like my exact pixels.

ThatIsAUsername:
On large maps (e.g. Europe) the white boxes in the team-select screen are not placed correctly.

src/UI/Art/SpriteArtist/MapSelectMenuArtist.java
```diff
@@ -108,7 +116,7 @@ private static void drawMapSelectMenu(Graphics g, MapSelectController gameSetup)
     int maxMiniMapHeight = drawableHeight - MapInfoPaneDrawHeight;
 
     // Find the center of the minimap display.
-    int miniMapCenterX = nameSectionDrawWidth + (maxMiniMapWidth / 2);
+    int miniMapCenterX = nameSectionDrawWidth + ((maxMiniMapWidth+1) / 2);
```
ThatIsAUsername, line 119: Why +1? Is half the map width not the center of the map frame for some reason?

ThislsAUsername: I think it had a chance (with odd widths?) of overlapping the border between the two sections at certain sizes.
So I just said "no".

ThislsAUsername:
> On large maps (e.g. Europe) the white boxes in the team-select screen are not placed correctly.

Good catch. Very amusing that the code was "right" aside from making the assumption that all scales are integer.

src/UI/Art/SpriteArtist/MapSelectMenuArtist.java
ThatIsAUsername: Strange. Makes me think we must be off on some of the math somewhere, but maybe not worth hunting down.

PR #363, opened 2022-03-14
J.A.K.E.M.A.N.

ThislsAUsername:
This AI doesn't cope well with weird maps, and you can beat it pretty easily by just building stuff it doesn't know how to counter... but the econ and general tactics seem solid.

### DIFF
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -94,7 +94,7 @@ public class AIUtils
 
     // Create the ActionType-indexed map, and ensure we don't have any null pointers.
     Map<UnitActionFactory, ArrayList<GameAction> > actionsByType = new HashMap<UnitActionFactory, ArrayList<GameAction> >();
-    for( UnitActionFactory atype : uc.calculatePossibleActions() )
+    for( UnitActionFactory atype : uc.calculateActionTypes() )
     {
       actionsByType.put(atype, new ArrayList<GameAction>());
     }
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -188,7 +188,9 @@ public class AIUtils
   /** Overload of {@link #moveTowardLocation(Unit, XYCoord, GameMap, Set)} **/
   public static GameAction moveTowardLocation(Unit unit, XYCoord destination, GameMap gameMap )
   {
-    return moveTowardLocation(unit, destination, gameMap, null);
+    Set<XYCoord> excludeDestinations = new HashSet<>();
+    excludeDestinations.add(new XYCoord(unit)); // Assume that the caller does, in fact, want to move.
+    return moveTowardLocation(unit, destination, gameMap, excludeDestinations);
   }
 
   /**
--- a/src/AI/AIUtils.java
+++ b/src/AI/AIUtils.java
@@ -153,7 +153,8 @@ public class AIUtils
       for( int y = 0; y < gameMap.mapHeight; ++y )
       {
         MapLocation loc = gameMap.getLocation(x, y);
-        if( loc.getResident() != null && army.isEnemy(loc.getResident().CO) )
+        final Unit resident = loc.getResident();
+        if( resident != null && resident.CO.isEnemy(army) )
         {
           unitLocs.add(new XYCoord(x, y));
         }
@@ -166,7 +167,7 @@ public class AIUtils
    * Creates a map of COs to the units they control, based on what can be seen in the passed-in map.
    * Units owned by allies are ignored - ours can be trivially accessed via Commander.units, and we don't control ally behavior.
    */
-  public static Map<Commander, ArrayList<Unit> > getEnemyUnitsByCommander(Army myCommander, GameMap gameMap)
+  public static Map<Commander, ArrayList<Unit> > getEnemyUnitsByCommander(Army army, GameMap gameMap)
   {
     Map<Commander, ArrayList<Unit> > unitMap = new HashMap<Commander, ArrayList<Unit> >();
 
@@ -174,7 +175,7 @@ public class AIUtils
       for( int y = 0; y < gameMap.mapHeight; ++y )
       {
         Unit resident = gameMap.getLocation(x, y).getResident();
-        if( (null != resident) && (myCommander.isEnemy(resident.CO)) )
+        if( (null != resident) && (resident.CO.isEnemy(army)) )
         {
           if( !unitMap.containsKey(resident.CO) ) unitMap.put(resident.CO, new ArrayList<Unit>());
           unitMap.get(resident.CO).add(resident);
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -123,21 +123,21 @@ public class CapPhaseAnalyzer implements Serializable
   {
     XYCoord goal = chain.get(0).coord;
 
-    boolean includeOccupiedSpaces = true; // Since we know how to shift friendly units out of the way
+    boolean includeOccupiedSpaces = false; // We assume that the area won't be cluttered
     ArrayList<XYCoord> destinations = Utils.findPossibleDestinations(unit, gameMap, includeOccupiedSpaces);
 
     // If we can get to our destination, go for it
     if( destinations.contains(goal) )
       return new CaptureLifecycle.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, goal, gameMap));
 
-    Utils.sortLocationsByTravelTime(goal, unit, destinations, gameMap);
+    Utils.sortLocationsByTravelTimeToCoord(unit, destinations, gameMap, goal);
 
     for( XYCoord moveCoord : destinations )
     {
       Unit resident = gameMap.getLocation(moveCoord).getResident();
       boolean spaceFree = null == resident;
-      if( !spaceFree )//&& ((unit.CO != resident.CO || resident.isTurnOver)) )
-        continue; // Bail if we can't clear the space
+      if( !spaceFree )
+        continue; // I don't know how we got here, but bail
 
       // Figure out how to get here.
       GamePath movePath = Utils.findShortestPath(unit, moveCoord, gameMap);
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -47,14 +47,18 @@ public class CapPhaseAnalyzer implements Serializable
   public String toString()
   {
     StringBuilder sb = new StringBuilder();
+    sb.append("Contested properties:");
+    for( XYCoord contested : contestedProps )
+      sb.append(String.format(" %s ", contested));
+    sb.append("\n");
     for( XYCoord factoryXYC : capChains.keySet() )
     {
-      sb.append(String.format("Cap chains for %s:", factoryXYC));
+      sb.append(String.format("Cap chains for %s:\n", factoryXYC));
       for( ArrayList<CapStop> chain : capChains.get(factoryXYC) )
       {
-        sb.append(String.format("  chain"));
+        sb.append(String.format("  chain\n"));
         for( CapStop stop : chain )
-          sb.append(String.format("    %s", stop));
+          sb.append(String.format("    %s\n", stop));
       }
     }
     return sb.toString();
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -130,7 +130,7 @@ public class CapPhaseAnalyzer implements Serializable
     if( destinations.contains(goal) )
       return new CaptureLifecycle.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, goal, gameMap));
 
-    Utils.sortLocationsByDistance(goal, destinations);
+    Utils.sortLocationsByTravelTime(goal, unit, destinations, gameMap);
 
     for( XYCoord moveCoord : destinations )
     {
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -14,7 +14,6 @@ import Engine.GamePath;
 import Engine.Utils;
 import Engine.XYCoord;
 import Engine.UnitActionLifecycles.CaptureLifecycle;
-import Engine.UnitActionLifecycles.WaitLifecycle;
 import Terrain.Environment;
 import Terrain.GameMap;
 import Terrain.TerrainType;
@@ -130,21 +129,7 @@ public class CapPhaseAnalyzer implements Serializable
     if( destinations.contains(goal) )
       return new CaptureLifecycle.CaptureAction(gameMap, unit, Utils.findShortestPath(unit, goal, gameMap));
 
-    Utils.sortLocationsByTravelTimeToCoord(unit, destinations, gameMap, goal);
-
-    for( XYCoord moveCoord : destinations )
-    {
-      Unit resident = gameMap.getLocation(moveCoord).getResident();
-      boolean spaceFree = null == resident;
-      if( !spaceFree )
-        continue; // I don't know how we got here, but bail
-
-      // Figure out how to get here.
-      GamePath movePath = Utils.findShortestPath(unit, moveCoord, gameMap);
-
-      return new WaitLifecycle.WaitAction(unit, movePath);
-    }
-    return null;
+    return AIUtils.moveTowardLocation(unit, goal, gameMap);
   }
 
   public CapPhaseAnalyzer(GameMap map, Army viewer)
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -196,7 +196,7 @@ public class CapPhaseAnalyzer implements Serializable
         inf.x = ownedFac.xCoord;
         inf.y = ownedFac.yCoord;
 
-        final GamePath infPath = Utils.findShortestPath(inf, neutralFac, map, true);
+        final GamePath infPath = findFeasiblePath(inf, neutralFac, map);
         if( null == infPath || infPath.getPathLength() < 1 )
           continue; // Can't reach
 
@@ -227,15 +227,15 @@ public class CapPhaseAnalyzer implements Serializable
         inf.x = ownedFac.xCoord;
         inf.y = ownedFac.yCoord;
 
-        final GamePath tankPath = Utils.findShortestPath(inf, propXYC, map, true);
-        if( null == tankPath || tankPath.getPathLength() < 1 )
+        final GamePath infPath = findFeasiblePath(inf, propXYC, map);
+        if( null == infPath || infPath.getPathLength() < 1 )
           continue; // Can't reach this city
 
         int oldDistance = Integer.MAX_VALUE;
         if( possibleOwners.containsKey(owner.army) )
           oldDistance = possibleOwners.get(owner.army);
 
-        int distance = tankPath.getFuelCost(inf, map) * 100 / inf.getMovePower(map);
+        int distance = infPath.getFuelCost(inf, map) * 100 / inf.getMovePower(map);
         if( distance < oldDistance )
           possibleOwners.put(owner.army, distance);
       }
@@ -285,7 +285,7 @@ public class CapPhaseAnalyzer implements Serializable
       inf.x = start.xCoord;
       inf.y = start.yCoord;
 
-      final GamePath infPath = Utils.findShortestPath(inf, dest, map, true);
+      final GamePath infPath = findFeasiblePath(inf, dest, map);
       if( null == infPath || infPath.getPathLength() < 1 )
         continue; // Can't reach
 
@@ -366,7 +366,7 @@ public class CapPhaseAnalyzer implements Serializable
           Utils.sortLocationsByTravelTime(inf, rightfulProps, map);
           XYCoord dest = rightfulProps.get(0);
 
-          final GamePath infPath = Utils.findShortestPath(inf, dest, map, true);
+          final GamePath infPath = findFeasiblePath(inf, dest, map);
           if( null == infPath || infPath.getPathLength() < 1 )
           {
             last.extraTurns = LOOKAHEAD_TURNS+1;
@@ -404,6 +404,14 @@ public class CapPhaseAnalyzer implements Serializable
       Collections.sort(chainList, new CapStopFundsComparator(infMove));
   }
 
+  public static GamePath findFeasiblePath(final Unit unit, final XYCoord destination, final GameMap map)
+  {
+    final boolean theoretical = true;
+    return Utils.findShortestPath(new XYCoord(unit), unit.getMoveFunctor(true, theoretical),
+                                  unit.getMovePower(map) * (LOOKAHEAD_TURNS),
+                                  destination.xCoord, destination.yCoord,
+                                  map);
+  }
 
   /**
    * Sort CapStop lists by potential total profit
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -24,10 +24,11 @@ import Units.UnitModel;
 public class CapPhaseAnalyzer implements Serializable
 {
   private static final long serialVersionUID = 7714433092190703028L;
+  static final int LOOKAHEAD_TURNS = 3;
 
   public static class CapStop
   {
-    public int extraTiles = 0; // Defines how many tiles we have already looked ahead to try to find another cap stop
+    public int extraTurns = 0; // Defines how many turns we have already looked ahead to try to find another cap stop
     public final XYCoord coord;
     public CapStop(XYCoord coord)
     {
@@ -36,7 +37,7 @@ public class CapPhaseAnalyzer implements Serializable
     @Override
     public String toString()
     {
-      return coord + "+" + extraTiles;
+      return coord + "+" + extraTurns;
     }
   }
   protected Map<XYCoord, ArrayList<ArrayList<CapStop>>> capChains = new HashMap<>();
@@ -292,7 +293,7 @@ public class CapPhaseAnalyzer implements Serializable
 
       ArrayList<CapStop> chain = new ArrayList<>();
       CapStop build = new CapStop(start);
-      build.extraTiles = distance - 13;
+      build.extraTurns = distance/inf.getMovePower(map) - 13;
       chain.add(build);
       CapStop cap = new CapStop(dest);
       chain.add(cap);
@@ -314,27 +315,35 @@ public class CapPhaseAnalyzer implements Serializable
 
   public void buildBaseCapChains(GameMap map, Army viewer, ArrayList<XYCoord> rightfulProps, ArrayList<XYCoord> startingFactories)
   {
+    ArrayList<XYCoord> remainingFactories = new ArrayList<>(startingFactories);
     // Build initial bits of capChains
-    for( XYCoord start : startingFactories )
+    for( XYCoord start : remainingFactories )
     {
-      ArrayList<ArrayList<CapStop>> chain = new ArrayList<>();
-      capChains.put(start, chain);
+      ArrayList<ArrayList<CapStop>> chainList = new ArrayList<>();
+      capChains.put(start, chainList);
     }
 
     boolean madeProgress = true;
     final Unit inf = new Unit(viewer.cos[0], viewer.cos[0].getUnitModel(UnitModel.TROOP));
-    int infMove = inf.getMovePower(map);
-    // Find the next stop or iterate extraTiles on all cap chains
+    if( startingFactories.size() > 0 )
+    {
+      inf.x = startingFactories.get(0).xCoord;
+      inf.y = startingFactories.get(0).yCoord;
+    }
+    final int infMove = inf.getMovePower(map);
+
+    // Find the next stop or iterate extraTurns on all cap chains
     while (madeProgress && !rightfulProps.isEmpty())
     {
       // Create new cap chains
-      for( XYCoord start : startingFactories )
+      for( XYCoord start : remainingFactories )
       {
         ArrayList<CapStop> chain = new ArrayList<>();
         CapStop build = new CapStop(start);
         chain.add(build);
         capChains.get(start).add(0, chain);
       }
+
       madeProgress = false;
       for( ArrayList<ArrayList<CapStop>> chainList : capChains.values() )
         for( ArrayList<CapStop> chain : chainList )
@@ -343,6 +352,13 @@ public class CapPhaseAnalyzer implements Serializable
             break;
 
           CapStop last = chain.get(chain.size() - 1);
+          if( last.extraTurns >= LOOKAHEAD_TURNS )
+          {
+            if (chain.size() == 1)
+              remainingFactories.remove(last.coord);
+            break;
+          }
+
           XYCoord start = last.coord;
 
           inf.x = start.xCoord;
@@ -352,11 +368,14 @@ public class CapPhaseAnalyzer implements Serializable
 
           final GamePath infPath = Utils.findShortestPath(inf, dest, map, true);
           if( null == infPath || infPath.getPathLength() < 1 )
+          {
+            last.extraTurns = LOOKAHEAD_TURNS+1;
             continue; // Can't reach
+          }
           madeProgress = true; // We have somewhere we can still get to
 
-          int distance = infPath.getFuelCost(inf, map);
-          final int currentTotalMove = last.extraTiles + infMove;
+          final int distance = infPath.getFuelCost(inf, map);
+          final int currentTotalMove = (last.extraTurns + 1) * infMove;
 
           if( distance <= currentTotalMove )
           {
@@ -365,7 +384,7 @@ public class CapPhaseAnalyzer implements Serializable
             chain.add(cap);
           }
           else
-            last.extraTiles = Math.min(infMove * 2, currentTotalMove);
+            last.extraTurns++;
         }
     }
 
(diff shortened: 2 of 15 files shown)


src/AI/AIUtils.java
```diff
@@ -166,15 +167,15 @@ public static ArrayList<XYCoord> findEnemyUnits(Army army, GameMap gameMap)
    * Creates a map of COs to the units they control, based on what can be seen in the passed-in map.
    * Units owned by allies are ignored - ours can be trivially accessed via Commander.units, and we don't control ally behavior.
    */
-  public static Map<Commander, ArrayList<Unit> > getEnemyUnitsByCommander(Army myCommander, GameMap gameMap)
+  public static Map<Commander, ArrayList<Unit> > getEnemyUnitsByCommander(Army army, GameMap gameMap)
```
ThatIsAUsername, line 170: Should this by `byArmy` now?

ThislsAUsername: Still no - individual units are assigned to a `Commander`, and their stats can differ within an `Army` as a result.

src/Test/TestProductionClearing.map
```diff
@@ -1,4 +1,4 @@
- 1FC  GR  GR  GR
+ 1FC  GR  GR  CT
```
ThatIsAUsername, line 1: Odd change, considering there are no other test changes. Is this just... for orientation purposes?

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+      boolean spaceFree = null == resident;
+      if( !spaceFree )
+        continue; // I don't know how we got here, but bail
+
+      // Figure out how to get here.
+      GamePath movePath = Utils.findShortestPath(unit, moveCoord, gameMap);
+
+      return new WaitLifecycle.WaitAction(unit, movePath);
+    }
+    return null;
```
ThatIsAUsername, line 147: Instead of all this, why not just call `AIUtils.moveTowardLocation()`?

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  public CapPhaseAnalyzer(GameMap map, Army viewer)
```
ThatIsAUsername, line 150: Y'know, most people would put the class constructor at the top of the class, below the class variables, but... you do you, I guess.

src/Test/TestProductionClearing.map
ThislsAUsername: Travel objectives were being added even when unreachable.
I fixed *that* bug, but that meant that the unit on the factory wouldn't move since it had nowhere useful to go.
I figured that's not really a representative use case, so I modified the test instead.

src/AI/ModularAI.java
```diff
@@ -183,4 +185,31 @@ public GameAction getNextAction(PriorityQueue<Unit> unitQueue, GameMap gameMap)
...
+      if( null == ai.capPhase
+          || !unit.model.baseActions.contains(UnitActionFactory.CAPTURE) )
+        return null;
+
+      // Follow the cap chain, if possible
+      final GameAction capAction = ai.capPhase.getCapAction(map, unit);
+      if( null != capAction && capAction.getType() == UnitActionFactory.CAPTURE )
+      {
+        ai.unownedProperties.remove(capAction.getMoveLocation());
+      }
```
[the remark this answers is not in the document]
ThislsAUsername: Yes, but `unownedProperties` is repopulated each turn - the intent is "things I don't have but want", and IDK that it's useful to allocate more dudes to go cap a property I'm already capturing.

src/AI/ModularAI.java
```diff
@@ -27,6 +28,7 @@ public abstract class ModularAI implements AIController
   protected ArrayList<AIModule> aiPhases;
   // Sets the ordering for units in the unit queue fed to the modules
   protected Comparator<Unit> unitOrderSetter = new AIUtils.UnitCostComparator(false);
+  protected CapPhaseAnalyzer capPhase;
```
ThatIsAUsername, line 31: Seems a little odd to have this as part of the base class here, instead of self-contained in the module. I guess that would require the Commander to initialize it once the map is available, but there's already an init step anyway.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: My reasoning was basically "put the basic API first, and the constructor goes later 'cause it's super long and involved"
Maybe I should make most of that a named function and make a smaller constructor that calls it.

src/AI/ModularAI.java
ThislsAUsername: Fair point.
The most immediate reason is "it seemed simpler this way, and I don't foresee making new AIs that don't want to have a good cap phase".
The other reason is that the `CapPhaseAnalyzer` could have relevant data for things other than just the cap phase - e.g. what cities will be contested could feed into other computation.
I suspect that the second reason there is something I *was* going to use for JakeMan, then didn't actually do anything with it because Lost&Found specified that it be calculated and... then didn't use it later.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: ...because I keep forgetting it exists?
Yeah, that sounds about right.

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+        final Commander owner = factoryOwnership.get(ownedFac);
+        if( owner == null )
+          continue; // Not yet owned
+
+        final Unit inf = new Unit(owner, owner.getUnitModel(UnitModel.TROOP));
+        inf.x = ownedFac.xCoord;
+        inf.y = ownedFac.yCoord;
+
+        final GamePath infPath = findFeasiblePath(inf, neutralFac, map);
+        if( null == infPath || infPath.getPathLength() < 1 )
```
ThatIsAUsername, line 185: Hm. Maybe we should fix `Utils.findShortestPath()` so we don't need a double failure check on the outside.
Obviously not a super priority for this PR.

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+
+      final GamePath infPath = findFeasiblePath(inf, dest, map);
+      if( null == infPath || infPath.getPathLength() < 1 )
+        continue; // Can't reach
+
+      int distance = infPath.getFuelCost(inf, map);
+
+      ArrayList<CapStop> chain = new ArrayList<>();
+      CapStop build = new CapStop(start);
+      build.extraTurns = distance/inf.getMovePower(map) - 13;
```
ThatIsAUsername, line 281: minus 13? Also it still seems odd to require a map to find a unit's move power.

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+      chain.add(build);
+      CapStop cap = new CapStop(dest);
+      chain.add(cap);
+      factoryCapChains.add(chain);
+
+      // Now that we're just in cap-chain land, don't worry about whether we start with this factory or not.
+      startingFactories.add(dest);
```
ThatIsAUsername, line 288: Huh. Interesting detail. So cap chains are planned from factories that are "rightfully" ours even if we don't have them yet, and just... hang around until infantry get built from those factories to take the assignment.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Minus 13 because that boosts the "assumed profit" from the cap chain, so it happens first.
This is necessary because cap chains to grab factories end at those factories (the cap chains from that factory are considered separately), and one property's income isn't superfantastic.

ThislsAUsername: As for the move power, you could argue either way.
The `extraTurns` metric won't be perfectly accurate in all cases with CO jank (since it doesn't plan the exact individual moves).
OTOH, it may give a better estimate, depending on the exact jank in question.

In summary, I'm hedging against my own future poor choices :v

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Technically, the infantry that snagged the factory will be doing the first chain.
But yes, this cap phase stuff definitely assumes the map is set up without contested production (and I'm not really sure how I *would* plan around contested production).

There's probably room here to modify the cap chain growth loop to account for these factories' chains starting later, but the added complexity probably isn't worth it.
Especially since this code should probably be expanded to handle bikes at some point...

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+        for( ArrayList<CapStop> chain : chainList )
+        {
+          if( rightfulProps.isEmpty() )
+            break;
+
+          CapStop last = chain.get(chain.size() - 1);
+          if( last.extraTurns >= LOOKAHEAD_TURNS )
+          {
+            if (chain.size() == 1)
+              remainingFactories.remove(last.coord);
```
ThatIsAUsername, line 343: This condition indicates that the factory at `last.coord` can't reach anything to capture within `LOOKAHEAD_TURNS`, so we just remove it from further consideration?

ThislsAUsername: Yeah. That's the "base case" if this were recursive.
We've determined that all the cap chains we *can* make from that factory have been made already, so there's no point spinning on it anymore.

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Couldn't this end up prioritizing paths from uncaptured factories to uncaptured bulidings that are more efficiently captured by an already-owned factory?

src/AI/CapPhaseAnalyzer.java
```diff
@@ -0,0 +1,454 @@
...
+          final int currentTotalMove = (last.extraTurns + 1) * infMove;
+
+          if( distance <= currentTotalMove )
+          {
+            rightfulProps.remove(dest);
+            CapStop cap = new CapStop(dest);
+            chain.add(cap);
+          }
+          else
+            last.extraTurns++;
```
ThatIsAUsername, line 372: Why do this iteratively instead of a single pass with `extraTurns == LOOKAHEAD_TURNS`?

ThislsAUsername: Because it's greedy.
If a city is within 2 turns of Factory B, and 3 turns of Factory A, then Factory A gets it under your scheme.

It *would* be pretty cool to find all "cap-chainable" properties and check all permutations for what's optimal, but I've spent a lot of time on this and I want to move on for now, lol

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Technically yes.
That's what I meant by `There's probably room here to modify the cap chain growth loop to account for these factories' chains starting later, but the added complexity probably isn't worth it.`

src/AI/ModularAI.java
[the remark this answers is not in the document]
ThatIsAUsername: Fair enough, but then the code is misleading as written. At least add a comment to explain what's going on here.

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: ... how?
"I need a way to move a unit towards some a location"
"Too bad that use case hasn't come up yet. Pretty sure we don't do that at all"
"Ah well, easy enough to add it."

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Might or might not help with clarity. Maybe not worthwhile if the new functions wouldn't be called from anywhere else.

src/AI/ModularAI.java
ThatIsAUsername: Hah. Did you follow up to see why Lost&Found wanted to calculate that thing?

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Why not just prioritize factory caps explicitly? I guess because a "sufficiently far away" factory should (probably) be worth less than a bunch of nearby cities? I just like to avoid fudge factors when possible.
I see that it matches `CapStopFundsComparator.turnLimit` - are those intentionally the same, or coincidentally (and otherwise unrelated)?

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Realistically, this is copied Wally code that doesn't use that function because it's integrated with the eviction loop and thus needs more insight into the details.
This code doesn't have anything like that since we assume there aren't units in the way when you're doing your initial capture runs.

I could conceive of maybe making a variant of the function that ignores friendlies so the caller can decide where to evict and/or snip.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Yeah. It'd only be called from there.
That's probably why I didn't do it.

It might be worth breaking that function down a bit, but I want readable function signatures and that's hard to do with so much reused state.
Could probably package things into structs to hold that specific data, but I'm not sure that actually makes things clearer.

src/AI/ModularAI.java
ThislsAUsername: Quoth he:
> The idea was to use neutral/opposing contested props as "where to go" tiebreakers pretty much - cities and the tiles next to them being extra spots to cover if it's possible to do that while not compromising the rest of your stuff
> But I suppose computers do what you tell them to and I am bad at talk

> Either what I just said or this was just step 1 of making your AI do my tax returns for me

I thought about trying to do that, but then I decided I already had something acceptable (copied from Wally) and he hadn't written up any sort of real algorithm on it.

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Speaking of contested... if a cap chain was built from a "rightful" factory, and then it was contested, and then halfway through the game we finally claimed it... would the cap chain just hang around until that factory built its first infantry? Or I guess the infantry that captured it could be immediately assigned that cap chain. Is there any statute of limitations on cap chains?

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Why would Factory A get it under that scheme? You could still see that A is farther away.
And yeah, not trying to get you to rewrite everything, just want to understand how it works.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: > Why not just prioritize factory caps explicitly?

Why have two lists when I can get away with one? Laziness.
> I guess because a "sufficiently far away" factory should (probably) be worth less than a bunch of nearby cities?

Uhhhh. I hadn't considered that at all. It's not wrong, but claiming that as my reasoning would be a lie.
> I just like to avoid fudge factors when possible.

Valid. This is supposed to be a big enough "fudge" that you'd expect to see effectively two sub-lists within the list.
> I see that it matches `CapStopFundsComparator.turnLimit` - are those intentionally the same, or coincidentally (and otherwise unrelated)?

Coincidence, probably. The gremlin that is my brain really likes it when I type `13` and `42`. You might have noticed this tendency.

ThatIsAUsername: OK. So factories are just assumed to be 13 turns closer to being captureable because reasons. Gotcha. I'd (as usual) advocate for adding a comment to explain what you are doing, since it's pretty nonobvious.

> Coincidence, probably. The gremlin that is my brain really likes it when I type 13 and 42. You might have noticed this tendency.

Not sure I have? Huh. Good thing I don't have anything against fudge factors. Or magic numbers. Or magic fudge factors.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: Yeah, that could happen.
JakeMan (and by extension this cap logic, which perhaps is worth fixing) is designed around conventional PvP scenarios.

If you load a save during a game, you'll also end up re-generating cap chains to any unclaimed/uncontested properties, which is also questionable.
I want to introduce some controlled state-saving for AIs down the line to fix that and similar stuff.
I guess we could just call the whole AI class `Serializable` and call it a day, but that feels questionable for some reason I can't quite put my finger on.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: >start
>initialize cap chain for each factory
>look at factory A, grab city 3 turns away
>look at factory B. Don't grab the city, because it's already claimed

The current scheme is a (very) low-fidelity simulation of a cap phase - each "inf wave" travels out from its factory until it hits a specific city, then it expands out from that point.
...I just used the wave/particle duality of light as a metaphor for my code. I don't know how to feel about that.

src/AI/CapPhaseAnalyzer.java
ThislsAUsername: That's fair.

There are 66 instances of `13` in our source, and 16 of `42`.

src/AI/CapPhaseAnalyzer.java
ThatIsAUsername: Well, it made sense. You may have just figured out light. Better submit that comment to a sci-mag somewhere and claim your Nobel.

ThislsAUsername: It could, yes. The iterative approach is simpler, though.
If I wanted to be really "correct" about all this, I'd build a(n a)cyclic graph of all the properties of interest and figure out what permutation of chains produces the best results.
That *might* be needed to properly handle bikes, since you don't know *a priori* where you ought to be building them to maximize funds.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+    @Override
+    public AIController create(Army army)
+    {
+      return new JakeMan(army);
+    }
+
+    @Override
+    public String getName()
+    {
+      return "JakeMan";
```
ThatIsAUsername, line 28: Where does this name come from, anyway?

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+    }
+
+    @Override
+    public String getDescription()
+    {
+      return
+          "Wants your free dudes.\n" +
+          "High-level design by @Lost&Found#6348, ID 824112693123612692.";
```
ThatIsAUsername, line 36: This... is not helpful.

src/AI/JakeMan.java
ThislsAUsername: Lost&Found frequently has similar things as his nickname in the AWBW server, and it's his decision tree.

src/AI/JakeMan.java
ThislsAUsername: "Free dude" is a technical term, standing for "insufficiently-protected unit".
Simple AI, simple description.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+
+  @Override
+  public AIMaker getAIInfo()
+  {
+    return info;
+  }
+
+  // What % base damage I'll ignore when checking safety
+  private static final int    INDIRECT_THREAT_THRESHHOLD = 7;
+  private static final int    DIRECT_THREAT_THRESHHOLD = 13;
```
ThatIsAUsername, line 49: https://grammarist.com/spelling/threshold-or-threshhold/

Not that it's like, important.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+    // look where all vehicles are and what their threat ranges are (yes, mechs are vehicles)
+    // take free dudes that you have more defenders for than them
+    // move your leftover vehicles so they cover the tiles attacking most relevant stuff (contested cities, own units)
+    // move infs as far towards nearest contested stuff as you can without underdefending, cap if possible
+
+    aiPhases = new ArrayList<AIModule>(
+        Arrays.asList(
+            new PowerActivator(army, CommanderAbility.PHASE_TURN_START),
+            new CapChainActuator(army, this),
+            new CaptureFinisher(army, this),
```
ThatIsAUsername, line 81: Doesn't `CapChainActuator` handle capturing cities? Or is that just for early game, and `CaptureFinisher` is needed for subsequent captures?

ThislsAUsername: The latter, yeah.
`CapChainActuator` stops doing anything once all the chains are burned.
`CaptureFinisher` is probably not *technically* needed, though you could get weird behavior because infantry would prefer to cap a different city than finish the one they're currently on.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+      ai.unitMapEnemy = new HashMap<UnitModel, Map<XYCoord, Double>>();
+      ai.unitMapFriendly = new HashMap<UnitModel, Map<XYCoord, Double>>();
+      Map<Commander, ArrayList<Unit>> unitLists = AIUtils.getEnemyUnitsByCommander(null, gameMap);
+        for( Commander co : unitLists.keySet() )
+        {
+        Map<UnitModel, Map<XYCoord, Double>> mapToFill;
+          if( myArmy.isEnemy(co) )
+          mapToFill = ai.unitMapEnemy;
+        else
+          mapToFill = ai.unitMapFriendly;
```
ThatIsAUsername, line 135: Indentation gets pretty wonky through here.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+    {
+      myArmy = co;
+      this.ai = ai;
+    }
+
+    @Override
+    public GameAction getNextAction(PriorityQueue<Unit> unitQueue, GameMap gameMap)
+    {
+      ai.unitMapEnemy = new HashMap<UnitModel, Map<XYCoord, Double>>();
+      ai.unitMapFriendly = new HashMap<UnitModel, Map<XYCoord, Double>>();
```
ThatIsAUsername, line 127: This completely rebuilds the threat map for every command the AI returns? Seems... inefficient.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+        GamePath movePath = Utils.findShortestPath(unit, moveCoord, gameMap);
+
+        // Figure out what I can do here.
+        ArrayList<GameActionSet> actionSets = unit.getPossibleActions(gameMap, movePath, includeOccupiedSpaces);
+        for( GameActionSet actionSet : actionSets )
+        {
+          Unit resident = gameMap.getLocation(moveCoord).getResident();
+          boolean spaceFree = null == resident;
+          if( !spaceFree )//&& ((unit.CO != resident.CO || resident.isTurnOver)) )
+            continue; // Bail if we can't clear the space
```
ThatIsAUsername, line 211: `canEvict` is never used in this function - should we check it here before bailing?

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+          continue;
+        double counterPowerTotal = 0;
+        final boolean counterIsMeAndIAmPeaceful = !amAttacking && (counter == unit.model);
+        for( XYCoord coord : counterCoords )
+        {
+          double counterPower = 0;
+          if( unitMapFriendly.get(counter).containsKey(coord) )
+            counterPower = unitMapFriendly.get(counter).get(coord);
+          if( counterIsMeAndIAmPeaceful )
+            counterPower -= PEACEFUL_SELF_THREAT_RATIO * unit.getHPFactor();
```
ThatIsAUsername, line 642: Why create a global constant for this? Seems like any other value wouldn't make much sense here.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+          counterPowerTotal += Math.max(0, counterPower);
+        }
+        final double counterPowerAverage = counterPowerTotal / counterCoords.size();
```
ThatIsAUsername, line 645: Huh. Taking the average is an interesting choice. If you have a 10HP unit and a 1HP unit of the same type, it would look much worse than if you just had the 10HP unit.

src/AI/JakeMan.java
ThislsAUsername: It appears I never got around to handling eviction properly here.
I imagine I punted because nothing like that was in the original writeup.

src/AI/JakeMan.java
ThislsAUsername: Because I was tweaking it and seeing what it did to the behavior.
I think I ended up finding the issue I was trying to solve somewhere else.

src/AI/JakeMan.java
ThislsAUsername: That takes the average of the *tiles*, not of the *units*.
Units are additive (should maybe cap to the top N units, where N is the number of available spaces? IDK), but each unit may not threaten all tiles that an enemy could attack me from... so I take the average.
Maybe it should instead only count units that can threaten all tiles a direct could attack me from?

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+            int damage = Math.min(defender.getHP(), (int)results.defender.getPreciseHPDamage());
+
+            // Convert to abstract value
+            int extraLoss = 0;
+            if( loss >= 1 && unit.getHP() == UnitModel.MAXIMUM_HP )
+              extraLoss += STAY_UNHURT_BIAS;
+            if( loss >= unit.getHP() )
+              extraLoss += STAY_ALIVE_BIAS;
+            loss *= unit.getCost();
+            loss += extraLoss;
```
ThatIsAUsername, line 579: These loss bias factors will have a much greater influence over cheap unit interactions than expensive ones.
The bias for staying unhurt is as much as the cost of an infantry, but piddling in the context of a Md Tank. Should this factor be scaling wrt unit cost?

ThislsAUsername: That was the intent, actually.
An unhurt inf is *massively* more useful than a chipped one, because unhurt infs are very difficult to oneshot.
Tech units can generally still contribute well when hurt, though.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+            damage *= defender.getCost();
+            if( isThreatenedBy(unit.model, defender.model) )
+              damage *= FIRSTSTRIKE_ON_THREAT_WEIGHT;
+            // Value damage to hurt units less
+            damage *= defender.getHPFactor();
+
+            return (double)(damage - loss);
+          }, (terrain, params) -> 0.01); // Attack terrain, but don't prioritize it over units
```
ThatIsAUsername, line 587: This will cause Jakeman to allocate infantry to shoot terrain (forever), when they could be doing more useful things.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+  /**
+   * Queue the first action required to move a unit out of the way
+   * For use after unit building is complete
+   * Can recurse based on the other functions it calls.
+   */
+  private GameAction evictUnit(
+                        GameMap gameMap,
+                        Unit evicter, Unit unit,
+                        boolean avoidProduction )
+  {
```
ThatIsAUsername, line 435: I ran JakeMan vs Muriel on TestRange to test the whole "will JakeMan shoot terrain with unsuitable units" hypothesis (well, first I ran JakeMan vs JakeMan, but the first one immediately capped the second one's Lab because the second one was too busy with cap chains to care, so then I put `Muriel` in slot 2. But anyway, that's a separate issue), and found that JakeMan was on the cusp of winning, but then choked himself with infantry, and slowly crumbled due to an apparent inability to clear production facilities to build more units despite banking tons of funds. Not sure what the deal was, because there were opportunities to clear production, but it seemed like he just stalled.
It may not even be an `evictUnit` problem if something was breaking before that; I haven't attempted to chase down specifically what was happening.

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+  }
+
+  /** Produces a list of destinations for the unit, ordered by their relative precedence */
+  private ArrayList<XYCoord> findTravelDestinations(
+                                  GameMap gameMap,
+                                  Unit unit,
+                                  boolean avoidProduction )
+  {
+    UnitContext uc = new UnitContext(gameMap, unit);
+    uc.calculatePossibleActions();
```
ThatIsAUsername, line 338: Is this just to populate the field inside the `UnitContext`? Intuitive.

How is that different/better than `AIUtils.getAvailableUnitActions()`? I feel like, at some level, we are just implementing parallel APIs.

ThislsAUsername: `calculatePossibleActions()` populates the set of action *types* your unit can do based on any `UnitModifier`s it has.
Is there a better name for it?

src/AI/JakeMan.java
```diff
@@ -0,0 +1,770 @@
...
+        if( unit.CO.unitProductionByTerrain.containsKey(loc.getEnvironment().terrainType)
+            && myArmy.isEnemy(loc.getOwner()) )
+        {
+          goals.add(coord);
+        }
+      }
+    }
+
+    if( goals.isEmpty() ) // If there's really nothing to do, go to MY HQ
+      goals.addAll(myArmy.HQLocations);
```
ThatIsAUsername, line 417: Why go to its own HQ instead of the enemy HQ locations?

ThislsAUsername: Because those should have been contemplated already.

src/AI/JakeMan.java
ThatIsAUsername: I mean, OK I guess. Might cause confusion with the similarly-named CO, but there's no actual affinity there. Not my baby though.

src/AI/JakeMan.java
ThatIsAUsername: "Simple" is subjective, and relative. The "free dudes"/contested-square logic is more sophisticated than the other AIs have, and in any case, the descriptor "Free Dudes" isn't "simple", it is simply obtuse.
Imagine you have played this game before, but were not in any way involved in inventing, writing, or reviewing this source code. What does this description tell you? Absolutely nothing.

src/AI/JakeMan.java
ThatIsAUsername: Hah. Silly infantry.

src/AI/JakeMan.java
ThatIsAUsername: Well. I'd advocate to clean this up one way or the other; clean up the unused nonsense, or actually use it. Then again, I don't care enough to make you, so you decide.

src/AI/JakeMan.java
ThatIsAUsername: > That takes the average of the tiles, not of the units.

Ah. Still an interesting choice. A 10HP unit with a single attack square is much more apparently dangerous than a 10HP unit with four attack options.
A "correct" solution might be to take into account the specific counter locations reachable by each countering unit and allocate slots, then add up the most potent permutation. But eh.

src/AI/JakeMan.java
ThatIsAUsername: Oh, gotcha. Maybe `calculatePossibleActionTypes()`. I might even go crazy and rename `UnitContext.possibleActions` to `UnitContext.actionFactories/Types` and change this function to `populateActionFactories/Types`.

src/AI/JakeMan.java
ThatIsAUsername: Ah, touché. So this is definitely a "we should never actually get here" kind of scenario. I normally like to throw in a warning when that happens, if for no other reason than that it's always fun when I see it pop up.

src/AI/JakeMan.java
ThislsAUsername: The contested square logic is more sophisticated, yes, but the building algorithm is greatly simplified and a fair chunk of the code *around* the contested tile logic is also simplified.
"Free dudes" is only as obtuse as e.g. "linked list". It's technical jargon, but technical jargon I'd kind of expect our effective audience to understand.
If I'm an AWBWian, I would probably know what a "free dude" is. I guess "exploits openings in your defenses" isn't exactly a super great description either, though.

How about I add: "Built specifically for AWBW-meta PvP"?

src/AI/JakeMan.java
ThislsAUsername: > A 10HP unit with a single attack square is much more apparently dangerous than a 10HP unit with four attack options.

Still no, unless I am really misremembering/misreading. The sums of the threats on each square are averaged.
So 2 tanks threatening one square is considered equivalent to 1 tank threatening 2 squares.
The above is kind of a fundamental problem with this method, but I don't see a way to resolve it without getting a lot more complicated.

src/AI/JakeMan.java
ThislsAUsername: > the second one was too busy with cap chains to care

I don't care about that at all. `JakeMan` isn't set up to deal with cheese.
> then choked himself with infantry, and slowly crumbled due to an apparent inability to clear production facilities to build more units despite banking tons of funds

Yeah, this is a problem common to both of my "fancy" AIs. Wally is complicated enough there's a high barrier to entry to start really messing with him, and I dunno that I'm really worried about this case for JakeMan since it's unlikely to show up in PvP maps.

src/AI/JakeMan.java
ThatIsAUsername: > technical jargon

Is "free dudes" really a common phrase that gets thrown around? At any rate, I'd still avoid technical jargon; at least anything that someone coming from the carts wouldn't know about (and they don't have much jargon anyway).

> How about I add: "Built specifically for AWBW-meta PvP"?

That may be overselling its capabilities. At least, I'm pretty sure artillery is pretty important in the AWBW meta. I'd probably just say it looks for favorable unit exchanges or something along that line.

src/AI/JakeMan.java
ThatIsAUsername: Fair enough.

src/AI/JakeMan.java
ThatIsAUsername: Probably also worth mentioning what it builds, since it has a pretty limited roster.

ThatIsAUsername:
Aight. If there isn't anything else you want to change, I think it's good to go.

PR #364, opened 2022-06-09
Fix Qis crash on modifyMovePower()

ThislsAUsername:


src/CommandingOfficers/Qis.java
```diff
@@ -106,7 +106,7 @@ public void modifyMoveType(UnitContext uc)
   @Override
   public void modifyMovePower(UnitContext uc)
   {
-    if( uc.map == null || uc.coord == null )
+    if( uc.map == null || !uc.map.isLocationValid(uc.coord) )
```
ThatIsAUsername, line 109: What was the crash? Was `uc.coord` somehow nonnull and also not valid?

PR #365, opened 2022-06-12
Tag gui

ThislsAUsername:
Just AWBW tags for this PR - those can be feature-complete without a bunch of added junk.


src/UI/PlayerSetupCommanderController.java
```diff
@@ -13,16 +13,29 @@ public class PlayerSetupCommanderController implements IController
...
 
     // Make sure we start with the cursor on the currently-selected Commander.
     cmdrSelector = new OptionSelector(infos.size());
     cmdrSelector.setSelectedOption(myPlayerInfo.currentCo);
+
+    // TODO: adjust for multi-CO
+    tagCmdrList = new ArrayList<>();
+    tagCmdrList.add(myPlayerInfo.currentCo);
+    // TODO: Check for tagging mode
+    tagCmdrList.add(noCmdrIndex); // Append a No CO
```
ThatIsAUsername: Are these TODOs still relevant?

ThatIsAUsername: Nope.

ThatIsAUsername:
If you select AWBW tags, then select strong for the first CO, then hit right, left, down (highlight second CO, then first CO, then move from Strong to No CO), it will scroll the CO list in the wrong direction to get from Strong to No CO.

Also there doesn't seem to be a way to have No CO as part of your team...

ThatIsAUsername:
How is this going to play nice when we get faction sections, since our existing prototype uses horizontal space for that?
I might move the tag boxes down and in front of the portrait to play nice with the default window size (at full screen it doesn't really matter, because space). Perhaps a problem for another time, though.

src/UI/PlayerSetupInfo.java
```diff
@@ -25,14 +26,38 @@
...
   {
-    return String.format("%s %s %s %s %s %s", currentCo, currentColor, currentFaction, flipUnits, currentTeam, currentAi);
+    StringBuilder coString = new StringBuilder();
+    // If we have more than one CO, build a parenthetical list within our list
+    if( coList.size() > 1 )
+      coString.append("( ");
+    for( int co : coList )
+      coString.append(co).append(" ");
+    if( coList.size() > 1 )
+      coString.append(")");
```
ThatIsAUsername, line 36: I probably would have just put this all in a single `if( coList.size() > 1)` for clarity, but this works too.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -44,14 +46,33 @@ public static void draw(Graphics g, IController controller, ArrayList<CommanderI
     BufferedImage tooltip = SpriteUIUtils.makeTextFrame("Press Q for more info", 3, 2);
     myG.drawImage(tooltip, myWidth - tooltip.getWidth(), 3, null);
 
+    /////////////// Tag Picker Panels //////////////////////
+    if( myControl.shouldSelectMultiCO || myControl.tagCmdrList.size() > 1 )
```
ThatIsAUsername, line 50: Uhhh, is it possible for `tagCmdrList.size()` to be greater than 1 if `!myControl.shouldSelectMultiCO`? Seems like just the first check should be sufficient (and the latter check an error case, if anything, assuming the first passes).

src/Engine/MapController.java
```diff
@@ -575,6 +548,10 @@ public void animationEnded(GameEvent event, boolean animEventQueueIsEmpty)
       }
       else
       {
+        if( event.shouldEndTurn() ) {
+          handleEndTurn();
+        }
```
ThatIsAUsername, line 553: I'm not necessarily opposed to having an event for the end of a turn, but adding a new base-class hook that is by definition used in only one subclass would cause me to question my approach.

src/Engine/StateTrackers/BuildCountsTracker.java
```diff
@@ -32,7 +32,7 @@ public GameEventQueue receiveCreateUnitEvent(Unit unit)
     return null;
   }
   @Override
-  public GameEventQueue receiveTurnInitEvent(Army army, int turn)
+  public GameEventQueue receiveTurnEndEvent(Army army, int turn)
```
ThatIsAUsername, line 35: Why switch this (and `TransformationTracker`) to work on end turn instead of init turn? Won't this increase the likelihood of having stale info, since stuff could change between the end of one turn and the beginning of the next?

src/Engine/MapController.java
ThatIsAUsername: Also, why move it into an event? Especially since the event doesn't actually *do* anything, which is what events were created to do.

src/Engine/GameEvents/SwapCOEvent.java
```diff
@@ -0,0 +1,93 @@
...
+    army.cos[0] = army.cos[swapIndex];
+    army.cos[swapIndex] = temp;
+
+    switch (army.gameRules.tagMode)
+    {
+      case AWBW:
+        for( Unit u : army.getUnits() )
+        {
+          if( u.CO == swapTarget )
+            continue;
```
ThatIsAUsername, line 48: Is it possible for this to be true?

ThatIsAUsername:
Is acquired experience split evenly among the inactive COs, or do they all just get a set fraction of whatever comes in? E.g., can I multiply my power intake by just bringing 10 copies of one CO? I assume I'll run across this in the code somewhere, but figured I should write it down to ensure I don't forget.

src/Engine/GameInput/SelectMetaAction.java
```diff
@@ -2,15 +2,20 @@
...
   {
-    CO_STATS, CO_INFO, DAMAGE_CHART, SAVE_GAME, CO_ABILITY, QUIT_GAME, END_TURN
+    CO_STATS, CO_INFO, DAMAGE_CHART, SAVE_GAME, CO_ABILITY, QUIT_GAME, SWAP, END_TURN
   }
-  private static final MetaAction[] NO_ABILITY = {MetaAction.CO_STATS, MetaAction.CO_INFO, MetaAction.DAMAGE_CHART, MetaAction.SAVE_GAME, MetaAction.QUIT_GAME, MetaAction.END_TURN};
+  // Four different cases isn't great, but I like the relative simplicity for now
+  // Will probably just make a pruned copy on-the-fly if we end up needing more cases
+  private static final MetaAction[] DEFAULT      = {MetaAction.CO_STATS, MetaAction.CO_INFO, MetaAction.DAMAGE_CHART, MetaAction.SAVE_GAME, MetaAction.QUIT_GAME, MetaAction.END_TURN};
+  private static final MetaAction[] ONLY_ABILITY = {MetaAction.CO_STATS, MetaAction.CO_INFO, MetaAction.DAMAGE_CHART, MetaAction.SAVE_GAME, MetaAction.CO_ABILITY, MetaAction.QUIT_GAME, MetaAction.END_TURN};
+  private static final MetaAction[] ONLY_SWAP    = {MetaAction.CO_STATS, MetaAction.CO_INFO, MetaAction.DAMAGE_CHART, MetaAction.SAVE_GAME, MetaAction.QUIT_GAME, MetaAction.SWAP, MetaAction.END_TURN};
```
ThatIsAUsername, line 18: Why have CO ability before Quit game and Swap after it? I'd probably put Swap right after CO ability.

src/Engine/Army.java
```diff
@@ -282,15 +306,36 @@ public GameEventQueue receiveMassDamageEvent(Commander attacker, Map<Unit, Integ
   {
     if( this == attacker.army )
       return null; // Punching yourself shouldn't make you angry
+    if( TagMode.AWBW == gameRules.tagMode
+        && getAbilityText().length() > 0 )
+      return null; // If any power is active, deny charge in AWBW mode
```
ThatIsAUsername, line 316: Could we put this check in one place instead of two? Maybe in `calculateCombatCharge`? I guess this way prevents superfluous calculations though. Hm.

src/Engine/GameAction.java
```diff
@@ -109,6 +109,7 @@ public GameEventQueue getEvents(MapMaster gameMap)
     {
       GameEventQueue buildEvents = new GameEventQueue();
 
+      buildEvents.addAll(who.cos[0].revertActiveAbility(gameMap));
```
ThatIsAUsername, line 112: Aww, that's mean. No lingering passive buffs if you pop a power and swap?

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThislsAUsername: Yes. You can get into that state via saved previous player configs for the map.
You could argue we should also save game rules for each map, but the player can just change those before going to the player setup screen anyway.
It is an error case, for sure.

src/Engine/MapController.java
ThislsAUsername: I feel like control flow is a bit of a special case all over the place, being the return values of functions and stuff. It seemed not too egregious here.
Swap actions are a thing, yes, and it also enables listening for end-of-turn.
I could have done swap actions without making an event, but I think it would have been messy without some sort of rework.

...It might not be such a bad idea to move the turn-end logic directly into the turn-end events, which could deduplicate code and remove the need for this hook. Hmm.

src/Engine/StateTrackers/BuildCountsTracker.java
ThislsAUsername: I'm not sure how a CO could end up building units when it's not his turn.
I guess I could see a CO that transforms *other* people's units, though that's kinda weird.
As a result, I'm not sure the stale data question is all that relevant.

Truth be told, the main reason I did it was to clean up Cinder's build count numbers so they don't show up during her next turn-init cycle.
There's probably a better way of doing that than throwing the data away early, but I haven't really thought about it and we don't currently have any cross-CO usecases for these trackers.

src/Engine/GameEvents/SwapCOEvent.java
ThislsAUsername: I don't think so. My hands are typing words.

src/Engine/GameInput/SelectMetaAction.java
ThislsAUsername: Swap is logically a different kind of end turn, not a kind of CO ability.
I don't recall feeling strongly about it, though.

src/Engine/Army.java
ThislsAUsername: I remember thinking about that and not coming up with anything I liked.

src/Engine/GameAction.java
ThislsAUsername: It's that way for every implementation of tags I'm aware of.
Flip side: the power is no longer active, so you can immediately get charge for your opponent's counterattack.

ThislsAUsername:
> If you select AWBW tags, then select strong for the first CO, then hit right, left, down (highlight second CO, then first CO, then move from Strong to No CO), it will scroll the CO list in the wrong direction to get from Strong to No CO.

Huh. That's uggo.
> Also there doesn't seem to be a way to have No CO as part of your team...

You can only have No CO if you don't have any COs. I didn't really see any point in letting you tag somebody and nobody, and it thus provides a decent UI affordance.

ThislsAUsername:
> How is this going to play nice when we get faction sections, since our existing prototype uses horizontal space for that?

Fantastic question. I don't have an answer, I just got struck with this UI idea and decided to throw it together.
I mostly didn't want to get tags implemented in-engine with the only way to access them being via hardcoding.

There's also the question of how to allocate multiple colors to a single player for persistent tags mode.
I have the feeling both of these will be wanting another button. Thankfully, we do have a spare... but I'm not sure how to make it intuitive.
> I might move the tag boxes down and in front of the portrait to play nice with the default window size (at full screen it doesn't really matter, because space). Perhaps a problem for another time, though.

Solid point.

ThislsAUsername:
> Edit: Looks like each CO gets 0.5 of what they would get as the prime. So just taking in 5 copies of Strong multiplies his effective power intake by 3. Taking in 11 copies multiplies his power intake by 6. Is this intended, or do we want to penalize using a giant stable of COs?

It's intended, yeah.
Looking at cart CO mechanics, adding a tag partner boosts your power in lots of ways with no drawbacks at all. So to me, it seems like "more COs = free power" is the most legit way to go. It's not like it's even close to balanced anyway.
AWBW tags are a lot tamer in most ways, but then go ahead and enable Colin/Colin tags so you can use Gold Rush 4 times in a row without even losing any defensive COP charge (since you can swap after activation).

For persistent tags, my plan is to divide all charge into [tag count + 1] shares, with the prime CO getting two.
This should help balance out the flexibility (and raw burst power of stacking COPs) of extra COs by making your powers come out later.

src/Engine/GameInput/SelectMetaAction.java
ThislsAUsername: The idea is:
menu->tap up->enter = end turn
menu->tap up twice->enter = swap CO

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThatIsAUsername: Oh. Yeah. If you start a tag game (so it saves the settings), then go to start a new game on that map without tags, it still shows the tag teams until you go into the CO-select screen. In fact, it shows the tags there until you back out, and then it fixes itself. That seems worth fixing.

ThislsAUsername:
Should I merge `master` back into here and try building out some base AI for tagging, or call that out of scope?

ThatIsAUsername:
> Should I merge `master` back into here and try building out some base AI for tagging, or call that out of scope?

I'd probably merge `master` into here just as good practice before merging. There are a couple comments above where you seem to consider additional tweaks as well; I'd recheck those at least, but I'd def say that AI modifications are beyond scope for this PR.

ThislsAUsername:
> ...It might not be such a bad idea to move the turn-end logic directly into the turn-end events, which could deduplicate code and remove the need for this hook. Hmm.

The actual logic I was thinking about is mostly already in `GameInstance.turn()`, and that seems like a pretty good home for it.
Could maybe use a better name, though.
It's also not really a good idea IMO to remove that handling from `MapController` for two reasons:
1) Handling turn protection/saving+quitting the game seems like it's outside the scope of what `GameEvent`s should be for
2) I would like to let all turn-end-listener-provided events churn through the animation queue before anybody thinks about starting the next turn, and that might be more awkward without the setup we have now.

> charge denial AWBW special case

This is an AWBW-exclusive thing. Cart tags let you charge the back CO while a front CO power is active, and I think I want it to work that way for persistent tags as well.
> I might move the tag boxes down and in front of the portrait to play nice with the default window size

I'm looking at this again, and it seems fine? Even with Patch, who's definitely the most "cluttered"
image

ThatIsAUsername:
>     1. Handling turn protection/saving+quitting the game seems like it's outside the scope of what `GameEvent`s should be for

I think I'd agree. I guess there's no reason you couldn't, probably, but it seems like more of a meta-control thing.

>     2. I would like to let all turn-end-listener-provided events churn through the animation queue before anybody thinks about starting the next turn, and that might be more awkward without the setup we have now.

Not sure what change you are suggesting here, or even what you mean specifically by "the setup".

> I'm looking at this again, and it seems fine? Even with Patch, who's definitely the most "cluttered"

Sure. Not suggesting to make a change now, but we can keep that option in mind if we implement CO team groupings and it increases the general clutter level.


Also, why does your "No CO" not have a portrait, but mine does?

ThislsAUsername:
> Not sure what change you are suggesting here, or even what you mean specifically by "the setup".

One way of moving away from "`MapController` handles turn switch logic" would be to do everything it does... in the turn end event/in some new associated events.
It would be technically feasible, I think, but turn init events would necessarily be pushed into the queue before everything from *end* turn finished chugging through.
"The setup" being "How our turn-switch logic works"
> Also, why does your "No CO" not have a portrait, but mine does?

Because I very cleverly forgot to commit those images?

ThatIsAUsername:
Sure, what could go wrong?

Issue #275, opened 2020-07-11
Slightly more unit info


ThislsAUsername:
We should show the materials and unit-cargo of units somewhere.
Maybe also their numerical weapon ranges?

ThatIsAUsername:
Also mobility. At minimum show max move distance, but ideally also have a way to see move cost, maybe a HUD display with all the terrain costs, or maybe adding footprint icons while choosing a destination to show the number of move points each tile takes.

PR #367, opened 2022-08-04
Kaiju Extras

ThislsAUsername:
I probably would have made some of these changes before, but you'd have said YAGNI

### DIFF
--- a/src/Engine/GameAction.java
+++ b/src/Engine/GameAction.java
@@ -7,7 +7,6 @@ import CommandingOfficers.Commander;
 import CommandingOfficers.CommanderAbility;
 import Engine.Combat.DamagePopup;
 import Engine.GameEvents.CommanderAbilityEvent;
-import Engine.GameEvents.ArmyDefeatEvent;
 import Engine.GameEvents.CreateUnitEvent;
 import Engine.GameEvents.GameEventQueue;
 import Engine.GameEvents.MassDamageEvent;
@@ -15,7 +14,6 @@ import Engine.GameEvents.ModifyFundsEvent;
 import Engine.GameEvents.SwapCOEvent;
 import Engine.GameEvents.TeleportEvent;
 import Engine.GameEvents.TurnEndEvent;
-import Engine.GameEvents.UnitDieEvent;
 import Engine.GameEvents.TeleportEvent.AnimationStyle;
 import Engine.GameEvents.TeleportEvent.CollisionOutcome;
 import Terrain.GameMap;
@@ -218,7 +216,8 @@ public abstract class GameAction
      * Creates a new `what` unit at location `where`, belonging to `who`. No funds will be taken. If `allowStomping` and a unit
      * is already present at `where`, the existing unit will be killed. If `unitIsReady`, then the new unit will be able to move immediately.
      */
-    public UnitSpawnAction(Commander who, UnitModel what, XYCoord where, CreateUnitEvent.AnimationStyle how, boolean allowStomping, boolean unitIsReady)
+    public UnitSpawnAction(Commander who, UnitModel what, XYCoord where, CreateUnitEvent.AnimationStyle how,
+        boolean allowStomping, boolean unitIsReady)
     {
       this.where = where;
       this.who = who;
@@ -253,17 +252,11 @@ public abstract class GameAction
           ArrayList<Unit> ary = new ArrayList<Unit>();
           ary.add(obstacle);
           boolean fatal = true;
-          MassDamageEvent mde = new MassDamageEvent(who, ary, obstacle.getHP()+1, fatal);
-          UnitDieEvent ude = new UnitDieEvent(obstacle);
+          MassDamageEvent mde = new MassDamageEvent(who, ary, obstacle.getHP() + 1, fatal);
           buildEvents.add(mde);
-          buildEvents.add(ude);
-
           // Poor sap died; Check if his CO lost the game. Stomping your own unit is silly, but won't cause a loss.
-          if( obstacle.CO.army.getUnits().size() == 1 && who != obstacle.CO )
-          {
-            ArmyDefeatEvent cde = new ArmyDefeatEvent(obstacle.CO.army);
-            buildEvents.add(cde);
-          }
+          final boolean canLose = who != obstacle.CO;
+          Utils.enqueueDeathEvent(obstacle, new XYCoord(obstacle), canLose, buildEvents);
         }
       }
       else
@@ -297,7 +290,7 @@ public abstract class GameAction
     {
       return null;
     }
-  } // ~UnitProductionAction
+  } // ~UnitSpawnAction
 
   // ===========  AbilityAction  =================================
   public static class AbilityAction extends GameAction
@@ -452,16 +445,8 @@ public abstract class GameAction
         ary.add(obstacle);
         boolean fatal = true;
         MassDamageEvent mde = new MassDamageEvent(unit.CO, ary, obstacle.getHP()+1, fatal);
-        UnitDieEvent ude = new UnitDieEvent(obstacle);
         subEvents.add(mde);
-        subEvents.add(ude);
-
-        // Poor sap died; Check if his CO lost the game.
-        if( obstacle.CO.army.getUnits().size() == 1 )
-        {
-          ArmyDefeatEvent cde = new ArmyDefeatEvent(obstacle.CO.army);
-          subEvents.add(cde);
-        }
+        Utils.enqueueDeathEvent(obstacle, subEvents);
       }
 
       // If our guy can't survive there, end him.
@@ -471,17 +456,8 @@ public abstract class GameAction
         ary.add(unit);
         boolean fatal = true;
         MassDamageEvent mde = new MassDamageEvent(unit.CO, ary, unit.getHP()+1, fatal);
-        UnitDieEvent ude = new UnitDieEvent(unit);
         subEvents.add(mde);
-        subEvents.add(ude);
-
-        // Our unit died; check if we are defeated.
-        if( unit.CO.units.size() == 1 )
-        {
-          // CO is out of units. Too bad.
-          ArmyDefeatEvent cde = new ArmyDefeatEvent(unit.CO.army);
-          subEvents.add(cde);
-        }
+        Utils.enqueueDeathEvent(unit, subEvents);
       }
       return subEvents;
     }
--- a/src/Engine/GameAction.java
+++ b/src/Engine/GameAction.java
@@ -163,7 +163,7 @@ public abstract class GameAction
       else
       {
         // We can't create this action. Leave the event queue empty.
-        System.out.println("WARNING! UnitProductionAction created with invalid arguments.");
+        System.out.println("WARNING! Building from a property failed.");
       }
       return buildEvents;
     }
@@ -262,7 +262,7 @@ public abstract class GameAction
       else
       {
         // We can't create this action. Leave the event queue empty.
-        System.out.println("WARNING! UnitProductionAction created with invalid arguments.");
+        System.out.println("WARNING! UnitSpawnAction cannot generate events.");
       }
       return buildEvents;
     }
--- a/src/Engine/GameEvents/CreateUnitEvent.java
+++ b/src/Engine/GameEvents/CreateUnitEvent.java
@@ -1,6 +1,7 @@
 package Engine.GameEvents;
 
 import CommandingOfficers.Commander;
+import Engine.Utils;
 import Engine.XYCoord;
 import Terrain.MapMaster;
 import UI.MapView;
@@ -13,6 +14,7 @@ public class CreateUnitEvent implements GameEvent
   private final Commander myCommander;
   private final Unit myNewUnit;
   private final XYCoord myBuildCoords;
+  private int myFudgeRadius = 0;
   private final AnimationStyle myAnimStyle;
   private final boolean myBoots;
 
@@ -25,6 +27,14 @@ public class CreateUnitEvent implements GameEvent
   public CreateUnitEvent(Commander commander, UnitModel model, XYCoord coords)
   {
     this(commander, model, coords, AnimationStyle.NONE, false, false);
+  }
+  /**
+   * Allows spawning the unit on a tile within a radius, in case the "correct" tile is blocked
+   */
+  public CreateUnitEvent(Commander commander, UnitModel model, XYCoord coords, AnimationStyle anim, boolean unitIsReady, int fudgeRadius)
+  {
+    this(commander, model, coords, anim, unitIsReady, false);
+    myFudgeRadius = fudgeRadius;
   }
   public CreateUnitEvent(Commander commander, UnitModel model, XYCoord coords, AnimationStyle anim, boolean unitIsReady, boolean allowStomping)
   {
@@ -60,7 +70,27 @@ public class CreateUnitEvent implements GameEvent
     if( null != myNewUnit )
     {
       myCommander.units.add(myNewUnit);
-      gameMap.addNewUnit(myNewUnit, myBuildCoords.xCoord, myBuildCoords.yCoord, myBoots);
+      if( myFudgeRadius < 1 || myBoots )
+        gameMap.addNewUnit(myNewUnit, myBuildCoords.xCoord, myBuildCoords.yCoord, myBoots);
+      else
+      {
+        boolean success = false;
+        // Check concentric rings outwards, to spawn as close to the original spot as possible
+        for( int radius = 0; !success && radius < myFudgeRadius; ++radius )
+          for( XYCoord xyc : Utils.findLocationsInRange(gameMap, myBuildCoords, radius, radius) )
+          {
+            Unit resident = gameMap.getResident(xyc);
+            if( resident == null )
+            {
+              gameMap.addNewUnit(myNewUnit, xyc.xCoord, xyc.yCoord, false);
+              success = true;
+              break;
+            }
+          }
+        if( !success )
+          System.out.println("Warning: Failed to find a place to spawn " + myNewUnit
+                            + " within " + myFudgeRadius + " tiles of " + myBuildCoords);
+      }
       myCommander.army.myView.revealFog(myNewUnit);
     }
     else
--- a/src/Engine/GameEvents/CreateUnitEvent.java
+++ b/src/Engine/GameEvents/CreateUnitEvent.java
@@ -76,7 +76,7 @@ public class CreateUnitEvent implements GameEvent
       {
         boolean success = false;
         // Check concentric rings outwards, to spawn as close to the original spot as possible
-        for( int radius = 0; !success && radius < myFudgeRadius; ++radius )
+        for( int radius = 0; !success && radius <= myFudgeRadius; ++radius )
           for( XYCoord xyc : Utils.findLocationsInRange(gameMap, myBuildCoords, radius, radius) )
           {
             Unit resident = gameMap.getResident(xyc);
--- a/src/Engine/GameEvents/CreateUnitEvent.java
+++ b/src/Engine/GameEvents/CreateUnitEvent.java
@@ -7,6 +7,7 @@ import Terrain.MapMaster;
 import UI.MapView;
 import UI.Art.Animation.GameAnimation;
 import Units.Unit;
+import Units.UnitContext;
 import Units.UnitModel;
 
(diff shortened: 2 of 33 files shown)


src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -590,15 +590,13 @@ public static int getAnimIndex()
   {
     // Calculate the sprite index to use.
     long thisTime = System.currentTimeMillis();
-    // Fun fact: casting long->int can produce negative numbers, for some reason.
-    return Math.abs((int) (thisTime / animIndexUpdateInterval));
+    return (int) ((thisTime / animIndexUpdateInterval) & Integer.MAX_VALUE);
```
ThatIsAUsername, line 593: Huh. Clever. Probably faster than the obvious modulus approach, too.

src/Engine/Utils.java
```diff
@@ -671,6 +675,15 @@ public static boolean enqueueMoveEvent(MapMaster gameMap, Unit unit, GamePath mo
...
       originalPathOK = false;
     }
+    if( unit.model.needsFuel ) // Check fuel.
+    {
+      int fuelBurn = movePath.getFuelCost(unit, gameMap);
+      if( fuelBurn > unit.fuel )
+      {
+        movePath.snipForFuel(gameMap, unit);
+        originalPathOK = false;
+      }
```
ThatIsAUsername, line 685: Are we ever calling `enqueueMoveEvent()` with paths that exceed the unit's range? I thought we just used this to handle the uncertainty of ambushes on FoW; fuel (and anything else we could know a priori) should be (and was, I thought) checked before this point.

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
```diff
@@ -150,7 +150,8 @@ public void performEvent(MapMaster gameMap)
       if( null != myNewUnit )
       {
         myCommander.army.money -= cost;
-        builder.materials -= 1;
+        if( builder.model.maxMaterials >= 0 )
+          builder.materials -= 1;
```
ThatIsAUsername, line 154: Why not use `hasMaterials()` here as well?

src/Units/UnitState.java
```diff
@@ -80,6 +80,11 @@ public void fire(WeaponModel weapon)
     }
   }
 
+  public boolean hasMaterials()
+  {
+    return materials > 0 || materials == model.maxMaterials;
```
ThatIsAUsername, line 85: Why `materials == model.maxMaterials`? Won't that result in `hasMaterials()` returning true for units that don't have any?

src/Engine/GameEvents/CreateUnitEvent.java
```diff
@@ -60,7 +70,27 @@ public void performEvent(MapMaster gameMap)
...
+        // Check concentric rings outwards, to spawn as close to the original spot as possible
+        for( int radius = 0; !success && radius <= myFudgeRadius; ++radius )
+          for( XYCoord xyc : Utils.findLocationsInRange(gameMap, myBuildCoords, radius, radius) )
+          {
+            Unit resident = gameMap.getResident(xyc);
+            if( resident == null )
+            {
+              gameMap.addNewUnit(myNewUnit, xyc.xCoord, xyc.yCoord, false);
+              success = true;
+              break;
```
ThatIsAUsername, line 87: Shouldn't this check for traversability as well?

src/Engine/GameAction.java
```diff
@@ -253,17 +252,11 @@ public GameEventQueue getEvents(MapMaster gameMap)
           ArrayList<Unit> ary = new ArrayList<Unit>();
           ary.add(obstacle);
           boolean fatal = true;
-          MassDamageEvent mde = new MassDamageEvent(who, ary, obstacle.getHP()+1, fatal);
-          UnitDieEvent ude = new UnitDieEvent(obstacle);
+          MassDamageEvent mde = new MassDamageEvent(who, ary, obstacle.getHP() + 1, fatal);
           buildEvents.add(mde);
-          buildEvents.add(ude);
```
ThatIsAUsername, line 259: If we no longer have a `UnitDieEvent` because mass damage covers it, we should remove it from the comment up at 236.

src/Engine/Utils.java
ThislsAUsername: That validation was in `MoveEvent`; I just moved most of it here so I can do terrible misdeeds with Kaiju movement (the preconditions will be satisfied by the time the `MoveEvent` *executes*, but not at the time it's created/queued).

There's also a test that fails if this validation isn't in the loop somewhere.

ThislsAUsername: The subject of whether that validation ought to happen at this point is harder to answer - the player probably can't violate anything here, but I think I'd prefer to not trust AIs with unbridled movement speed.

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
ThislsAUsername: This commit adds support for units to have unlimited materials via a max value of -1.
I don't want to decrement materials past that sentinel value.

src/Units/UnitState.java
ThislsAUsername: That's... true, actually. What are you doing giving material-using actions to unit types with 0 `maxMaterials`, though?

src/Engine/GameAction.java
ThislsAUsername: The death+defeat event(s) are added by `Utils.enqueueDeathEvent()`.
As far as I can tell, the comment is still correct.

src/Engine/GameEvents/CreateUnitEvent.java
ThislsAUsername: Uh, probably yes.
That does bring to my attention that the animation for "fuzzy spawns" can be wrong, though I'm not hugely worried about it?
I don't see an obvious way to guarantee (by construction) that the animation and actual event go to the same place.

You can kinda assume it'll work if you just update the destination coordinate after the animation, probably, but that turns a getter into a secret setter.
Recalculating should work, assuming the iteration ordering is the same.

Calculating the actual drop point at event creation time doesn't work because other events can happen between creation and execution.

src/Engine/Utils.java
```diff
@@ -681,16 +694,26 @@ public static boolean enqueueMoveEvent(MapMaster gameMap, Unit unit, GamePath mo
    * @param unit The unit who is to die.
    * @param eventQueue Will be given the new MoveEvent(s).
    */
-  public static void enqueueDeathEvent(Unit unit, GameEventQueue eventQueue)
+  public static void enqueueDeathEvent(Unit unit, XYCoord grave, boolean canLose, GameEventQueue eventQueue)
```
ThatIsAUsername, line 697: `canLose` is never used in this function. Though if we were stomping our own unit, wouldn't the new unit then exist and prevent a loss without needing a flag to prevent it? I don't know that we have any relevant cases to test this.

src/Engine/GameEvents/CreateUnitEvent.java
```diff
@@ -60,7 +71,29 @@ public void performEvent(MapMaster gameMap)
...
+          for( XYCoord xyc : Utils.findLocationsInRange(gameMap, myBuildCoords, radius, radius) )
+          {
+            Unit resident = gameMap.getResident(xyc);
+            if( resident == null && uc.moveType.canTraverse(gameMap.getEnvironment(xyc)) )
+            {
+              gameMap.addNewUnit(myNewUnit, xyc.xCoord, xyc.yCoord, false);
+              success = true;
+              break;
+            }
+          }
```
ThatIsAUsername, line 92: Don't we typically create Events for effects that have already been determined? E.g., TechDrop finds locations, generates the event, and we can know where they are gonna land. This is figuring out where to create a unit in the Event's `performEvent` call itself. It seems like that logic should be pre-event-creation.

src/Engine/UnitMods/UnitModifier.java
```diff
@@ -53,4 +55,7 @@ public interface UnitModifier extends Serializable
   void modifyCargoCapacity(UnitContext uc);
   void modifyActionList(UnitContext uc);
 
+  /** Convenience hook for StateTracker injection */
+  default void registerTrackers(GameInstance gi) {}
```
ThatIsAUsername, line 59: Should this do anything? It doesn't look like any subclasses populate the method either.

src/Units/UnitModelScheme.java
```diff
@@ -38,15 +41,22 @@ public final GameReadyModels getGameReadyModels()
   }
   protected abstract GameReadyModels buildGameReadyModels();
 
+  public void registerStateTrackers(GameInstance gi)
+  {
+    for( UnitModel um : gi.rules.unitModelScheme.getGameReadyModels().unitModels )
+      for( UnitModifier mod : um.getModifiers() )
+        mod.registerTrackers(gi);
+  }
```
ThatIsAUsername, line 49: Dare I ask why a `UnitModelScheme` would need to register state trackers? Isn't a scheme immutable? Why would it need to know what's going on in the game? Maybe I just haven't gotten there yet, but it feels like we are breaking some assumptions here.
One of which might be: Vanilla units in a `UnitModelScheme` not having modifiers.

Also, why `gi.rules.unitModelScheme.getGameReadyModels()` instead of just `getGameReadyModels()`?
For that matter, maybe this should be moved to `GameInstance.init()` or something.

src/Engine/Utils.java
ThislsAUsername: Goood point. Dunno how I missed that.
The new unit doesn't exist until the event finishes, while the `ArmyDefeatEvent` (once queued) doesn't care whether you've got units - you're just gonna lose.
I guess I probably should add a testcase or two.

src/Engine/GameEvents/CreateUnitEvent.java
ThislsAUsername: What I'm using it for is the resurrection thing - If your hovercraft or fancy fighter dies, it leaves the board entirely and shows up "near" its original location.
At the time that event is created (via listener), the unit will be dead... but because Kaiju step on things to kill them, it's not guaranteed the space you were on before is available.

src/Engine/UnitMods/UnitModifier.java
ThislsAUsername: It's used in the next PR.

src/Units/UnitModelScheme.java
ThislsAUsername: Sure. If a particular unit type needs to have extra information about the gamestate in order to handle its functionality, then it needs a `StateTracker` and can't depend on a CO or something to initialize it. The UMS knows about the unit and can initialize the tracker.
Schemes are immutable, yes. This isn't really for the *scheme* so much as the units in it. That's probably not obvious when there's no implementation, sorry.
> Vanilla units in a `UnitModelScheme` not having modifiers.

So uh. I have been planning for a while to introduce something like this. That's why the `UnitModel` supports its own list of modifiers even though COs aren't allowed to mess with it.
An example I've been thinking on is the lose-range-with-movement effect on the Bship. It *could* be a property on the weapon that every piece of code that touches it has to enforce, or it could be a `UnitModifier` and be handled the same way CO effects are.

Uhhhh, I originally had this code in a static function in another class, I think? Copy/paste error from there.
I'm not super opposed to pulling it up to `GameInstance.init()`, but I think the way it is now is also useful.
For example, all Kaiju types use the same tracker for tracking ability cooldowns - who registers that tracker? You could register it in each Kaiju unit definition, or you could just register it in the UMS.

src/Engine/Utils.java
ThislsAUsername: While we're talking deathy edge cases, consider:
Tank shoots my last unit
My last unit, on death, produces a new unit via event listener.
`ArmyDefeatEvent` still goes through and I lose.

This, I think, highlights the utility of pushing victory/defeat conditions out of event generation and into a bit of code that gets run when the event queue is empty.

src/Units/UnitState.java
ThatIsAUsername: And you need the `==` comparison for the -1/inf case, of course. I think I've grown towards using dedicated booleans instead of sentinel values just to avoid obtuse logic like this.

src/Engine/Utils.java
ThatIsAUsername: > the preconditions will be satisfied by the time the MoveEvent executes, but not at the time it's created/queued

Might have to unpack this one for me. Terrible misdeeds, eh? Also, if the preconditions aren't satisfied at creation time, won't this snip the route so it doesn't do the full move anyway?

> The subject of whether that validation ought to happen at this point is harder to answer
I don't think I object to moving the validation here from `MoveEvent` though. Now that it's the main entrypoint for creating moves, I think that makes sense.

src/Engine/Utils.java
ThatIsAUsername: Yeah, I can see the logic for that. Just check victory/defeat conditions after each action or something. Probably less efficient, because we know what conditions to check (and for whom) right now, but simpler.

src/Engine/Utils.java
ThislsAUsername: > Might have to unpack this one for me. Terrible misdeeds, eh?

Since I know I will be stomping away any unit on a tile I move through, I can build a `MoveEvent` onto or even from a tile that already has a different unit on it.
> Also, if the preconditions aren't satisfied at creation time, won't this snip the route so it doesn't do the full move anyway?

It would, if I was using this helper function. Which is why I moved the validation here - so I can circumvent it when I have other ways of ensuring the gamestate stays legal.

src/Units/UnitState.java
ThislsAUsername: You say boolean - can I raise you a `moveFuelBurn` integer (matching `idleFuelBurn`) that can be 0-1 (or go up to 2 to support DS Olaf or shenanigans units)?

src/Engine/GameEvents/CreateUnitEvent.java
ThatIsAUsername: If you are creating a respawn event due to a kaiju movement, wouldn't you know where the kaiju ends up, and be able to find a suitable landing pad for the crushee at event-creation time?

ThislsAUsername: If I was creating the respawn event inside of the movement logic... yeah, probably. It'd be a pain because edgecases, but yeah.
However, the resurrection logic is done by a listener for `UnitDieEvent` injecting the new unit creation into the event stream. I don't really want to force any code that might kill a unit to check for whether it resurrects.

src/Engine/Utils.java
ThatIsAUsername: Are you creating/enqueuing `MoveEvent`s somewhere besides this helper function? I'm not seeing where, if so. I'm not seeing it. Or is that just later?

src/Units/UnitState.java
ThatIsAUsername: > FuelBurn

Sounds like a separate discussion, but: Do we have units that burn fuel differently than the movement values? Depending on the use though, sure.

src/Engine/GameEvents/CreateUnitEvent.java
ThatIsAUsername: > You can kinda assume it'll work if you just update the destination coordinate after the animation, probably, but that turns a getter into a secret setter.

You mean, recalculating the spawn location in the animation right before actually drawing it, and then updating the event with that result as well? Wouldn't work for previews, of course, but again, maybe not the end of the world.

src/Engine/Utils.java
ThatIsAUsername: Could even get fancy and add flags to know which conditions to check. No units died? Don't need to check if we are out of units. No buildings captured? No need to check if it was an HQ, or if there are none left, etc.

Some conditions would be harder to check ex post facto though; if an HQ gets captured, and I have another one, it's not obvious I should lose without knowing that history.

src/Engine/Utils.java
ThislsAUsername: That's in the Kaiju movement function, yes.

src/Units/UnitState.java
ThislsAUsername: Fuel is burned 3 ways:
Drake powers
Idle burn (sea/air)
1 fuel per move point spent

Snow in Dual Strike increases fuel burn to 2 per move point spent.

src/Engine/Utils.java
ThislsAUsername: OTOH, we're already tracking all HQ locations for each `Army`, so that's probably not a hard check to write... though that would result in the occasional false positive if there's no HQ/Lab (with the current codebase)

We will probably want flags in the `GameRules` that determine whether those things are loss conditions, though.

src/Engine/GameEvents/MoveEvent.java
```diff
@@ -49,8 +49,7 @@ public void performEvent(MapMaster gameMap)
         gameMap.moveUnit(unit, endpoint.x, endpoint.y);
       unit.isTurnOver = true;
 
-      if( unit.model.needsFuel )
-        unit.fuel = Math.max(0, unit.fuel - fuelBurn); // Don't prevent zero-distance "moves" when out of fuel.
+      unit.fuel = Math.max(0, unit.fuel - fuelBurn);
```
ThatIsAUsername, line 52: I assume you verified this doesn't break zero-distance moves?

src/Units/UnitState.java
```diff
@@ -82,7 +82,7 @@ public void fire(WeaponModel weapon)
 
   public boolean hasMaterials()
   {
-    return materials > 0 || materials == model.maxMaterials;
+    return materials > 0 || !model.needsMaterials;
```
ThatIsAUsername, line 85: If a unit has no materials, and also doesn't need materials, then it `hasMaterials()`?

src/Engine/GameEvents/MoveEvent.java
ThislsAUsername: Seems fine to me, boss.

src/Units/UnitState.java
ThislsAUsername: Should it be `canBuild()`, then? That requires a little bit of logic to take you to "this cares about materials", and doesn't necessarily account for other uses we may have down the line.
Could call it `canUseMaterials()`, but that's roughly equivalent and not as nice to say.

ThatIsAUsername: I think I'd take `canBuild()` over `canUseMaterials()` since this function is also checking whether materials are present, not just whether they would be usable.

Or just change the logic so that it returns `false` if `model.needsMaterials` is `false`, which you'd have to do anyway for `canBuild()` to make sense as the name.

ThislsAUsername: Eh? `needsMaterials = false` means `canBuild()` should always return true.
If I don't need materials, then I can build regardless of whether I have any.

ThatIsAUsername: Do we have any units that produce things without materials? I tried to explore how materials are actually handled, and all I found is that carriers are broken on the KaijuOnly branch: `WARNING! UnitProduction Action created with invalid arguments.`

ThislsAUsername: Oh, goody. I guess I should make a test on that, then :/
...OTOH, it's not like we'll be touching this code again. What possibilities exist aside from "no materials", "finite materials", and "unlimited materials"?

I decided the Sky Carrier shouldn't have a materials limit (since there aren't any other limits like that in its source game), which is why I added this behavior.

ThislsAUsername: Uhhh, I dunno where you're getting that Carriers are broken.
I just tested on both `kaijuOnly` and `kaijuExtras`, and DoR Carriers build stuff just fine.

Also also, that error message only comes from trying to build from a property (or from `UnitSpawnAction`... I guess I should fix that)

ThatIsAUsername: Turns out I was just trying to build it with insufficient funds. Really need better visuals for that. Silly me.

As for possibilities; I think that covers the bases (though: what's the difference between none and infinity?). Also, we should probably show the current materials count in the unit preview overlay (unless it's infinite).

ThislsAUsername: None means you can't build anything, though I guess that's not really a different *case* from finite materials - it's just that the finite quantity is currently 0. Blame tired brains.
Infinity means you can build as much as you like.

Also, we totally have an issue for that already: #275

PR #368, opened 2022-08-04
Add Kaiju Wars units

ThislsAUsername:


src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+
+      KaijuStateTracker kaijuTracker = StateTracker.instance(gameMap.game, KaijuStateTracker.class);
+      // Setting the tracker state here feels wrong
+      if( longCooldown )
+        kaijuTracker.abilityUsedLong(actor, abilityKey);
+      else
+        kaijuTracker.abilityUsedShort(actor, abilityKey);
+
+      return eventSequence;
+    }
```
ThislsAUsername, line 358: Tracker information is properly part of the game state, I think... which makes setting that in `getEvents()` feel a little weird.
On the other hand, this *works* just fine since `getEvents()` is only ever called once per action and has explicit `MapMaster` access.
Also worth noting is that we do a lot of similar side-effect-y stuff in event listener hooks.

That said, I feel like maybe all changes to the game state proper should go through events, so nothing outside of explicit mutators has side effects.
This would have the nice property that you can reproduce any game state with the start and a stream of events - no listeners required at all.

There are two approaches I can see to achieve the above dream:
Add a bunch of events, one for each potential gamestate mutation
Add a generic event you can throw "whatever" in, i.e. my `LambdaEvent` from a while back (we could throw in a `name` field or something for easier debugging?)

Another reasonable approach would be to just accept that event listen hooks (and possibly `getEvents()` itself) can have side effects on the game state.
I would probably want some nomenclature change to make that more clear.
There is one notable drawback over the above two possibilities, as well: you don't get to order your changes arbitrarily - if I do something that I don't have an event for, it must happen before any events are executed.

src/Units/KaijuWarsKaiju.java
```diff
@@ -0,0 +1,850 @@
...
+        return null;
+
+      XYCoord end = unitPath.getEndCoord();
+      // Check for any hostile kaiju in range to study
+      for( Unit kaiju : kaijuAbilityTier.keySet() )
+      {
+        if( kaiju.CO.isEnemy(unit.CO) && Radar.PIERCING_VISION >= end.getDistance(unit) )
+        {
+          // TODO: This is definitely wrong
+          unit.CO.modifyAbilityPower(2);
```
ThatIsAUsername, line 561: Hmmm.

ThislsAUsername: See my above comment on this.
Though, there probably should be a `ModifyAbilityEnergy` event or such - for animation/audio cues down the line if nothing else.

src/Units/KaijuActions.java
ThatIsAUsername: > Tracker information is properly part of the game state, I think... which makes setting that in getEvents() feel a little weird.

Yeah - is there any reason you can't set the cooldown when the events are executed?

> That said, I feel like maybe all changes to the game state proper should go through events, so nothing outside of explicit mutators has side effects.

That is the direction I was trying to take things when events were initially introduced.

> Add a bunch of events, one for each potential gamestate mutation

I mean, that's basically what we do, right? There are a couple of combo ones, like I think the unit-healing event also handles the payment portion and maybe shouldn't.

> generic event

I think I'd lean against this by default. Not sure what would necessitate events with unknown effects.

> Another reasonable approach would be to just accept that event listen hooks (and possibly getEvents() itself) can have side effects on the game state.

Event listen hooks (by design) can inject new events into the stream - this happens after the listened event is executed though. Granted, a CO could potentially change internal state without an event based on receiving one, but I have less heartburn about that than about using `getEvents` to generate side effects.

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,783 @@
...
+    for (UnitModel um : airportModels)
+      kjwModels.unitModels.add(um);
+    for (UnitModel um : kaijuModels)
+      kjwModels.unitModels.add(um);
+    for (UnitModel um : extras)
+      kjwModels.unitModels.add(um);
+
+    // Multiplies all movement (and ranges) by N
+    for( UnitModel um : kjwModels.unitModels )
+      um.baseMovePower *= KaijuWarsWeapons.KAIJU_SCALE_FACTOR;
```
ThatIsAUsername, line 119: Why? Was this just for prototyping with different values before deciding that 1 is best and we don't need the variable after all? It is used as a multiplier for a number of seemingly-unrelated things.

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,783 @@
...
+        continue;
+
+      // We have a type that resurrects; find the type it rezzes to
+      for( UnitModel rezToModel : grms.unitModels )
+      {
+        if( !rezToModel.name.equals(umCast.resurrectsAs) )
+          continue;
+        rezzer.resurrectionTypeMap.put(um, rezToModel);
+        break;
+      }
```
ThatIsAUsername, line 146: Why make `resurrectAs` a `String` instead of a `UnitModel`? Avoid this loop and might streamline other things too.

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,783 @@
...
+   * fighter: +2 counter on empty(plains?)/water
+   *  - Adding
+   * ground: +2 ATK hills, +2 counter on forest/jungle
+   *  - Adding the first
+   * air: +3 move on airport
+   *  - Adding
+
+   * Hover Drives: Gives all ground +1 move and the ability to move on water (Super Z2 doesn't benefit)
+   * Ultimate Autobahn: Gives all ground +2 move if starting on road/bridge
+   */
```
ThatIsAUsername, line 191: These seem like dev notes: are they still relevant, or can we clean this up?

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,783 @@
...
+
+   * Freezer slows on attack
+   * Guncross Wing becomes Guncross Robot on death
+   * Super Z2 also lives on death, but just goes from 4 to 1 counter power
+   * Super Z2 can go on water, but is a land unit for all other purposes
+   * Kaputnik has a suicide attack
+   * Shark Jet can move (but not shoot) again after shooting
+   * Sky Carrier lets you build air units adjacent to it
+   * Cannon Of Boom cannot move normally, but can teleport between labs as its turn
+   */
```
ThatIsAUsername, line 215: More dev notes? Didn't you decide not to add Freezer and Cannon? At least according to prior comments.

src/Units/KaijuWarsUnits.java
ThislsAUsername: Yeah, it was just for prototyping.
I wasn't sure whether I liked 2 or 1 better, but it can probably go away now due to factors like "Kaiju exist now" and "Tactic bonuses make 2x less good"

src/Units/KaijuWarsUnits.java
ThislsAUsername: I was worried about whether that would work after save/load.

src/Units/KaijuWarsUnits.java
ThislsAUsername: They totally are dev notes, but they're relevant since it's the "source" info and might be relevant comparisons if KaijuWars itself ever gets updated and we want to match.

src/Units/KaijuWarsUnits.java
ThislsAUsername: Yeah. I did decide that. Same with Shark Jet.
Freezer (and Shark Jet) is complicated, and really bad in a world with counterattacks.
Cannon is hard to say how you would even implement it (I guess you could say "it's the only unit Ports build" instead of Kaiju, but ehhhh)

src/Units/KaijuActions.java
ThislsAUsername: > Yeah - is there any reason you can't set the cooldown when the events are executed?

There's no event for it, and no unique event that the cooldown tracker could listen to.
> That is the direction I was trying to take things when events were initially introduced.

Should that also include event listeners publishing their own events to make changes to the gamestate?
> I mean, that's basically what we do, right? There are a couple of combo ones, like I think the unit-healing event also handles the payment portion and maybe shouldn't.

Yeah. There's so much boilerplate, though.
> I think I'd lean against this by default. Not sure what would necessitate events with unknown effects.

Nothing *necessitates* it. But it'd be really nice to have a facility for doing one-off "change this number in a `StateTracker`" stuff without writing a whole new class. If it'd make you feel better, we could require a description for the `ToString()` for `LambdaEvent` so it's easier to debug.
If we didn't have this and did move to "listeners may only make changes by publishing more events", then every `StateTracker` would have to define at least one new type of event.
> Event listen hooks (by design) can inject new events into the stream - this happens after the listened event is executed though.

Two fine points here:
They initially could not inject new events, so a number of listeners were written without that assumption.
The event is executed before it is published - a listener "direct effect" takes place *immediately after* the event, while injected events will go to the end of the *event sequence* from the last `GameAction` taken.
> Granted, a CO could potentially change internal state without an event based on receiving one, but I have less heartburn about that than about using `getEvents` to generate side effects.

Yeah. I would still like to do something with nomenclature to make it clearer that that's kosher, though (or decide it isn't kosher).

src/Units/KaijuWarsWeapons.java
```diff
@@ -0,0 +1,264 @@
...
+    public double getDamage(KaijuWarsUnitModel defender)
+    {
+      int counterPower = defender.kaijuCounter;
+      if( !negateCounterBonuses )
+      {
+        if( isAirWeapon && defender.slowsAir )
+          counterPower += SLOW_BONUS;
+        if( !isAirWeapon && defender.slowsLand )
+          counterPower += SLOW_BONUS;
+        if( defender.resistsKaiju )
```
ThatIsAUsername: Why is it `resistsKaiju` instead of `slowsKaiju` like the others?

src/Units/KaijuWarsWeapons.java
```diff
@@ -0,0 +1,264 @@
...
+          counterPower += SLOW_BONUS;
+      }
+
+      int attack = 0;
+      if( defender.isAirUnit() )
+        attack = vsAir;
+      else
+        attack = vsLand;
```
ThatIsAUsername: Aww, missed a perfect opportunity for a ternary operator.

src/Units/KaijuWarsWeapons.java
```diff
@@ -0,0 +1,432 @@
...
+    int finalPower = attack - durability;
+    if( finalPower > -3 )
+      damage += finalPower * 10;
+    else if( finalPower > -8 )
+      damage = damage - 10 + finalPower * 5;
+    else
+      damage = 2 * (12 + finalPower);
+
+    return damage;
+  }
```
ThatIsAUsername, line 306: This is the simpler version? It's made out of edge cases.

src/Units/KaijuWarsWeapons.java
```diff
@@ -0,0 +1,432 @@
...
+
+    public Bomber()
+    {
+      super(2, 0);
+      isAirWeapon = true;
+    }
+  }
+
+  // experimental weapons
+  // TODO: slow logic?
```
ThatIsAUsername, line 164: Is this comment still relevant?

src/Units/KaijuWarsWeapons.java
ThislsAUsername: `resistsKaiju` means you're immune to special attacks other than the stomp (exists for OGR Platform and Kaiju themselves).
Slowing is just "-1 move if I get stepped on"

src/Units/KaijuWarsWeapons.java
ThislsAUsername: Good point. I don't actually remember what I was doing with that, I guess.

src/Units/KaijuWarsWeapons.java
ThislsAUsername: Technically, yes. The answer is "probably not, because I'm probably not implementing the unit"
I will rewrite it to just be a note to future me instead of a TODO, I think.

src/Units/KaijuWarsWeapons.java
ThislsAUsername: I considered it, FWIW.

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,783 @@
...
+    }
+  }
+
+  /*
+   * Units I do not plan to implement at this time
+
+   * Freezer (0-counter unit with 1 damage that slows targets by a tile; too complicated and bad)
+   * Food Cart (attracts kaiju within 3 tiles; not implementing kaiju path/targeting mechanics)
+   * Shark Jet (faster Fighter that moves after shooting; messy and either OP or UP)
+   * Cannon of Boom (immobile Rocket equivalent that teleports between labs; ew)
```
ThatIsAUsername, line 166: Freezer and Cannon are both defined in the unit set. Is this comment outdated?

ThislsAUsername: Their weapons are defined, but not the units themselves.

src/Units/KaijuWarsUnits.java
```diff
@@ -0,0 +1,791 @@
...
+  {
+    return super.toString() + "Kaiju Wars";
+  }
+
+  @Override
+  public String getIconicUnitName()
+  {
+    // TODO: Does the infrastructure around this need to change, or should this go away?
+    return "Infantry";
+  }
```
ThatIsAUsername, line 38: Is this TODO still relevant?

src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+      {
+        KaijuWarsUnitModel victimType = (KaijuWarsUnitModel) victim.model;
+        int counter = 0;
+        int stompDamage = victim.getHP();
+        if( victimType.isKaiju )
+        {
+          // When stomping another kaiju, at least one of us dies
+          counter += victim.getHP();
+          stompDamage = kaijuState.getHP();
+        }
```
ThatIsAUsername, line 142: So a 10 HP Kaiju can kill a 1HP Kaiju for 1 damage, and a 1 HP Kaiju can murder/suicide a 10 HP Kaiju?

src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+  public static class KaijuCrushAction extends GameAction
+  {
+    private final KaijuCrushFactory type;
+    private final GamePath movePath;
+    private final XYCoord waitLoc;
+    private final Unit actor;
+
+    public KaijuCrushAction(Unit unit, GamePath path, KaijuCrushFactory pType)
+    {
+      type = pType;
```
ThatIsAUsername, line 71: Why does the constructor need to accept the `KaijuCrushFactory` parameter? Is it just so we have the object to return from `getType()`? For that matter, why do all the other game actions just return `null` from getType()`?

src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+                && victimType.slowsAir )
+              --kaijuState.movePower;
+          }
+        }
+        final boolean isLethal = true;
+
+        ArrayList<Unit> stompable = new ArrayList<>();
+        stompable.add(victim);
+        crushEvents.add(new MassDamageEvent(kaijuState.CO, stompable, stompDamage, isLethal));
+        kaijuState.damageHP(counter);
```
ThatIsAUsername, line 178: Directly changing HP in `getEvents()`?

src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+        if( isCrushable(location) )
+        {
+          Environment oldEnvirons = location.getEnvironment();
+          Environment newEnvirons = Environment.getTile(oldEnvirons.terrainType.getBaseTerrain(), oldEnvirons.weatherType);
+          crushEvents.add(new MapChangeEvent(location.getCoordinates(), newEnvirons));
+          if( stomperType.regenOnBuildingKill )
+            crushEvents.add(new HealUnitEvent(kaijuState.unit, 2, null, true));
+          if( stomperType.chargeOnBuildingKill && location.isCaptureable() )
+            // TODO: This is definitely wrong
+            kaijuState.CO.modifyAbilityPower(1);
```
ThatIsAUsername, line 220: Any plans to change this?

src/Units/KaijuActions.java
ThatIsAUsername: > Should that also include event listeners publishing their own events to make changes to the gamestate

As in, do we need  meta-event for "I just created more events"? Probs not. Do you just mean "Should publishing events on receive be allowed"? Pretty sure we already do that.

> a listener "direct effect" takes place immediately after the event, while injected events will go to the end of the event sequence from the last GameAction taken.

Is that a problem? I guess you could just immediately execute/publish a sub-event instead of simply queuing it if it makes a big difference.

ThislsAUsername: > As in, do we need meta-event for "I just created more events"? Probs not. Do you just mean "Should publishing events on receive be allowed"? Pretty sure we already do that.

That's not where I was going with that.
To rephrase: Should event listeners be able to make changes to the gamestate _without_ publishing an event?
Current usecases of that are COs and all StateTrackers.
The former probably should move to publishing events for ability charge.
The latter probably shouldn't have to make a new type of event for any given change, which implies either "allow it" or "add a generic event".
> > a listener "direct effect" takes place immediately after the event, while injected events will go to the end of the event sequence from the last GameAction taken.
> 
> Is that a problem? I guess you could just immediately execute/publish a sub-event instead of simply queuing it if it makes a big difference.

It may or may not be a problem in the future. It isn't one right now. I just wanted to clarify.

src/Units/KaijuActions.java
ThislsAUsername: Nope.
10 HP can run over 1 HP kaiju for one damage.
1 HP kaiju can kill itself to deal 1 damage to a 10 HP kaiju.

src/Units/KaijuActions.java
ThislsAUsername: Yeah, it's just to have something to return for `getType()`.
`getType()` is used by AIs (for organizing) and by the game input machinery (for control flow and its `toString()`)

`GameAction`s that aren't taken by a unit (i.e. put into a `GameActionSet`) don't need one because they aren't constructed via `UnitActionFactory`.
`GameAction` types defined in a lifecycle file have a real `getType()` because they're produced in groups, and the caller has to have some way of categorizing/interrogating them.

I had considered renaming it `getActionType()`, but that provides no extra clarity.
`getActionFactory()` could work, I guess?

src/Units/KaijuActions.java
ThislsAUsername: `kaijuState` is a scratch struct.

src/Units/KaijuActions.java
ThislsAUsername: The correct fix is to make an event to change `Commander` energy levels.
So yes, but also that's been kicked down the road for quite a while now - I can fix it here, but I dunno if it's in-scope for this PR.

ThislsAUsername: This is part of why I started the discussion above about handling gamestate changes in event listeners.
Specifically:
If gamestate changes in listeners are a-ok, then I should add an event for this and leave the `Commander` charge listeners alone until we want to add animations.
If not, then I need to touch `Commander` and should probably hunt down all the rest of the instances and correct them - that sounds like a separate PR to me.

src/Units/KaijuWarsWeapons.java
ThislsAUsername: For future reference (have added comments, but they're not visible here):
The intent was relatively consistent addition/subtraction from a base value - the difference matters, not how big each side is.
The edge cases are to make it decay to 0 less abruptly:
55 is the base value. Any +1 in KW = +10 here.
The first two -1s give -10
The next 5 give -5
Any further give -2 (and I didn't lower-bound since the lowest difference is -8)

src/Units/KaijuActions.java
```diff
@@ -0,0 +1,943 @@
...
+      boolean isValid = true;
+      isValid &= null != actor && !actor.isTurnOver;
+      isValid &= (null != movePath) && (movePath.getPathLength() > 0);
+      isValid &= (null != gameMap);
+      isValid &= (actor.model instanceof KaijuUnitModel);
+
+      // Generate events.
+      if( isValid )
+      {
+        Utils.enqueueMoveEvent(gameMap, actor, movePath, crushEvents);
```
ThatIsAUsername, line 107: The comment above says that the MOVE comes last, but it looks like this enqueues a full-length move up front before enqueuing the individual CRUSHes.

src/Units/KaijuActions.java
ThatIsAUsername: Agree not in scope for this PR.

src/Units/KaijuWarsUnits.java
```diff
@@ -727,7 +727,7 @@ public static class GuncrossWing extends KaijuWarsUnitModel
     private static final int UNIT_COST = PREP_COST * 5;
     static final String REZ_TO_NAME = "Guncross Robot";
 
-    private static final MoveType moveType = new Hovercraft();
+    private static final MoveType moveType = new Flight();
```
ThatIsAUsername, line 736: Why are some move types created inside models and others are created once globally (~292-293) and shared?

src/Units/KaijuActions.java
ThatIsAUsername: Gotcha. Is there an advantage to passing it in here instead of referencing the type statically like we do elsewhere? Just curious why it's handled differently here.

src/Units/KaijuWarsUnits.java
ThislsAUsername: Yes, but probably out of scope for this PR

src/Units/KaijuActions.java
ThatIsAUsername: Ooooooor I guess we can add an energy event now.

src/Units/KaijuWarsUnits.java
ThislsAUsername: Because I'm a hack and it doesn't really matter anyway - the `UnitModel` constructor is gonna clone whatever it gets passed.
I guess we could clean that up at some point if we care about checking equality on `moveTypes` and/or saving some bytes in our save files.

src/Units/KaijuActions.java
ThislsAUsername: Making the energy event go all the places it wants to go isn't in scope, but adding the event itself seems fine.

src/Units/KaijuActions.java
ThislsAUsername: `KaijuAttackAction` is a generic parent for basically everything in here, so it doesn't know which class to point to.
So I'm left choosing between:
Make each action type implement that static linkage (function override -> 4 lines)
`KaijuAttackAction` handles it, and the type object is passed in (add params to a bunch of constructors -> +~7 characters on a bunch of lines)

I defaulted to the second, I think, since I was copying code from `TransformLifecycle` (which needs the specific type object so it actually shows the user the right thing).

It might also be possible to elide a few concrete `Action` classes entirely (and make `KaijuAttackAction` not abstract). I'd have to look a bit closer at it.

src/Units/KaijuActions.java
ThislsAUsername: The uh... the next line deletes it. I think that just got left in there when I did the iterative-move thing.

ThatIsAUsername: So it does. Sounds like two lines we don't need anymore then.

ThislsAUsername: Something seemed off to me about that, and I looked again - `enqueueMoveEvent()` actually has a side-effect in that it snips the path if it will run into something.
OTOH, Kaiju are freely able to run through units, so I'm not sure it's needed?

I dunno. New code that breaks all the rules is hard to fit into your framework of the rules that it breaks.

src/Units/KaijuActions.java
```diff
@@ -214,25 +250,45 @@ public static void enqueueSingleTileEvents(MapMaster gameMap, int startingMP, Un
         {
           Environment oldEnvirons = location.getEnvironment();
           Environment newEnvirons = Environment.getTile(oldEnvirons.terrainType.getBaseTerrain(), oldEnvirons.weatherType);
-          crushEvents.add(new MapChangeEvent(location.getCoordinates(), newEnvirons));
+          crush.events.add(new MapChangeEvent(location.getCoordinates(), newEnvirons));
+
+          // Overwrite existing popup, but don't spawn one
+          if( null != tileToast )
+            tileToast = "WRECK";
```
ThatIsAUsername, line 257: Why only flag terrain destruction if we are also killing a unit?

src/Units/KaijuActions.java
```diff
@@ -214,25 +250,45 @@ public static void enqueueSingleTileEvents(MapMaster gameMap, int startingMP, Un
...
+            crush.events.add(new CommanderEnergyChangeEvent(kaijuState.CO, 1));
+            tileToast = "ENERGY";
           }
+
+          // Icky, but I don't really wanna change this function signature
+          if( gameMap instanceof MapMaster )
+            if( Utils.willLoseFromLossOf((MapMaster) gameMap, location) )
+            {
+              crush.events.add(new ArmyDefeatEvent(location.getOwner().army));
+            }
```
ThatIsAUsername, line 276: This is icky. And you are changing the function signature; it accepted a `MapMaster` before.

src/Units/KaijuActions.java
ThislsAUsername: The popup looked bad to me at the end of the movement, and it's always at the end if you're wrecking a building (unless Rampage is on).
(specifically, it looked weird 'cause we have that down-arrow thing, which shows on top of the movement arrow+menu)
image

src/Units/KaijuActions.java
ThislsAUsername: I didn't change the function signature of `Utils.willLoseFromLossOf()`, and I don't really want to because it's used to decide when to end the game - you want complete information for that.

src/Engine/Utils.java
```diff
@@ -719,7 +719,7 @@ public static void enqueueDeathEvent(Unit unit, GameEventQueue eventQueue)
   /**
    * @return Whether the owner of this property will lose if the property is lost
    */
-  public static boolean willLoseFromLossOf(MapMaster map, MapLocation capLoc)
+  public static boolean willLoseFromLossOf(GameMap map, MapLocation capLoc)
```
ThatIsAUsername, line 722: I thought you didn't want to change the signature of this function, since we want to have a `MapMaster` if we are making win/loss decisions.

ThislsAUsername: I didn't.
However, I realized that there are valid usecases for checking that sort of question without immediately generating events - for example, this is probably more reliable than "is this a lab or HQ?" for AIs to make decisions on.

ThatIsAUsername: But there are places where this is invoked with just a `GameMap` - doesn't this mean that if a Kaiju crushes a lab, and doesn't know about any other labs, then in `KaijuActions::271` it'll decide there must not be any other labs and enqueue the defeat event?

ThislsAUsername: Yes and no.
Yes, that code will be executed. The event will be generated and queued.
The event will *not* be executed, because `getEvents()` requires a `MapMaster`, and the event queue generated by this function is only used by `getEvents()`.

On reflection, I don't think it's the job of a util function to make sure you do the right thing with its output - it should only be concerned with giving you the right answer to the question you asked.

ThatIsAUsername: OK, I may buy that. Or maybe I'm just tired of seeing the word "Kaiju". Either way, go ahead and merge this stuff if you're happy with it.

PR #369, opened 2022-08-04
Add some small tweaks so JakeMan can sorta participate

ThislsAUsername:


src/AI/JakeMan.java
```diff
@@ -691,14 +691,23 @@ private Map<XYCoord, UnitModel> queueUnitProductionActions(GameMap gameMap)
...
     wantedTypes.add(infantry);
-    wantedTypes.add(allTanks.get(0));
-    wantedTypes.add(copter);
-    wantedTypes.add(allTanks.get(1));
+    if( capPhase.capChains.isEmpty() )
+    {
+      wantedTypes.add(allTanks.get(0));
+      wantedTypes.add(copter);
+      wantedTypes.add(allTanks.get(1));
+    }
```
ThatIsAUsername, line 702: Only allowed to build infantry until all cap chains are finished? Sounds risky.

ThislsAUsername: Building only tanks so you can never tech up is even riskier.
...realistically, this AI isn't designed to work with this unit set at all and IDK if I even want to stretch it as much as this PR does.

ThatIsAUsername: Right. I mean, this substantially changes the behavior in the non-KW case as well, and (based on some very light testing) not for the better.

PR #370, opened 2022-10-23
CO pick grid

ThislsAUsername:
I don't think this is the final form I want - grouping by game as well seems like it will probably be pretty necessary.
However, this is functional... and adding more would require a lot of scope creep.

The main candidate feature to add to this in the short term would be "behave well when squished below normal size", but I don't rate that high priority.

### DIFF
--- a/src/UI/PlayerSetupColorFactionController.java
+++ b/src/UI/PlayerSetupColorFactionController.java
@@ -7,8 +7,11 @@ import UI.InputHandler.InputAction;
 public class PlayerSetupColorFactionController implements IController
 {
   private PlayerSetupInfo myPlayerInfo;
+  public final boolean shouldSelectMultiCO;
   OptionSelector colorSelector;
   OptionSelector factionSelector;
+  public int[] colorPicks, factionPicks;
+  public int tagIndex;
   public final String iconicUnitName;
 
   public PlayerSetupColorFactionController(PlayerSetupInfo playerInfo, String unitName)
@@ -16,11 +19,22 @@ public class PlayerSetupColorFactionController implements IController
     colorSelector = new OptionSelector(UIUtils.getCOColors().length);
     factionSelector = new OptionSelector(UIUtils.getFactions().length);
 
+    final int coCount = playerInfo.coList.size();
+    colorPicks   = new int[coCount];
+    factionPicks = new int[coCount];
+    for(int i = 0; i < coCount; ++i)
+    {
+      colorPicks[i]   = playerInfo.coList.get(i).color;
+      factionPicks[i] = playerInfo.coList.get(i).faction;
+    }
+
     // Start the selectors at the initial values.
     myPlayerInfo = playerInfo;
-    colorSelector.setSelectedOption(myPlayerInfo.coList.get(0).color);
-    factionSelector.setSelectedOption(myPlayerInfo.coList.get(0).faction);
-    
+    shouldSelectMultiCO = coCount > 1; // Don't need the TagMode since we only assign stuff for as many COs as we have
+    tagIndex = 0;
+    colorSelector.setSelectedOption(colorPicks[tagIndex]);
+    factionSelector.setSelectedOption(factionPicks[tagIndex]);
+
     iconicUnitName = unitName;
   }
 
@@ -31,10 +45,26 @@ public class PlayerSetupColorFactionController implements IController
     switch(action)
     {
       case SELECT:
-        // Apply change and return control.
-        myPlayerInfo.coList.get(0).color = colorSelector.getSelectionNormalized();
-        myPlayerInfo.coList.get(0).faction = factionSelector.getSelectionNormalized();
-        done = true;
+        colorPicks[tagIndex]   = colorSelector  .getSelectionNormalized();
+        factionPicks[tagIndex] = factionSelector.getSelectionNormalized();
+        ++tagIndex;
+        // If we have more tag COs, stick around and handle those...
+        if( tagIndex < colorPicks.length )
+        {
+          // Navigate the cursor to the current value of that index's position
+          // Is this good UX? Nobody knows!
+          colorSelector  .setSelectedOption(colorPicks  [tagIndex]);
+          factionSelector.setSelectedOption(factionPicks[tagIndex]);
+        }
+        else // Otherwise, apply change and return control.
+        {
+          for( int i = 0; i < colorPicks.length; ++i )
+          {
+            myPlayerInfo.coList.get(i).color   = colorPicks  [i];
+            myPlayerInfo.coList.get(i).faction = factionPicks[i];
+          }
+          done = true;
+        }
         break;
       case SEEK:
         // Flip units
@@ -49,8 +79,15 @@ public class PlayerSetupColorFactionController implements IController
         colorSelector.handleInput(action);
         break;
       case BACK:
-        // Don't apply changes; just return control.
-        done = true;
+        // Don't apply changes
+        --tagIndex;
+        if( tagIndex < 0 )
+          done = true;
+        else // If we're not backing out, point the cursor at the new-old selection
+        {
+          colorSelector  .setSelectedOption(colorPicks  [tagIndex]);
+          factionSelector.setSelectedOption(factionPicks[tagIndex]);
+        }
         break;
       default:
         // Do nothing.
--- a/src/UI/PlayerSetupColorFactionController.java
+++ b/src/UI/PlayerSetupColorFactionController.java
@@ -11,6 +11,7 @@ public class PlayerSetupColorFactionController implements IController
   OptionSelector colorSelector;
   OptionSelector factionSelector;
   public int[] colorPicks, factionPicks;
+  public boolean flipUnits;
   public int tagIndex;
   public final String iconicUnitName;
 
@@ -31,6 +32,7 @@ public class PlayerSetupColorFactionController implements IController
     // Start the selectors at the initial values.
     myPlayerInfo = playerInfo;
     shouldSelectMultiCO = coCount > 1; // Don't need the TagMode since we only assign stuff for as many COs as we have
+    flipUnits = playerInfo.flipUnits;
     tagIndex = 0;
     colorSelector.setSelectedOption(colorPicks[tagIndex]);
     factionSelector.setSelectedOption(factionPicks[tagIndex]);
@@ -63,12 +65,13 @@ public class PlayerSetupColorFactionController implements IController
             myPlayerInfo.coList.get(i).color   = colorPicks  [i];
             myPlayerInfo.coList.get(i).faction = factionPicks[i];
           }
+          myPlayerInfo.flipUnits = flipUnits;
           done = true;
         }
         break;
       case SEEK:
         // Flip units
-        myPlayerInfo.flipUnits = !myPlayerInfo.flipUnits;
+        flipUnits = !flipUnits;
         break;
       case UP:
       case DOWN:
@@ -107,6 +110,6 @@ public class PlayerSetupColorFactionController implements IController
 
   public boolean getShouldFlip()
   {
-    return myPlayerInfo.flipUnits;
+    return flipUnits;
   }
 }
--- a/src/UI/PlayerSetupColorFactionController.java
+++ b/src/UI/PlayerSetupColorFactionController.java
@@ -18,8 +18,8 @@ public class PlayerSetupColorFactionController implements IController
 
     // Start the selectors at the initial values.
     myPlayerInfo = playerInfo;
-    colorSelector.setSelectedOption(myPlayerInfo.currentColor);
-    factionSelector.setSelectedOption(myPlayerInfo.currentFaction);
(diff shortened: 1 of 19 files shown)


ThatIsAUsername:
Initial impressions:
- Not a huge fan of the "pick NO_CO to exit" mechanic - unnatural and not easily discoverable. It's also odd to select an option in the CO grid to confirm CO selections and exit.
- Having the BACK button simply remove a CO is confusing as well. If I enter CO selection, realize I don't want to change anything, and try to back out, it just removes my current CO. If I enter CO selection on accident, remember I can't hit back to exit, and hit enter instead, it adds a second CO to my list. Could I get the hang of this? Sure, but it's not intuitive. The whole operation of the CO selection screen changes if you hit AWBW tags, and this is pretty disorienting.
- There is no obvious indication that you can select more than one CO until you actually do.
- Suggestions:
  - Add a fat "CONFIRM" button or similar on the right; moving all the way to the right off a row of COs drops you onto that, and hit that to exit. Moving left from that puts you back on the grid.
  - Don't jump the cursor around after each CO selection.
  - Show the "empty slot" of the CO being selected, with the face changing as you move around the CO grid.

Actually, a possibly better (but more different scheme) could be:
- Start with the cursor on the top row where the selected CO's go, on the first slot.
- Selecting a slot on the top row drops you onto the CO grid. Selecting back cancels CO selection.
- Selecting a CO in the CO grid populates the current slot and moves the cursor focus back up. Selecting back moves the cursor back up without selecting a CO.
- If a CO is selected, create a new blank slot to the right (if an empty slot doesn't exist) and place the cursor there.
- Include the confirm/exit button on the top row at the far right.

src/UI/UIUtils.java
```diff
@@ -260,8 +270,53 @@ public static COSpriteSpec instance(Commander co)
         spec = new COSpriteSpec(co.faction, co.myColor);
       return spec;
     }
+
+    /**
+     * Attempts to pull the key color for this canon faction from loaded resources.
+     * <p>Falls back on Salmon
```
ThatIsAUsername, line 276: We have a Salmon?

src/UI/UIUtils.java
```diff
@@ -260,8 +270,53 @@ public static COSpriteSpec instance(Commander co)
...
+      for( Entry<Color, String> pair : paletteNames.entrySet() )
+      {
+        if( pair.getValue().equalsIgnoreCase(col) )
+        {
+          spec = new COSpriteSpec(facInst, pair.getKey());
+        }
+      }
+
+      if( null == spec )
+        spec = new COSpriteSpec(facInst, Color.PINK);
```
ThatIsAUsername, line 293: We should probably create a custom color constant (a la `PURPLE`) for this to call it `ROSE` instead to avoid confusion with the external pink.
For that matter, `PURPLE` should probably be renamed to `VIOLET` since that's what we actually call it.

src/UI/PlayerSetupInfo.java
```diff
@@ -17,61 +15,74 @@
...
 public class PlayerSetupInfo
 {
-  public ArrayList<Integer> coList = new ArrayList<>();
-  public int currentColor;
-  public int currentFaction;
+  public static class CODeets
+  {
+    int co, color, faction;
+  }
+  public ArrayList<CODeets> coList = new ArrayList<>();
```
ThatIsAUsername, line 22: Did we just not support saving and loading of multi-CO teams before this?

src/UI/PlayerSetupInfo.java
```diff
@@ -17,61 +15,74 @@
...
+    for( int i = 0; i < coCount; ++i )
+      if( s.length > si )
+        coList.get(i).faction = Integer.valueOf(s[si++]) % availableFactions.length;
+
+    if( s.length > si )
+      flipUnits = Boolean.valueOf(s[si++]);
+    if( s.length > si )
+      currentTeam = Integer.valueOf(s[si++]);
+    if( s.length > si )
+      currentAi = Integer.valueOf(s[si++]) % availableAis.length;
```
ThatIsAUsername, line 85: So each member of a tag team can have their own art and color? Having everything change color when you hit `SWAP` is... strange, to say the least. I'd probably prefer to have one color/faction per player for standard tag teams (though I know you have ideas for other tag modes).

src/UI/PlayerSetupColorFactionController.java
```diff
@@ -31,10 +45,26 @@ public boolean handleInput(InputAction action)
...
-        myPlayerInfo.coList.get(0).faction = factionSelector.getSelectionNormalized();
-        done = true;
+        colorPicks[tagIndex]   = colorSelector  .getSelectionNormalized();
+        factionPicks[tagIndex] = factionSelector.getSelectionNormalized();
+        ++tagIndex;
+        // If we have more tag COs, stick around and handle those...
+        if( tagIndex < colorPicks.length )
+        {
+          // Navigate the cursor to the current value of that index's position
+          // Is this good UX? Nobody knows!
```
ThatIsAUsername, line 57: It's as good a starting point as any. I'm not sure what I'd do differently on this one.

src/UI/PlayerSetupColorFactionController.java
```diff
@@ -63,12 +65,13 @@ public boolean handleInput(InputAction action)
             myPlayerInfo.coList.get(i).color   = colorPicks  [i];
             myPlayerInfo.coList.get(i).faction = factionPicks[i];
           }
+          myPlayerInfo.flipUnits = flipUnits;
```
ThatIsAUsername, line 68: Oh, snap. It is not clear at all from the UI that this is a change that will carry into the game.
I sorta like having it be based on start position (because that's simple and sensible and transparent), but uh. I guess.
Maybe use the old way as the default (if you aren't already)?
Also maybe improve the tooltip to clarify this is an in-game change, not just so you can see the sprite both ways.

src/UI/PlayerSetupInfo.java
ThislsAUsername: Nah, I just aggregated the CO data into a class.
What was supported was "Each player has N COs, 1 color, and 1 faction", but now it's N of each.

src/UI/PlayerSetupColorFactionController.java
ThislsAUsername: I feel like the facing being different in the main player setup screen is a solid hint, but there's probably more I can do.

ThatIsAUsername: It is, but it's a hint you only see after you've made the change. The wording tweak feels like enough to make it more obvious.

ThatIsAUsername:
This works pretty well; Initial impressions (having yet to look at the code):

Nitpicks:
Possibly put "ADD" in the empty block on the top row, or just dotted-line the box? Though really it's clear enough as is.
Once a CO is selected, I'd bump the focus on the top bar over to the new blank slot (though possibly not if re-selecting a CO that isn't at the end of the row? Have to play with that).
Selecting too many COs goes behind the tooltips... and then off the screen. I'd probs start compressing/overlapping the faces, maybe even to avoid tooltip.

Complaints:
Hitting "back" when focus is down in the grid should probably just jump back to the top bar, not exit to the previous menu altogether. Back out one layer at a time.
It's annoying to remove a CO once added. Tracking down "No CO" down in the "MISC" section is unintuitive and cumbersome. Since we have `E` as a button now, maybe use that to insta-kick an already-selected CO? I know that's supposed to be `VIEWMODE` though. Maybe even swap-em (`Q` to kick, `E` to see CO details) if that feels more natural? Obviously add a top-right tooltip for the extra control.
Can probably compress the tool-tip text too, in light of the above space considerations, e.g., `Q - Kick CO` & `E - CO Info`

Altogether this already feels a lot better than the previous version; good stuff.

ThislsAUsername:
> Nitpicks: Possibly put "ADD" in the empty block on the top row, or just dotted-line the box? Though really it's clear enough as is.

I think I had a + sign in there at some point. I can probably bring that back.
> Once a CO is selected, I'd bump the focus on the top bar over to the new blank slot (though possibly not if re-selecting a CO that isn't at the end of the row? Have to play with that).

Wha? You can't add a new CO without focusing the empty one yourself.
I'm not a super fan of how nothing on the top bar changes when you're in the submenu, but I dunno what a good indicator would be. A dark patch behind your portrait/color selected?
> Selecting too many COs goes behind the tooltips... and then off the screen. I'd probs start compressing/overlapping the faces, maybe even to avoid tooltip.

I dunno if I'm too worried about that - 2 COs fits, and you can make the screen bigger if you want to do silly things.
> Complaints: Hitting "back" when focus is down in the grid should probably just jump back to the top bar, not exit to the previous menu altogether. Back out one layer at a time.

I intended to do that, and brainfarted. Whoops.
> It's annoying to remove a CO once added. Tracking down "No CO" down in the "MISC" section is unintuitive and cumbersome.

Agreed, but we're running low on buttons. I could put it back in the start of each row, but that's also weird.
Could give it its own "faction", so it's always separate?
> Since we have `E` as a button now, maybe use that to insta-kick an already-selected CO? I know that's supposed to be `VIEWMODE` though. Maybe even swap-em (`Q` to kick, `E` to see CO details) if that feels more natural?

But then we don't have a button to change filter stuff later on.
...we *can* just keep adding buttons for quite a while yet, but it has to stop at some point.
> Obviously add a top-right tooltip for the extra control. Can probably compress the tool-tip text too, in light of the above space considerations, e.g., `Q - Kick CO` & `E - CO Info`

Sounds good, provided the previous comment is addressed.
> Altogether this already feels a lot better than the previous version; good stuff.

Thanks.

src/UI/PlayerSetupCommanderController.java
```diff
@@ -99,40 +176,36 @@ public boolean handleInput(InputAction action)
...
         }
-        else
+        else // Yes
         {
-          if( tagCmdrList.size() == 0 || noCmdr != selectedCO )
-            tagCmdrList.add(selectedCO);
+          amPickingTagIndex = true;
+
+          final int selTagIndex = tagIndexSelector.getSelectionNormalized();
+          // This index should never be too high, because of the structure of handleTagChoiceInput()
```
ThatIsAUsername, line 191: What constitutes "too high"?

ThislsAUsername: Out of the array bounds.
It's documented extra-good (and by that I mean "not"), but `tagIndexSelector` goes up to 1 more than `tagCmdrList`'s size, since you have to be able to hit "done".

ThatIsAUsername:
> Wha? You can't add a new CO without focusing the empty one yourself.

Not sure I understand. That's true at the moment. No reason it has to stay that way.

> I'm not a super fan of how nothing on the top bar changes when you're in the submenu, but I dunno what a good indicator would be. A dark patch behind your portrait/color selected?

Dark overlay on top of the top bar (except the current empty slot we are trying to fill? Or just collapse the whole thing up off the screen.

> But then we don't have a button to change filter stuff later on.

Right. Forgot about filters. Hm. Really prefer to avoid the "NO CO" nonsense. Maybe that button can swap the filter when in the grid and kick when not? The contexts of those actions don't overlap. Might be terrible. Might not.

ThislsAUsername:
> > Wha? You can't add a new CO without focusing the empty one yourself.
> 
> Not sure I understand. That's true at the moment. No reason it has to stay that way.

There is surely a reason, and his name is "buttons".
I guess we could go back to "picking a CO always adds it", but then the bimodal-ness isn't super useful anymore.
> Right. Forgot about filters. Hm. Really prefer to avoid the "NO CO" nonsense. Maybe that button can swap the filter when in the grid and kick when not? The contexts of those actions don't overlap. Might be terrible. Might not.

Agreed.
That seems tolerable, if a little prone to muscle memory trolling.

ThatIsAUsername:
> > > Wha? You can't add a new CO without focusing the empty one yourself.
> > 
> > 
> > Not sure I understand. That's true at the moment. No reason it has to stay that way.
> 
> There is surely a reason, and his name is "buttons". I guess we could go back to "picking a CO always adds it", but then the bimodal-ness isn't super useful anymore.

I think we are somehow talking past each other here. I am simply describing an automatic focus-change after selecting a CO, which should not involve any extra buttons. E.g., after opening the CO select screen:
- RIGHT to focus on the '+'
- ENTER to focus the selection grid
- ENTER to select a CO. The selected CO is placed where the '+' currently is. The '+' is accordingly moved over a slot. The focus jumps back up to the top bar, **focusing the CO that was just selected**.
What if we replace that bold text with **focusing the '+' again**? This would make adding another CO take one less button press, and put the focus closer to DONE as well. These are probably the most common next two actions, and this focus-change benefits both. It is absolutely not a mission-critical change though.

ThatIsAUsername:
Hah, hitting "E" on the plus sign takes you to the CO info page for NoCO. Classic.

ThatIsAUsername:
Looks good. A couple comments:
1) It's still odd to allow multiple different colors/sprites on the same team, but I guess users can do what they want.
2) Hitting 'E' on the plus sign brings up the "no CO" bio.
3) Starting a match currently results in victory for all players.

PR #371, opened 2022-10-30
Teletiles

ThislsAUsername:
I dunno why I hadn't done this sooner.


res/map/Blizzard_Battle.map
```diff
@@ -11,3 +11,26 @@
...
+1, rockets,   14, 12
+1, md_tank,   15, 4
+1, infantry,  15, 9
+1, infantry,  16, 1
+1, infantry,  16, 2
+1, tank,      16, 7
+1, md_tank,   16, 11
+1, infantry,  17, 9
+1, rockets,   18, 12
+
```
ThatIsAUsername, line 36: Gotta say, was not expecting this random addition to the teletiles terrain-type addition pull request.

Do we also need to create a separate "predeployed" map category now?

src/Terrain/TerrainType.java
```diff
@@ -158,6 +160,11 @@ private TerrainType(int attributeFlags, int defense, String name, TerrainType ba
   private static final String ROAD_NAME = "ROAD";
   public static final TerrainType ROAD = new TerrainType( ROAD_FLAGS, ROAD_DEFENSE, ROAD_NAME, SHOAL );
 
+  private static final int TELETILE_FLAGS = UNWEATHERABLE;
+  private static final int TELETILE_DEFENSE = 0;
+  private static final String TELETILE_NAME = "TELETILE";
+  public static final TerrainType TELETILE = new TerrainType( TELETILE_FLAGS, TELETILE_DEFENSE, TELETILE_NAME, null );
+
```
ThatIsAUsername, line 167: Excellent. Mag tubes are definitely what we were still missing. Except these are better, since they work on air and sea units as well.

All we need now is a sample map that contains them.

res/map/Blizzard_Battle.map
ThislsAUsername: Blizzard Battle had been converted by an earlier conversion script, so it had teletiles in it... and it's a map designed around the predeployed scenario (it's from the AW1 campaign).
I was fixing the map anyway, so...

Map categories in general would be good, probably - seems to me that the best option is just to let the user make a folder structure and mirror that in the UI.

src/Terrain/TerrainType.java
ThislsAUsername: By your command.

ThatIsAUsername: It is... Glorious. Better than I could have imagined.

ThatIsAUsername:
Good stuff. Merge when ready.

PR #372, opened 2022-10-31
AIs should probably not move ally units

ThislsAUsername:
Ran into this during testing. Was very confused.
I'm not sure this is catching all the cases, or that it's the right solution.


ThatIsAUsername:
Good stuff. Nice catch.

ThatIsAUsername:
I don't think I had any issues with this one, so merge whenever.

PR #377, opened 2023-02-03
Movement fixes

ThislsAUsername:
I should probably playtest more often.

### DIFF
--- a/src/Engine/Utils.java
+++ b/src/Engine/Utils.java
@@ -92,14 +92,13 @@ public class Utils
   {
     ArrayList<XYCoord> locations = findLocationsInRange(map, moveLoc, 1);
     ArrayList<XYCoord> dropoffLocations = new ArrayList<XYCoord>();
-    final FloodFillFunctor cargoMoveType = new UnitContext(cargo).calculateMoveType().getUnitMoveFunctor(cargo, false, false);
-    if( MoveType.IMPASSABLE > cargoMoveType.getTransitionCost(map, moveLoc, moveLoc) )
+    final MoveType cargoMoveType = new UnitContext(cargo).calculateMoveType();
+    if( cargoMoveType.canTraverse(map.getEnvironment(moveLoc)) )
       for( XYCoord loc : locations )
       {
         // Add any location that is empty and supports movement of the cargo unit.
         if( (map.isLocationEmpty(loc) || map.getLocation(loc).getResident() == transport)
-            && MoveType.IMPASSABLE > cargoMoveType.getTransitionCost(map, loc, loc)
-            && cargoMoveType.canEnd(map, loc) )
+            && MoveType.IMPASSABLE > cargoMoveType.getMoveCost(map.getEnvironment(loc.xCoord, loc.yCoord)) )
         {
           dropoffLocations.add(loc);
         }
--- a/src/Engine/Utils.java
+++ b/src/Engine/Utils.java
@@ -98,7 +98,7 @@ public class Utils
       {
         // Add any location that is empty and supports movement of the cargo unit.
         if( (map.isLocationEmpty(loc) || map.getLocation(loc).getResident() == transport)
-            && MoveType.IMPASSABLE > cargoMoveType.getMoveCost(map.getEnvironment(loc.xCoord, loc.yCoord)) )
+            && cargoMoveType.canTraverse(map.getEnvironment(loc.xCoord, loc.yCoord)) )
         {
           dropoffLocations.add(loc);
         }
--- a/src/Engine/Utils.java
+++ b/src/Engine/Utils.java
@@ -92,13 +92,14 @@ public class Utils
   {
     ArrayList<XYCoord> locations = findLocationsInRange(map, moveLoc, 1);
     ArrayList<XYCoord> dropoffLocations = new ArrayList<XYCoord>();
-    final MoveType cargoMoveType = new UnitContext(cargo).calculateMoveType();
-    if( cargoMoveType.canTraverse(map.getEnvironment(moveLoc)) )
+    final FloodFillFunctor cargoMoveType = new UnitContext(cargo).calculateMoveType().getUnitMoveFunctor(cargo, false, false);
+    if( MoveType.IMPASSABLE > cargoMoveType.getTransitionCost(map, moveLoc, moveLoc) )
       for( XYCoord loc : locations )
       {
         // Add any location that is empty and supports movement of the cargo unit.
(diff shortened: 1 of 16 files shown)


src/Engine/Utils.java
```diff
@@ -92,13 +92,14 @@ public static ArrayList<XYCoord> findUnloadLocations(GameMap map, Unit transport
   {
     ArrayList<XYCoord> locations = findLocationsInRange(map, moveLoc, 1);
     ArrayList<XYCoord> dropoffLocations = new ArrayList<XYCoord>();
-    final MoveType cargoMoveType = new UnitContext(cargo).calculateMoveType();
-    if( cargoMoveType.canTraverse(map.getEnvironment(moveLoc)) )
+    final FloodFillFunctor cargoMoveType = new UnitContext(cargo).calculateMoveType().getUnitMoveFunctor(cargo, false, false);
+    if( MoveType.IMPASSABLE > cargoMoveType.getTransitionCost(map, moveLoc, moveLoc) )
```
ThatIsAUsername, line 96: This new code is not as idiomatic as what it replaces.
I'm also not sure why this change is needed? We already couldn't drop across invalid transitions (e.g. a lander dropping an inf onto land via a cliff).
Also don't like having two different classes responsible for movement. Is there a way to use `MoveType` as the interface and have it use `FloodFillFunctor` under the hood to simplify usage?

src/UI/Art/SpriteArtist/SpriteMapView.java
```diff
@@ -103,7 +104,13 @@ public SpriteMapView(GameInstance game)
...
     mapViewWidth = SpriteLibrary.baseSpriteSize * game.gameMap.mapWidth;
     mapViewHeight = SpriteLibrary.baseSpriteSize * game.gameMap.mapHeight;
-    SpriteOptions.setScreenDimensions(mapViewWidth * SpriteOptions.getDrawScale(), mapViewHeight * SpriteOptions.getDrawScale());
+
+    if( !Driver.getInstance().isMaximized() )
+    {
+      SpriteOptions.setScreenDimensions( mapViewWidth  * SpriteOptions.getDrawScale(),
+                                         mapViewHeight * SpriteOptions.getDrawScale() );
+      Driver.getInstance().updateView();
+    }
```
ThatIsAUsername, line 113: Amazing. Completely unrelated to unit movement, but amazing.

How'd you track this one down? It looks like the behavior should be mostly the same in the non-maximized case, which is incidentally also where it would go crazy. I'm assuming `Driver.updateView()` was indirectly invoking `SpriteOptions.setScreenDimensions()` or something silly?

ThislsAUsername: Oh. The maximized check is just so starting a map with the game maximized doesn't un-maximize the game on you (I noticed this behavior seeming to bother some users and never bothered to write it down because I'm a silly goose)

I probably bundled it and the fix into one commit because I was testing the changes together.

ThislsAUsername: The real fix, for what it's worth, is that all of our `IView`s don't end up calling `Driver.updateView()`.
That function invokes the API to resize the view, which is a questionable plan when you just got an update *from* the API.

I don't know why it wasn't broken earlier, but mostly I just looked at the code and decided that seemed like the right thing to tweak.

src/Engine/Utils.java
ThislsAUsername: You're right.
The issue I was fixing is "You can unload units onto Teletiles"
The meat of this change is in the `cargoMoveType.canEnd(map, loc)` check below. I looked at the logic of `canTraverse()`, and it doesn't check whether it's a valid tile to end a move on.... and I assumed I shouldn't touch that.
`FFF.canEnd()` checks explicitly for Teletile and for there being a unit present.
Looking at the current usecases for `canTraverse()` (Wally's "usable area", teleport/create's valid destination check, unload), I think it's a good idea to add `canEnd()`'s logic in there.

I'm thinking maybe `MoveType` can function as an FFF with `includeOccupied` and `canTravelThroughEnemies` set to false?
Or maybe there's a better idiom for how we handle `includeOccupied` and `canTravelThroughEnemies`, like having them as optional parameters for `MoveType` and removing FFF entirely.

I guess for this PR, I'll shuffle that logic into `canTraverse()` and then tackle that refactor later? Or I can just do it here, probably.

src/UI/Art/SpriteArtist/SpriteMapView.java
ThatIsAUsername: The resize jankiness has been an issue for... ever, I think (at least on Linux). So good job tracking it down.

ThislsAUsername: Huh. I think I only started noticing it after I got a second monitor, and I never had it on Windows.
I think I did get the game to *crash* on Windows by fiddling with the resizing, but that only ever happened once.

Thanks.

src/Engine/Utils.java
ThatIsAUsername: The new change is indeed more idiomatic, but it unfortunately prevents all movement across teletiles. [Hah](https://github.com/ThislsAUsername/DefendPeace/pull/377#issue-1569182892)

ThislsAUsername: I absolutely did test it, though - I don't see this bug on my end.

ThatIsAUsername: Huh. On second inspection, you are correct. I think I am guilty of attempting code review after midnight.
Though in my defense, a plain and simple reading of MoveType.java lends itself to that interpretation:
```java
  /** Returns whether the unit can travel in the specified environment. */
  public boolean canTraverse(Environment tile)
  {
    if( tile.terrainType == TerrainType.TELETILE )
      return false;
    return getMoveCost(tile) < IMPASSABLE;
  }
```
Which makes me wonder just what this code is doing, if it doesn't actually relate to a unit's ability to traverse through tiles. Things seem to have gotten muddy somewhere.

ThislsAUsername: `canTraverse()` has basically been the check used for "can I put this unit here?". The movement calc code doesn't use it because it needs to handle the precise movecost anyway.
So... yeah, I'd agree it's a little muddy.
Should I rename it to `canEnd()` to match the FFF's name for the concept?

ThatIsAUsername: Yes, or maybe `canStandOn()` or `canEndMovement/TurnOn()` or something that will be less misleading.

ThatIsAUsername:
Sweet, looks good, let's ship it.

PR #379, opened 2023-03-05
Scale crush counterattacks with HP

ThislsAUsername:


src/Units/UnitState.java
```diff
@@ -138,14 +138,14 @@ public int damageHP(double damage, boolean allowOverkill)
    * When healing, rounds health up to a whole HP (e.g. 2.5 + 2 = 4.5 -> 5.0)
    * @return the change in HP
    */
-  public int alterHP(int change)
+  public int alterHP(double change)
```
ThatIsAUsername, line 141: I feel like we had this whole health-vs-HP discussion already once before. HP is supposed to be non-fractional, and I'm not thrilled with the idea of making it so.
Why not just round to the nearest HP before creating the mass damage event? I get that sometimes this would negate (or accentuate) damage, but this way introduces changes to code that is exposed to other unit sets.

ThislsAUsername: That's a good point, and I agree (which is why this is a separate PR)
I really don't like the idea of rounding.

Thought: add a new function, `alterHealthPercent()` to serve this usecase?

ThatIsAUsername: Wouldn't `alterHealthPercent()` just be identical to `DamageHP()` (except for possibly allowing bidirectional adjustments)?

ThislsAUsername: I'd give it an int argument (out of 100 instead of 10), and it would be non-lethal by default.

If you're not down with that, I can just close the PR, though.

ThatIsAUsername: And then `MassDamageEvent` use the new call? Still not sure what benefit that would provide over `DamageHP`, since `DamageHP` already accepts a double argument.
I mean, what you are doing definitely breaks established precedent, and I'm still not sure why you need to do it to accomplish your current goals, but if I was worried about polluting the game with superfluosities, I should have stopped you before Kaiju became a thing. Just try not to upset your online fan club with collateral breaking changes I guess.

ThislsAUsername: `DamageHP` allows killing and doesn't allow healing. I don't think I want to change either of those properties, lol
 - Sidenote: I just realized `MassDamageEvent` will happily pipe healing into `damageHP()` when that will just cause an exception. So that's neat.

`AlterHP` doesn't kill, but does round up when healing

It'd be intuitive IMO for a function that takes fractional HP inputs to not round up.
The new `alterHealthPercent()` would therefore give us an endpoint to accurately model an annoying inconsistency in AW2 - Hawke's healing doesn't round fractional HP up. (None of AWBW's healing does either, heh)

src/Units/UnitState.java
```diff
@@ -134,26 +134,68 @@ public int damageHP(double damage, boolean allowOverkill)
...
+      final int capHP = Math.max(oldHP, UnitModel.MAXIMUM_HP);
+      // Apply the cap as needed
+      final int newHP = Math.min(capHP, oldHP + changeHP);
+      // Figure out whether that reduces our healing
+      realPercentChange = Math.min(realPercentChange, healthFromHP(newHP) - health);
+    }
+
+    health = Math.max(1, health + realPercentChange);
+
+    return getHP() - oldHP;
```
ThatIsAUsername, line 198: A bit strange to take in a percent and return an HP value, no? Like writing a function that accepts degrees and return radians by default.

ThatIsAUsername:
Cool. Now that we have three separate ways to modify a Unit's health, I'm guaranteed never to know which one to use. I might lobby at this point to just always use health (0-100) behind the scenes and only render to HP (0-10) for display just to simplify things, but hey, as long as it works I guess.

src/Units/UnitState.java
ThislsAUsername: The displayed HP is what matters in most cases, though.
Agreed that it's a little weird.

ThislsAUsername:
> Cool. Now that we have three separate ways to modify a Unit's health, I'm guaranteed never to know which one to use.

They have documentation. That's what it's there for.
Should I add the general cases where each should be used to the docs?
> I might lobby at this point to just always use health (0-100) behind the scenes and only render to HP (0-10) for display just to simplify things, but hey, as long as it works I guess.

The HP is used for all damage calcs (barring a new exception I've learned since I wrote the code - AW1 counterattacks don't round up to HP)
IDK. I don't hate the idea, and it would clean up a lot of floating point stuff that doesn't need to be floating point.

ThatIsAUsername:
> Should I add the general cases where each should be used to the docs?

Might be helpful for future me if it isn't too much of a pain, but I'm not gonna sweat it.

ThatIsAUsername:
Looks marvy.

PR #381, opened 2023-05-02
Sprite fixes

ThislsAUsername:
We should have move animations for all the AWBW units now, so that's neat.


ThatIsAUsername:
I don't understand. It looks like you added a ton of duplicate sprites for factions that should just use `basis.txt`.
Also many of the sprites you added are marked as executable, so that's a thing.

ThislsAUsername:
White Nova has a bunch of units that are hue-shifted on their palette from normal.
Teal Galaxy has different colors for their gun metal/treads; I did, however, forget to flip their sprites.
The rest is just pixel tweaks/adding infantry

ThatIsAUsername:
> White Nova has a bunch of units that are hue-shifted on their palette from normal. Teal Galaxy has different colors for their gun metal/treads; I did, however, forget to flip their sprites. The rest is just pixel tweaks/adding infantry

Seems like the Nova thing could be fixed by shifting the palette (and maybe just the infs) instead of replacing all the other units. And I don't think tread is noticeable at all (and being nonstandard is more likely an oversight than a purposeful choice). Plus they still got unique (but still not noticeably) sprites for things like the BBomb, which has neither treads nor gun.

So again, I'm not sure I see the need, but I'm also not sure the added footprint is big enough to care, so I guess I don't object to merging. Animated movement will be cool to have. Poor Thorn is kinda left out though.

ThislsAUsername:
Shifting the WN palette (and thus the infantry) would mean the lightest infantry color would need to be hardcoded, which IMO looks terrible in other colors (as illustrated in the first image here).
image
image
And not all the other units are replaced - it's mostly air units (with Mds and subs).

As for Galaxy, the tread being a different color is IMO a noticeable difference - the palette of the units is overall a lot cooler with the blue-gray than with the normal brown. And they're the only faction AFAIK that replaces the accent colors while keeping the shades/shapes/etc distinct.
Their animations are out of sync, too, so it's harder to use a visual diff tool to find duplicates.
And I missed "blackboat" vs "bboat" too, even on the second pass. Derp.

Sorry about Thorn. I'm a lot better at stealing art than making it, heh.

ThatIsAUsername:
All right then, good 'nuff.

And I wouldn't expect you to fix Thorn (though I don't suppose adding a basis.txt to Thorn would make it "work" if you don't look too closely?)

Anyway, go ahead and hit the button if there aren't any more minor oversights to remedy.

ThislsAUsername:
> add basis to Thorn

Nah. The logic will only grab movement sprites from the same folder as the base sprite.
We could change that, but it'd look pretty weird.

There are probably more, but I don't feel like looking again.

PR #382, opened 2023-05-24
Add luck damage

ThislsAUsername:
Ohey, information hiding and display choices

### DIFF
--- a/src/CommandingOfficers/Venge.java
+++ b/src/CommandingOfficers/Venge.java
@@ -8,6 +8,7 @@ import Engine.Combat.StrikeParams;
 import Engine.Combat.StrikeParams.BattleParams;
 import Engine.Combat.BattleSummary;
 import Engine.Combat.CombatContext;
+import Engine.Combat.CombatContext.CalcType;
 import Engine.GameEvents.GameEventQueue;
 import Engine.GameEvents.MassDamageEvent;
 import Engine.UnitActionLifecycles.JoinLifecycle.JoinEvent;
@@ -133,6 +134,12 @@ public class Venge extends Commander
 
         instance.defender = instance.attacker;
         instance.attacker = minion;
+
+        // Since we're swapping the combatants, we also need to swap the prediction polarity
+        if( instance.calcType == CalcType.PESSIMISTIC )
+          instance.calcType = CalcType.OPTIMISTIC;
+        else if( instance.calcType == CalcType.OPTIMISTIC )
+          instance.calcType = CalcType.PESSIMISTIC;
       }
     }
   }
--- a/src/Engine/Combat/StrikeParams.java
+++ b/src/Engine/Combat/StrikeParams.java
@@ -18,7 +18,7 @@ public class StrikeParams
 {
   public static BattleParams buildBattleParams(
       UnitContext attacker, UnitContext defender,
-      GameMap gameMap, int battleRange,
+      CombatContext combatContext,
       boolean isCounter)
   {
     List<UnitModifier> aMods = new ArrayList<>(attacker.mods);
@@ -26,8 +26,10 @@ public class StrikeParams
 
     BattleParams params = new BattleParams(
         buildStrikeParams(attacker, defender.model,
-                          gameMap, battleRange, defender.coord,
+                          combatContext.gameMap, combatContext.battleRange,
+                          defender.coord,
                           isCounter),
+        combatContext,
         defender);
 
     for( UnitModifier mod : aMods )
@@ -46,8 +48,8 @@ public class StrikeParams
     List<UnitModifier> aMods = new ArrayList<>(attacker.mods);
 
     StrikeParams params = new StrikeParams(
-        attacker, gameMap,
-        battleRange, target,
+        attacker, gameMap, battleRange,
+        target,
         ( null == attacker.weapon ) ? 0 : attacker.weapon.getDamage(defender),
         isCounter);
 
@@ -67,6 +69,9 @@ public class StrikeParams
   public double baseDamage;
   public double attackerHP;
   public double attackPower;
+  public int luckBase = 0; // Luck value if you roll 0
+  public int luckRolled = 0; // The number we plug into the RNG for luck damage
+  public int luckRolledBad = 0; // The number we plug into the RNG for negative luck damage
   public final boolean isCounter;
 
   public double defenderHP = 0;
@@ -88,6 +93,7 @@ public class StrikeParams
     this.baseDamage = baseDamage;
     this.attackerHP = attacker.getHP();
     this.attackPower = attacker.attackPower;
+    this.luckRolled = attacker.CO.luck;
     this.isCounter = isCounter;
 
     this.targetCoord = target;
@@ -102,6 +108,7 @@ public class StrikeParams
     this.baseDamage = other.baseDamage;
     this.attackerHP = other.attackerHP;
     this.attackPower = other.attackPower;
+    this.luckRolled = other.luckRolled;
     this.isCounter = other.isCounter;
 
     this.targetCoord = other.targetCoord;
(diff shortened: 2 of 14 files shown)


src/Engine/GameInstance.java
```diff
@@ -283,7 +291,18 @@ public void endGame()
       army.deInitForGame(this);
     }
   }
-  
+
+  /**
+   * Gives you a random number between 0 (inclusive) and the input (exclusive)
+   * <p>Call this in GameAction.getEvents()/event constructors and nowhere else
```
ThatIsAUsername, line 297: Could we enforce this access, e.g. by making it a protected static method of the `GameEvent` base class or something?
I can easily see someone misusing this call; as a member of `GameInstance`, it looks like it should be more broadly used, javadoc comment notwithstanding.

src/CommandingOfficers/Venge.java
```diff
@@ -133,6 +134,12 @@ public void changeCombatContext(CombatContext instance)
 
         instance.defender = instance.attacker;
         instance.attacker = minion;
+
+        // Since we're swapping the combatants, we also need to swap the prediction polarity
+        if( instance.calcType == CalcType.PESSIMISTIC )
+          instance.calcType = CalcType.OPTIMISTIC;
+        else if( instance.calcType == CalcType.OPTIMISTIC )
+          instance.calcType = CalcType.PESSIMISTIC;
```
ThatIsAUsername, line 142: Swapping attacker/defender is a common enough use-case - would it make more sense to create a `CombatContext` function that handles all this nonsense so that individual COs don't have to worry about it?

src/UI/InputOptionsController.java
```diff
@@ -19,11 +19,25 @@ public String toString()
       return super.toString().replace("_", " ");
     }
   }
+  /**
+   *  Compare with {@link CombatContext.CalcType}
+   */
+  public static enum InputCalcType
+  {
+    NO_LUCK, PESSIMISTIC, OPTIMISTIC;
```
ThatIsAUsername, line 27: Huh. It seems odd to make this user-configurable instead of providing the range. Yes, that's harder UI, so barring that, the `NO_LUCK` answer is probably the right one, especially since you can't change this setting in-game.

Admittedly, the "just let users configure what they want, it doesn't hurt anything" argument sounds compelling, but that can quickly turn into having too much user-configurable stuff that nobody (or even worse: a vocal 5% of users) actually uses.

src/Engine/Combat/StrikeParams.java
```diff
@@ -67,6 +69,9 @@ public static StrikeParams buildStrikeParams(
   public double baseDamage;
   public double attackerHP;
   public double attackPower;
+  public int luckBase = 0; // Luck value if you roll 0
+  public int luckRolled = 0; // The number we plug into the RNG for luck damage
+  public int luckRolledBad = 0; // The number we plug into the RNG for negative luck damage
```
ThatIsAUsername, line 74: Why are there three different combat luck variables? It doesn't look like `luckBase` is ever set to anything besides zero (so why have a separate variable), and I also don't understand the need for separate positive/negative luck variables, and calculating positive/negative separately.

Is separate positive/negative calculations a detail of how the carts do it? It seems like a single roll over a defined range is simpler and gives a more correct distribution over the range.

src/Engine/UnitActionLifecycles/BattleLifecycle.java
```diff
@@ -174,7 +176,9 @@ public Collection<DamagePopup> getDamagePopups(GameMap map)
     {
       ArrayList<DamagePopup> output = new ArrayList<DamagePopup>();
 
-      BattleSummary summary = CombatEngine.simulateBattleResults(attacker, defender, map, movePath);
+      // This feels like an uggo hack, but this would be kind of out of place as a parameter
+      CalcType calcType = CalcType.values()[InputOptionsController.damagePreviewTypeOption.getSelectedObject().ordinal()];
+      BattleSummary summary = CombatEngine.simulateBattleResults(attacker, defender, map, movePath, calcType);
```
ThatIsAUsername, line 181: Uggo for sure, but at least understandable here.

src/Units/UnitState.java
```diff
@@ -162,7 +162,7 @@ public int alterHP(int change, boolean allowOver)
 
     health = Math.max(1, health + healthFromHP(realChange));
     // Round HP up, if healing
-    if( change > 0 )
+    if( change >= 0 )
       health = healthFromHP(getHP());
```
ThatIsAUsername, line 166: OK, I'm curious. What significant effect does this change have?

src/Engine/GameInstance.java
ThislsAUsername: That seems sensible. That does mean the RNG state itself has to be public, which could lead to stuff getting out of sync if you mess with it.
Of the two (exposing private state, exposing a method you shouldn't call), I am not sure which is the greater evil.
The "exposing state" probably could be done via getters/etc, but ugh.

Ideally, any solution should allow powers to directly access RNs as well, since Sturm at least needs one.

I suppose it's possible to provide public access via injection, but it seems like a lot of boilerplate: https://stackoverflow.com/a/14226229

src/CommandingOfficers/Venge.java
ThislsAUsername: I am mildly amused that we went from condemning this as weird jank to "a common usecase".
I do like the DRY you propose, though.

src/UI/InputOptionsController.java
ThislsAUsername: Wargroove was criticized by a fair number of people for having the displayed damage not be the minimum, but the middle of the range. That's probably mostly coming from people expecting AW things, sure, but we're more like AW than WG is.

IMO it makes most sense to show you the *actual* minimum by default, rather than something that only *might* be the minimum. Yes, this will give Sonja/Flak weird damage numbers... and I think that's a good thing.
(You have correctly identified cart precedent, though, and that's also a fair argument)

Agreed, but unfortunately the annoying "vocal 5%" is part of the dev team in this case.
I strongly considered grabbing e.g. VIEWMODE to toggle between damage prediction types while previewing attack damage, but I didn't want to just throw it in willy-nilly without thinking through what I want the end-state of the controls to look like.

src/Engine/Combat/StrikeParams.java
ThislsAUsername: `luckBase` is for providing a one-RN solution to positive-negative luck ranges, as you propose. AW1 Sonja will use it ~~and it's not uncommon in Nell rebalance ideas~~, though AW1 has some other silly subtleties to luck.
AW2+ uses a second RN for negative luck.

Two-RN makes extreme results less common; YMMV on whether that's a positive, heh.

src/Units/UnitState.java
ThislsAUsername: I wanted to use `alterHP(0)` to round up the health to the maximum for that HP value.
I could write a new function instead, but we already have too many health functions.

src/CommandingOfficers/Venge.java
ThatIsAUsername: I mean, it kinda is weird jank, but it's also cart-approved weird jank.

src/Engine/GameInstance.java
ThatIsAUsername: If it's a protected method of `GameEvent`, then it's not public, so I'm not sure what you mean. It would at least be less public than having it as a public member of `GameInstance`, which is maybe about as public and accessible as you can get.

src/Engine/Combat/StrikeParams.java
ThatIsAUsername: Two-RN reduces the impact of luck overall and turns the quoted stats into lies (though I guess AW doesn't tell the player luck percentages). Like how the two-RN FE system shifts the probability so that unlikely events happen even less often than they should.

src/Engine/GameInstance.java
ThislsAUsername: The *method* wouldn't be public, sure, but the members of `GameInstance` that it acts on would then have to be public.
Getting an RN from the game state seems like a pretty obviously gamestate-altering thing, and the only reason I want to restrict calls to the method is... that it alters the game state.
The API is about as clean as it can get.
Exposing the RNG-related state publicly for everyone to mess with opens you up to new categories of bad things (vs "I advanced the game's RNG accidentally")

If you mean to also push the RNG *state* into `GameEvent`, that's a non-starter IMO since I want the RNG to be consistent for a given seed and set of actions.

src/Engine/Combat/StrikeParams.java
ThislsAUsername: It does reduce the impact of luck. That's arguably a good thing.

The quoted stats aren't lies. You really can reach those damage upper and lower bounds. It's just less likely to happen than it would be under a uniform distribution.

Two-RN in FE does indeed lie to you about the true chance of a hit connecting. I'd argue that's good for gamefeel because humans are dumb, but that's not really relevant here.
AW luck damage percent doesn't directly implicate the percent *chance* of e.g. finishing off a unit - it has to line up with your normal damage vs the enemy's HP/chip damage taken in a specific way for it to matter for that at all.
AW does like to throw "%" on anything related to damage (since health is natively a percentage), and that's a little deceptive, but I wasn't planning to duplicate that issue.

src/Engine/GameInstance.java
ThatIsAUsername: Hmm, fair enough. Plus we've tried to push all action into events, so anything that would want to pull a random number should be in an event anyway.

src/Engine/Combat/StrikeParams.java
ThatIsAUsername: OK, that all tracks. I'd probably leave a comment to explain that the multiple variables are to support different games' different luck implementations, because it's unexpected without that context.

ThatIsAUsername:
Sweet. Now it'll be ever so slightly annoying to merge in other COs who use the swapping mechanic, since we'll have to remember to switch them to use the common utility. Great.
I think this is looking good now.

ThislsAUsername:
Most of the API is at least a little different in various ways, and I did a fair number of questionable things originally.
Suffice to say, that's nothing new.

PR #383, opened 2023-06-02
Some bugs

ThislsAUsername:
Squished

### DIFF
--- a/src/Engine/UnitActionLifecycles/UnloadLifecycle.java
+++ b/src/Engine/UnitActionLifecycles/UnloadLifecycle.java
@@ -3,6 +3,7 @@ package Engine.UnitActionLifecycles;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import Engine.GameAction;
 import Engine.GameActionSet;
@@ -18,7 +19,6 @@ import Terrain.MapMaster;
 import UI.MapView;
 import UI.Art.Animation.GameAnimation;
 import Units.Unit;
-import Units.UnitContext;
 
 public abstract class UnloadLifecycle
 {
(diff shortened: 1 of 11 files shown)


src/Engine/UnitActionLifecycles/UnloadLifecycle.java
```diff
@@ -30,28 +30,32 @@ public static class UnloadFactory extends UnitActionFactory
...
+      if( actor.model.unloadExclusionTerrain.contains(map.getEnvironment(moveLocation).terrainType) )
+        return null;
 
-          if( !unloadActions.isEmpty() )
-          {
-            return new GameActionSet(unloadActions);
-          }
+      ArrayList<GameAction> unloadActions = new ArrayList<GameAction>();
+
+      // TODO: Consider using ignoreResident for dropoff points as well
```
ThatIsAUsername, line 44: Only if the space is in fog of war I'd think, but you already handled that case. Why use `ignoreResident` for dropoff points otherwise?

src/Units/DoRUnits.java
```diff
@@ -672,6 +672,7 @@ public GunboatModel()
       baseCargoCapacity = 1;
       carryableMask = TROOP;
       carryableExclusionMask = TANK; // Can't carry Bikes
+      unloadExclusionTerrain.add(TerrainType.BRIDGE);
```
ThatIsAUsername, line 675: So only DoR units are unable to unload on bridges? Your dedication to accurately recreating the nuances of cart differences is impressive.

src/Engine/UnitActionLifecycles/UnloadLifecycle.java
ThislsAUsername: AI logic - "I can unload this infantry on that mountain if I move my other infantry that's already there"

src/Units/DoRUnits.java
ThislsAUsername: Trilogy Landers can't get on bridges. I think they also couldn't if you hacked them onto the bridge, but I only care about implementation details like that if they tickle my fancy.
I do think I probably want to make DoR-style bridges a separate terrain type eventually, but "We don't have a new terrain type" is probably not a bug.
(And they will totally be called `DRAWBRIDGE`s)

And thanks, I think.

src/Engine/UnitActionLifecycles/UnloadLifecycle.java
ThatIsAUsername: Ah, right.

ThatIsAUsername:
This looks ready to go. Does GitHub not have an "approve" button anywhere? I guess text works too.

ThislsAUsername:
I think you can give an "official" yae/nae if you "start a review", but yes, text works.

PR #384, opened 2023-06-02
FloodFillFunctor refactor

ThislsAUsername:
Where "refactor" means "delete"


src/Units/MoveTypes/MoveType.java
```diff
@@ -68,9 +69,27 @@ public int getMoveCost(Weathers weather, TerrainType terrain)
...
+      return MoveType.IMPASSABLE;
+
+    // note to self: extend this if we ever support moving to non-adjacent tiles
+    int cost = getMoveCost(map.getEnvironment(to.xCoord, to.yCoord));
+
+    // if there is an enemy unit in that space
+    if( !canTravelThroughEnemies
+        && (map.getLocation(to).getResident() != null)
+        && mover.CO.isEnemy(map.getLocation(to).getResident().CO) )
+      cost = MoveType.IMPASSABLE;
```
ThislsAUsername, line 90: This code will currently blow up IFF there are units in the way, you pass in a `null` unit, and `canTravelThroughEnemies` is unset.
Should a `null` unit input simply count *everyone* as enemies?
Is the above just an error condition we should avoid?

I've been sort of treating it as a malformed call, but IDK if that's the right idea since that seems like something that *could* blow up at some point.
Maybe the argument should be an allegiance thing, so you input your team number and there's no option for `null`?

src/Engine/Utils.java
```diff
@@ -267,7 +269,7 @@ public static GamePath findShortestPath(XYCoord start, Unit unit, int x, int y,
    * @param y Final Y coordinate
    * @param map The map to search over.
    */
-  public static GamePath findShortestPath(XYCoord start, FloodFillFunctor fff, int initialFillPower, int x, int y, GameMap map)
+  public static GamePath findShortestPath(XYCoord start, Unit unit, MoveType fff, int initialFillPower, int x, int y, GameMap map, boolean canTravelThroughEnemies)
```
ThatIsAUsername, line 272: Here and a few others - why require both a `Unit` and a `MoveType` when every unit has a `MoveType` already?

src/Units/MoveTypes/MoveType.java
ThatIsAUsername: Are there any cases where a `null` unit could be passed in? Though it might make more sense to pass in a `Commander` instead, since we only use it for a friendliness check.

I think it's fair to either treat `null` as an exception (in which case, raise something) or by treating everyone as enemies.

src/Units/MoveTypes/MoveTypeFey.java
```diff
@@ -26,29 +24,21 @@ public MoveType clone()
...
+    // Mandatory sanity check
+    if( !map.isLocationValid(to) )
+      return IMPASSABLE;
+
+    Environment endEnv = map.getEnvironment(to);
+    // Fey units cannot enter enemy-controlled spaces.
+    if( null != mover && mover.model.healableHabs.contains(endEnv.terrainType)
+        && endEnv.terrainType.isCapturable()
+        && mover.CO.isEnemy(map.getLocation(to).getOwner()) )
+      return IMPASSABLE;
```
ThatIsAUsername, line 39: Do we need the `healableHabs` check? I would think the `isEnemy` check would be sufficient. Seems odd to ban units from entering a space just because it could heal them (if friendly).

src/Units/MoveTypes/MoveType.java
ThislsAUsername: Yes. AI might (will) want to calculate paths without having an actual unit yet. That said, it's also easy to make an ephemeral `Unit` for the call.
I think I agree; passing in an `Army` makes it intuitive that a `null` says you're enemies with everyone.

src/Engine/Utils.java
ThislsAUsername: `Unit` can be null, as discussed above.

src/Engine/Utils.java
```diff
@@ -113,20 +113,25 @@ public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gam
...
+   * @param team Affiliation of the unit that's moving; may be null
+   * @param fff Determines the cost to move from one tile to another.
+   * @param initialFillPower How much move juice we start with
    * @param gameMap The map to search over.
+   * @param includeOccupiedSpaces Whether to include places that have units in them already; generally true if canTravelThroughEnemies is
+   * @param canTravelThroughEnemies Whether to allow pathing through enemies
    * @return The list of XYCoords in gameMap reachable by the FloodFillFunctor from the start coord, given initialFillPower.
    */
-  public static ArrayList<XYCoord> findFloodFillArea(XYCoord start, FloodFillFunctor fff, int initialFillPower, GameMap gameMap)
+  public static ArrayList<XYCoord> findFloodFillArea(XYCoord start, Unit mover, Army team, MoveType fff, int initialFillPower, GameMap gameMap, boolean includeOccupiedSpaces, boolean canTravelThroughEnemies)
```
ThatIsAUsername, line 134: Why accept a `Unit`, its `Army`, and its `MoveType` all as separate arguments? If you have a valid Unit, it should already have the `Army` and the 'MoveType`.

src/Engine/Utils.java
```diff
@@ -199,75 +201,57 @@ public static boolean isPathValid(XYCoord start, FloodFillFunctor fff, int initi
...
-  /** Alias for {@link #findShortestPath(XYCoord, Unit, int, int, GameMap, boolean) findShortestPath(XYCoord, Unit, int, int, GameMap, boolean=false)} **/
-  public static GamePath findShortestPath(Unit unit, XYCoord destination, GameMap map)
-  {
-    return findShortestPath(unit, destination.xCoord, destination.yCoord, map, false);
-  }
-  /** Alias for {@link #findShortestPath(XYCoord, Unit, int, int, GameMap, boolean) findShortestPath(XYCoord, Unit, int, int, GameMap, boolean=false)} **/
   public static GamePath findShortestPath(Unit unit, int x, int y, GameMap map)
   {
-    return findShortestPath(unit, x, y, map, false);
+    return findShortestPath(unit, new XYCoord(x, y), map, false);
```
ThatIsAUsername, line 221: Yeah, I tended to use overloads that take `XYCoord` and separate `x` and `y` values just to avoid this. Not that it's like, a huge deal, and there is an argument for standardization.

src/Engine/Utils.java
ThislsAUsername: It's providing different levels of specificity.
The Unit is there for identity purposes: for checking whether it can stand the tile it currently occupies.
The Army is necessary for logic around pathing through other units, and can also work with null (enemy of everyone, as usual).
The MoveType is necessary to do anything; it can't be null.

Thinking about it, we don't need the Unit param if we remove `includeOccupiedSpaces`.
We would make that always true (units' tiles are put into the output set)., so we'd have to adapt some AI functions. Maybe it's a change worth making.

src/Engine/Utils.java
ThislsAUsername: I was annoyed by having too many overloads, and most usecases already have an XYCoord handy (and going XYCoord->X/Y is annoying because the var names are kinda long)

src/Engine/Utils.java
ThatIsAUsername: Fair enough I guess.

ThatIsAUsername:
Think that's it for this one. Fire away.

PR #385, opened 2023-06-05
Brenner

ThislsAUsername:
I may have gotten carried away. I haven't even gotten to the mistake yet.


src/CommandingOfficers/BrennerWolves/BrennerDoR.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private static final CommanderInfo coInfo = new instantiator();
+  private static class instantiator extends CommanderInfo
+  {
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Brenner", UIUtils.BW);
+      infoPages.add(new InfoPage(
+          "Captain of the 12th Battalion. He has an unshakable faith in humanity’s goodness, and will aid any in need.\n"));
```
ThatIsAUsername, line 24: Just armies in need? I assume you just stole this text from somewhere, but it seems amusingly specific, especially since most of the other armies are actively fighting against him.

src/CommandingOfficers/Commander.java
```diff
@@ -49,7 +49,7 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
   public static final int CHARGERATIO_FUNDS = 9000; // quantity of funds damage to equal 1 unit of power charge
   public static final int CHARGERATIO_HP = 100; // Funds value of 1 HP damage dealt, for the purpose of power charge
   public int incomeAdjustment = 0; // Commander subclasses can increase/decrease income if needed.
-  private double myAbilityPower = 0;
+  private int myAbilityPower = 0;
```
ThatIsAUsername, line 52: Why ints now? Moving us over to use proper stars? I did notice that term in the Ave change (though I haven't looked at most of this yet).

src/CommandingOfficers/BrennerWolves/BrennerDoR.java
ThislsAUsername: I copied it from the cart. I guess it makes more sense in the context of the campaign there, lol

src/CommandingOfficers/Commander.java
ThislsAUsername: Funds, actually. My thinking is that "real charge" will be in funds (since funds are the smallest fraction of a "star"), and we can just define conversion rates to and from that.
I think I ran into another case of rounding error. I hate rounding error.

src/CommandingOfficers/Commander.java
```diff
@@ -251,22 +251,27 @@ public void activateAbility(CommanderAbility ability, MapMaster map)
...
       myAbilityPower = 0;
-    double maxPower = getMaxAbilityPower();
+    int maxPower = getMaxAbilityPower();
     if( myAbilityPower > maxPower )
       myAbilityPower = maxPower;
   }
+  public void modifyAbilityStars(int amount)
+  {
+    // TODO: Make this scale with fatigue
+    modifyAbilityPower(amount * CHARGERATIO_FUNDS);
```
ThatIsAUsername, line 266: Is.. fatigue a thing in AW? I'm admittedly less familiar with DoR ability mechanics, but I don't remember that.

src/CommandingOfficers/DeployableCommander.java
```diff
@@ -0,0 +1,263 @@
...
+  public long canDeployMask = Long.MAX_VALUE;
+  public final boolean canDeployOn(UnitModel type) {return type.isAny(canDeployMask);};
+
+  public boolean resetCOUsEveryTurn = false;
+
+  /** Can I deploy my COU here? Ignored if resetCOUsEveryTurn==true */
+  protected boolean eligibleDeployLocation(Unit actor, MapLocation loc)
+  {
+    if( loc.getOwner() != actor.CO )
+      return false; // No COing on neutral factories
```
ThatIsAUsername, line 52: So if `resetCOUsEveryTurn==true`, then we are allowed to CO on neutral factories?

src/CommandingOfficers/DeployableCommander.java
```diff
@@ -0,0 +1,263 @@
...
+      COUs.add(join.unitRecipient);
+    }
+    return super.receiveUnitJoinEvent(join);
+  }
+
+  //////////////////////////////////////////////////////////
+  // Action definition happens after this point
+  //////////////////////////////////////////////////////////
+
+  private static class DeployCOUFactory extends UnitActionFactory
```
ThatIsAUsername, line 129: Do we want to keep spreading action/factory/event classes everywhere (especially given that we have dedicated packages for those things already? I mean there's global search, but if I don't remember the proper name for a given thing it may not help much.

src/CommandingOfficers/Commander.java
ThislsAUsername: Uhhh, people started calling it "power fatigue" recently and it got stuck in my brain.
It's the thing where power costs scale up by 0.2x of the base cost each use for 9 or 10 uses.
DoR actually doesn't have it, since you can only charge in your zone (which shrinks after COP anyway).

src/CommandingOfficers/DeployableCommander.java
ThislsAUsername: If `resetCOUsEveryTurn==True`, you can put your "COU" on any unit you want.
Maybe "COU anywhere" shouldn't be tied to "COU every turn", since I guess you could pull off some funny designs with that... but all the current usecases did it that way.

Limiting your COUs to one turn away from your production seems pretty bad, lol

src/CommandingOfficers/DeployableCommander.java
ThislsAUsername: I wasn't under the impression they were being spread "everywhere".
This is code for this specific subtype of CO, and isn't applicable to anyone else. Ditto with Meridian.
I also don't foresee other pieces of code needing/wanting to use the Kaiju actions for anything - they're uh, very unique (and only work if every unit around is from the KW set).

It seems more logical to me for this code to be put with the CO/unit set that uses them, rather than in the general "actions for units" bin.
IIRC I moved some of the transform stuff out of Meridian when it started being used by subs/stealths.
I suppose I don't mind relocating this code super strongly, but I would mind it a little bit.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Ahh, gotcha. Just haven't heard that term before.

src/CommandingOfficers/DeployableCommander.java
ThatIsAUsername: Fair point. Are there COs that have to redeploy themselves every turn? That seems... odd.

src/CommandingOfficers/DeployableCommander.java
ThatIsAUsername: Fair enough.

src/CommandingOfficers/Meridian.java
```diff
@@ -64,7 +64,7 @@ public Commander create(GameScenario.GameRules rules)
   final VehicularCharge myVehicularCharge = new VehicularCharge(this);
   private static final int POST_REFRESH_STAT_ADJUSTMENT = -30;
   final UnitModel tank, arty;
-  final UnitActionFactory tankAction, artyAction;
+  final UnitActionFactory tankAction, artyAction; // Has to be one per CO, so units get the same instance over multiple action searches (mostly for the AI)
```
ThatIsAUsername, line 67: Asplain. Would one static global instance not satisfy that requirement? What other alternative was there besides that and one per CO?

src/CommandingOfficers/Commander.java
```diff
@@ -290,15 +295,15 @@ public ArrayList<GameOverlay> getMyOverlays(GameMap gameMap, boolean amIViewing)
   }
 
   // Note: Cart charge only uses whole HP, so that's what we're doing, too.
-  public double calculateCombatCharge(UnitDelta minion, UnitDelta enemy)
+  public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
```
ThatIsAUsername, line 298: Why add parameter and not use?

ThatIsAUsername: Never mind. Found it in `RuinedCommander`. So only when the attacker is in the CO zone do we get charge.

src/CommandingOfficers/DeployableCommander.java
ThislsAUsername: That was the initial draft of the Tabitha conversion, and still is Lin's.
My idea is basically: "Lin has a zone of 1, so she should be more flexible than Tabitha. But having a real zone mechanic is too much... so we can just sort of assume Lin has her zone where it needs to be to boost one critical unit each turn".
Whether that's a good conversion or even actually simpler than setting up zones for real? Good question :v

src/Engine/StateTrackers/KillCountsTracker.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Tracks a simple kill count for each unit; joined units use the higher value.
+ */
+public class KillCountsTracker extends StateTracker
+{
+  private static final long serialVersionUID = 1L;
+
+  // I wrote a generic handler for this and I'm going to use it
+  private CountManager<GameInstance, Unit> killCounts = new CountManager<>();
```
ThatIsAUsername, line 18: Sigh. I think this class has been overkill everywhere I've actually seen it used. Cinder uses it, but only ever uses her own army as the list key. Now we have one where the GameInstance is the only key.

I mean, I guess it works.

src/CommandingOfficers/Commander.java
ThislsAUsername: Correct.
I'm considering a game option to make it work as you'd expect (i.e. when *your* unit is in the zone), but also that sounds like effort.

src/CommandingOfficers/RuinedCommander.java
```diff
@@ -0,0 +1,436 @@
...
+import Units.UnitDelta;
+
+public abstract class RuinedCommander extends DeployableCommander
+{
+  private static final long serialVersionUID = 1L;
+  public static final InfoPage DOR_MECHANICS_BLURB = new InfoPage(
+            "Days of Ruin mechanics:\n"
+          + "You have a CO zone that covers the area around your COU, and grants bonuses to your units inside.\n"
+          + "The zone expands 1 tile when your meter is half full, and again at full charge.\n"
+          + "All units get +10 attack and +10 Damage Division (DD) while in the zone.\n"
```
ThatIsAUsername, line 39: Is "Damage Division" like, a real term?

src/CommandingOfficers/RuinedCommander.java
```diff
@@ -0,0 +1,436 @@
...
+  // Tell MapController and friends that we never have abilities ready, since those go through the COU
+  @Override
+  public ArrayList<CommanderAbility> getReadyAbilities()
+  {
+    return new ArrayList<CommanderAbility>();
+  }
+  @Override
+  public void onCOULost(Unit minion)
+  {
+    modifyAbilityStars(-42);
```
ThatIsAUsername, line 104: Okay, but hear me out: what if one day we have a Commander with an ability that requires more than 42 stars?

src/Engine/StateTrackers/KillCountsTracker.java
ThislsAUsername: It is global per `GameInstance`, so you need *some* distinguishing factor... unless it really is global.
It might be doable to not require the first key.

src/CommandingOfficers/RuinedCommander.java
ThislsAUsername: It is now.

src/CommandingOfficers/RuinedCommander.java
ThislsAUsername: Then he will be the chosen one.

src/CommandingOfficers/Meridian.java
ThislsAUsername: I initially created the action type instance any time the modifier was invoked. (You can see that in the `DeployableCommander` diff on the last commit)

I worry about how a static/global instance would work with savegames, though I guess I could do the thing I did with terrain and make it "deserialize" by pointing to the static instance. Having it in the CO is a small cost of memory for a simpler implementation, though.
...I guess I could test it, but (de)serialization is so full of caveats and black magic that I don't like touching it if I don't have to.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -151,17 +151,17 @@ private static int getAnimIndex()
...
    * Generate an ability-power bar for the given Commander at 1x size. The requester is responsible for applying any scale factors.
    */
-  public static BufferedImage buildCoPowerBar(Commander co, double[] abilityPoints, double currentPower, boolean leftSide)
+  public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, int currentPower, boolean leftSide)
   {
     if( 0 == abilityPoints.length )
       return SpriteLibrary.createTransparentSprite(1, 1);
 
-    final double pixelsPerPowerUnit = 3.0;
+    final double pixelsPerPowerUnit = 3.0 / Commander.CHARGERATIO_FUNDS;
```
ThatIsAUsername, line 159: Am I missing something, or are we calculating 3/9000==0 pixels per power unit? I mean, it still draws, so clearly not, but still.

ThislsAUsername: 3.0 is a double value, so...

src/CommandingOfficers/RuinedCommander.java
ThatIsAUsername: You gotta define your terms, man. You always just invent them and then expect people to understand what you mean. The obvious meaning is that incoming damage is now divided by 10, but that can't be right.

ThislsAUsername: It's Advance Wars, so everything has a base of 100.
You're right, I should write up that readme, though.
Should probably move most of this to that, too? It's uh... kind of a lot for the CO page.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThatIsAUsername: Is each "power unit" a fund now? It used to be a star equivalent, but that clearly isn't what you're doing here.

ThatIsAUsername:
> I may have gotten carried away. I haven't even gotten to the mistake yet.

Have you gotten to the mistake yet?

src/CommandingOfficers/RuinedCommander.java
ThatIsAUsername: So "Damage Division" is just "Defense, but calculated according to Days of Ruin rules". Yeah, it's confusing that it's calculated differently in the different games, and more confusing that we are combining them both in one game. It's probably good to use a separate term for clarity, though I might even just spell it out since we don't have a glossary, e.g.: units get +10% attack and take 10% less damage while in the zone." It really isn't more verbose, and I don't think I'd stress about repeating it on all of them.

ThislsAUsername: The problem with that phrasing is that it applies equally well to either type of defense.
You will note that I generally leave of percents for attack as well, since +10% attack could just as easily be a 1.1x multiplier to your damage as a +10 to your attack value.

Trilogy defense makes you take X% less of the opponent's entire modified damage output, regardless of your previous stat (there are AW1-shaped asterisks there, but bleeeh)
DoR defense makes you take X% of your existing defense stat less of the opponent's already-modified damage output. (that's not quite an exact description, I think but you get the gist)

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThislsAUsername: Yes. I did mention that on another thread, I think.
I didn't make the interface as obvious as e.g. HP since I was trying to limit how much I touched (oops?).
This does mean "1 star of charge" can always be 1 star as the carts define it, though, since the scaling is underneath that level (in funds).

ThislsAUsername:
I have made the mistake, but I made another mistake which kind of scuttles the whole operation.
A fix for that mistake is pending.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
ThatIsAUsername: Were we defining power units differently than stars before? I know we do scaling differently, but otherwise it was just a naming difference AFAIK.

ThislsAUsername: Sort of.
The power costs were defined in *base* stars, so scaling either had to bring you to a non-integer cost or increase by an integer number of stars... and there was no way to easily refer to "stars as scaled with fatigue", while now it wouldn't be too hard.

src/CommandingOfficers/RuinedCommander.java
ThatIsAUsername: Admittedly the language is overloaded, which creates confusion. This is why it's important to clarify. Maybe even just "DoR Defense" might be clear enough since the nerds will know what that means and the normies won't care. Making up new terms could be helpful, but only if they have discoverable definitions (and I have little hope that anyone is gonna bust out the README - I sure wouldn't think to look there). Might not even hurt to rename _both_ quantities just to create distance from the overloaded term.

Also, for your amusement, and because LLMs are fun:
https://www.perplexity.ai/search/f0e12951-2024-4559-abce-7de865255750?s=u

ThislsAUsername: Yeah. I was thinking I ought to write a README on the damage calculations, but it doesn't really matter if nobody reads it.
Jay said it sounded like a good idea, but my level of confidence in that solution is not high.

That is amusing, but those proposed names aren't super clearly related to either defense model either.

ThatIsAUsername: Yeah, and yeah. A game glossary kinda has to be in-game to be useful, especially nowadays.

And yes; this is a hard naming problem, since they are very related calculations. It might not hurt to just call it DoR defense instead of trying to come up with fancy names.

ThislsAUsername: Yeah, fair enough.

src/CommandingOfficers/RuinedCommander.java
```diff
@@ -0,0 +1,437 @@
...
+          + "  (This means an enemy in your zone attacking outside your zone can give you charge from your counterattack)\n"
+          + "However, your CO ability does not increase in cost as you use it.\n"
+          + "Veterancy:\n"
+          + "On making a kill, your unit will level up.\n"
+          + "There are 3 veterancy levels; their stat boosts are +5/0, +10/0, and +20/20 attack/defense.\n"
+          + "Your COU is always max level.\n"
+          + "Veterancy bonuses do not apply if this CO is tagged out, but kills are always tracked.\n"
+          );
+  public static final int DOR_STAR_VALUE = 10000;
+  public static final int CHARGERATIO_HP =  1000; // Funds value of 1 HP damage dealt, for the purpose of power charge
```
ThatIsAUsername, line 55: This doesn't vary based on the unit? Or wait, that changed between the OT and DoR, didn't it; they used destroyed funds, and DoR used straight HP?

ThislsAUsername: Correct.
DoR does straight HP damage *dealt* only, and only if the combat initiator is inside your zone (rather than, y'know, your unit).

src/UI/UIUtils.java
```diff
@@ -320,6 +320,11 @@ public static COSpriteSpec fromDisk(String col, String fac)
   public static final COSpriteSpec AR = COSpriteSpec.fromDisk("Acid"  , "Rain");
   public static final COSpriteSpec WN = COSpriteSpec.fromDisk("White" , "Nova");
 
+  public static final COSpriteSpec BW = COSpriteSpec.fromDisk("Brenner's"     , "Wolves");
+  public static final COSpriteSpec NW = COSpriteSpec.fromDisk("New Rubinelle" , "Wolves");
+  public static final COSpriteSpec LA = COSpriteSpec.fromDisk("Lazurian"      , "Army");
+  public static final COSpriteSpec IA = COSpriteSpec.fromDisk("IDS"           , "Army");
```
ThatIsAUsername, line 326: Two wolves and two armies? Duplicate names isn't ideal, but I guess maybe it's hard to avoid. Also, wasn't Rubinelle also an army?

src/UI/Art/SpriteArtist/OverlayArtist.java
```diff
@@ -59,15 +59,22 @@ public static void drawHighlights(Graphics bigG,
         overlays.add(ov);
 
     // Check if input is the same as last time
-    ByteBuffer params = ByteBuffer.allocate(Integer.BYTES * (8 + inputOverlays.size()) );
+    ByteBuffer params = ByteBuffer.allocate(Integer.BYTES * (8 + inputOverlays.size()*3) );
```
ThatIsAUsername, line 62: Magic 3?

ThatIsAUsername: Also, magic 8?

src/UI/UIUtils.java
ThislsAUsername: It's an artifact of how we set up unit sprites.
In the lens of unit sprites, there are only two "real" armies in DoR: Rubinelle-based and Lazuria-based. IDS uses the Lazuria sprites.
Because of that, "Wolves" has the right infantry sprites for the NRA.

src/UI/Art/SpriteArtist/OverlayArtist.java
ThislsAUsername: The magic numbers are how many integers I'm storing of each type, in this function.
Potentially 8 ints non-scaling, and potentially 3 ints per overlay that's requested for display.
I don't think defining constants or whatever will make the code much more clear?

ThatIsAUsername: What are we actually calculating? The number of overlays? The number of tiles that can be overlaid? And why did you have to introduce the magic three when we didn't have it before? What quantity are we calculating? Even after reading your description I have no idea.

PR #386, opened 2023-06-07
Actually revert the Overcharge overhealing

ThislsAUsername:
I broke it and didn't notice, derp

### DIFF
--- a/src/CommandingOfficers/Tech.java
+++ b/src/CommandingOfficers/Tech.java
@@ -194,10 +194,15 @@ public class Tech extends Commander
       if( unitsOverCharged.containsKey(myCommander) )
       {
         // End Overcharge. Any units who still have > MAXIMUM_HP get reset to max.
+        final int maxPercent = UnitModel.MAXIMUM_HP*10;
         for( Unit u : unitsOverCharged.get(myCommander) )
         {
-          if( u.getPreciseHP() > UnitModel.MAXIMUM_HP )
-            u.alterHP(10);
+          final int preciseHP = (int) (u.getPreciseHP() * 10);
+          if( preciseHP > maxPercent )
+          {
+            int deltaPercent = Math.min(preciseHP - maxPercent, healAmount * 10);
+            u.alterHealthPercent(-1 * deltaPercent);
+          }
         }
         unitsOverCharged.remove(myCommander);
       }


ThislsAUsername:
This does mean she can't boost Kaiju healing to the moon, but I guess I'm not worried about that yet.

src/CommandingOfficers/Tech.java
```diff
@@ -194,10 +194,15 @@ protected void revert(MapMaster gameMap)
       if( unitsOverCharged.containsKey(myCommander) )
       {
         // End Overcharge. Any units who still have > MAXIMUM_HP get reset to max.
+        final int maxPercent = UnitModel.MAXIMUM_HP*10;
         for( Unit u : unitsOverCharged.get(myCommander) )
         {
-          if( u.getPreciseHP() > UnitModel.MAXIMUM_HP )
-            u.alterHP(10);
+          final int preciseHP = (int) (u.getPreciseHP() * 10);
+          if( preciseHP > maxPercent )
```
ThatIsAUsername, line 201: Comparing an HP value with a percent value looks wrong. Looks like just a variable-naming thing, not an actual mis-comparison, but it smells as is.

src/Engine/GameInstance.java
```diff
@@ -283,7 +291,18 @@ public void endGame()
...
+  /**
+   * Gives you a random number between 0 (inclusive) and the input (exclusive)
+   * <p>Call this in GameAction.getEvents()/event constructors and nowhere else
+   */
+  public int getRN(int upperBound)
+  {
+    upperBound = Math.abs(upperBound);
+    ++rngNumbersGenerated;
+    return rng.nextInt(upperBound);
+  }
```
ThatIsAUsername, line 304: I have a bit of heartburn creating a public, effectively global method that is only supposed to be used by one component. How do we either a) prevent misuse or b) detect it if it happens?

src/CommandingOfficers/Tech.java
ThislsAUsername: "Precise HP" and "percent health" are the same thing.
On that note, getting a double, multiplying by 10, and casting to int kinda gives me heartburn. Can/should I just introduce a `getHealthPercent()` that returns the raw value?

Another option that's 100% outside the scope of this PR is to move to using % health everywhere but display, and leave the accessors in place to round it where needed (almost everywhere).

src/Engine/GameInstance.java
ThislsAUsername: We already had that discussion on the other PR, remember? I just forgot to rebase.

src/CommandingOfficers/Tech.java
ThatIsAUsername: I mean, we could. And it would frankly make sense to just move to the 0-100 health value everywhere except in the display. I've had the heebie-jeebies about how we've been handling health and HP as two different quantities for a while.

src/Engine/GameInstance.java
ThatIsAUsername: I do remember the discussion - I don't remember which PR it was, or if I asked exactly these questions, or what the outcome was.

ThislsAUsername: That'll be this boi: https://github.com/ThislsAUsername/DefendPeace/pull/382#discussion_r1252276588

ThatIsAUsername:
Leaving aside the standardization of health-related calculations as out of scope, this PR is probably good to go.

PR #388, opened 2023-06-11
CO Filtering

ThislsAUsername:
I ended up pushing the actual manipulation logic into the view, but (as I comment) I think we're gonna be moving the control flow there properly at some point.
Control flow through the GUI would be quite useful for anything outside keyboard control, and I think doing the visual bin wrapping could only be done with worse violations of separation of concerns otherwise.

### DIFF
--- a/src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
+++ b/src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
@@ -31,8 +31,8 @@ public class PlayerSetupCommanderArtist
   private static SlidingValue tagPickerOffsetX = new SlidingValue(0);
 
   static final BufferedImage kickTip = SpriteUIUtils.makeTextFrame("Q: Kick CO", 3, 2);
-  static final BufferedImage toGameTip = SpriteUIUtils.makeTextFrame("Q: To Game", 3, 2);
-  static final BufferedImage toArmyTip = SpriteUIUtils.makeTextFrame("Q: To Army", 3, 2);
+  static final BufferedImage toGameTip = SpriteUIUtils.makeTextFrame("Q: By Game", 3, 2);
+  static final BufferedImage toArmyTip = SpriteUIUtils.makeTextFrame("Q: By Army", 3, 2);
   static final BufferedImage etip = SpriteUIUtils.makeTextFrame("E: CO Info", 3, 2);
 
   static final int textWidth = SpriteLibrary.getLettersSmallCaps().getFrame(0).getWidth();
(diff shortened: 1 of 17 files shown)


ThatIsAUsername:
Oh, weird. When in "game" mode, it's only showing COs from a single army in that game.
This also results in many empty categories - e.g., if I focus Brenner and hit "To Game", I'm shown a screen where his is the only selectable army, even though the other games are shown. This is... confusing.
I would have expected to filter by one or the other as the primary key, but still show everything.
The left-side caption is also unhelpful - it shows either a game or a faction, but for example, in "AW4", I can still see DefendPeace/AWBW factions.

OK, I think I figured it out. In Game Mode, it shows the COs from each faction in that game, and in Army mode, it shows that faction in each game. But it still shows empty categories; I'd skip those. Oh, except then it would be impossible to get out of the Brenner's Wolves/AW4 hole once you got in there. I see...

I might also change to "By Game/Army" instead of "To Game/Army" since we sort/filter by that thing.

I can see the intent; in execution it feels pretty unintuitive (so far, though I'm sure I'll get used to it). This is the kind of thing user testing would help to refine. I'd probably expect to still see everything, and just re-branch the tree based on primary/secondary filter. E.g., in sort-by-game-first Mode, have a side panel for *each* game (Not just one as it currently does), with only the relevant factions/COs inside, and vice versa. I think that would make it more intuitive to find things and also fix the BW/AW4 trap mentioned above.

Sorry if that was kind of train-of-thought, but that was my train of thought while trying this out.

The other issue is we've got conflicts with `brenner`, but one thing at a time.

ThislsAUsername:
> OK, I think I figured it out. In Game Mode, it shows the COs from each faction in that game, and in Army mode, it shows that faction in each game. But it still shows empty categories; I'd skip those. Oh, except then it would be impossible to get out of the Brenner's Wolves/AW4 hole once you got in there. I see...

Yeah...
> I might also change to "By Game/Army" instead of "To Game/Army" since we sort/filter by that thing.

Solid point, particularly if we also show all top-level categories... which is probably necessary.
> I can see the intent; in execution it feels pretty unintuitive (so far, though I'm sure I'll get used to it). This is the kind of thing user testing would help to refine. I'd probably expect to still see everything, and just re-branch the tree based on primary/secondary filter. E.g., in sort-by-game-first Mode, have a side panel for _each_ game (Not just one as it currently does), with only the relevant factions/COs inside, and vice versa. I think that would make it more intuitive to find things and also fix the BW/AW4 trap mentioned above.

You are right. I did have that in mind.
That's just more code than I wanted to tackle all at once without getting at least a little of a sanity check.
> Sorry if that was kind of train-of-thought, but that was my train of thought while trying this out.

Not a problem at all. It was kinda gratifying, honestly.
> The other issue is we've got conflicts with `brenner`, but one thing at a time.

That's as easy as a rebase I just forgot to do.

ThislsAUsername:
So uh, I got the thing working. However, there's a new problem.
image

ThatIsAUsername:
> So uh, I got the thing working. However, there's a new problem.

Yeah. I'd almost say to reorient the text when filtering by faction because a) faction names are longer, and b) factions have fewer COs than games do, so you have fewer mugshots to fit. Can stack the two words in two lines also.

There's also a slight graphical thing where the tab header on the lower left is drawn as terminating if it's at the bottom of the screen, even if the category extends beyond the screen boundary (sometimes, but not always).

ThislsAUsername:
> Yeah. I'd almost say to reorient the text when filtering by faction because a) faction names are longer, and b) factions have fewer COs than games do, so you have fewer mugshots to fit. Can stack the two words in two lines also.

That'd work, I guess. Though I dunno how to... I can just stick it before the game name, actually.
> There's also a slight graphical thing where the tab header on the lower left is drawn as terminating if it's at the bottom of the screen, even if the category extends beyond the screen boundary (sometimes, but not always).

Oh, that's fun. It's actually the next category down drawing itself on top
image

ThatIsAUsername:
That looks better. Though in the "by army" page the factions are now drawn with their native colors (but they are in the "by game" page, oddly enough).

image

ThislsAUsername:
The factions' colors *are* drawn... on the side panel.
Maybe it makes sense to only use "game" colors for the case where the games are the outer category.

ThatIsAUsername:
Oh, is that's what's going on? The games have assigned colors too? Also I didn't know Rose Thorn and Cyan Ocean were in AW2...

I was talking before about how extra info would be helpful when reviewing certain PRs; this one is a good example. You've got a nontrivial mechanism for sorting CO faces in two different ways, and some way to assign colors by faction (an apparently by game). I haven't run across a high level description yet, but knowing the overall strategy/data structures involved would provide context for the lines of code I'm reading without my having to build the forest from the trees, so to speak.

src/UI/UIUtils.java
```diff
@@ -323,8 +323,18 @@ public static COSpriteSpec fromDisk(String col, String fac)
   public static final COSpriteSpec BW = COSpriteSpec.fromDisk("Brenner's"     , "Wolves");
   public static final COSpriteSpec NW = COSpriteSpec.fromDisk("New Rubinelle" , "Wolves");
   public static final COSpriteSpec LA = COSpriteSpec.fromDisk("Lazurian"      , "Army");
-  public static final COSpriteSpec IA = COSpriteSpec.fromDisk("IDS"           , "Army");
+  public static final COSpriteSpec IDS = COSpriteSpec.fromDisk("IDS"          , "Army");
 
+  public enum SourceGames
+  {
+    AW1(OS), AW2(BM), AW3(GE), AW4(IDS), DEFEND_PEACE(RT); // More? :v
```
ThislsAUsername, line 330: Right, so I defined this game enum here, and arbitrarily picked which faction they'd take their color from.
Having a game enum is sort of required to have filtering by game, and having a color for each game is nice for GUI things.

This GUI is mostly to make it easier to pick through an unreasonably large cast of COs, since that's super where we're headed (and arguably already hit on the test branch)

I figured having the ability to toggle between the filter strategies would do two important things:
You can more easily limit what you're looking at to what you care about ("I'm looking for an OS CO" vs "I only want to get AW2 COs").
It lets you navigate to specific things a little faster/easier... albeit in a way that isn't super intuitive.

I moved all the "current state" tracking logic into the view, since it has to handle wrapping concerns.
The basic structure is:
View owns a list of outer bins
Outer bins own a list of inner bins
Inner bins can have multiple lines, since the user can fiddle with the window size

Everything is re-generated in place when the filter strategy flips.
Filter strategy determines what is "outer" vs "inner" (with "inner" being the category for each row)

I'm not super thrilled with the way this divides the code (stealing from the short file to make the longer one longer), but the division of responsibilities feels better.


I guess I kind of assumed "You can see the GUI, GUI tells you what it is" would be good enough. This is a silly assumption because a lot of GUI stuff you wrote confused me when I first read it...

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -26,9 +30,14 @@ public class PlayerSetupCommanderArtist
...
   private static SlidingValue tagPickerOffsetX = new SlidingValue(0);
 
-  static final BufferedImage qtip = SpriteUIUtils.makeTextFrame("Q: Kick CO", 3, 2);
+  static final BufferedImage kickTip = SpriteUIUtils.makeTextFrame("Q: Kick CO", 3, 2);
+  static final BufferedImage toGameTip = SpriteUIUtils.makeTextFrame("Q: By Game", 3, 2);
+  static final BufferedImage toArmyTip = SpriteUIUtils.makeTextFrame("Q: By Army", 3, 2);
   static final BufferedImage etip = SpriteUIUtils.makeTextFrame("E: CO Info", 3, 2);
 
+  static final int textWidth = SpriteLibrary.getLettersSmallCaps().getFrame(0).getWidth();
+  static final int textHeight = SpriteLibrary.getLettersSmallCaps().getFrame(0).getHeight();
```
ThatIsAUsername, line 39: We should make a less painful way to get this info. Also it hurts me a little bit that we handle the standard vs small-caps fonts in _completely_ different ways.

ThislsAUsername: Kerning isn't terribly important with a monospace font, to be fair.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -305,4 +333,498 @@ public BufferedImage update(CommanderInfo coInfo, Color color)
...
+        {
+          if( sortByGameThenFaction )
+          {
+            palette = UIUtils.getMapUnitColors(spriteSpec.color).paletteColors;
+            canonName = UIUtils.getCanonicalFactionName(spriteSpec);
+          }
+          else // Our inner category is games, so steal stuff from the outer category
+          {
+            palette   = outerPalette;
+            canonName = outerSpec.faction.name + ": " + spriteSpec.faction.name;
```
ThatIsAUsername, line 587: Interesting choice to use the canonical name ("Green Earth") when sorting by game, but just using the faction name ("Earth") when sorting by faction. Any particular reason?

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThatIsAUsername: Obviously out of scope for this MR, I'm just complaining because I do that.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThislsAUsername: Uhhh. Being blind, probably

ThatIsAUsername:
Looks good. I haven't tried it out with the full list of COs from the playtest branch, but it handles well for the current roster here at least.

ThislsAUsername:
I haven't either, to be fair. It'd be a lot of effort to do that just for a "mockup"

ThislsAUsername:
Herf derf, my brain turned off and I forgot that this is pointing at `brenner`.

PR #389, opened 2023-06-16
JakeMan with COUs

ThislsAUsername:
He doesn't use DoR powers, because logic for that sounds hard.

### DIFF
--- a/src/AI/COParser.java
+++ b/src/AI/COParser.java
@@ -1,0 +1,59 @@
+package AI;
+
+import java.util.ArrayList;
+
+import CommandingOfficers.Commander;
+import CommandingOfficers.DeployableCommander;
+import CommandingOfficers.RuinedCommander;
+import Engine.Army;
+import Units.Unit;
+
+/**
+ * Figures out which of your COs have a special sub-type you can glean more info from.
+ */
+public class COParser
+{
+  public final ArrayList<DeployableCommander> deployableCOs;
+  // Will necessarily be in the above list, but there may be more flavors later
+  public final ArrayList<RuinedCommander> dorCOs;
+
+  public COParser(Army army)
+  {
+    if( null == army )
+      throw new NullPointerException();
+
+    deployableCOs = new ArrayList<>();
+    for( Commander co : army.cos )
+    {
+      if( !(co instanceof DeployableCommander) )
+        continue;
+      deployableCOs.add((DeployableCommander) co);
+    }
+
+    dorCOs = new ArrayList<>();
+    for( Commander co : army.cos )
+    {
+      if( !(co instanceof RuinedCommander) )
+        continue;
+      dorCOs.add((RuinedCommander) co);
+    }
+  }
+
+  /**
+   * @return All COUs in your army
+   */
+  public ArrayList<Unit> getAllCOUs()
+  {
+    ArrayList<Unit> cous = new ArrayList<>();
+    for( DeployableCommander co : deployableCOs )
+    {
+      for( Unit u : co.COUs )
+        cous.add(u);
+      for( Unit u : co.COUsLost )
+        cous.remove(u);
+    }
+
+    return cous;
+  }
+
+}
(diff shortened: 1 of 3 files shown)


ThatIsAUsername:
This appears to have a bunch of duplicate commits due to force-pushing on `brenner`. Can we clean this up (e.g. recreate the branch and cherry-pick the relevant commits)? Otherwise review is gonna be difficult.

src/AI/COParser.java
```diff
@@ -0,0 +1,59 @@
...
+import Units.Unit;
+
+/**
+ * Figures out which of your COs have a special sub-type you can glean more info from.
+ */
+public class COParser
+{
+  public final ArrayList<DeployableCommander> deployableCOs;
+  // Will necessarily be in the above list, but there may be more flavors later
+  public final ArrayList<RuinedCommander> dorCOs;
```
ThatIsAUsername, line 18: I'm... not sure if I should ask.

src/AI/JakeMan.java
```diff
@@ -74,8 +75,13 @@ public JakeMan(Army army)
     // move your leftover vehicles so they cover the tiles attacking most relevant stuff (contested cities, own units)
     // move infs as far towards nearest contested stuff as you can without underdefending, cap if possible
 
+    // Comment from L&F/Rize, on the idea of engaging e.g. a Lin COU on a city:
+    // IF the dude is free & there's multiple attackers & you can do multiple attacks (since having 2 tanks for 1 tile won't help),
+    //   check the total calc after shooting with every firepower thing that can shoot and not the normal one
```
ThatIsAUsername, line 80: Is this English?

src/AI/COParser.java
ThislsAUsername: The Tabitha conversion, for example, won't be a "real" `RuinedCommander` - she uses Trilogy defense, has global charge, etc.

src/AI/JakeMan.java
ThislsAUsername: Yes?
On the subject of attacking a unit that would be unappealing to hit with one unit (city tank with +stats)...
If I think I have local force superiority, and I can hit with multiple units, evaluate the entire sequence of shots rather than just a single attack.
(since the more hits you can do, the more favorable the overall outcome is liable to be)

ThatIsAUsername: > "check the total calc after shooting with every firepower thing that can shoot and not the normal one"

Aside from the generally atrocious verbiage, it isn't even clear. Which one is the normal one?

ThislsAUsername: The normal calc to see if an attack is a good idea.
I don't recall if he ever laid that out precisely, but I came up with something decent, I think.

ThatIsAUsername: I have a mild preference for well-formed English, and a strong preference for unambiguous prose in any case (and I may note these two often correlate). At minimum I'd at least like an antecedent for any pronouns in use (and you'll say there was one--"calc", presumably--but then we are back to the "unambiguous" requirement).

I'm admittedly impressed that you managed to derive an implementable requirement from what he wrote; I don't think I could have.

Anyway, guess I'll climb back down off this high horse of mine (until next time!)

ThislsAUsername: Very fair. It's also kind of plucked out of the larger context of the discussion, since "a chatlog" is questionable requirement text in its own ways.

Hah, thanks.

ThatIsAUsername: Groovy.

ThatIsAUsername:
This is good to go whenever you want to merge.

ThislsAUsername:
I think I'll let this hang out and rebase/merge once `brenner` goes through, since it seems like it deserves its own mainline commit?

ThatIsAUsername:
> I think I'll let this hang out and rebase/merge once `brenner` goes through, since it seems like it deserves its own mainline commit?

Okay. I was planning to try and get through filters and brenner this weekend, but now it's looking like I may just have overtime instead, so no promises. But hopefully next weekend...

ThislsAUsername:
Ouch. Good luck.

PR #391, opened 2023-07-20
Schmovement

ThislsAUsername:
Teletiles zoom now

### DIFF
--- a/src/Engine/GameEvents/MoveEvent.java
+++ b/src/Engine/GameEvents/MoveEvent.java
@@ -45,8 +45,7 @@ public class MoveEvent implements GameEvent
       GamePath.PathNode endpoint = unitPath.getEnd();
       int fuelBurn = unitPath.getFuelCost(unit, gameMap);
 
-      if( null == gameMap.getLocation(endpoint.x, endpoint.y).getResident() ) // Just avoid triggering a warning.
-        gameMap.moveUnit(unit, endpoint.x, endpoint.y);
+      gameMap.moveUnit(unit, endpoint.x, endpoint.y);
       unit.isTurnOver = true;
 
       unit.fuel = Math.max(0, unit.fuel - fuelBurn);
--- a/src/Terrain/MapMaster.java
+++ b/src/Terrain/MapMaster.java
@@ -271,36 +271,31 @@ public class MapMaster extends GameMap
       return;
     }
 
+    final MapLocation newLoc = getLocation(x, y);
     if( !isLocationEmpty(unit, x, y) )
     {
       if( force ) // Force is set; the user *must* know what he's doing.
       {
-        removeUnit(getLocation(x, y).getResident());
-      }
-      else
-      {
-        System.out.println("ERROR! Attempting to move unit to an occupied MapLocation!");
-        return;
+        removeUnit(newLoc.getResident());
       }
     }
 
     // Update the map
-    MapLocation priorLoc = getLocation(unit.x, unit.y);
-    if( null != priorLoc && priorLoc.getResident() == unit )
+    removeUnit(unit);
+
+    // If we're joining/loading, we don't want to mess with the destination unit
+    if( null == newLoc.getResident() )
     {
-      priorLoc.setResident(null);
+      newLoc.setResident(unit);
+      unit.x = x;
+      unit.y = y;
     }
-    getLocation(x, y).setResident(unit);
 
     // Reset capture progress, since we moved.
     if( unit.getCaptureProgress() > 0 )
     {
       unit.stopCapturing();
     }
-
-    // Update the Unit location.
-    unit.x = x;
-    unit.y = y;
   }
 
   /** Removes the Unit from the map, if the map agrees with the Unit on its location. */
@@ -310,7 +305,7 @@ public class MapMaster extends GameMap
     {
       if( getLocation(u.x, u.y).getResident() != u )
       {
-        System.out.println("WARNING! Trying to remove a Unit that isn't where he claims to be.");
+        // Assume it's already gone
       }
       else
       {
--- a/src/Test/TestGameEvent.java
+++ b/src/Test/TestGameEvent.java
@@ -244,17 +244,18 @@ public class TestGameEvent extends TestCase
     GameEventQueue events = new GameEventQueue();
     // Validation is now performed primarily in the enqueue function since there's too much complexity to handle it nicely in the constructor
     Utils.enqueueMoveEvent(testMap, mech, path, events);
-    events.getFirst().performEvent(testMap); // This should not execute. Water is bad for grunts.
+    events.pop().performEvent(testMap); // This should not execute. Water is bad for grunts.
     testPassed &= validate(7 == mech.x && 6 == mech.y, "    Mech does not think he is at (7, 6), but should.");
     testPassed &= validate(testMap.getLocation(7, 6).getResident() == mech, "    Mech is not still at (7, 6), but should be.");
     testPassed &= validate(testMap.getLocation(7, 0).getResident() == null, "    MapLocation (7, 0) should still be empty.");
 
     path.addWaypoint(7, 5); // New endpoint to move apc over infantry.
-    new MoveEvent(apc, path).performEvent(testMap); // This should not execute. Treads are bad for grunts.
+    Utils.enqueueMoveEvent(testMap, apc, path, events);
+    events.pop().performEvent(testMap); // APC shouldn't go the full movement. Treads are bad for grunts.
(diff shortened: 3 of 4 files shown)


src/Engine/GameEvents/MoveEvent.java
```diff
@@ -45,8 +45,7 @@ public void performEvent(MapMaster gameMap)
       GamePath.PathNode endpoint = unitPath.getEnd();
       int fuelBurn = unitPath.getFuelCost(unit, gameMap);
 
-      if( null == gameMap.getLocation(endpoint.x, endpoint.y).getResident() ) // Just avoid triggering a warning.
-        gameMap.moveUnit(unit, endpoint.x, endpoint.y);
+      gameMap.moveUnit(unit, endpoint.x, endpoint.y);
```
ThatIsAUsername, line 48: If this PR is for animating teletile movement faster, why are we also removing a stomp check?

src/Test/TestGameEvent.java
```diff
@@ -244,17 +244,18 @@ boolean testMoveEvent()
...
     path.addWaypoint(7, 5); // New endpoint to move apc over infantry.
-    new MoveEvent(apc, path).performEvent(testMap); // This should not execute. Treads are bad for grunts.
+    Utils.enqueueMoveEvent(testMap, apc, path, events);
+    events.pop().performEvent(testMap); // APC shouldn't go the full movement. Treads are bad for grunts.
     testPassed &= validate(7 == inf.x && 5 == inf.y, "    Infantry should still think he is at (7, 5).");
     testPassed &= validate(testMap.getLocation(7, 5).getResident() == inf, "    Infantry should still be at (7, 5).");
-    testPassed &= validate(3 == apc.x && 2 == apc.y, "    APC should still think it is at (3, 2).");
-    testPassed &= validate(testMap.getLocation(3, 2).getResident() == apc, "    APC should still be at (3, 2)");
+    testPassed &= validate(3 != apc.x && 2 != apc.y, "    APC should have moved.");
+    testPassed &= validate(testMap.getLocation(3, 2).getResident() == null, "    APC should not be at (3, 2)");
```
ThatIsAUsername, line 258: Ah. And reversing the test case to ensure stomping DOES happen.
A) What does this have to do with teletile movement animations
B) Why reverse this precedent and allow inadvertent stomping?

src/Engine/GameEvents/MoveEvent.java
ThislsAUsername: It's for visual movement stuff, and I also removed the warning it was trying to avoid.

src/Terrain/MapMaster.java
```diff
@@ -310,7 +305,7 @@ public void removeUnit(Unit u)
     {
       if( getLocation(u.x, u.y).getResident() != u )
       {
-        System.out.println("WARNING! Trying to remove a Unit that isn't where he claims to be.");
+        // Assume it's already gone
```
ThatIsAUsername, line 308: Huh. Do we hit this case now?

src/Test/TestGameEvent.java
ThatIsAUsername: Huh. Given that the use case (queuing a normal move onto a friendly unit) isn't something that can/should ever happen, I think the original behavior (rejecting the move altogether) is more correct than just freestyling something different. A more reasonable approach might be "best effort" - to get as close as possible to the original destination, but even then I'd rather just throw an error than change the command and pretend nothing is wrong.

Also, given that this isn't something you can actually do in-game: why? I'm assuming Kaiju factor in here somewhere.

As for the visual issue - what visual issue? I'll admit I haven't scrutinized this too closely (at least not recently), but I'm not seeing a difference between master and this branch.

ThislsAUsername: > Huh. Given that the use case (queuing a normal move onto a friendly unit) isn't something that can/should ever happen, I think the original behavior (rejecting the move altogether) is more correct than just freestyling something different.

I think you're right, but then we'd have to have two fully separate flavors of move - those that end off the board, and those that don't.
...framing it like that, I guess that does sorta make sense, but I would have to chew on how to make that DRY/clear/consistent enough for my liking.
> A more reasonable approach might be "best effort" - to get as close as possible to the original destination, but even then I'd rather just throw an error than change the command and pretend nothing is wrong.

I believe that's what happens under this PR - the inf basically works like a blocker you didn't see in the fog. It works slightly differently (invalid end tile vs can't even enter the tile), but that's the blueprint.
> Also, given that this isn't something you can actually do in-game: why? I'm assuming Kaiju factor in here somewhere.
> As for the visual issue - what visual issue? I'll admit I haven't scrutinized this too closely (at least not recently), but I'm not seeing a difference between master and this branch.

Kaiju are entirely irrelevant, actually. They don't have the visual bug because they kill their victims *before* they move (and it'd work to let them stomp in most cases anyway).
Like I said above, both load and join (until this PR) have a 1-frame window where the unit you're moving reappears at the starting tile (movement has ended, and animating the load/join action hasn't started).
I suppose another valid approach would be to rework that whole MapController/JPanel/animation song and dance, but I don't really have ideas for making it better.

src/Terrain/MapMaster.java
ThislsAUsername: Yeah. When you load or join, the _movement_ now removes the unit from the board. (So that when the "between event" frame is drawn the unit isn't sitting there at the start position)
So when the LOAD/JOIN code comes along, it also tries to `removeUnit`, and it's already off the board.

src/Test/TestGameEvent.java
ThatIsAUsername: > I believe that's what happens under this PR

Looking at the code (and not trying, because there's no way to trigger it in-game and I'm lazy), it looks like it just teleports the moving unit off the board instead of doing "best effort" and treating it like FOW. Either way though, I like the former rejection behavior better; maybe even turn it into an exception.

ThislsAUsername: Seems you are correct - I just stepped through the test case in question and `GamePath.snipCollision()` didn't do what I expected. Not shocking, I suppose.

What function should do that rejection, though? If it's in `MoveEvent`, then we need a new kind of move event(?)
I don't like it in `Utils.enqueueMoveEvent()` because it's already handling the fog case in a different way.
Putting it at the `GameAction` level seems wrong.
`GameMap` could reject it, I guess, but we'd probably need to modify `MoveEvent` to tread around that case anyway.

I guess we could have `MoveEvent` do it, and add a bool param to it that unconditionally just removes the unit from the map.
Then if there's a collision and the param isn't set, crash the game?
Call it `moveOffMap`, I guess

ThatIsAUsername: The whole reason for doing this (introducing an illegal action) is just to remove a 1-frame graphical discontinuity, right? Seems like the right place to fix it is in the rendering code, not in the game logic.

ThislsAUsername: I guess that depends what your view of the role of the rendering code is.
For that frame:
There's no active animation (or if there is one, it's empty)
The unit is still on its starting tile

If the rendering code's role is to patch up any and all differences between the actual game state and what the player expects to see, sure.
If the rendering code's role is to show only what the game state is telling it to show, no.

The "simple" fix, I guess, would be to just not draw the active unit directly at all, but only through animations.
That's not ideal because now all the "normal" movement cases would have the unit disappear for one frame.
You could fix that maybe by drawing the unit where the cursor is, but... ew.

ThatIsAUsername: I thought the approach we took was already to not render the active unit except via animations. That's what I remember implementing anyway, plus otherwise movement anims wouldn't work at all.

But yeah, that still leaves us with the one-frame update issue. How hard would it be to apply an action's events after the animation finishes and before the map renders again? I feel like there was a reason I didn't do it the other way (apply actions and then animate), but it may be a potato potato thing.

ThislsAUsername: Yeah, that's more or less what happens.
I was suggesting to consider an animation active any time there are events waiting in the queue to be executed, though.

We... already do apply an _event's_ effects before the map renders again (which is why move animations more generally don't have this issue).
The problem here is that the move event doesn't actually move the unit if it's moving onto another unit.
Another possibility might be to keep executing events (with no intermediate frames) until we find an animation to play, so there are no gap frames.

ThislsAUsername: > keep executing events (with no intermediate frames) until we find an animation to play, so there are no gap frames.

Done in https://github.com/ThislsAUsername/DefendPeace/pull/392

PR #392, opened 2023-10-14
Schmoview

ThislsAUsername:
This is an alternate implementation of the Schmovement PR, with a different final commit.
I believe you may find it more to your taste.

### DIFF
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -45,6 +45,8 @@ public class MapController implements IController, GameInputHandler.StateChanged
 
   private InputMode inputMode;
 
+  // Holds events that are yet to be executed
+  // If we are currently animating, the top of the queue is the event being animated
   GameEventQueue activeEventQueue;
   private boolean isTurnEnding;
   private boolean isGameOver;
@@ -536,6 +538,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
       GameEventQueue listenerEvents = GameEventListener.publishEvent(event, myGame);
 
       activeEventQueue.addAll(listenerEvents);
+      event = null;
     }
 
     return event;
@@ -663,7 +666,7 @@ public class MapController implements IController, GameInputHandler.StateChanged
     GameEvent toAnimate = executeSilentEvents();
 
     // Kick off the animation cycle
-    if( !activeEventQueue.isEmpty() )
+    if( null != toAnimate )
       myView.animate(toAnimate);
     else // If there's nothing to animate, cut out the middleman
       animationEnded();
--- a/src/Engine/MapController.java
+++ b/src/Engine/MapController.java
@@ -45,6 +45,8 @@ public class MapController implements IController, GameInputHandler.StateChanged
 
   private InputMode inputMode;
 
+  GameEventQueue activeEventQueue;
+  private boolean isTurnEnding;
   private boolean isGameOver;
 
   public MapController(GameInstance game, MapView view)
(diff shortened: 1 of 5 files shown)


src/Engine/MapController.java
```diff
@@ -503,23 +509,49 @@ private boolean executeGameAction(GameAction action)
...
+  }
+
+  public void animationEnded()
+  {
+    GameEvent evt = executeSilentEvents(true);
+    boolean animEventQueueIsEmpty = activeEventQueue.isEmpty();
+    if( !animEventQueueIsEmpty )
+    {
+      myView.animate(evt);
+      return;
```
ThatIsAUsername, line 551: I assume `activeEventQueue.isEmpty()` ends up being equivalent to `null == evt`?

ThislsAUsername: I deliberately avoided returning null from `executeSilentEvents()` as much as possible (it will return the event to be animated or the last-executed event).
However, `activeEventQueue` only holds events that haven't been animated/executed yet, so yes: If there's nothing in `activeEventQueue`, there's nothing to animate.

...I guess that feels like a comment to throw in the file somewhere.

ThatIsAUsername: Hm. I'm not sure that this return behavior for `executeSilentEvents()` is any better than just returning `null`. As is, it will always return an event, but you don't know if it's the previous one or the next one? Sounds similar to the `null` option in that you can't just blindly use it, but different in that there is no way to check if it is valid to use (except to check the `activeEventQueue` side-channel for that information).

ThislsAUsername: That's fair. I feel like I did that specifically so I didn't have to think too hard about how the `SpriteMapView` piece worked:
In the case where all the events are non-animatable, they'll be executed immediately, and the last one will be pushed to `MapView`. The `MapView` will then animate nothing for 1 frame and return control to `MapController`.
...at least, that's how I remember it. It's been like 3 weeks.

ThatIsAUsername: > The MapView will then animate nothing for 1 frame and return control to MapController

I thought the whole point of this branch was to change things so we don't waste 1 frame animating nothing?

ThislsAUsername: Yes and no. (Also "technically no", in that there are other changes)
The point is to not have a frame where we display a bad gamestate (my tank just got in the lander, so why's it back over there?")
If *all* the events are already complete, there's no such problem.

And I recall there being a problem at some point where the AI would do a bunch of non-animated actions at once and blow up the stack (since `MapController` would just fetch more AI actions in `AnimationEnded()` or something).
I think wasting one frame per action, where we show the finished gamestate anyway, is excusable.

ThatIsAUsername: Fair enough, I can accept that. The inconsistent return values and the side-channel check are still a bit odd though. I mean I guess it works.

ThislsAUsername: I made `SpriteMapView` behave fine when fed `null` anyway, so have it your way, lol

src/UI/MapView.java
```diff
@@ -34,10 +34,8 @@ public InGameMenu<? extends Object> getCurrentGameMenu()
     return mapController.getCurrentGameMenu();
   }
 
-  /**
-   * Adds the new events to the queue so they can be animated.
-   */
-  public abstract void animate( GameEventQueue newEvents );
+  public abstract boolean shouldAnimate( GameEvent toAnimate );
+  public abstract void animate( GameEvent toAnimate );
```
ThatIsAUsername, line 38: I suspect you have uncommitted changes - InfoView is whining that it needs to inherit the redefined `animate(GameEvent)` from `MapView`.

ThislsAUsername: No, actually, somehow. It complains there's an error when you try to run it, but the `InfoView` itself works.
I guess I didn't notice because I have a bunch of untracked files in there that Eclipse whines are errors.

ThatIsAUsername:
Sweet, this is probs good to go then (and we can kill t'other).

PR #393, opened 2023-10-28
Add the new teletile ID as an option

ThislsAUsername:
One line PR, woo?


ThatIsAUsername:
I thought we had teletiles already? Or was there a change on the AWBW side to officially support them? I'm assuming the latter, in which case, woo?

ThislsAUsername:
They are indeed adding stuff to officially support them, with that tile ID.
So sure, "woo".
I'm just amused because I haven't released any builds with teletile support yet, so AWBW's beating me to making the feature official.

ThatIsAUsername:
Nothing to nitpick here if you want to pull the trigger.

PR #394, opened 2023-10-29
A Mistake

ThislsAUsername:
I enjoy this CO thematically, but it is mechanically not very good.


src/CommandingOfficers/DocLight.java
```diff
@@ -0,0 +1,104 @@
...
+    public instantiator()
+    {
+      super("Doctor Light", UIUtils.MISC);
+      infoPages.add(new InfoPage(
+            "Even here, it is not safe. Even this grave has been defaced.\n"
+          + "Someone has written on this stone: 'Hope rides alone'\n"));
+      infoPages.add(new InfoPage(
+          "Base Zone: 0\n" +
+          "Can deploy your COU only twice, with different stats each time (Protoman is first)\n" +
+          "Protoman Boost: +"+BLUES_POWER+"/"+BLUES_POWER+" stats and +0/"+BLUES_SHIELD+" vs indirect attack ("+(BLUES_POWER+130)+"/"+(BLUES_POWER+BLUES_SHIELD+130)+" vs indirects)\n" +
```
ThatIsAUsername, line 26: Admittedly I'm not terribly well versed in Megaman lore, but why prefix Protoman's variables with "BLUES"?

src/CommandingOfficers/DocLight.java
```diff
@@ -0,0 +1,104 @@
...
+          "Megaman Boost: +"+ROCK_POWER+"/"+ROCK_POWER+" stats ("+(ROCK_POWER+130)+"/"+(ROCK_POWER+130)+" total)\n"));
```
ThatIsAUsername, line 27: Is the Megaman boost on top of the Protoman boost? The wording is kinda confusing: +50, but 180 total.

src/CommandingOfficers/DocLight.java
ThatIsAUsername: Also the wording here isn't very clear: +30/30 makes sense, +140 defense if attacked by an indirect, but then also a boost for attacking indirects, with an even bigger defensive boost (which isn't needed because indirects don't counter)?

res/co/Doctor Light.png
ThatIsAUsername, line 1: Poor Protoman gets cut off even when I go to fullscreen. Does it look awful if you just scale it when drawing?

src/CommandingOfficers/DocLight.java
ThislsAUsername: It's his original name. (Megaman was originally called Rockman, or sometimes just Rock... which makes Roll's name make a little more sense)

src/CommandingOfficers/DocLight.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Doctor Light", UIUtils.MISC);
+      infoPages.add(new InfoPage(
+            "Even here, it is not safe. Even this grave has been defaced.\n"
+          + "Someone has written on this stone: 'Hope rides alone'\n"));
+      infoPages.add(new InfoPage(
+          "Base Zone: 0\n" +
+          "Can deploy your COU only twice, with different stats each time (Protoman is first)\n" +
```
ThatIsAUsername, line 25: I read this as being able to deploy both in tandem at first (as is seemingly implied by the artwork) - but it appears that it actually means only twice ever, full stop. Would be worth emphasizing that it means only two CO deployments over an entire game, unlike other deployable COs (so choose wisely).

src/CommandingOfficers/DocLight.java
ThislsAUsername: Generic zone bonus: +10/10
Veterancy bonus (COUs are auto-vet, even if veterancy is disabled): +20/20
Megaman boost (which matches Tabitha and Caulder): +50/50
COU Statline: 180/180
Tabi/Caulder zone statline: 160/160

res/co/Doctor Light.png
ThislsAUsername: Scale it to what? Squishing it down so it fits horizontally would make it super short.
And I like my crispy pixels.
...that said, it could be worth setting up individual integer scaling for the CO art...

src/CommandingOfficers/DocLight.java
ThislsAUsername: I thought about having that be the mechanic, but it's too strong for what I'm going for, hah.
IMO that's a weird reading of what I wrote, since the simple way to express that is "Can deploy two COUs at once".

src/CommandingOfficers/DocLight.java
ThislsAUsername: I just didn't write out the total stats for both cases, for Protoman, since defense matches attack for direct combat cases.
And nah, he's just super duper tough vs ranged attacks. I don't think the bonus applies to Anti-Tank (counter)attacks at 1 range.

src/CommandingOfficers/DocLight.java
ThatIsAUsername: Right, the issue is just that the implemented mechanic differs from a) what the art seems to indicate and b) what I therefore naturally expected. And I like to word things precisely and explicitly.

src/CommandingOfficers/DocLight.java
ThatIsAUsername: Sure, I know Megaman was Rockman, but Protoman is Blues? He's very not blue at all.

> And nah, he's just super duper tough vs ranged attacks

Ok, but are we talking +140 defense, or +300? It seems to claim both.

src/CommandingOfficers/DocLight.java
ThatIsAUsername: OK, so this Megaman has higher stats than the Tabi/Caulder COU (180 vs 169), but doesn't affect any other units? The phrasing is still strange for anyone not intimately familiar with the DoR mechanics. Again I'd lean towards being a bit more verbose for clarity.

src/CommandingOfficers/DocLight.java
ThislsAUsername: Megaman = Rock
Megaman's sister = Roll
Megaman's older brother = Blues
It's a music joke.

It is an extra 140 on top of the rest of his stats.
+30 from generic DoR stuff (base zone boost+veterancy)
+30 from his normal stats
+140 vs ranged
So that's 140 + 30 + 30 = 200 total bonus
So his final stats when being shot at by arty are 160/300

src/CommandingOfficers/DocLight.java
ThislsAUsername: No, he has the same stats.
+10/10 generic
+50/50 specific (universal boost)
This accounts for the 160/160 Tabi/Caulder zone stats
Max veterancy adds an extra +20/20, and always applies to the COU so:
180/180
This statline applies to Megaman, Caulder, and Tabitha.
Caulder and Tabitha can have zone units that aren't veterans, though, which is why I added that line.

I mean, that's why I wrote that big ol' blurb on DoR mechanics that gets appended to the descriptions.
It's just a lot to unpack if you're going to write out everything explicitly.

src/CommandingOfficers/DocLight.java
ThislsAUsername: "Cannot deploy a COU if your COU dies twice. Uses the Megaman statline after your COU dies.", then?

src/CommandingOfficers/DocLight.java
ThatIsAUsername: > It's a music joke

... Oh. Of course it is. I'm sure I knew that at one point too. Sigh.

I think DoR just lists the CO-specific boost in the CO power description, and leaves you to apply other bonuses separately. This may be a case where adding partial information creates more confusion until you explain where everything comes from. I think I'd either remove the totals or clarify that along "with Zone and veterancy boosts, total is blah". Or run it by the Discord folks as your representative audience to see what is clear to them.

src/CommandingOfficers/DocLight.java
ThatIsAUsername: I'd just add parens: "Can deploy your COU only twice (per game!), with..."

src/CommandingOfficers/DocLight.java
ThislsAUsername: > I think DoR just lists the CO-specific boost in the CO power description

You are correct.
It lists the set of boosted units, and then 1 star for each CO-specific +10 in each of attack and defense.
Adding a generic +10/10 gets you the zone boost.
Adding vet stats gets you the final D2D picture ~sans comm towers~

The only CO that boosts attack/defense on power is Waylon, and no specifics on the number is given.

> This may be a case where adding partial information creates more confusion until you explain where everything comes from

I mean, I have the big ol' blurb at the bottom that gives you the source of everything.
image

ThislsAUsername: FWIW, a lot of the reason I didn't give totals for Brenner is that he has two most-relevant totals, and another two niche ones.
110/130 base zone
130/150 COU/vet

src/CommandingOfficers/DocLight.java
ThatIsAUsername: Right. So I'd probably either keep it DRY (only mention DoR mechanics in the DoR section, and CO stat boosts in the CO stat boost section, and let the user add it all up), or specifically state how a +50 magically turns into a 180 total ("... with the DoR zone and veterancy bonuses..."). Currently you are just giving half of the story, which is arguably more confusing than skipping it altogether.

But you probably know your target audience better than I do, so if this is standard parlance and perfectly clear to people who are not me, I'll try not to nitpick too much. I guess up to you if it's worth clarifying, but other than that I guess it's ready to go (unless you want to redesign the CO's mechanics first...).

PR #395, opened 2023-11-04
Pathing Fun

ThislsAUsername:
In which a pathing fun is introduced to make the pathing more fun.

The big takeaway here is returning a list of `SearchNode` (which is now public, and a subclass of `XYCoord`) for your "all places I can go" searches.
The `SearchNode`s form a set of linked-lists back to the start, which you can exploit to create a new shortest `GamePath` in `O(n)` time rather than doing a whole new search.

As an additional point of fun, I decided to try something not-entirely-unlike-a-builder-pattern for the interface, which could slaughter a bunch of `findShortestPath()` overloads while making the code more explicit.

I don't think there's much question as to whether we want to use the optimization (it's pretty massive, and opens up new options for metadata I want to play with for Wally).
I think we probably do want to have only one version of the function to find the unit's reachable area... which begs the question as to whether this build-and-call interface is an improvement. I kind of like it, but I don't want to do a bunch and hear "bleeeh", thus the experiment.
I think `SearchNode` being public and a subclass of `XYCoord` is cool, though I wonder if it belongs in `Utils` at that point...
I think I want to take this opportunity to replace the `XYCoord` variable names with just `x` and `y` finally, as well.

The big "however" here is that we have like 7 potential merges in the near future, so I don't want to do any widespread refactoring with the Sword of Damocles waiting to make me suffer if I do.

~~just don't ask me how long I've been sitting on this idea~~


src/Engine/Utils.java
```diff
@@ -357,9 +407,12 @@ private static void expandSearchNode(Army team, MoveType fff, GameMap map, Searc
...
    *   reference to a parent node for path reconstruction.
    */
-  private static class SearchNode
+  public static class SearchNode extends XYCoord
   {
-    public int x, y;
+    private static final long serialVersionUID = 2637721435469761667L;
+    // Both XYCoord's and these are final, so no desync is possible.
+    // However, TODO: delete one pair or the other
+    final public int x, y;
```
ThatIsAUsername, line 415: TODO When?

src/Engine/Utils.java
```diff
@@ -175,6 +175,75 @@ public static ArrayList<XYCoord> findFloodFillArea(XYCoord start, Unit mover, Ar
...
+      start = new XYCoord(unit);
+      moverIdentity = unit;
+      team = unit.CO.army;
+      mt = unit.getMoveFunctor();
+      initialMovePower = Math.min(unit.getMovePower(gameMap), unit.fuel);
+      includeOccupiedSpaces = true;
+      canTravelThroughEnemies = false;
+    }
+
+    public ArrayList<SearchNode> findAllPaths()
```
ThatIsAUsername, line 202: It looks like a lot of this is copies from `findFloodFillArea` - do you intend to consolidate?

ThatIsAUsername:
Good stuff.

src/Engine/Utils.java
ThislsAUsername: Yes, but I was waiting for the "go ahead" on replacing the old API with the new one.

src/Engine/Utils.java
ThislsAUsername: I'm thinking I want to get everything else merged, and then just do it in this PR.

src/Engine/Utils.java
ThatIsAUsername: Cool. I don't object.

src/Engine/GamePath.java
```diff
@@ -19,16 +19,30 @@ public class GamePath
...
   public GamePath()
   {
-    waypoints = new ArrayList<PathNode>();
+    waypoints = new ArrayList<>();
+  }
+  public static GamePath getStatic(Unit unit)
+  {
+    return getStatic(new XYCoord(unit));
+  }
+  public static GamePath getStatic(XYCoord xyc)
```
ThatIsAUsername, line 28: I might call this `stayPut` or `to[Now]here` or 'dummy` or something just because "static" can mean things besides "stationary", especially in a programming context.

src/AI/AICombatUtils.java
```diff
@@ -54,8 +54,8 @@ public static double scoreAttackAction(Unit unit, GameAction action, GameMap map
     }
     else
     {
-      StrikeParams params = CombatEngine.calculateTerrainDamage(unit,
-          Utils.findShortestPath(unit, action.getMoveLocation(), map), targetLoc, map);
+      GamePath path = new Utils.PathCalcParams(unit, map).findShortestPath(action.getMoveLocation()); // Hmm. TODO: Add unit path getter to GameAction?
```
ThatIsAUsername, line 57: Since you kept the function in `Utils`: why instantiate a `PathCalcParams` instead of just calling the utility function to do it under the hood?

For that matter, is there an argument for putting that logic somewhere else so that you don't have to create an PCP object for the simple case where you just need a single path?

src/Engine/Utils.java
```diff
@@ -98,81 +98,163 @@ public static ArrayList<XYCoord> findUnloadLocations(GameMap map, Unit transport
...
         {
           dropoffLocations.add(loc);
         }
       }
     return dropoffLocations;
   }
 
-  /** Alias for {@link #findPossibleDestinations(XYCoord, Unit, GameMap) findPossibleDestinations()} **/
-  public static ArrayList<XYCoord> findPossibleDestinations(Unit unit, GameMap gameMap, boolean includeOccupiedSpaces)
+  public static class PathCalcParams
```
ThatIsAUsername, line 109: This is complex enough that we should probably push it into its own file.

src/Engine/Utils.java
```diff
@@ -531,11 +511,13 @@ public int getCachedDistance(XYCoord xyc)
...
-      int distance = path.getFuelCost(myUnit, myMap);
-      if( 0 == path.getPathLength() )
-        distance = Integer.MAX_VALUE;
+      Utils.PathCalcParams pcp = new Utils.PathCalcParams(myUnit, myMap);
+      pcp.setTheoretical();
+      pcp.start = start;
+      final GamePath path = pcp.findShortestPath(end);
+      int distance = Integer.MAX_VALUE;
+      if( null != path )
+        distance = path.getFuelCost(myUnit, myMap);
```
ThatIsAUsername, line 520: Did our zero-length path behavior change to return a null `Path` instead of an empty one?

ThatIsAUsername: Okay, based on the next commit it looks like it.

src/Engine/GamePath.java
ThislsAUsername: I was coming up with `getNull`, which doesn't really work.
I like `getDummy` better, but `stayPut` is real nice.

src/AI/AICombatUtils.java
ThislsAUsername: I intended to remove it entirely, but you cleverly wrote a bunch of test code I didn't feel like touching.
I guess it *is* a lot simpler at point-of-call to just call a function if you aren't doing anything funky.

There's an argument for almost anything, but the PCP is a nice bundle of params and a big part of this project was getting rid of our blob of overloads.
I don't love putting an extra object on the heap for bookkeeping we don't really "need", but the language makes it hard to avoid doing that without devolving into unlimited overload works.

I've also found I can pass the PCP into `expandSearchNode()` to save a few params.

src/Engine/Utils.java
ThislsAUsername: Yeah, I was tired of having two failure modes.

src/Engine/GamePath.java
ThatIsAUsername: >> I like `getDummy` better
>
> Implements `stayPut`

-\\_(^ ^)_/-

ThislsAUsername: The referent of "better" was my existing solution

ThatIsAUsername:
Cool. looks good.

PR #396, opened 2023-11-06
Game Start QoL

ThislsAUsername:
Defaulting High Funds and Fog seems useful, as is storing any shenanigans the user set up.
Also, I once watched a guy set up like 10 teams, then back out to tweak a game option (and lose all his picks), and I don't want to see that again.

I opted not to maintain selections if you back all the way out to the map picker menu, since I feel like that's less intuitive.

### DIFF
--- a/src/UI/GameOptionSetupController.java
+++ b/src/UI/GameOptionSetupController.java
@@ -1,6 +1,11 @@
 package UI;
 
 import Engine.GameScenario.TagMode;
+
+import java.util.HashMap;
+import java.util.Scanner;
+
+import Engine.ConfigUtils;
 import Engine.IController;
 import Engine.OptionSelector;
 import Terrain.Environment.Weathers;
@@ -9,6 +14,8 @@ import Units.UnitModelScheme;
 
 public class GameOptionSetupController implements IController
 {
+  public final int GAME_OPTIONS_CONFIG_KEY = -1; // Player indices start at 0, so -1 should be safe.
+
   private GameOption<Boolean> fowOption = new GameOptionBool("Fog of War", false);
   private GameOption<Integer> startingFundsOption = new GameOptionInt("Starting Funds", 0, 50000, 1000, 0);
   private GameOption<Integer> incomeOption = new GameOptionInt("Income", 250, 20000, 250, 1000);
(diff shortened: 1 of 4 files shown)


src/UI/GameOptionSetupController.java
```diff
@@ -40,6 +50,32 @@ public GameOptionSetupController(GameBuilder aGameBuilder)
...
+      while (s[si].isEmpty()) ++si; // Skip any sneaky empty splits
+
+      for( int i = 0; i < gameOptions.length; ++i )
+        if( s.length > si )
+          gameOptions[i].setSelectedOption(Integer.valueOf(s[si++]));
+    }
+    else
+    {
+      if( gameBuilder.mapInfo.mapName.startsWith("HF") )
+        incomeOption.setSelectedOption(3 + 4); // 250 = 0, increment 250
```
ThatIsAUsername, line 75: Why 3+4 instead of 7? Is 4 the normal setting and 7 is the normal HF setting?

src/UI/GameOptionSetupController.java
```diff
@@ -9,6 +14,8 @@
 
 public class GameOptionSetupController implements IController
 {
+  public final int GAME_OPTIONS_CONFIG_KEY = -1; // Player indices start at 0, so -1 should be safe.
```
ThatIsAUsername, line 17: Player indices? This class handles game options, not players, right?

ThatIsAUsername:
This is currently busted when you try to change COs.

src/UI/GameOptionSetupController.java
ThislsAUsername: 250*(1+3) is the normal 1k setting, so yes.

src/UI/GameOptionSetupController.java
ThislsAUsername: Yessir, but it goes in the same file as the player config options. They're numbered from 0, so -1 is safe.

src/UI/GameOptionSetupController.java
ThatIsAUsername: There you go with the 1+3 instead of 4 thing again. makes no sense. If you want to break it out this way, say why.

In fact, if you want to use magic numbers at all, say why.

ThatIsAUsername:
Coooooool

PR #397, opened 2023-11-06
DoR Pathing Fixes

ThislsAUsername:
Found some weirdness in a HF test game.

Launched planes couldn't actually pick actions in the UI, since carried units are at (-1,-1).
`isPathValid()` is only used by `SelectMoveLocation`, and "path valid" doesn't scream "unit is at start of path", to me... so I figured just moving the check to the code that wants it was a sensible plan.

Since `GamePath` wasn't telling the `MoveType` which team the unit was on, the BShip/Carrier-movetype was assuming it wasn't allowed in the port, and thus returned a fuel cost of 99.
This wasn't technically a big deal (it won't sink since it's at port), but it would prevent the BShip from shooting.

### DIFF
--- a/src/Engine/Utils.java
+++ b/src/Engine/Utils.java
@@ -177,9 +177,9 @@ public class Utils
 
   public static boolean isPathValid(Unit unit, GamePath path, GameMap map, boolean includeOccupiedSpaces)
   {
-    return isPathValid(new XYCoord(unit), unit, unit.CO.army, unit.getMoveFunctor(), Math.min(unit.getMovePower(map), unit.fuel), path, map, includeOccupiedSpaces);
+    return isPathValid(unit, unit.CO.army, unit.getMoveFunctor(), Math.min(unit.getMovePower(map), unit.fuel), path, map, includeOccupiedSpaces);
   }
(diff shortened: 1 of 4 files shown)


src/Engine/Utils.java
```diff
@@ -177,17 +177,17 @@ public static ArrayList<XYCoord> findFloodFillArea(XYCoord start, Unit mover, Ar
...
+  public static boolean isPathValid(Unit mover, Army team, MoveType fff, int initialFillPower, GamePath path, GameMap map, boolean includeOccupiedSpaces)
   {
     if( (null == path) || (null == fff) )
     {
       return false;
     }
 
     // Make sure the first waypoint is under the Unit.
-    if( path.getPathLength() <= 0 || !path.getWaypoint(0).GetCoordinates().equals(start) )
+    if( path.getPathLength() <= 0 )
```
ThatIsAUsername, line 190: Comment out of date

ThatIsAUsername: Also won't this check fail now for actions that a unit performs in situ? Units still "move" on a length-1 path to its current location before performing the action.

ThislsAUsername: Nope. A length-zero (or less) path is totally invalid, since the first point is the unit's starting position.

ThatIsAUsername:
Cool. This seems fine, in that DoR BShips can now shoot after parking

PR #398, opened 2023-11-12
Correct movement table definitions

ThislsAUsername:
It appears I had a little too much fun when I set these up initially.


ThatIsAUsername:
And to think I just trusted you to put in the correct values for the movement table. I mean, I'm gonna do it again, but still.

PR #400, opened 2023-11-24
Detail Pane updates + oversized sprite fixes

ThislsAUsername:
Implemented as part of the Fire Emblem jazz

### DIFF
--- a/src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
+++ b/src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
@@ -4,6 +4,7 @@ import java.awt.Color;
 import java.awt.Graphics;
 import java.awt.image.BufferedImage;
 import java.util.ArrayList;
+import java.util.Map;
 
 import Engine.GameEvents.GameEventListener.CacheInvalidationListener;
 import Engine.GameInstance;
@@ -11,6 +12,9 @@ import Engine.XYCoord;
 import Terrain.GameMap;
 import Terrain.MapLocation;
 import Terrain.TerrainType;
+import UI.UnitMarker;
+import UI.Art.SpriteArtist.MarkArtist.MarkingCache;
+import UI.UnitMarker.CustomStatData;
 import Units.Unit;
 
 /**
@@ -92,6 +96,17 @@ public class MapTileDetailsArtist
       if( unit.getCaptureProgress() > 0)
         unitAttrs.add(new AttributeArtist(SpriteLibrary.getCaptureIcon(unit.CO.myColor),
             map.getEnvironment(coord).terrainType.getCaptureThreshold()-unit.getCaptureProgress()));
+
+      final MarkingCache cache = MarkingCache.instance(map.game);
+      cache.setupMarkers();
+      for( UnitMarker m : cache.markers )
+      {
+        CustomStatData stat = m.getCustomStat(unit);
+        if (null == stat)
+          continue;
+        BufferedImage symbol = SpriteLibrary.getColoredMapTextSprites(stat.markColor).get(stat.mark);
+        unitAttrs.add(new AttributeArtist(symbol, stat.textColor, stat.text));
+      }
     }
 
     ///////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


ThislsAUsername:
Aight, so I think fixing oversized sprites in the various places is a bugfix for our current codebase: we don't prevent using bigger than 16x16 sprites for the main map sprite, and they already display correctly on the map, so...

I've also added the custom stats and cargo display features I added in the other branch, since we'll want the cargo eventually.
Custom stats I can maybe take or leave, to be fair. Maybe there's a better solution there.


I think the design for how the custom stats get accessed is a little icky, but I don't see an obvious DRY solution except "pull the implementation of the cache's `setupMarkers()` function out into a util function", which itself sort of begs to be cached...

I'm not super huge on the idea of displaying all the cargo in a line (it can get pretty silly pretty fast if you start stacking), but it's way better than not having the info.

ThislsAUsername:
It is indeed busted on `master`.
Seems I broke it with 936715fecd93c2f08ddd9ce68c42a477329eb058 ("Some bugs") when I refactored stuff while preventing sea units unloading from bridges and "simplified" the check. Derp.
True to form, I have pushed the fix to the wrong/other active branch.

That's why I made sure to do a DFS - if I see a lander with two APCs and one inf in it, I don't have any ambiguity.
You could construct a unit set that would have ambiguity, but I'm going to pretend that won't happen until I do it.

ThatIsAUsername:
We should probs add a test case for "can unload from current tile" - I'm a bit surprised we don't have one already.

I'm not gonna sweat the ambiguity thing for now I think since it's fairly easy to look at unload options, and that should mostly resolve ambiguities.

ThislsAUsername:
We had one for "unload to the tile the APC was on", but even having a test of the flavor you mean wouldn't have caught that problem - it's in the factory layer, not the action/event layer.
Pushed, and I checked that the test fails on `master`.

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -206,12 +206,13 @@ public void drawUnit(Graphics g, AnimState state, int imageIndex, int drawX, int
     // Draw the unit, facing the appropriate direction.
     if( flipImage && isStateFlippable(state) )
     {
-      g2d.drawImage(frame, drawX - shiftX + (frame.getWidth()), drawY + shiftY, -frame.getWidth(), frame.getHeight(), null);
+      int flippedDrawX = drawX + SpriteLibrary.baseSpriteSize;
+      g2d.drawImage(frame, flippedDrawX - shiftX, drawY + shiftY, -frame.getWidth(), frame.getHeight(), null);
```
ThatIsAUsername, line 210: Huh. Looking at the initialization of `shiftX`, it appears that it will always be zero, and does not match the comment ("center X") because the enforced order of operations is wrong.

That said - it looks like `flippedDrawX` assumes that sprites have the standard width. Otherwise shouldn't it be `drawX + SpriteLibrary.baseSpriteSize + ((frame.getWidth()-SpriteLibrary.baseSpriteSize) / 2)`? i.e., draw from the right bound for normal-sized sprites, and shift others by half the difference between normal and actual sprite width?

ThatIsAUsername: Maybe I should just make up some irregular sprites and try it to see.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
```diff
@@ -77,34 +80,29 @@ private static void generateOverlay(GameMap map, XYCoord coord)
...
+    int columnW = bufferPx + iconSize+ATTR_TEXT_SPACING + tileSize + bufferPx*2;
+    int panelW = columnW * (unitImageCache.size() + 1);
+    if( unitImageCache.size() > 0 )
+      panelW += bufferPx*2; // The panel gets N-1 right buffers
+
+    int numAttrs = terrainAttrs.size();
+    for( ArrayList<AttributeArtist> attrList : unitAttrCache )
+      numAttrs = Math.max(numAttrs, attrList.size());
+    //         upper buffer   tile   lower buffer   each attribute with a 1-px buffer
+    int panelH = iconSize + tileSize + bufferPx  +  numAttrs*(iconSize+1);
```
ThatIsAUsername, line 105: It seems like including `iconSize`and also `numAttrs*(iconSize+1)` should put us over by one, but it doesn't. Is there a "default" attribute that we draw but isn't counted in `numAttrs`?

ThislsAUsername: You're putting the +1 on the wrong thing.
It's `iconsize+1`, i.e. "the icon with 1 pixel of buffer"

ThislsAUsername: Oh, I misread. The extra `iconSize` is (as mentioned in the comment) the upper buffer of the panel (above the unit/terrain image).

src/UI/UnitMarker.java
```diff
@@ -43,4 +43,23 @@ public MarkData(char mark, Color color)
...
+    {
+      this.mark      = mark;
+      this.markColor = markColor;
+      this.textColor = textColor;
+      this.text      = text;
+    }
+  }
+  public default CustomStatData getCustomStat(Unit unit)
+  {
+    return null;
```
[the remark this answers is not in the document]
ThislsAUsername: All that's required is an implementing class, yes.
Returning `null` is correct behavior in other cases (much like the `\0` we have for the marking).

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
```diff
@@ -155,30 +170,40 @@ private static void drawColumn(Graphics g, BufferedImage image, ArrayList<Attrib
...
+      this(image, Color.white, ""+Math.min(99, value));
+    }
+    public AttributeArtist(BufferedImage image, Color textColor, String value)
     {
       icon = image;
-      value = quantity;
+      Map<Character, BufferedImage> allChars = SpriteLibrary.getColoredMapTextSprites(textColor);
+      text = allChars.get('A');
+      int charWidth = text.getWidth();
+      text = SpriteLibrary.createTransparentSprite(charWidth*ATTR_TEXT_COUNT, text.getHeight());
```
ThatIsAUsername, line 203: always three characters?

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
```diff
@@ -155,30 +170,40 @@ private static void drawColumn(Graphics g, BufferedImage image, ArrayList<Attrib
...
+      Graphics g = text.getGraphics();
+
+      int drawX = 0, drawY = 0;
+      int maxIter = Math.min(2, value.length());
```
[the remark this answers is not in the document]
ThislsAUsername: Because that's the spacing we had before and it looks aight?

src/UI/Art/SpriteArtist/UnitSpriteSet.java
ThatIsAUsername: Ah, okay. I don't know why I didn't think of the wider movement sprites, or the fact that I was reinventing the wheel I was complaining about. Sigh.

src/UI/Art/SpriteArtist/MapTileDetailsArtist.java
ThatIsAUsername: Oooh, okay. Right. The tile art just overlaps into that space, which makes it less apparent, and is also why we need it.

ThatIsAUsername:
Cool. looks good.

PR #401, opened 2023-12-09
Integer HP

ThislsAUsername:
We discussed changing HP to always be denominated in %, so here it is.

### DIFF
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -258,7 +258,7 @@ public class AICombatUtils
       neededAttacks.put(xyc, null);
     }
 
-    double damage = findMultiHitKill(gameMap, target, attackers, neededAttacks, 0);
+    int damage = findMultiHitKill(gameMap, target, attackers, neededAttacks, 0);
     if( damage >= target.getHP() )
     {
       // Prune excess attacks and empty attacking spaces
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -270,7 +270,7 @@ public class AICombatUtils
           neededAttacks.remove(space);
           continue;
         }
-        double thisShot = CombatEngine.simulateBattleResults(attacker, target, gameMap, space, CalcType.PESSIMISTIC).defender.getPreciseHPDamage();
+        int thisShot = CombatEngine.simulateBattleResults(attacker, target, gameMap, space, CalcType.PESSIMISTIC).defender.getPreciseHPDamage();
         if( target.getHP() <= damage - thisShot )
         {
           neededAttacks.remove(space);
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -293,19 +293,19 @@ public class AICombatUtils
    * @param pDamage The cumulative base damage done by those mandatory attacks
    * @return The cumulative base damage of all attacks already in the neededAttacks
    */
-  public static double findMultiHitKill(
+  public static int findMultiHitKill(
                                 GameMap gameMap, Unit target,
                                 Collection<Unit> attackCandidates,
                                 Map<XYCoord, Unit> neededAttacks,
-                                double pDamage)
+                                int pDamage)
   {
     // Base case; we found a kill
-    if( pDamage >= target.getPreciseHP() )
+    if( pDamage >= target.getHP() )
     {
       return pDamage;
     }
 
-    double damage = pDamage;
+    int damage = pDamage;
     // Iterate through the attack spaces, and try filling all spaces recursively from each one
     for( XYCoord xyc : neededAttacks.keySet() )
     {
@@ -330,12 +330,12 @@ public class AICombatUtils
         if( movePath != null )
         {
           neededAttacks.put(xyc, unit);
-          double thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc, CalcType.PESSIMISTIC).defender.getPreciseHPDamage();
+          int thisDamage = CombatEngine.simulateBattleResults(unit, target, gameMap, xyc, CalcType.PESSIMISTIC).defender.getPreciseHPDamage();
 
           thisDamage = findMultiHitKill(gameMap, target, attackCandidates, neededAttacks, damage + thisDamage);
 
           // If we've found a kill, we're done
-          if( thisDamage >= target.getPreciseHP() )
+          if( thisDamage >= target.getHP() )
           {
             damage = thisDamage;
             break;
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -80,7 +80,7 @@ public class AICombatUtils
     ArrayList<Utils.SearchNode> destinations = pcp.findAllPaths();
     for( WeaponModel wep : unit.model.weapons )
     {
-      int damage = (null == target)? 1 : wep.getDamage(target) * unit.getHP() / UnitModel.MAXIMUM_HP;
+      int damage = (null == target)? 1 : wep.getDamage(target) * unit.getHealth() / UnitModel.MAXIMUM_HP;
       if( null == target || unit.canTarget(target) )
       {
         if( !wep.canFireAfterMoving )
@@ -177,7 +177,7 @@ public class AICombatUtils
                                    Collection<Unit> attackCandidates,
                                    Collection<XYCoord> excludedSpaces)
   {
-    if( target.getHP() < 1 ) // Try not to pick fights with zombies
+    if( target.getHealth() < 1 ) // Try not to pick fights with zombies
       return null;
     if( attackCandidates.size() < 1 )
       return null;
@@ -259,7 +259,7 @@ public class AICombatUtils
     }
 
     int damage = findMultiHitKill(gameMap, target, attackers, neededAttacks, 0);
-    if( damage >= target.getHP() )
+    if( damage >= target.getHealth() )
     {
       // Prune excess attacks and empty attacking spaces
       for( XYCoord space : new ArrayList<XYCoord>(neededAttacks.keySet()) )
@@ -271,7 +271,7 @@ public class AICombatUtils
           continue;
         }
         int thisShot = CombatEngine.simulateBattleResults(attacker, target, gameMap, space, CalcType.PESSIMISTIC).defender.getPreciseHPDamage();
-        if( target.getHP() <= damage - thisShot )
+        if( target.getHealth() <= damage - thisShot )
         {
           neededAttacks.remove(space);
           damage -= thisShot;
@@ -300,7 +300,7 @@ public class AICombatUtils
                                 int pDamage)
   {
     // Base case; we found a kill
-    if( pDamage >= target.getHP() )
+    if( pDamage >= target.getHealth() )
     {
       return pDamage;
     }
@@ -335,7 +335,7 @@ public class AICombatUtils
           thisDamage = findMultiHitKill(gameMap, target, attackCandidates, neededAttacks, damage + thisDamage);
 
           // If we've found a kill, we're done
-          if( thisDamage >= target.getHP() )
+          if( thisDamage >= target.getHealth() )
           {
             damage = thisDamage;
             break;
--- a/src/AI/AICombatUtils.java
+++ b/src/AI/AICombatUtils.java
@@ -66,21 +66,21 @@ public class AICombatUtils
   /**
    * @return The area and severity of threat from the unit, against the specified target type
    */
-  public static Map<XYCoord, Double> findThreatPower(GameMap gameMap, Unit unit, UnitModel target)
+  public static Map<XYCoord, Integer> findThreatPower(GameMap gameMap, Unit unit, UnitModel target)
   {
     XYCoord origin = new XYCoord(unit.x, unit.y);
     return findThreatPower(gameMap, unit, origin, target);
   }
-  public static Map<XYCoord, Double> findThreatPower(GameMap gameMap, Unit unit, XYCoord origin, UnitModel target)
+  public static Map<XYCoord, Integer> findThreatPower(GameMap gameMap, Unit unit, XYCoord origin, UnitModel target)
   {
-    Map<XYCoord, Double> shootableTiles = new HashMap<XYCoord, Double>();
+    Map<XYCoord, Integer> shootableTiles = new HashMap<>();
     PathCalcParams pcp = new PathCalcParams(unit, gameMap);
     pcp.start = origin;
     pcp.includeOccupiedSpaces = true; // We assume the enemy knows how to manage positioning within his turn
     ArrayList<Utils.SearchNode> destinations = pcp.findAllPaths();
     for( WeaponModel wep : unit.model.weapons )
     {
-      double damage = (null == target)? 1 : wep.getDamage(target) * unit.getHPFactor();
+      int damage = (null == target)? 1 : wep.getDamage(target) * unit.getHP() / UnitModel.MAXIMUM_HP;
       if( null == target || unit.canTarget(target) )
       {
         if( !wep.canFireAfterMoving )
@@ -88,7 +88,7 @@ public class AICombatUtils
           UnitContext uc = new UnitContext(gameMap, unit, wep, null, origin);
           for (XYCoord xyc : Utils.findLocationsInRange(gameMap, origin, uc))
           {
-            double val = damage;
+            int val = damage;
             if (shootableTiles.containsKey(xyc))
               val = Math.max(val, shootableTiles.get(xyc));
             shootableTiles.put(xyc, val);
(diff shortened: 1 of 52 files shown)


src/Units/UnitState.java
```diff
@@ -87,64 +87,62 @@ public boolean hasMaterials()
...
-    return ((double) input) / 10;
-  }
-  public static int healthFromHP(double input)
-  {
-    return (int) (input * 10);
+    if( health >= 0 )
+      //     "round up", then kill the last digit
+      return (health + 9) / 10 * 10;
+    // Truncation rounds toward zero, so we need to round down for negative values
+    return (health - 9) / 10 * 10;
```
ThatIsAUsername, line 106: Is "units with negative health" a common use case?

ThislsAUsername: I didn't think so either, but then I remembered we use `UnitContext`->`UnitState` as part of `UnitDelta` to store combat damage.
If this didn't work right, we'd potentially get pretty weird results from simulated damage calcs (which allow overkill)

src/Units/UnitState.java
```diff
@@ -87,64 +87,62 @@ public boolean hasMaterials()
...
       health = Math.max(0, health);
     return getHP() - before;
   }
 
   /**
    * Increases HP by the specified amount.
-   * <p>Enforces a minimum of 0.1, and a maximum (optional) of MAXIMUM_HP.
-   * <p>When healing, rounds health up to a whole HP (e.g. 2.5 + 2 = 4.5 -> 5.0)
+   * <p>Enforces a minimum of 1, and a maximum (optional) of MAXIMUM_HP.
+   * <p>When healing, rounds health up to a whole HP (e.g. 25 + 20 = 45 -> 50)
```
ThatIsAUsername, line 133: Is there anywhere we define the relationship between health and HP? If I didn't already know what I was looking at, this comment could be pretty confusing.

ThatIsAUsername: Ah, back on line 27-28 I guess, where `health` is defined.

ThislsAUsername: They're uh... not really different anymore? I guess I should say "GUI HP" or something?
I dunno. Nomenclature is hard

src/CommandingOfficers/Ave.java
```diff
@@ -644,7 +644,7 @@ private static class OblidoAbility extends CommanderAbility
     private static final int OBLIDO_COST = 8;
     private static final int OBLIDO_BUFF = 20;
     private static final int OBLIDO_RANGE = 2;
-    private static final int OBLIDO_DAMAGE = 2;
+    private static final int OBLIDO_DAMAGE = 20;
```
ThatIsAUsername, line 647: So is the rule "Always work in terms of health (0-100, typically) and only convert to HP when rendering"?

src/CommandingOfficers/Commander.java
```diff
@@ -300,17 +300,17 @@ public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCo
     if( minion == null || enemy == null )
       return 0;
 
-    int myHPLoss  = minion.getHPDamage();
-    int myHPDealt = enemy.getHPDamage();
+    int guiHPLoss  = minion.getHPDamage() / 10;
+    int guiHPDealt =  enemy.getHPDamage() / 10;
```
ThatIsAUsername, line 304: Shouldn't `getHPDamage()` already be returning the (0-10) number, thus no need to divide by 10?

src/CommandingOfficers/Ave.java
ThislsAUsername: Soooooort of.
The problem is that we have to handle (almost) everything as if the rounded value is the real one.
But yes, everything refers to raw percents now.

src/CommandingOfficers/Commander.java
```diff
@@ -300,17 +300,17 @@ public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCo
...
 
     int power = 0; // value in funds of the charge we're getting
 
     // Add up the funds value of the damage done to both participants.
-    power += myHPLoss / UnitModel.MAXIMUM_HP * minion.unit.getCost();
+    power += guiHPLoss * minion.unit.getCost() / 10;
```
ThatIsAUsername, line 309: Is this `10` the assumed maximum HP value, or the conversion factor between health and HP, or something different? In any case, there's a better way to represent this.

src/CommandingOfficers/RuinedCommander.java
```diff
@@ -139,11 +139,11 @@ public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCo
     if( !isInZone(chargeSource) )
       return 0;
 
-    double myHPDealt = enemy.getHPDamage();
+    int guiHPDealt = enemy.getHPDamage() / 10;
```
ThatIsAUsername, line 142: Already in HP so why divide, also maaaagic.

src/CommandingOfficers/Commander.java
ThislsAUsername: Nope, that's the base damage from the damage chart.
The only thing in (0,10] now is `getHPFactor()` and whatever is explicitly converted.

src/CommandingOfficers/Tech.java
```diff
@@ -197,7 +197,7 @@ protected void revert(MapMaster gameMap)
         final int maxPercent = UnitModel.MAXIMUM_HP*10;
         for( Unit u : unitsOverCharged.get(myCommander) )
         {
-          final int preciseHP = (int) (u.getPreciseHP() * 10);
+          final int preciseHP = u.health;
```
ThatIsAUsername, line 200: Maybe I should stop complaining about places where `health` and  `HP` are interchanged, despite being off from each other by an order of magnitude. Maybe.

src/CommandingOfficers/Commander.java
ThislsAUsername: 10 is the assumed maximum *GUI* HP value, and it's GUI HP because it was just converted above.
I could multiply by 10 and then divide by `UnitModel.MAXIMUM_HP`, but bleeeeeh.

src/CommandingOfficers/RuinedCommander.java
ThislsAUsername: Because it's in [0,100]

src/CommandingOfficers/Tech.java
ThislsAUsername: They're not anymore. That's the point.

ThatIsAUsername: Okay. If they are the same now (which is good!) then the random magic factors of ten I see littered around are perhaps more confusing.

src/Units/UnitState.java
ThatIsAUsername: Oh. Then the aforementioned comment should probably be updated?

src/AI/AICombatUtils.java
```diff
@@ -66,29 +66,29 @@ public static double scoreAttackAction(Unit unit, GameAction action, GameMap map
...
-    Map<XYCoord, Double> shootableTiles = new HashMap<XYCoord, Double>();
+    Map<XYCoord, Integer> shootableTiles = new HashMap<>();
     PathCalcParams pcp = new PathCalcParams(unit, gameMap);
     pcp.start = origin;
     pcp.includeOccupiedSpaces = true; // We assume the enemy knows how to manage positioning within his turn
     ArrayList<Utils.SearchNode> destinations = pcp.findAllPaths();
     for( WeaponModel wep : unit.model.weapons )
     {
-      double damage = (null == target)? 1 : wep.getDamage(target) * unit.getHPFactor();
+      int damage = (null == target)? 1 : wep.getDamage(target) * unit.getHP() / UnitModel.MAXIMUM_HP;
```
ThatIsAUsername, line 83: `getHPFactor` still exists. Why break it out this way?

src/AI/JakeMan.java
```diff
@@ -730,7 +731,7 @@ private boolean isDudeFree(GameMap gameMap, Unit unit, XYCoord xyc, boolean amAt
           if( unitMapFriendly.get(counter).containsKey(coord) )
             counterPower = unitMapFriendly.get(counter).get(coord);
           if( counterIsMeAndIAmPeaceful )
-            counterPower -= PEACEFUL_SELF_THREAT_RATIO * unit.getHPFactor();
+            counterPower -= (PEACEFUL_SELF_THREAT_RATIO * unit.getHPFactor()) / 10;
```
ThatIsAUsername, line 734: Isn't HP Factor already [0, 1]? Why divide by 10 again?

src/UI/COStateInfo.java
```diff
@@ -44,7 +44,7 @@ public COStateInfo(GameMap map, Army thisArmy)
           unitCount++;
           if( resident.model.isTroop() )
             vehCount++;
-          unitFunds += resident.getCost() * resident.getHPFactor();
+          unitFunds += resident.getCost() * resident.getHPFactor() / 10;
```
ThatIsAUsername, line 47: Same question about HP Factor and diving again by ten

src/CommandingOfficers/Tech.java
ThislsAUsername: The issue is that a lot of our code really wants numbers in [0,10], so now I have to divide by 10. It feels weird to define a constant for "ten percent" or something.
I dunno, I feel like it's clear enough when HP/health is known to be nominally in [0,100]

src/AI/AICombatUtils.java
ThislsAUsername: I changed it to [0,10] instead of [0,1], which was probably ill-advised... but made a few things more convenient.
And it banishes floating point from a few calcs, which I see as a win.

src/AI/JakeMan.java
ThislsAUsername: It got changed too, because I rage against IEE754

src/Units/UnitState.java
```diff
@@ -157,48 +155,27 @@ public int alterHP(int change, boolean allowOver)
...
+      realChange = Math.min(change, newHP - health);
     }
 
-    health = Math.max(1, health + healthFromHP(realChange));
+    health = Math.max(1, health + realChange);
     // Round HP up, if healing
-    if( change >= 0 )
-      health = healthFromHP(getHP());
+    if( roundUp && change >= 0 )
+      health = getHP();
```
ThatIsAUsername, line 166: Strange to use `getHP` here instead of `roundHP`. Shrug.

src/Units/UnitState.java
```diff
@@ -22,10 +22,11 @@ public abstract class UnitState implements Serializable
...
   /**
-   * HP determines the current actual durability of a unit.
-   * It's typically in range [1-10]. A unit at 0 HP is dead.
-   * Health is HP value as a percentage, thus ~10x the HP value.
-   * When determining HP, health must always be rounded up.
+   * HP/health determines the current durability of a unit.<p>
+   * It's typically in range [1,100]. A unit at 0 HP is dead.<p>
+   * HP is used to scale unit effectiveness, and normally rounds up to the next 10 for this purpose.<p>
+   * HP is also shown only in this [1,10] scale, so the last digit of the health number is considered hidden information.<p>
+   * As such, direct access of this value should only be used when you have a specific reason to choose it over getHP()<p>
```
ThatIsAUsername, line 29: What is "It" that is typically in range [1,100]? `HP/health', or one or or the other? Pronouns are the worst.
You are using two apparent synonyms in an equation without defining either one (or rather, defining one of them and not specifying which).

You can't say "in this [1,10] scale" when you haven't yet introduced the scale. You do mention rounding in the previous sentence, but that is in the context of the [1,100] range.

Make sure I have this straight:
`health` is just public with no accessor, and is typically in [1,100], and represents the true/precise value.
`getHP()` returns a number in [1,100], but ceilinged to the nearest 10.
`getHPFactor()` returns a number in [1,10]? If so that's just named wrong (unless health goes to 1000%)
To get the health value displayed in the UI, we have to `getHP()/10` (or I guess `getHPFactor()`, but I refuse)

I don't like having a `getFactor()` that doesn't return a factor, or a `getHP()` that returns something different than what is called HP in the UI. I do like keeping health in the range [1,100] until it is drawn, but names matter.

src/CommandingOfficers/Commander.java
ThatIsAUsername: Oh. Well that's just confusing naming all around.

src/Units/UnitState.java
```diff
@@ -157,48 +155,27 @@ public int alterHP(int change, boolean allowOver)
...
   /**
-   * Increases *fractional health* by the specified amount.
-   * <p>Enforces a minimum of 0.1, and a maximum (optional) of MAXIMUM_HP.
+   * Increases health by the specified amount.
+   * <p>Enforces a minimum of 1, and a maximum (optional) of MAXIMUM_HP.
    * <p>Does not round.
    * <p>Use this when you want precise non-combat health changes, or want to heal without rounding up.
    * @return the change in HP
    */
   public int alterHealthPercent(int percentChange)
```
ThatIsAUsername, line 178: Is there any reason to have both `alterHP` and `alterHealthPercent`? I could see it if `alterHP` accepted and returned [1,10] and `alterHealthPercent` accepted and returned [1,100] maybe, but right now they are just clones.

src/Units/UnitState.java
ThislsAUsername: HP/health. Both. I think(?) I've settled on "GUI HP" for what we see in the GUI. Fair criticism.
The intent was to show that they are indeed synonyms. Whoops.

A number in [0,100] that doesn't have a ones digit can be encoded in [0,10]. But fair enough.

Yes.
Yes.
Yes. I would rather delete it entirely than bring floating point back.
Yes, and fair enough.

Fair enough.
I was under the impression that HP and health being different things was a source of confusion.

So... do I just rename all the functions to talk about "health" and then we bring back the health/HP divide?
And my new `getHPFactor()` can become just `getHP()`?

ThislsAUsername: We could also bring back `getPreciseHealth()`, I guess, though I kind of hate having nothing-getters.
~~clearly, it should take a string argument that justifies why you're calling it~~

src/Units/UnitState.java
ThislsAUsername: It's effectively an overload with different connotations, as is documented.
It doesn't round up HP when healing.
The single-bool overload of `alterHP()` allows overhealing, not the rounding toggle.

src/Units/UnitState.java
ThislsAUsername: Same math either way, lol

src/Units/UnitState.java
ThatIsAUsername: I think it's good to standardize on [0, 100] for calculations to avoid floating point nonsense. I'd just call it `health` though all around so we don't create ambiguity with `HP`, which is overloaded for the GUI context (and it feels silly to have `HP` and `GUI HP` be different things).

I'm fine with removing  `getHPFactor`; the name is ambiguous at best. We can keep `getHP` if convenient I guess? Or banish that too, update calcs that currently use it to work in terms of health instead, and just divide by 10 before drawing.

src/Units/UnitState.java
ThatIsAUsername: Okay, I see the rationale for having both. I still feel like the naming could be tightened (maybe it's just a me problem that I can't look at `HP` and see [0-100]).
Also seems odd that it takes and applies a precise change, but returns a rounded value. Though I guess the rounded value is what always matters external to `Unit`.

Aside: Documentation here refers to an optional cap which isn't supported (no Boolean toggle param).

ThislsAUsername: Yeah. "HealthPercent" is a non-terse way to write it, which signals to me personally that it's an alternate function that does a non-expected thing (which, in this case, is not rounding up).
It's probably not fair to expect others to uh... read the tea leaves like that.

I've renamed it to [`alterHealthNoRound`](https://github.com/ThislsAUsername/DefendPeace/pull/401/commits/bbd41a202f5a35258c243b21d0e21630e3a1671a), and deleted the documentation since it looked like a restatement of the function name.

ThatIsAUsername:
Sweet. There are technically still quite a few possibly-erroneous mentions of HP floating around, but I think I care a lot less about that when they aren't part of the interface defining the terms. Might be good to chase some of that down if convenient, but otherwise I think it's in pretty good shape.

ThislsAUsername:
Probably take a gander at the Muriel change. I noticed the numbers were inconsistent (which I assumed was my fault), so I switched her back to using HP.

src/Engine/GameEvents/HealUnitEvent.java
```diff
@@ -44,21 +44,22 @@ public GameEventQueue sendToListener(GameEventListener listener)
   public void performEvent(MapMaster gameMap)
   {
     if (null == payer)
-      unit.alterHealth(repairPowerHP, canOverheal);
-    else if( unit.isHurt() )
```
ThislsAUsername, line 48: Yeeted this check since the only thing it can do is make me fail to round up when I should sometimes.

ThatIsAUsername:
> Probably take a gander at the Muriel change. I noticed the numbers were inconsistent (which I assumed was my fault), so I switched her back to using HP.

Inconsistent how?

ThatIsAUsername:
Also it looks like Wally forgot how to move landers. Seems to be a problem on `master` too.

ThislsAUsername:
> Inconsistent how?

Some of the HP numbers were dividing by 100.0, some by 10.0.
I made them all match in the most recent change to her file.
> Also it looks like Wally forgot how to move landers. Seems to be a problem on `master` too.

Yeah. He's in a bad state partly because I'm trying to avoid merge conflicts with my ongoing changes.

PR #402, opened 2023-12-16
Magic is dumb

ThatIsAUsername:
Not sure what you mean about behavior being inconsistent, and I'm not sure how a shift from HP->Health values would change much (since most of what Muriel does is in terms of ratios and multiplied factors, but I did notice some magic HP values hanging around.

### DIFF
--- a/src/AI/Muriel.java
+++ b/src/AI/Muriel.java
@@ -91,6 +91,13 @@ public class Muriel implements AIController
   private final double COST_EFFECTIVENESS_HIGH = 1.25;
   private final double INFANTRY_PROPORTION = 0.5;
 
+  private class UnitHealthThresholds
+  {
+    static final int HEALTHY = 80;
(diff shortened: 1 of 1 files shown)


src/AI/Muriel.java
```diff
@@ -495,7 +502,7 @@ private boolean queueUnitAction(GameMap gameMap, Unit unit, boolean allowDeferri
 
     //////////////////////////////////////////////////////////////////
     // See if there's something to capture (but only if we are moderately healthy).
-    if( unit.getHealth() > 7 )
+    if( unit.getHealth() >= UnitHealthThresholds.ACTION_READY )
```
ThatIsAUsername, line 505: Technically a change in logic, but consistent with line 560 (and my remembered intent that the unit can start a capture if it it would take three or fewer turns to complete).

PR #403, opened 2023-12-17
Triple Andybiotic

ThislsAUsername:
Adds the 3 Andys (Andies?) that act roughly the same as one another (AW2/3/BW).

To do this, I added the explicit star dealie and added a single source of truth for star cost across powers.

As of this PR, we no longer assume ability energy is denoted in funds - it just uses whatever backend energy calc the CO wants to (and the abilities just need to get with the program).
This isn't super amazing w.r.t. the possibility of these things not being in sync, but at least it'd be pretty obvious one way or the other.

I didn't link Ave's star costs since she has different scaling for each power.


ThatIsAUsername:
> As of this PR, we no longer assume ability energy is denoted in funds - it just uses whatever backend energy calc the CO wants to (and the abilities just need to get with the program).
This isn't super amazing w.r.t. the possibility of these things not being in sync, but at least it'd be pretty obvious one way or the other.

What, uh. What _is_ "the program", if it's no longer funds? And sure, individual CO's can override power charge if they want, but most don't, so there has to be some kind of standard.

Note that I haven't actually looked at the code yet, but I'm already more confused than before I got here.

ThislsAUsername:
> What, uh. What _is_ "the program", if it's no longer funds?

"The program" here being: The CO's code determines how much charge you get from a given event.
This is not necessarily in sync with the scale of values the abilities are using for their costs (but the CO also chooses that, so they _should_ be).
> And sure, individual CO's can override power charge if they want, but most don't, so there has to be some kind of standard.

There are 3.5 cases:
Funds (base of 9k/star or AW1's increments of 5k)
AWDS charge (arbitrary 100 points/star since I don't know the internal numbers)
HP (DoR, 10 points/star)

In this branch, AW2 and AWDS charge logic are added in new parent CO classes.
The main CO class still implements our "default" charge method, with funds + HP -> funds.
It also adds code for the UI charge bar, so it can handle different base star costs.

Before, we were converting all charge into the scale of funds... which works, but isn't the most readable.
The DoR charge branch off of this one shows 1 energy for each HP you need to kill for your zone, which is very nice QoL (rather than having to ignore a bunch of zeroes).
I think readability probably helps more than having the option to do other silly things in the future, so yeah.

> Note that I haven't actually looked at the code yet, but I'm already more confused than before I got here.

Sorry.

ThatIsAUsername:
> The DoR charge branch off of this one

So that change is _not_ in this PR? Is that next in the queue?

ThislsAUsername:
> > The DoR charge branch off of this one
> 
> So that change is _not_ in this PR? Is that next in the queue?

Correct. The PR is up, and pointed at this branch.
FWIW, I can probably split the three Andys into separate PRs if that would be easier for you to review.

src/CommandingOfficers/Commander.java
```diff
@@ -50,8 +50,9 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
   public static final int CHARGERATIO_HP = 100; // Funds value of 10 HP damage dealt, for the purpose of power charge
   public int incomeAdjustment = 0; // Commander subclasses can increase/decrease income if needed.
   private int myAbilityPower = 0;
+  public boolean roundUpRepairs = true; // I blame AWBW
```
ThatIsAUsername, line 53: Does AWBW not? That sounds like a bug on them.

src/CommandingOfficers/CommanderAbility.java
```diff
@@ -52,18 +98,15 @@ public String toString()
    */
   protected void adjustCost()
   {
-    // Increase the cost of this ability for next time to mitigate spam and
-    // accommodate the presumably-growing battlefront.
-    // Cost grows by at least one, and at most 10% of the current cost.
-    myPowerCost += Math.max(myPowerCost*0.1, 1);
+    costBasis.numCasts += 1;
   }
```
ThatIsAUsername, line 102: Is there a reason to keep  `adjustCost` around at all now that it isn't actually adjusting the cost? Seems like you could just inline the contents in `activate`.

ThislsAUsername: Ave overloads it.

src/CommandingOfficers/Ave.java
```diff
@@ -655,14 +653,10 @@ private static class OblidoAbility extends CommanderAbility
...
       AIFlags = PHASE_TURN_START | PHASE_TURN_END;
-    }
-
-    @Override
-    protected void adjustCost()
-    {
       // One of the big benefits of this power is deforestation, since trees get in Ave's way.
       // Trees are only removed once, so we'll increase cost more slowly to counteract the decreased utility.
-      myPowerCost += 0.5 * CHARGERATIO_FUNDS;
+      costBasis.starRatioPerCast = 555; // 8 stars * 555 = +4440 per cast, vs the old 9000/2=4500 per cast
```
ThatIsAUsername, line 658: Obvious question - why make it different-but-almost-the-same?

src/CommandingOfficers/Commander.java
ThislsAUsername: It's been known for years. They ain't fixing it. Same deal with the way unloading works (which I may or may not try to accommodate)

src/CommandingOfficers/Ave.java
ThislsAUsername: 4500/8 = 562.5, which doesn't seem representable in the integers.
I guess I could have done something like what Nix does - half the star ratio, double the star count, and increase stars by 1 per cast.
I figured you'd have an opinion either way, and it's your CO, and you're reviewing the code. I just picked something so I could move on.
(and I changed everyone else to have linked ability costs, so you may or may not want to adjust her stuff too)

src/CommandingOfficers/Commander.java
```diff
@@ -251,6 +253,7 @@ public void activateAbility(CommanderAbility ability, MapMaster map)
     myActiveAbility = ability;
   }
 
+  /** Do not use this unless you ask the Commander how much energy to input */
```
ThatIsAUsername, line 256: example?

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,64 @@
...
+import Engine.UnitMods.UnitDamageModifier;
+import Engine.UnitMods.UnitModifier;
+import Terrain.MapMaster;
+import Units.Unit;
+import Units.UnitDelta;
+import java.util.ArrayList;
+import CommandingOfficers.CommanderInfo.InfoPage;
+
+/** Base class for AW2/3 COs (AW1 is special) */
+public abstract class TrilogyCommander extends Commander
```
ThatIsAUsername, line 13: Is `TrilogyCommander` appropos if it only represents 2&3? Just splitting hairs here ofc.

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,85 @@
...
+{
+  private static final long serialVersionUID = 1L;
+  public static final InfoPage TRILOGY_MECHANICS_BLURB = new InfoPage(
+      "CO powers never charge while a power is active.\n"
+    + "On each power activation, each star in your meter costs 1/5 more of the base star cost.\n"
+    + "This continues for 9 activations, after which the star cost settles at double the base star cost.\n"
+    + "CO power activation deducts the *new* power cost rather than the old one.\n"
```
ThatIsAUsername, line 20: Wait, what? That seems wrong. Do the carts actually do that? Might explain some things.

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,85 @@
...
+  {
+    private static final long serialVersionUID = 1L;
+    UnitModifier genericAttack;
+    UnitModifier genericDefense;
+
+    protected TrilogyAbility(int genericAtt, int genericDef, Commander commander, String name, int cost, CostBasis basis)
+    {
+      super(commander, name, cost, basis);
+      genericAttack  = new UnitDamageModifier(genericAtt);
+      genericDefense = new UnitDamageModifier(genericDef);
```
ThatIsAUsername, line 64: copypasta?

src/CommandingOfficers/Commander.java
ThislsAUsername: `Army`'s listeners for combat and mass damage. They call the functions in `Commander` that assess a charge value to add.

src/CommandingOfficers/TrilogyCommander.java
ThislsAUsername: This is also a known AWBW difference.
And yes, it's very trippy that Hachi can COP twice with no extra charge for his 9th and 10th casts (on his 3/5 meter)

src/CommandingOfficers/TrilogyCommander.java
ThislsAUsername: Copied from what?

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,85 @@
...
+      modList.add(genericAttack);
+      modList.add(genericDefense);
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      // Change my CO's energy such that it's as if this power cost the new cost rather than the actual cost
+      int oldCost = baseStars * costBasis.calcCostPerStar(costBasis.numCasts - 1);
+      int newCost = costBasis.calcCost(baseStars);
```
ThatIsAUsername, line 81: I rarely like using "old" and "new" as modifiers - unless the reader is intimately familiar with what is going on, it can be hard to know when "new" becomes "old". In this case, "old" I think is the cost at the time the power is activated, rather than the cost at the prior activation. So I'd probably lobby for more annoyingly verbose names, such as `costAtActivation` and `actualCostBeingChargedBecauseTrilogyCartsAreBugged` just to make things super clear.

src/CommandingOfficers/TrilogyCommander.java
ThatIsAUsername: But only on the 9th and 10th casts? That seems odd either way, since this nuance should always work against the player.

res/co/Andy_3.png
ThatIsAUsername, line 1: This looks identical to Andy_BW except for having extra blank space.

src/CommandingOfficers/AW2/AW2Commander.java
```diff
@@ -0,0 +1,32 @@
...
+    super(info, rules);
+  }
+
+  protected abstract static class AW2Ability extends TrilogyAbility
+  {
+    private static final long serialVersionUID = 1L;
+
+    protected AW2Ability(Commander commander, String name, int cost, CostBasis basis)
+    {
+      super(0, 10, commander, name, cost, basis);
```
ThatIsAUsername, line 29: magic bad

res/co/Andy_3.png
ThislsAUsername: Yeah. I just didn't crop it since this is the one Jay made.

src/CommandingOfficers/AW2/AW2Commander.java
ThislsAUsername: These ain't ever changing, and the only other reference is in the mechanics blurb.

src/CommandingOfficers/TrilogyCommander.java
ThislsAUsername: Do you want to read `AWTwoThreeCommander`?

src/CommandingOfficers/AW2/AW2Commander.java
```diff
@@ -0,0 +1,32 @@
...
+import CommandingOfficers.CommanderInfo;
+import CommandingOfficers.TrilogyCommander;
+import CommandingOfficers.CommanderInfo.InfoPage;
+import Engine.GameScenario;
+
+public abstract class AW2Commander extends TrilogyCommander
+{
+  private static final long serialVersionUID = 1L;
+  public static final InfoPage AW2_MECHANICS_BLURB = new InfoPage(
+      "Gain +10 defense on activating a power.\n"
```
ThatIsAUsername, line 13: It's strange to have this separate from the advertised power stats. Like, is it whatever the power claims, plus an additional 10 defense? And I guess now we might have to ask what kind of defense it's adding, too.

src/CommandingOfficers/AW2/OS/Andy.java
```diff
@@ -0,0 +1,122 @@
...
+  {
+    private static final long serialVersionUID = 1L;
+    private static final String NAME = "Hyper Repair";
+    private static final int COST = 3;
+    private static final int HEAL = 2;
+
+    HyperRepair(Andy commander, CostBasis basis)
+    {
+      super(commander, NAME, COST, basis);
+      AIFlags = 0; // Why one would ever use this is beyond me
```
ThatIsAUsername, line 71: lolz

src/CommandingOfficers/AW2/AW2Commander.java
ThislsAUsername: OTOH, it's kind of ugly to glue that on every power description ever (Reboot Camp does this, FWIW).
Yup. +10/10 for AW3, x1.1/0.9 for AW1 (Did I mention it's special?)

Non-DoR gives subtractive defense, unless stated otherwise.
I guess I could blurb everything, but bleeeeeeh

src/CommandingOfficers/AW2/OS/Andy.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      for( Unit u : myCommander.army.getUnits() )
+      {
+        u.alterHP(HEAL*10);
```
ThatIsAUsername: Any reason to do a magic multiplication here instead of specifying the healing power in terms of health? I thought we were moving that way.
Also, didn't we standardize on the `HP` methods working on [0,10] HP, and changing `*Health` methods to work on [0,100] health?

src/CommandingOfficers/TrilogyCommander.java
ThislsAUsername: Yes. Cast 9 costs 2.8x base, cast 10 costs 2.0x base. It's a bug that the cost ever goes down.
So you get to keep more than (for cast 10) two stars' worth of charge, then you get a refund on the energy you _did_ spend.
Cast 11 costs exactly the same as cast 10 (so Hachi keeps exactly 2* if he charged SCOP before COPing)

...and I know how this happened because I accidentally did it myself.

src/CommandingOfficers/TrilogyCommander.java
ThatIsAUsername: Also maybe specify that this is only deducting the delta because the parent ability logic is already handling deducting the advertised cost, and we are just adjusting here to match cart logic here (unless that isn't true).

src/CommandingOfficers/AW2/OS/Andy.java
ThislsAUsername: Because I'm also referencing the `HEAL` value in the description.
Yes, this code is a compile error and will be fixed in another commit.

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,86 @@
...
+
+    return power;
+  }
+  @Override
+  public int calculateMassDamageCharge(Unit minion, int lostHP)
+  {
+    return 0;
+  }
+
+  protected abstract static class TrilogyAbility extends CommanderAbility
```
ThatIsAUsername, line 55: Since this is referenced/extended by other classes, part of me thinks it should live in its own file.

Then again, all the one that do inherit the definition, so maybe not so urgent.

Might not hurt to eventually rearrange the directory in general though. We have a bunch of COs together with abstract base classes, COAbility, and some other utility stuff. Maybe you already have a plan though, since you are creating some new directories. Got a long term vision here already?

ThislsAUsername: Right, and it's `protected` to enforce that.

That's a good idea. No, I don't currently have a long-term plan. (And if I did, it's slightly schizo since Andy is packaged by game.faction and Brenner is packaged only by faction)

I do kind of like packaging by game.faction.
That would imply:
Ave/Strong/etc -> CommandingOfficers.DefendPeace
RuinedCommander -> COs.AW4
Brenner -> COs.AW4.BrennerWolves

Then I guess the rest are generic extension points/etc used by the rest, so they can stay?

src/UI/UIUtils.java
```diff
@@ -274,6 +274,14 @@ public static COSpriteSpec instance(Commander co)
       return spec;
     }
 
+    @Override
+    public String toString()
+    {
+      if( paletteNames.containsKey(color) )
+        return getCanonicalFactionName(this);
+      return color.toString() + " " + faction.name;
```
ThatIsAUsername, line 282: Only get the canonical name if the color is in `paletteNames`? Aren't they all?

ThislsAUsername: `MISC` is not.

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
```diff
@@ -565,7 +567,7 @@ public static void initBins(final int selectedCO)
       }
       OuterCategoryPanel outerCategory = new OuterCategoryPanel(outerPalette, outerName);
 
-      for( int innerIndex = 0; innerIndex < allCmdrLists.length; ++innerIndex )
+      for( int innerIndex = 0; innerIndex < innerMax; ++innerIndex )
```
ThatIsAUsername, line 570: Curious: This looks like the only thing that is actually different after the `innerMax` change, but I don't see a difference in behavior in the menu. What was the actual problem here?

ThislsAUsername: Yessir. It was using the number of outer categories as the bound for the inner category's length.
This resulted in not finding some COs in the collection, so they wouldn't show up in the GUI (but did sometimes affect control, which was weird).
The CO(s) that didn't show up was semi-random per run due to hash ordering, which was pretty weird to diagnose.

src/UI/Art/SpriteArtist/CommanderOverlayArtist.java
```diff
@@ -156,7 +156,9 @@ public static BufferedImage buildCoPowerBar(Commander co, int[] abilityPoints, i
     if( 0 == abilityPoints.length )
       return SpriteLibrary.createTransparentSprite(1, 1);
 
-    final double pixelsPerPowerUnit = 3.0 / Commander.CHARGERATIO_FUNDS;
+    // Just grab the first ability's star cost since I don't really want to consider what to do under variant star costs.
+    int starCost = co.myAbilities.get(0).costBasis.baseStarRatio;
+    final double pixelsPerPowerUnit = 3.0 / starCost;
```
ThatIsAUsername, line 161: Remember when `pixelsPerPowerUnit` was just `3`? I remember.

It might be more clear/less magical to have `pixelsPerStar(or star equivalent)=3` and also `pixelsPerPowerPlanck=pixelsPerStar/starCost`. Or like, something, I dunno. The now-magic 3 just makes me double-take every time I see it now.

src/Units/KaijuWarsKaiju.java
```diff
@@ -83,7 +83,7 @@ public class KaijuWarsKaiju
 
   private static final int KAIJU_COST = 0;
   private static final int MOVE_POWER = 2; // Dummy value
-  private static final double STAR_VALUE = 10.0;
+  private static final int STAR_VALUE = 50;
```
ThatIsAUsername, line 86: Not that it matters much (because Kaiju) but why not 100? You did a x10 everywhere else.

src/CommandingOfficers/Tech.java
```diff
@@ -571,7 +571,7 @@ private static UnitModel createBattleMechModel(GameReadyModels grms)
     BattleMech.name = BATTLEMECH_NAME;
     BattleMech.role = BattleMech.role | UnitModel.SURFACE_TO_AIR;
     BattleMech.costBase = mdTank.costBase*2 + (antiAir.costBase/2);
-    BattleMech.abilityPowerValue = 2.0;
+    BattleMech.abilityPowerValue = 20;
```
ThatIsAUsername, line 574: So with the change in definitions, does that mean this is now worth 1/5 of a star instead of 2 stars? Arguably it should just be worth 0 stars as a free unit; I forget why it was 2 initially.

res/co/Andy_3.png
ThatIsAUsername: I guess disk space is cheap.

src/Units/KaijuWarsKaiju.java
ThislsAUsername: Kaiju can have a lot more than 10 HP, and I really didn't think about it much the first time.

src/CommandingOfficers/AW3/AW3Commander.java
```diff
@@ -0,0 +1,54 @@
...
+import CommandingOfficers.CommanderInfo.InfoPage;
+import Engine.GameScenario;
+import Units.UnitDelta;
+
+public abstract class AW3Commander extends TrilogyCommander
+{
+  private static final long serialVersionUID = 1L;
+  public static final InfoPage AW3_MECHANICS_BLURB = new InfoPage(
+      "Gain +10 attack and defense on activating a power.\n"
+    + "Power charge is not based on funds, but is a specific stat for each unit type.\n"
```
ThatIsAUsername, line 15: So like, power could be awarded based on HP for infantry, move distance for recons, armor rating for bopters, etc?

src/CommandingOfficers/Commander.java
ThatIsAUsername: Is their unloading broken too?

src/CommandingOfficers/Commander.java
ThatIsAUsername: I meant to give an example of how to do what the comment says to do. And also what the preferred alternative is. If I, an ignorant code monkey, come across this a year from now, how will I know what to do? All this tells me is "don't use this unless you _really_ know what you're doing."

src/CommandingOfficers/TrilogyCommander.java
ThatIsAUsername: I want to read _accurate_ things, of course.

src/CommandingOfficers/TrilogyCommander.java
ThatIsAUsername: ...and now we are doing it on purpose for verisimilitude. Cool I guess.

src/CommandingOfficers/AW2/AW2Commander.java
ThatIsAUsername: Sounds like Reboot Camp learned that sometimes ugly/repetitive/verbose == clear. So I think I agree with them, and I'd co-locate the power stuff, but you are the one doing the work, so \_( ' ' )_/.

src/UI/UIUtils.java
ThatIsAUsername: Right, the infamous "No CO".

src/UI/Art/SpriteArtist/PlayerSetupCommanderArtist.java
ThatIsAUsername: Huh. Guess I've just gotten lucky.

src/CommandingOfficers/AWBW/AWBWCommander.java
```diff
@@ -0,0 +1,71 @@
...
+import CommandingOfficers.CommanderInfo;
+import CommandingOfficers.CommanderInfo.InfoPage;
+import Engine.GameScenario;
+import Engine.UnitMods.UnitDamageModifier;
+import Engine.UnitMods.UnitModifier;
+import Terrain.MapMaster;
+import Units.Unit;
+import Units.UnitDelta;
+
+public abstract class AWBWCommander extends Commander
```
ThatIsAUsername, line 16: Seems odd that AWBW commander isn't based on TrilogyCommander, (or even AW2Commander) since that is the source material.
I guess this gets +10/10 like AW3, but gains charge like AW2? (I still don't understand the blurb on AW3 power charge).
The scaling also claims to go for 10 activations here, instead of the 9 for Trilogy.
Also things in the blurb are out of order compared to Trilogy, though it's probably hard to fully align across commanders due to how the blurbs are spread across the hierarchy.
So I guess uh, carry on, and it still surprises me how much nuance there is between the various AW implementations.

src/CommandingOfficers/AWBW/AWBWCommander.java
```diff
@@ -0,0 +1,71 @@
...
+    //     (8.1HP unit on city will heal to 10HP, costing 1HP of repair).
+    roundUpRepairs = false;
+  }
+
+  @Override
+  public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
+  {
+    if( null != getActiveAbility() )
+      return 0;
+    // TODO: Copy/paste super's code and delete the HP-based bit once the parent PR settles down.
```
ThatIsAUsername, line 42: Does this still have a parent PR? Or does the child PR address this?

src/CommandingOfficers/AWBW/OS/Andy.java
```diff
@@ -0,0 +1,117 @@
...
+import CommandingOfficers.AWBW.AWBWCommander;
+import Engine.GameScenario;
+import Engine.UnitMods.UnitDamageModifier;
+import Engine.UnitMods.UnitModifier;
+import Engine.UnitMods.UnitMovementModifier;
+import UI.UIUtils;
+import Terrain.MapMaster;
+import Units.Unit;
+
+public class Andy extends AWBWCommander
```
ThatIsAUsername, line 16: Some part of my brain is _bothered_ that we have an AWBW Andy. It's not even an AW game. Why is it different? Are all the other COs different in some subtle way such that we need a third (fourth?) nearly-indistinguishable variant of each of them? Is anyone going to even notice the differences in gameplay? Is this what happens when I review code after midnight? Or just when I review your design decisions specifically?

src/CommandingOfficers/AWBW/AWBWCommander.java
ThatIsAUsername: Looks like it came from Trilogy, which has addressed it already. So we can probably just delete the function here? And the only reason Trilogy needed to reimplement it is because our Commander is nonconformant to the carts, because it counts both HP and funds-based charge accruals, right?

src/CommandingOfficers/TrilogyCommander.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      // Change my CO's energy such that it's as if this power cost the activation cost of the next power rather than the actual cost
+      int costAtActivation = baseStars * costBasis.calcCostPerStar(costBasis.numCasts - 1);
+      int costForNextCast  = costBasis.calcCost(baseStars);
+      int extraEnergyCost  = costForNextCast - costAtActivation; // Will be negative on cast 9
+      myCommander.modifyAbilityPower(-1 * extraEnergyCost);
```
ThatIsAUsername, line 83: I might say "cost of the next _activation_ rather than the _advertized_ cost".

Also I think I'd still like a "why" in the comments here since... it doesn't make any logical sense at face value.

src/CommandingOfficers/BuggedEnergyCostCommander.java
```diff
@@ -0,0 +1,60 @@
...
+import Engine.GameScenario;
+import Engine.UnitMods.UnitDamageModifier;
+import Engine.UnitMods.UnitDefenseModifier;
+import Engine.UnitMods.UnitModifier;
+import Terrain.MapMaster;
+import java.util.ArrayList;
+import CommandingOfficers.CommanderInfo.InfoPage;
+
+/** Base class for AW2/3 COs (AW1 is special) */
+public abstract class BuggedEnergyCostCommander extends Commander
```
ThatIsAUsername, line 12: This is just gratuitous. The slightly-inaccurate `Trilogy` was a better prefix I think, especially as you aren't even implementing the AW1 version (yet?)

src/CommandingOfficers/AW2/AW2Commander.java
```diff
@@ -0,0 +1,59 @@
...
+    + TRILOGY_MECHANICS_BLURB.info
+    );
+
+  public AW2Commander(CommanderInfo info, GameScenario.GameRules rules)
+  {
+    super(info, rules);
+  }
+
+  @Override
+  public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
```
ThatIsAUsername, line 25: You moved this down from `TrilogyCommander` - doesn't that mean that `AW3Commander` now inherits the Default `Commander`'s method for this (which counts both HP and cost for generating power)?

ThatIsAUsername: Never mind, found it. It just didn't show up in the other diff because it was already present.

res/co/Andy_2.png
ThatIsAUsername, line 1: What did you do to optimize these? It clearly made a difference.

src/CommandingOfficers/Ave.java
ThatIsAUsername: The balance here was never exactly fine tuned, so I'm not going to lose sleep over it.

src/CommandingOfficers/BuggedEnergyCostCommander.java
ThislsAUsername: Yes, "yet" is correct.
You did say you want to read _accurate_ things.

ThislsAUsername (commented):
I have discovered why you didn't reply to any of my comments.

src/CommandingOfficers/Commander.java
ThislsAUsername: Yes. Unloading can be done by waited units, and doesn't take an action.
An APC can load, unload, move, resupply, load, and unload again all in one turn.
A Cruiser can kill a copter and then unload a copter into its space.

src/CommandingOfficers/AW3/AW3Commander.java
ThislsAUsername: I mean, kind of.
Recons have the highest move on a ground unit, and give the most charge-per-funds of anything but Infantry.

src/CommandingOfficers/AWBW/OS/Andy.java
ThislsAUsername: Forgive me being laconic, as Github helpfully deleted my comment when I deleted the extra copy.

It's different because AWBW was implemented before a lot of the various specific mechanics were well-known, then inertia.

There are various stat differences (Sturm, Drake, Flak)
The COP cost bug doesn't exist (benefiting Adder a lot)
Universal +10/10 like AW3, except it's taken out of universal attack boosts for some reason.
Charge isn't like AW3, as well, so you can't just use that one.

I can't really speak to whether most will notice, but someone actually implemented no healing round-up to have a more accurate playtest.

res/co/Andy_2.png
ThislsAUsername: Like I said, I unchecked stuff like EXIF data in the GIMP export dialog.
I think I left "save resolution" checked, but that's it.

src/CommandingOfficers/Commander.java
ThatIsAUsername: OP

src/CommandingOfficers/CommanderAbility.java
ThatIsAUsername: Oh, I thought I commented on that. I'm fine with the approximate fix you did; it's not like the numbers were fine-tuned for balance before.

src/CommandingOfficers/AWBW/AWBWCommander.java
ThatIsAUsername: And they don't all agree with how much it takes to make a star, either.

Whatever, DefendPeace COs OP.

res/co/Andy_2.png
ThatIsAUsername: There's an option to "save resolution"? And if you don't, what'll it do, save a 1D line of pixels? I'm also surprised that EXIF would be present at all in these images. I was expecting some kind of fancy compression settings, but OK.

src/CommandingOfficers/Tech.java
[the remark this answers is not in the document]
ThatIsAUsername: > likely to eat a lot of damage

Well, maybe. They might also just kill anything that might hurt them when they land.
Playing as Tech always felt like you could hit a point where you were primed for a new Steel Hail almost before the old one died off, and it just feels way OP. Which is admittedly not a driving design concern just yet.

src/CommandingOfficers/BuggedEnergyCostCommander.java
ThatIsAUsername: ...sigh. I mean yes, but I also prefer to read meaningful things. This new name gives no indication of why someone would want to use this class (Hmm, I want my new commander to have incorrect power costs. Oh look, a convenient base class!). Trilogy does, though it would also create confusion given that AW1/4 have their own base classes already (and so do 2 and 3 for that matter), but a little skulking around would clarify things. As usual, my vote is going to be for something annoyingly wordy like `AW2And3CommanderBase` or some such tomfoolery, but I'd live with `TrilogyCommander` too.

src/CommandingOfficers/CommanderAbility.java
ThislsAUsername: You totally did. I wrote that days ago; it just showed up now because I was "writing a review" until today.

src/CommandingOfficers/Tech.java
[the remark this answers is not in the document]
ThislsAUsername: > Well, maybe. They might also just kill anything that might hurt them when they land.

Fair point.
> Playing as Tech always felt like you could hit a point where you were primed for a new Steel Hail almost before the old one died off, and it just feels way OP.

It's also 14* and you have a D2D that doesn't do much, so you're probably in a very bad position before you get to pop it. Compare it with Lightning Strike, for example.

> And as for the Oozium comparison; that's a campaign unit, unusable in competitive matches, so balance logic doesn't apply at all anyway.

Hey now, you can unlock the ability to place them in the map editor.
...I mean, it's still useless and bad, but it's not completely exclusive to campaign.

res/co/Andy_2.png
ThislsAUsername: Yes. I dunno, and I didn't especially feel like testing it at the moment.
Nah, the only compression setting is "compression level", which defaults to the max of 9.

src/CommandingOfficers/AWBW/OS/Andy.java
ThatIsAUsername: So uh, why did Andy1 miss the boat?

ThatIsAUsername:
Looks like we ran out of comments on this PR and that one, so I guess that means they are good to go.

src/CommandingOfficers/AWBW/OS/Andy.java
ThislsAUsername: Because AW1 is special (i.e. will require more changes), and I am trying to keep PRs not too gigantic.
These were mostly already supported by the engine.

PR #404, opened 2023-12-17
DoR Charge

ThislsAUsername:
Moves DoR charge to be in range [0,60] since that's the most readable of all.

### DIFF
--- a/src/CommandingOfficers/RuinedCommander.java
+++ b/src/CommandingOfficers/RuinedCommander.java
@@ -51,8 +51,8 @@ public abstract class RuinedCommander extends DeployableCommander
           + "Your COU is always max level.\n"
           + "Veterancy bonuses do not apply if this CO is tagged out, but kills are always tracked.\n"
           );
-  public static final int DOR_STAR_VALUE = 10000;
-  public static final int CHARGERATIO_HP =  1000; // Funds value of 1 HP damage dealt, for the purpose of power charge
+  public static final int DOR_STAR_VALUE = 10;
+  public static final int CHARGERATIO_HP =  1; // Funds value of 10 HP damage dealt, for the purpose of power charge
 
   public static final int GENERIC_STAT_BUFF = 10;
   /** If the unit type matches any flag in this mask, it receives my zone stat boosts */
--- a/src/Test/TestCommanderBrenner.java
+++ b/src/Test/TestCommanderBrenner.java
@@ -1,0 +1,164 @@
+package Test;
+
+import java.util.ArrayList;
+
+import CommandingOfficers.Commander;
+import CommandingOfficers.DeployableCommander;
+import CommandingOfficers.Patch;
+import CommandingOfficers.RuinedCommander;
+import CommandingOfficers.RuinedCommander.UseAbilityFactory;
+import CommandingOfficers.BrennerWolves.BrennerDoR;
+import Engine.Army;
+import Engine.GameActionSet;
+import Engine.GameInstance;
+import Engine.GamePath;
+import Engine.GameScenario;
+import Engine.UnitActionFactory;
+import Engine.Utils;
+import Engine.XYCoord;
+import Engine.UnitActionLifecycles.BattleLifecycle;
+import Engine.UnitActionLifecycles.LoadLifecycle;
+import Terrain.MapLibrary;
+import Terrain.MapMaster;
+import Units.Unit;
+import Units.UnitModel;
+
+public class TestCommanderBrenner extends TestCase
+{
+  private Commander Patch;
+  private BrennerDoR Brenner;
+  private MapMaster testMap;
+  private GameInstance game;
+
+  private void setupTest()
+  {
+    GameScenario scn = new GameScenario();
+    Brenner = new BrennerDoR(scn.rules);
+    Patch = new Patch(scn.rules);
+    Army[] cos = { new Army(scn, Brenner), new Army(scn, Patch) };
+
+    testMap = new MapMaster(cos, MapLibrary.getByName("Firing Range"));
+
+    game = new GameInstance(cos, testMap);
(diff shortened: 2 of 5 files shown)


src/Test/TestCommanderBrenner.java
```diff
@@ -0,0 +1,164 @@
...
+    Unit aa   = addUnit(testMap, Brenner, UnitModel.SURFACE_TO_AIR, 2, 2);
+
+    // And our victim
+    Unit inf = addUnit(testMap, Patch, UnitModel.TROOP, 1, 2);
+
+    Brenner.initTurn(testMap);
+    Brenner.army.money = 9001;
+    boolean testPassed = true;
+    
+    performGameAction(new DeployableCommander.DeployCOUAction(Brenner.deployAction, COU), game);
```
ThatIsAUsername, line 111: The transport/COU is at (4, 2) which is a grass tile. Shouldn't this action fail? Or does this just skip the "can I?" step and take Shia's advice?

src/CommandingOfficers/RuinedCommander.java
```diff
@@ -51,8 +51,8 @@ public abstract class RuinedCommander extends DeployableCommander
           + "Your COU is always max level.\n"
           + "Veterancy bonuses do not apply if this CO is tagged out, but kills are always tracked.\n"
           );
-  public static final int DOR_STAR_VALUE = 10000;
-  public static final int CHARGERATIO_HP =  1000; // Funds value of 1 HP damage dealt, for the purpose of power charge
+  public static final int DOR_STAR_VALUE = 10;
+  public static final int CHARGERATIO_HP =  1; // Funds value of 10 HP damage dealt, for the purpose of power charge
```
ThatIsAUsername, line 55: All right, so DoR COs now get 1 energy per HP lost, and need 10 energy per star. Yet the energy display in the CO Info pane still shows values in the tens of thousands. How are you reconciling that with the differing power cost bases?

src/Test/TestCommanderBrenner.java
ThislsAUsername: Yeah, I've been depending a lot lately on validation at the `UnitActionFactory` level, since that is what determines what goes up on the GUI (i.e. an invalid action that's constructed and executed will simply fail with no explanation, while an action that doesn't get constructed just won't show up)

src/CommandingOfficers/RuinedCommander.java
ThislsAUsername: Eh? It showed up as 60 for me, and I don't think I have uncommitted changes.

src/Test/TestCommanderBrenner.java
```diff
@@ -0,0 +1,164 @@
...
+  {
+    setupTest();
+
+    // Add our combatants
+    Unit mech = addUnit(testMap, Brenner, UnitModel.MECH, 1, 1);
+    Unit COU  = addUnit(testMap, Brenner, UnitModel.TRANSPORT, 4, 2);
+    Unit aa   = addUnit(testMap, Brenner, UnitModel.SURFACE_TO_AIR, 2, 2);
+
+    // And our victim
+    Unit inf = addUnit(testMap, Patch, UnitModel.TROOP, 1, 2);
```
ThatIsAUsername, line 105: It might be nice to drop in a high-level description of this test in a comment, so future readers know what's going on. Just to outline that `inf` (enemy) and `aa` (Brenner's) are inside the CO zone, but `mech` is not. We expect to not get charge when mech attacks (not in zone), but we should get charge when `aa` attacks.

src/CommandingOfficers/RuinedCommander.java
ThatIsAUsername: No, you are right, I just forgot to switch to the new branch when I moved to the new PR.

Though it does seem a bit odd to see three orders of magnitude difference in the "Power to go" measurements for different COs. Not super sure what to do about that though. Maybe change it to display in stars so we aren't comparing the underlying nonstandard units?

ThislsAUsername: It's odd, yes, but IMO it's good for the mechanics to be obvious.
DoR charge goes "Hey look, it changed by the HP damage I just dealt", which is super obvious.

ThatIsAUsername: That leaves the non-DoR four-digit magnitudes confusing though, potentially. I guess a universal "stars to go" isn't necessarily more clear either.

src/Test/TestCommanderBrenner.java
```diff
@@ -92,22 +92,27 @@ private boolean testCOU()
...
 
     // And our victim
     Unit inf = addUnit(testMap, Patch, UnitModel.TROOP, 1, 2);
 
     Brenner.initTurn(testMap);
     Brenner.army.money = 9001;
     boolean testPassed = true;
-    
+
+    // Note: COUing on this tile is not normally valid, but we're skipping the logic that enforces that.
```
ThatIsAUsername, line 115: Could just swap the attack command so the inf attacks the mech instead to test that case. Shrug.

ThislsAUsername: I don't think that comment applies to this line?
Either way, I'm not worried about having a test for the charge bug, especially since I may be putting in an option for it at some point.

src/CommandingOfficers/RuinedCommander.java
ThislsAUsername: Four-digit magnitudes were specifically requested, since they refer to the funds of damage directly (so you can grok them if you know what's going on).
Three-digit AWDS magnitudes are also good IMO since it is clearly neither of the other options, and that's about as clear as it's getting without adding a stats screen/page that shows the energy-per-HP stat.

PR #406, opened 2023-12-28
Most DoR COs

ThislsAUsername:
Adds more real DoR COs. Wow, Caulder is busted.

Skipping Lin because vision stuff, Penny because weather/vision stuff, and Tabitha because I'm away from my notes.
On the note of being away from my notes, I don't have an image editor on hand. I plan to add the mugs to this PR when I get back.

I realized I need a bit more flexibility on the masking, since Will's boosts require land AND direct and e.g. Gage requires indirect OR sea.
...and I copypasted the "if" check for the defender badly, but I fixed it.

On the note of masking, this PR includes a thing I was avoiding (encoding cart-style direct/indirect) and a thing that's very uggo (putting a CO-specific flag in `UnitModel`).
It'd be possible to avoid both by making a filter that uses code, but I didn't feel like it.
(Note: I don't really mind changing the approach if you want, but this is what I path-of-least-resistance'd)


src/Units/UnitModel.java
```diff
@@ -53,6 +53,13 @@ public abstract class UnitModel implements Serializable, ITargetable, UnitModLis
   public static final long AIR_TO_AIR      = 1 << 22;
   public static final long TRANSPORT       = 1 << 23;
 
+  // Calculated properties so I can be lazy
+  public static final long DIRECT          = 1l << 50; // Has only weapons for 1 range
+  public static final long INDIRECT        = 1l << 51; // Has weapons for 2+ range
+  // Sins against game design
+  public static final long GREYFIELD_LIKES = 1l << 62; // For Seaplanes
```
ThatIsAUsername, line 60: Sigh. Well, at least you are going in with both eyes open.

src/Units/UnitModel.java
```diff
@@ -131,6 +140,24 @@ private UnitModel(String pName, long pRole, int cost, int pAmmoMax, int pFuelMax
...
+        continue;
+      isDirect = false;
+      isIndirect = true;
+      break;
+    }
+    if( isDirect )
+      this.role |= DIRECT;
+    if( isIndirect )
+      this.role |= INDIRECT;
+  }
```
ThatIsAUsername, line 159: This is good. Not sure why we didn't have something like it before.

src/CommandingOfficers/AW4/BrennerWolves/BrennerDoR.java
```diff
@@ -24,7 +24,7 @@ public instantiator()
           "Captain of the 12th Battalion. He has an unshakable faith in humanity's goodness, and will aid any in need.\n"));
       infoPages.add(new InfoPage(
           "Base Zone: 3\n" +
-          "Zone Boost: All units +20 DD\n"));
+          "Zone Boost: All units +20 defense\n"));
```
ThatIsAUsername, line 27: Calling him Brenner"DOR" is kinda redundant now. Shrug.

src/CommandingOfficers/AW4/RuinedCommander.java
```diff
@@ -59,7 +59,9 @@ public abstract class RuinedCommander extends DeployableCommander
 
   public static final int GENERIC_STAT_BUFF = 10;
   /** If the unit type matches any flag in this mask, it receives my zone stat boosts */
-  public long canBoostMask = Long.MAX_VALUE;
+  public long boostMaskAny = Long.MAX_VALUE;
+  /** Boost if all flags match; you should customize boostMaskAny if you use it */
+  public long boostMaskAll = Long.MAX_VALUE;
```
ThatIsAUsername, line 64: Looks like you'd also have to set `boostMaskAny` to zero if using `boostMaskAll`. Oh, that's probably what you are saying. It wasn't clear what "it" refers back to, especially since the comment is on `boostMaskAll`. Changing "it" to "this" at a minimum (and add "also" before "customize") might make it more clear, but it also wouldn't be out of line to clarify more (or at least add a comment to `boostMaskAny`; clarify that it boosts everything by default, and that both conditions are OR'd to get the final boost mask.

src/CommandingOfficers/AW4/BrennerWolves/Will.java
```diff
@@ -0,0 +1,84 @@
...
+    UnitTypeFilter moveMod;
+
+    protected RallyCry(RuinedCommander commander)
+    {
+      super(commander, "Rally Cry");
+      if( null == commander )
+        return; // This isn't a "real" ability, just a scratch struct for the info page
+      moveMod = new UnitTypeFilter(new UnitMovementModifier(2));
+      moveMod.allOf = commander.boostMaskAll;
+    }
```
ThatIsAUsername, line 75: Rally Cry is not a real ability?

src/CommandingOfficers/AW4/NRA/Waylon.java
```diff
@@ -0,0 +1,83 @@
...
+    public instantiator()
+    {
+      super("Waylon", UIUtils.SourceGames.AW4, UIUtils.NW);
+      infoPages.add(new InfoPage(
+          "A member of the Rubinelle Army who seeks only to fulfill his own pleasures and desires.\n"));
+      infoPages.add(new InfoPage(
+          "Base Zone: "+RADIUS+"\n"
+          + "Zone Boost: Air units +"+POWER+"/"+DEFENSE+".\n"));
+      infoPages.add(new InfoPage(new Wingman(null),
+          "Boost air unit defense by 270 (410 + veterancy, total).\n"));
```
ThatIsAUsername, line 34: 410 "with" veterancy?

src/CommandingOfficers/AW4/BrennerWolves/Will.java
ThatIsAUsername: Oh, I see. You are only creating the unit mods if we are going in-game with a commander instance. We don't do that for the rest of our commanders. Maybe there's a way to statically separate out the description part somehow so we don't even need an instance? Pretty low priority though.

src/CommandingOfficers/AW4/NRA/Greyfield.java
```diff
@@ -0,0 +1,79 @@
...
+      super(commander, "Supply Chain");
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      super.perform(gameMap);
+      for( Unit unit : myCommander.army.getUnits() )
+      {
+        unit.resupply(); // the missing HP has to go somewhere...
```
ThatIsAUsername, line 73: Missing HP, eh? He doesn't do self-damage or anything, does he?

src/Units/UnitModel.java
ThislsAUsername: The only other real option is filtering on the name, and that seems bad too.
...I guess you could look at fuel/ammo numbers or something, but bleh.

src/Units/UnitModel.java
ThislsAUsername: I wanted to do silly things instead.

src/CommandingOfficers/AW4/BrennerWolves/BrennerDoR.java
ThislsAUsername: Wally**AI** is also redundant. Again, shrug.

src/CommandingOfficers/AW4/BrennerWolves/Will.java
ThislsAUsername: There totally is, but it requires more boilerplate... and the enterprise here is to reduce that.

src/CommandingOfficers/AW4/NRA/Waylon.java
ThislsAUsername: 410 minimum, up to 430 with veterancy's +20.

src/CommandingOfficers/AW4/NRA/Greyfield.java
ThislsAUsername: Whoops.

src/Units/UnitModel.java
ThislsAUsername: I've realized also that Seaplanes on cart can load the COU from an airport (but that's not true here).
Should I just reuse this flag to fix that?
Or maybe I can make the check "can build any unit type, and can repair this unit"?

ThislsAUsername:
Sidenote: these are the options I was talking about in that other PR
image

src/Units/UnitModel.java
ThatIsAUsername: > filtering by name

As a one-off for a CO with a special affinity for a particular unit, I'd probs allow it.

I'm also fine with adapting the condition to allow for properties that repair to avoid more heinous hijinks.

src/CommandingOfficers/AW4/RuinedCommander.java
```diff
@@ -116,10 +117,20 @@ public ArrayList<CommanderAbility> getReadyAbilities()
     return new ArrayList<CommanderAbility>();
   }
   @Override
-  public void onCOULost(Unit minion)
+  public GameEventQueue onCOULost(Unit minion)
   {
     modifyAbilityStars(-42);
     zoneIsGlobal = false;
+
+    // Stolen from Commander.revertActiveAbility(), with the caveat that we know all our subclass's powers have no revert events.
```
ThislsAUsername, line 125: I had to add this because Waylon's powers specifically do actually matter while his COU is dead.
It's kind of icky that it doesn't use the main revert function, but I didn't feel like touching the declaration of `GameEvent.sendToListener()` to daisy-chain the master map down.

ThatIsAUsername: You mention doing this for Waylon specifically, but this is in the base class. Will it not also cause other CO's abilities to persist after COU death?

src/CommandingOfficers/DeployableCommander.java
```diff
@@ -102,13 +102,20 @@ public void modifyActionList(UnitContext uc)
...
     {
       COUsLost.add(victim);
-      onCOULost(victim);
+      GameEventQueue couLossEvents = onCOULost(victim);
+      if( null != couLossEvents )
+      {
+        if( null == events )
+          events = new GameEventQueue();
+        events.addAll(couLossEvents);
+      }
```
ThatIsAUsername, line 116: Any reason `onCOULost` returns a `GameEventQueue` instead of just accepting one to push events into? This could avoid the null-checking here. I guess `ReceiveUnitDieEvent` already doesn't conform to that paradigm too.

src/CommandingOfficers/AW4/RuinedCommander.java
ThislsAUsername: The correct behavior for DoR COs is "if the COU is killed during COP, immediately end the power".
Most of this is fulfilled by "simply revoke the global zone", since move boosts/instant healing/etc don't matter much when it's not your turn (and you obviously can't charge with no zone).

Waylon COP adds combat stats, which apply whether or not his zone is in the picture.
I guess I could add a specific exception there, but that seems a lot less clean than "just end the power".
The compromise here is there's no nice way to get at a `MapMaster` instance, so we can't get revert events for the power... but none supply any. (We can just steal it from the CO's `MapPerspective`, but I like to pretend that's not an option because spaghetti.)

src/CommandingOfficers/DeployableCommander.java
ThislsAUsername: I guess taking a queue to use is maybe a more-obvious way of returning a list of side-effects.
That feels like bait for a larger refactor, though, yeah.

I've favored return values over out params a lot since it makes the function "more pure".
I guess that rule doesn't make as much sense when the function mutates state on its instance, so I'm down with making `GameEventListener` hooks returning `void`.
We can also throw the game or map in as a param for them too, to fix the thing in the thread above.

src/CommandingOfficers/AW4/RuinedCommander.java
ThatIsAUsername: > The correct behavior for DoR COs is "if the COU is killed during COP, immediately end the power"

Ok, sure. when COU dies, revert the power. Simple. But you also say.

> Waylon's powers specifically do actually matter while his COU is dead

Except it appears this unconditionally reverts powers on COU death (including Wingman's `moveMod` that grants defense). I don't see the difference between Waylon's ability and say, Gage's (whose `moveMod` grants indirect range - this appears to be a common unit modifier variable name regardless of effect).

I feel like I'm missing something - how does this treat Waylon differently? Why would his powers outlive his COU when others don't?

src/CommandingOfficers/DeployableCommander.java
ThatIsAUsername: > That feels like bait for a larger refactor, though, yeah

Maybe, maybe not. I already follow the "pass in the queue to populate" strategy in other (most?) places where we call functions that produce GameEvents.

> I've favored return values over out params a lot since it makes the function "more pure"

Not sure I buy that; I'd tend to think the method of receiving a function's output is a question of syntax rather than purity, and creating additional containers just to concatenate them is just inefficient (not that this is a driving concern for us).

src/CommandingOfficers/AW4/RuinedCommander.java
ThislsAUsername: > > Waylon's powers specifically do actually matter while his COU is dead
> 
> Except it appears this unconditionally reverts powers on COU death (including Wingman's `moveMod` that grants defense).

Yes. It does that because otherwise Waylon's ability *would* still work after he dies. Which I don't want.
Brenner COP doesn't matter whether it's on or off if his COU is dead, since there's no zone and the only "real" COP effect is the instant heal.
Waylon COP matters because it has a defense boost that isn't tied to the zone explicitly. So I added this code to turn it off when he dies.
> I don't see the difference between Waylon's ability and say, Gage's (whose `moveMod` grants indirect range - this appears to be a common unit modifier variable name regardless of effect).

Good catch, that is bad pasta and should be fixed.
Gage's indirect range doesn't impact combat when it's not his turn, so just turning his CO zone off when his COU dies is technically enough (assuming you don't step on your own COU with a Kaiju, I guess).
It would look weird that he had the "power active" effect when no stat boosts are in play, but mechanically it'd be the same as if his power was turned off.
By contrast, Waylon's stat boost will affect combat when it's not his turn.

src/CommandingOfficers/DeployableCommander.java
ThislsAUsername: > > That feels like bait for a larger refactor, though, yeah
> 
> Maybe, maybe not. I already follow the "pass in the queue to populate" strategy in other (most?) places where we call functions that produce GameEvents.

Every listener hook returns `null` right now. Changing that's a fairly large refactor, I feel like.
> > I've favored return values over out params a lot since it makes the function "more pure"
> 
> Not sure I buy that; I'd tend to think the method of receiving a function's output is a question of syntax rather than purity, and creating additional containers just to concatenate them is just inefficient (not that this is a driving concern for us).

If I call `enqueueMoveEvent()` on the same event queue twice, I get different results than if it simply provided a queue as a return value. Calling the function changes the state of the process, and you would have to know the internals of `enqueueMoveEvent()` or redo any previous work to the queue to undo that change.

A "return-pattern `createMoveEvent()`" would allow me to produce the first's output (by adding each return value to my own queue immediately), but it isn't built into calling the function at all anymore.
Likewise, I could produce the output of the return-patterned one with the existing function by explicitly creating a new queue for each call to operate on.

I suppose it's _mostly_ semantics either way, but I do slightly prefer not modifying input variables directly since that means I can reset a given function multiple times during debugging without producing aberrant behavior.

src/CommandingOfficers/AW4/RuinedCommander.java
ThatIsAUsername: > Yes. It does that because otherwise Waylon's ability would still work after he dies. Which I don't want.

Oh. Okay. I interpreted your initial statement "Waylon's powers specifically do actually matter while his COU is dead" to mean that his stat boosts _should_ outlive his CO unit. Now I get it: this is the correct behavior for all DoR COs, but you could technically have gotten away with not doing it, if it wasn't for that meddling Waylon.

ThatIsAUsername:
Sweet. I think this is probs good to go.

ThislsAUsername:
Gimmie a sec to merge this and then rebase `awone`, then

PR #408, opened 2024-01-01
Fix infantry movement animations

ThislsAUsername:
I noticed GS's movement sprites weren't being recolored, and then found more problems.


ThatIsAUsername:
Cool

PR #409, opened 2024-01-03
AW One

ThislsAUsername:
Sonja, Sami, and Sturm get their own PRs since they add new mechanics.
Note: This is only based on `mostdor` since it uses the new calculated role logic.


src/CommandingOfficers/AW1/BM/Grit.java
```diff
@@ -87,6 +87,7 @@ private static class SnipeAttack extends AW1BasicAbility
     @Override
     protected void enqueueUnitMods(MapMaster gameMap, ArrayList<UnitModifier> modList)
     {
+      super.enqueueUnitMods(gameMap, modList);
       modList.add(rangeMod);
       modList.add(atkMod);
```
ThislsAUsername, line 92: I'm thinking it might be a good idea to make the `AW1BasicAbility` declare this `final` in the root and call a new function it defines for its children to overload?
Ditto for the rest of the games, 'cause I apparently keep forgetting to add these.

src/Engine/Combat/StrikeParams.java
```diff
@@ -100,6 +109,9 @@ protected StrikeParams(
     this.attackPower = attacker.attackPower;
     this.luckRolled = attacker.CO.luck;
     this.isCounter = isCounter;
+    aw1Luck = attacker.CO.aw1Combat;
+    if( aw1Luck && isCounter )
+      this.attackerHealth = attacker.health;
```
ThatIsAUsername, line 114: This makes it use precise health instead of rounded health for the calculation? Might toss that in a comment. Yes, that's basically what the code says, to the discerning eye, but it's also an odd thing to switch to exact HP just for AW1 and only for counterattacks, and I like having a comment to affirm non-obvious intent.

src/Engine/Combat/StrikeParams.java
```diff
@@ -115,14 +127,22 @@ protected StrikeParams(StrikeParams other)
...
   public int calculateDamage()
   {
     int luckDamage = getLuck();
-    final int rawDamage = baseDamage * attackPower * damageMultiplier / 100 / 100;
+    if( aw1Luck && isCounter )
+      luckDamage = 0;
+    final int rawDamage = baseDamage * attackPower * attackerDamageMultiplier / 100 / 100;
+    int hpScalingDamage = rawDamage;
+    if( !aw1Luck )
+      hpScalingDamage += luckDamage;
```
ThatIsAUsername, line 145: Changing luck damage to zero for counters is part of what FFFox talked about, but he also said that luck damage in AW1 does not scale with HP, but in later games it does. Are you capturing that also (or don't care)?

ThislsAUsername: Yup. It's a few lines down in the damage formula.

ThislsAUsername: I threw a nice doc comment on the `aw1Luck` variable.

src/CommandingOfficers/AW1/AW1Commander.java
```diff
@@ -0,0 +1,92 @@
...
+    // Add up the funds value of the damage done to both participants.
+    power += guiHPLoss * minion.unit.getCost() / 10;
+    power += guiHPDealt * enemy.unit.getCost() / 10 / 4;
+
+    return power;
+  }
+  @Override
+  public int calculateMassDamageCharge(Unit minion, int lostHP)
+  {
+    return 0;
```
ThatIsAUsername, line 55: No charge from damage from other CO's powers? Sad day.

ThislsAUsername: _waves cane_ Back in my day, the only sources of map damage were Drake and Sturm.
Drake is pretty bad (80A air, he does 1 damage for 40k charge, and no fuel thing), and Sturm is the final boss.
In that light, it makes a lot of sense.

src/Engine/UnitMods/DirectDamageModifier.java
```diff
@@ -0,0 +1,21 @@
...
+  public DirectDamageModifier(int firepowerChange)
+  {
+    attackModifier = firepowerChange;
+  }
+
+  @Override
+  public void modifyUnitAttack(StrikeParams params)
+  {
+    if( params.battleRange < 2 )
+      params.attackPower += attackModifier;
```
ThatIsAUsername, line 19: Does AW1 use both multipliers and modifiers in its damage calculations?

ThislsAUsername: Technically, literally everything is multiplicative except for COP stats overwriting normal stats.
Since additive buffs work just fine for that case and the mathematical result should be identical (modulo rounding error maybe), IMO it makes most sense to use this value for Max.
Max's COP stats are like 150 (base) +10 (power boost) and then multiply that all by 1.1x.
I probably could have used division defense for Sami (in the next PR), now that I think about it, but I gave her inf the Kanbei treatment (replace D2D buff with COP buff) since I didn't.

Rounding error is likely beyond what I can find it in myself to care about.

src/CommandingOfficers/AW1/OS/Max.java
```diff
@@ -0,0 +1,115 @@
...
+  public void modifyUnitAttack(StrikeParams params)
+  {
+    if( params.battleRange < 2 )
+      params.attackPower += 50;
+    if( params.attacker.model.isAny(UnitModel.INDIRECT) )
+    {
+      params.attackerDamageMultiplier *=  90;
+      params.attackerDamageMultiplier /= 100;
+    }
+  }
```
ThatIsAUsername, line 72: Is there a reason to prefer these modify functions over just attaching a permanent `UnitModifier` to the Commander? Looks like we still use that approach in other places.

ThislsAUsername: The Commander _is_ a `UnitModifier`.
But like, other than that? Fewer object types.

src/CommandingOfficers/AW1/BM/Grit.java
ThatIsAUsername: I could go either way I think. Making this final and creating a new override just means you'd have to remember to include the base bonuses for the description text along with any additions you make. Leaving it like this means subclasses have to remember to include the base bonuses at all with their additions.

src/CommandingOfficers/AW1/OS/Max.java
ThislsAUsername: To elaborate a bit:
In places where we want a temporary effect, applying a temporary `UnitModifier` makes sense (since ideally it should work across COs for my jank tag mechanics).

Very basic code for those is ideal, since then you can reuse them.
But oh no, I want to apply the bonus only to some units!
A role filter can usually do the job. It's a bit more abstract, but it makes modifiers composable in nice ways.

As for the direct attack modifier itself, it gets a pass because it will have 4 "unique" users with each one using it more than once.


I guess I can see having Max just use a permanent `UnitModifier` for his direct attack buff.
That creates an arguably unnecessary object, isn't much less code than overriding a method, and he'd still need code for his debuff.
So I feel like standardizing COs on custom code for their D2Ds makes the most sense.

src/Engine/UnitMods/DirectDamageModifier.java
ThatIsAUsername: Sure, for Max you could use either this or a `DamageMultiplierOffense` of 1.2, and the combat math should work out the same, yeah? Though if literally everything is multiplicative except for stat overwriting, I question that we have multiple methods of stat augmentation.

src/CommandingOfficers/AW1/OS/Max.java
ThatIsAUsername: > As for the direct attack modifier itself, it gets a pass

Not sure how that breaks your rules - just because it does its own filtering?

Otherwise fair enough. Potato potato, I was just curious.

src/Engine/UnitMods/DirectDamageModifier.java
ThislsAUsername: You are missing that `DamageMultiplierOffense` *multiplies* by 1.2, which would also affect his D2D buff. (and the generic COP buff of 1.1x scales both)
I don't want to have things adding to the damage multiplier because that creates order-dependence for your buffs.
(The "downside" of this is that a few COs get bigger power buffs for tag partners than they "should", but I'm not too torn up about it since AW1 COs are mostly weak and balance in tags is going to not exist anyway.)

Everything (in AW1) is multiplicative aside from COP stats replacing D2D stats.

src/CommandingOfficers/AW1/OS/Max.java
ThislsAUsername: Yeah, it does its own filtering so it's not composable... but the set of buffs that make sense for direct combat is like 3, so it'd be a much more minimal filter class.
I guess if we add DS skills we could break it apart, but I really don't plan to do that.

src/Engine/UnitMods/DirectDamageModifier.java
ThatIsAUsername: > You are missing that `DamageMultiplierOffense` multiplies by 1.2, which would also affect his D2D buff

Ok, just to lay this out more plainly to make sure I understand. The D2D is +50 for a base attack power of 150. Max force adds a base multiplier of 1.1 plus a direct boost of 20.
150 * 1.1 + 20 = 185 power. Whereas if we used a multiplier here it would be
150 * 1.1 * 1.2 = 198 power. Right?
I guess the obvious question; why not represent the AW1 base buffs as damage modifiers instead of multipliers to avoid order dependence and narrow to having one right and obvious way to do things?

> Everything (in AW1) is multiplicative aside from COP stats replacing D2D stats

Does that mean the D2D should not be active while the COP is?

src/CommandingOfficers/AW1/OS/Max.java
ThatIsAUsername: > Yeah, it does its own filtering so it's not composable

Dumb question then - why do that instead of composing it with the dedicated filter class?

src/Engine/UnitMods/DirectDamageModifier.java
ThislsAUsername: > The D2D is +50 for a base attack power of 150.
> Max force adds a base multiplier of 1.1 plus a direct boost of 20.
> 150 * 1.1 + 20 = 185 power.

It's actually (150 + 20) * 1.1 = 187 power, but yes.
> Whereas if we used a multiplier here it would be 150 * 1.1 * 1.2 = 198 power.

Yes.
> Right? I guess the obvious question; why not represent the AW1 base buffs as damage modifiers instead of multipliers to avoid order dependence and narrow to having one right and obvious way to do things?

Because that doesn't mathematically work. I can't add one constant number to the attack value for Andy and Max to get the right buff amount - it's 1.1x whatever the stat is.
Also, I can't even add one constant number to Eagle, Drake, and (next PR) Sami because all 3 of them have different statlines for different units.
The buff has to be a multiplier because that provides the correct result (and is also order-independent, as long as you group the mutually-commutative operations).
> Does that mean the D2D should not be active while the COP is?

I get around that by using addition, since that works for attack.
This also lets me use the same buff type that I'll use for the other 3 Max variants, and it gives correct results in all cases.

Kanbei does work like this (replacing D2D with powers) because he subtracts 10 from his defensive multiplier and "multiplies damage taken by 7/8ths" is a super janky thing to put in a power description.

src/CommandingOfficers/AW1/OS/Max.java
ThislsAUsername: Because the dedicated filter class can't filter by "at range 1", and I want to use my own silly interpretation of how Max works.

src/Engine/UnitMods/DirectDamageModifier.java
ThatIsAUsername: Okay, it sounds like the answer to my initial question was a simple "yes". Is Kanbei the only one who deactivates his D2D during COP? and that matches cart behavior? That's pretty strange.

src/CommandingOfficers/AW1/OS/Max.java
ThatIsAUsername: ooooookay

ThatIsAUsername: I mean, it's the same, right? I can't think of when these wouldn't be the same. But okay.

src/Engine/UnitMods/DirectDamageModifier.java
ThislsAUsername: In my implementation, yes.
The original uses tables of stats which multiply with the CO's global attack and defense multipliers (which Kanbei and Sturm notably modify during D2D).

Kanbei deactivates his D2D during COP to replace it with the stronger COP stats, because math for 0.8x to 0.7x would be weird.
It might be doable with DoR-style defense, but I didn't think of that.

Sturm and Sami will also have similar effects:
Sturm (like Kanbei) has a global stat multiplier that he makes integer changes to.
Sami has an infantry attack/defense boost that she makes integer changes to (in addition to the default power boost).

It'd be possible to encode their final effects explicitly, but I like modeling the math in a similar way to the original.
The reason it's not exactly like the original is I don't want to mix multiplication and addition on a single value because that makes order important.

src/CommandingOfficers/AW1/OS/Max.java
ThislsAUsername: Anti-Tanks.
I am a very silly goose.

PR #410, opened 2024-01-03
Sami One

ThislsAUsername:
Add that capture boost mechanic so we can have nice things.


src/Units/Unit.java
```diff
@@ -86,6 +86,28 @@ public UnitContext getRangeContext(GameMap map, WeaponModel weapon)
     return new UnitContext(map, this, weapon);
   }
 
+  public boolean capture(MapLocation target)
```
ThatIsAUsername, line 89: Why move `capture` from `UnitState` to `Unit`?

ThislsAUsername: There's no `UnitContext` constructor that takes a `UnitState`, and I didn't feel like making one.

ThatIsAUsername: Fair enough I guess. This PR is pretty straightforward otherwise, so fire when ready.

PR #411, opened 2024-01-03
Meteor Mania

ThislsAUsername:
Adds all the nukey COs from the carts.


src/CommandingOfficers/AW1/BH/Sturm.java
```diff
@@ -0,0 +1,174 @@
...
+  private static class instantiator extends CommanderInfo
+  {
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Sturm_1", UIUtils.SourceGames.AW1, UIUtils.BH);
+      infoPages.add(new InfoPage(
+            "Sturm (AW1)\n"
+          + "A riddle within a shadow, revealing nothing.\n"
+          + "???\n"
```
ThatIsAUsername, line 44: Are we just tossing question marks in here for appearance's sake and then explaining everything anyway?

ThislsAUsername: Hey, man. That's the original description. I dunno what you want from me.

src/CommandingOfficers/MeteorParams.java
```diff
@@ -0,0 +1,116 @@
...
+import Engine.GameEvents.MassDamageEvent;
+import Terrain.GameMap;
+import Terrain.MapMaster;
+import Units.Unit;
+
+public class MeteorParams
+{
+  public XYCoord target; // This is allowed to be null. If it is, then this class will produce a no-op MassDamageEvent.
+  public int power = 3; // Silo
+  public int radius = 2; // ditto
```
ThatIsAUsername, line 16: Silo?

ThislsAUsername: Rachel and VB's damage matches Silo damage?

src/CommandingOfficers/MeteorParams.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+    }
+    return new MeteorParams(maxTarget, radius);
+  }
+
+  /**
+   * Restricts targeting to always center on enemy units.
+   * @param targetingMap The map to use to find units to center on.
+   * @param scoringMap The map to use to score valid strike locations.
```
ThatIsAUsername, line 90: Why would these be different?

ThislsAUsername: AW2 Sturm is a tricky boi.
He can center on any unit, but he only counts units he can see for damage... and also excludes steathed units even if he can see them.

src/CommandingOfficers/AW1/BH/Sturm.java
```diff
@@ -0,0 +1,174 @@
...
+          + "(1.3x/1.2x damage dealt/taken while COP isn't active, perfect movement outside snow.)"));
+      infoPages.add(new InfoPage(new MeteorStrike(null),
+            "???\n"
+          + "(1.5x/1.1x damage dealt/taken.)\n"
+          + MeteorStrike.POWER+" HP meteor with a 1/3 chance of targeting HP, funds, or funds with doubled indirects.\n"
```
ThatIsAUsername, line 49: I thought it targeted max funds? AWW says "The target will be chosen by finding the biggest amount of fund damage the meteor can do".

ThislsAUsername: I would not implicitly trust anything on the AW Wiki.
We don't have datamined info on it, so the targeting may be different, but I figure I'll match AW2's logic (sans fog jank) until such time as "otherwise" is proven since I've heard it's the same.
I guess I could test it, but I dunno how RN rolling works.

AWBW Wiki is usually right re: AWBW.
WarsWorldNews is probably the most reliable for the carts.

src/CommandingOfficers/AW1/BH/SturmVS.java
```diff
@@ -0,0 +1,174 @@
...
+    {
+      super("Sturm_VS", UIUtils.SourceGames.AW1, UIUtils.BH);
+      infoPages.add(new InfoPage(
+            "Sturm (AW1 VS)\n"
+          + "A riddle within a shadow, revealing nothing.\n"
+          + "???\n"
+          + "(0.8x/0.8x damage dealt/taken while COP isn't active, perfect movement outside snow.)"));
+      infoPages.add(new InfoPage(new MeteorStrike(null),
+            "???\n"
+          + "(0.9x/0.7x damage dealt/taken.)\n"
```
ThatIsAUsername, line 48: His attacks are worse than standard even during his power?

ThislsAUsername: Same with Grimm's defense, yeah.

src/CommandingOfficers/MeteorParams.java
```diff
@@ -0,0 +1,116 @@
...
+      {
+        currentValue += evaluator.getValue(attacker, found);
+      }
+    }
+
+    return currentValue;
+  }
+
+  /**
+   * @return A MassStrikeParams with the best target, or no target if there are no values above 0.
```
ThatIsAUsername, line 65: Return type is wrong. Same for `findValueOnEnemy` below.

Also these names are not... ideal. `getScore` and `findValue` just sound like synonyms, and `findValueOnEnemy` is just confusing.

src/CommandingOfficers/SturmValueFinders.java
```diff
@@ -0,0 +1,46 @@
+package CommandingOfficers;
+
+import Units.Unit;
+import Units.UnitModel;
+
+public abstract class SturmValueFinders // Until such time as we find out AW1 works substantially differently
+{
+  public static class HPValueFinder implements IValueFinder
+  {
+    public boolean countHidden = true; // Ignore cloaked units even if we can see them
```
ThatIsAUsername, line 10: Value says "yes, we will count hidden units", but comment seems to indicate "no, we shouldn't count them."

They say that if the code and comments disagree, they are both wrong. Since this is a Boolean expression, that's literally impossible here, but y'know, it's the principle of the thing.

ThislsAUsername: Comment is missing a few words of context. Fair enough.

src/CommandingOfficers/AW2/BH/Sturm.java
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    public GameEventQueue getEvents(MapMaster map)
+    {
+      MeteorParams[] targets = findTargets(map);
+      int rand = map.game.getRN(targets.length);
+      MeteorParams meteor = targets[rand];
+      GameEventQueue events = super.getEvents(map);
+
+      if( null == meteor.target )
+        return events;
```
ThatIsAUsername, line 122: Unnecessary null check.

src/CommandingOfficers/AW3/AW3ValueFinders.java
```diff
@@ -0,0 +1,73 @@
...
+      health = Math.min(health, maxDamage * 10);
+
+      int value = unit.getCost() * health / UnitModel.MAXIMUM_HEALTH;
+
+      if( !unit.CO.isEnemy(attacker) )
+        return -value;
+      return value;
+    }
+  }
+  public static class HealthValueFinder implements IValueFinder
```
ThatIsAUsername, line 30: Sturm's is called `HPValueFinder`. Minor nit.

src/CommandingOfficers/AW3/AW3ValueFinders.java
```diff
@@ -0,0 +1,73 @@
...
+
+public abstract class AW3ValueFinders
+{
+  public static class CostValueFinder implements IValueFinder
+  {
+    int maxDamage = 3;
+    @Override
+    public int getValue(Commander attacker, Unit unit)
+    {
+      // Ignore Oozium, if we ever add those
```
ThatIsAUsername, line 17: Why? Also, should Sturm's version have a similar note?

src/CommandingOfficers/AW3/AW3ValueFinders.java
```diff
@@ -0,0 +1,73 @@
...
+      int health = unit.getHealth();
+      if( health < 10 )
+        return 2;
```
ThatIsAUsername, line 20: Value is always 2 if health is less than 10? Whereas full-health units start at 1000? Is this right?

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: Because the carts do that, and Oozium is immune to mass damage of all flavors.
Sturm never coexisted with Oozium, so eh.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: Don't ask me, I just work here.
It's from Xenesis's meteor datamine thread.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: And if you look closely, you'll see he uses HP instead of health.

src/CommandingOfficers/AW3/AW3ValueFinders.java
```diff
@@ -0,0 +1,73 @@
...
+    int maxDamage = 3;
+    public int getValue(Commander attacker, Unit unit)
+    {
+      // Ignore Oozium, if we ever add those
+      int health = unit.getHealth();
+      if( health < 10 )
+        return 1;
+      health = Math.min(health, maxDamage * 10);
+
+      for( UnitActionFactory action : unit.model.baseActions ) // Does it make sense for this to be dynamic?
```
ThatIsAUsername, line 57: If we have or ever intend to have units that can capture but only sometimes, then yes.
Also, do we care about capture power? Might want to feel more threatened by Sami's infs than others. Though you can't really factor in her SCOP capture power, since if you can see that it's active, it's too late to stop it. I guess you could also poll the other CO's star meter...

ThislsAUsername: We don't currently, and it's a very jank design space that I dunno if I like.
No. That has no basis in the cart logic.
That... also has no basis in cart logic, but I have very vague plans to use the mod enqueue hook at some point to let AIs plan around powers and/or implement a fancy damage calc screen.

src/CommandingOfficers/MeteorParams.java
ThatIsAUsername: Oh. I guess I probably should have made the connection, but for once, I'm not sure we need a comment to explain this. I guess it doesn't hurt though. Unless the reader just gets confused.

src/CommandingOfficers/MeteorParams.java
ThatIsAUsername: Oh, so you use the map master for targeting, but the map view for scoring. Interesting.

src/CommandingOfficers/AW1/BH/Sturm.java
ThatIsAUsername: Huh. Then should his Hit/Miss be Invasions/Peace to match https://www.warsworldnews.com/wp/aw12r/co-aw12r/sturm-2/ ? Or was that a retcon in RBC?
That link agrees with the targeting logic anyway.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThatIsAUsername: I didn't look closely. I assumed that two functions to do apparently the same job would work the same way. Why don't they work the same way?

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThatIsAUsername: Sturm can mass damage Oozium, got it. Sturm OP.

src/CommandingOfficers/MeteorParams.java
ThatIsAUsername: So now we can plan meteor strikes against the enemy specifically, or just plan them? Is the difference just in the use of two maps to make AW2 Sturm work right? Is one a special case of the other? I could try to read the code and find out, but it's a weeknight.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThatIsAUsername: Drop a short explanation and link in a comment, maybe? That's super non-obvious at best, and looks like a bug on a normal day.

src/CommandingOfficers/MeteorParams.java
ThislsAUsername: Except I also use the map perspective for targeting (in the case of previewing the targets), because nothing is allowed to be simple.

src/CommandingOfficers/AW1/BH/Sturm.java
ThislsAUsername: image

src/CommandingOfficers/MeteorParams.java
ThislsAUsername: "On enemy" means "centered on a particular enemy unit", so it is indeed a special case of the other one. I guess the docstring could say that.
One map is used to check whether an enemy is on the center tile (so you can hit there), the other is used to calculate the expected value of the strike.

I thought about making that logic live in the value finder interface, but figured it fit better this way.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: Because AW3's targeting has special casing for units under 1.0 HP.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: Done.

ThatIsAUsername: And you didn't share your newfound knowledge back with the rest of humanity? for shame...

src/CommandingOfficers/MeteorParams.java
ThatIsAUsername: So then the preview would not match the actual shot? Fun fun.

ThislsAUsername: Of course. The other guy might have put a stealthed/forest unit in a very convenient targeting spot, or something.

src/CommandingOfficers/AW3/AW3ValueFinders.java
ThislsAUsername: I did tell Xenesis. What she puts in her thread is her prerogative, lol
I could make my own account, ooooor I could be lazy.

At any rate, this code is open sauce.

src/CommandingOfficers/MeteorParams.java
ThatIsAUsername: > "On enemy" means "centered on a particular enemy unit"

Ah, I didn't make that connection. Just looked like it meant "at the enemy" which seems... obvious?

> EDIT: Past me agreed

So you did. Must have glossed over it, maybe because the comment on the second just says it restricts targeting, and doesn't otherwise connect itself to the first (I mean the names are similar, but they also aren't overloads). Code monkey like to be hit in the face with extremely unsubtle comments, like "This function is the same as `planMeteor`, except it restricts targeting...", and then probably put a parallel/reversed comment on the other one stating that it doesn't have the restriction.

> it is indeed a special case of the other one

Could there just be an optional Boolean to restrict targeting? I guess that might be cumbersome given the map treatment differences. Probably fine as is, I'm sure you considered consolidating things.

ThislsAUsername: > Ah, I didn't make that connection. Just looked like it meant "at the enemy" which seems... obvious?

Yeah, "On" isn't the most disambiguated word.
> Code monkey like to be hit in the face with extremely unsubtle comments

I suppose that's fair.
> Could there just be an optional Boolean to restrict targeting? I guess that might be cumbersome given the map treatment differences. Probably fine as is, I'm sure you considered consolidating things.

Not super conveniently.
The "on" variant needs a separate map for finding units to center on, which isn't needed at all on the other version.

I was also originally planning to have the "on" variant directly iterate over the enemy's unit collection, but then I was lazy when I remembered transported units exist (at -1,-1).

You could probably do it, but it kinda makes sense to leave them separate IMO.

src/CommandingOfficers/AWBW/BH/Sturm.java
```diff
@@ -127,7 +127,10 @@ private MeteorParams findTarget(GameMap map)
     {
       AWBWValueFinders.CostValueFinder finder = new AWBWValueFinders.CostValueFinder();
       finder.maxDamage = power;
-      return MeteorParams.planMeteorOnEnemy(map, map, myCommander, 2, finder);
+      MeteorParams meteor = MeteorParams.planMeteorOnEnemy(map, map, myCommander, 2, finder);
+      if( meteor.target == null )
+        meteor.target = new XYCoord(0, 0);
+      return meteor;
```
ThatIsAUsername, line 133: Could this logic be shuffled inside the `planMeteor...` calls so that you don't have to remember to manually reset it after planning?

Also does this happen? Is AWBW behavior different than the carts?

ThislsAUsername: Probably, though I dunno if it really makes sense to go into the function signature there.
As far as I can tell, yes. Testing is like pulling teeth, but it does work like that in the Move Planner.

Also, the AWBW meteor targeting is different in a bunch of ways, so I dunno why you're surprised.

ThatIsAUsername: What do you mean, "go into the function signature"? Seems like it would be straightforward to either a) modify the `MeteorParams` constructor to replace `null` with `(0, 0)`, or b) change `planMeteor...` to default the target to `(0, 0)` instead of `null` at the outset, so it's already the right thing if it doesn't get replace.

> I dunno why you're surprised.

Ignorance has that effect on people.

ThatIsAUsername: Oh, gotcha. the (0, 0) variant is only for the AWBW dudes. My brain is still adjusting to thinking of that as an independent AW game/category.

ThatIsAUsername:
All right, fire when ready, I guess.

PR #412, opened 2024-01-04
Add Bukie's infs, with permission

ThislsAUsername:
We already have an Azure, which does make adding an "Azure Asteroid" pretty awkward.
But the sprites are nice.


ThatIsAUsername:
A second Azure that is almost indistinguishable from the first? I dunno, man. I guess people can do what they want.

I guess it's cool to have an Oozium faction though, and I like the roaches too.

ThislsAUsername:
Yeah. I'm not a huge fan of throwing away either palette, and also not a huge fan of providing this faction without its "canon" colors.
Either way, I think I will let this sit for a while since they've already gotten a V2 and may get more edits later.

PR #413, opened 2024-01-05
Fancy Fog

ThislsAUsername:
Aight, so this one's a little involved.
The core thing here is adding support for both DoR-style and Trilogy-style fog.

I ran into a crash while testing some of the other PRs, so I simplified the code and made it stop doing bad things and maybe that fixed it.
The crash was on line 56 of CommanderProductionInfo's constructor, where it was trying to get an entry for another army's CO. Dunno exactly how that ended up happening, but I suspect I solved it.

You'll notice there's an "off" setting for both types of fog.
There will be powers that bring up FoW, so we need to know ahead of time which type it will be.

I also got rid of the separate piercing vision range - it seems to me that you could get into bad situations with that.
Say Sonja uses a power, then there's another vision boost that doesn't set piercing. Now you have a piercing vision mismatch during a Sonja power, which seems wrong.

I've added two COs that mess with vision (Lin and AW1 Sonja).
I also added handling for air units in forests and rain's vision malus.


src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -284,7 +285,9 @@ public void drawUnitIcons(Graphics g, GameInstance game, Unit u, int animIndex,
     }
 
     // Transport icon.
-    if( u.heldUnits != null && !u.heldUnits.isEmpty() )
+    boolean fogOn = map.game.isFogEnabled();
+    boolean hideCargo = fogOn && u.CO.isEnemy(map.viewer);
+    if( !hideCargo && !u.heldUnits.isEmpty() )
```
ThislsAUsername, line 290: The carts replace the envelope icon with an always-active dash icon, which I'm not a super fan of.
I'm not a huuuuge fan of just hiding the icon, but bleh.

Also, adding an icon now would be annoying with #417 pending

src/Terrain/MapLocation.java
```diff
@@ -40,19 +40,13 @@ public Commander getOwner()
     return owner;
   }
 
+  /**
+   * If the actual map state is changing, make sure to update the ownedProperties lists.<p>
+   * MapMaster.setOwner() is a convenient way to do so.
+   */
```
ThatIsAUsername, line 46: Are there times we might call this where the actual map state is _not_ changing?

src/Terrain/MapMaster.java
```diff
@@ -237,6 +238,29 @@ public boolean isLocationEmpty(Unit unit, int x, int y)
...
+    setOwner(owner, xyc.x, xyc.y);
+  }
+  public void setOwner(Commander owner, int x, int y)
+  {
+    MapLocation loc = getLocation(x, y);
+    Commander prevOwner = loc.getOwner();
+    if( prevOwner == owner )
+      return;
+
+    XYCoord xyc = loc.getCoordinates();
```
ThatIsAUsername, line 252: Seems a bit wonky to have the (x, y) overload be the "base" one when it just reconstitutes the `XYCoord` halfway through. Not that this is like, important.

src/Terrain/MapLocation.java
ThislsAUsername: Yup, `MapPerspective` does this. Quoth commit 1:
```
Move CO.ownedProperties management up to MapMaster

MapPerspective calls location.setOwner() a lot when it's creating
 the ephemeral capturable location objects.
I don't know for certain that that's the smoking gun on a crash
 in CommanderProductionInfo's constructor, but it seems messed up.
It also made the crash pretty well impossible to debug.
```

src/Terrain/MapMaster.java
ThislsAUsername: `loc.getCoordinates()` doesn't create a coord, so it's one less total allocation.
Truly, the pinnacle of optimizations.

src/Engine/GameScenario.java
```diff
@@ -46,6 +46,19 @@ private TagMode(boolean multiCO)
...
+  {
+    OFF_TRILOGY(false, false), ON_TRILOGY(true, false), OFF_DOR(false, true), ON_DOR(true, true);
+
+    public final boolean fogDefaultsOn;
+    public final boolean dorMode;
+    private FogMode(boolean isOn, boolean DoR)
+    {
+      this.fogDefaultsOn = isOn;
+      this.dorMode = DoR;
+    }
```
ThatIsAUsername, line 59: Whoah, are we saying we can't have two different kinds of fog active at the same time? I'm kinda surprised.

src/Terrain/MapPerspective.java
```diff
@@ -241,29 +257,65 @@ public void revealFog(Unit scout, GamePath movepath)
...
-      confirmedVisibles.add(loc.getResident());
+    Unit resident = loc.getResident();
+
+    TerrainType tt = loc.getEnvironment().terrainType;
+    boolean shouldSee = piercing || !tt.isCover(master.game.rules.fogMode);
+    if( null != resident )
+    {
+      if( piercing )
+        confirmedVisibles.add(resident);
+      else if( !resident.model.hidden ) // Non-invisible aircraft reveal cover that can't repair them.
```
ThatIsAUsername, line 314: invisible aircraft do not?

ThatIsAUsername: Oh, this is going the other way. An aircraft that is seen will also reveal the terrain underneath it, even if it is (otherwise) cover. Makes sense.

src/Units/UnitContext.java
```diff
@@ -208,6 +237,25 @@ public int calculateCapturePower()
...
+  {
+    visionRange = model.visionRange;
+    visionPierces = model.visionPierces;
+    if( null != env )
+    {
+      // if it's a surface unit, give it the boost the terrain would provide, so long as it's not adjacent-only vision
+      if( model.isSurfaceUnit() )
+        visionRange += env.terrainType.getVisionBoost();
+      if( env.weatherType == Weathers.RAIN )
+        --visionRange;
```
ThatIsAUsername, line 250: This code doesn't appear to have an accommodation for "adjacent-only" vision, and I'm not sure when that's even relevant.

src/CommandingOfficers/AW1/YC/Sonja.java
```diff
@@ -0,0 +1,88 @@
...
+
+    addCommanderAbility(new EnhancedVision(this));
+  }
+  @Override
+  public void modifyVision(UnitContext uc)
+  {
+    uc.visionRange += 1;
+  }
+
+  private static class EnhancedVision extends AW1BasicAbility
```
ThatIsAUsername, line 61: I notice Sturm, Eagle, and Kanbei all inherit from `AW1Ability` instead of `AW1BasicAbility` - is that on purpose?

src/UI/Art/SpriteArtist/InfoView.java
```diff
@@ -92,7 +92,12 @@ public void render(Graphics g)
       return; // Continue no further if we don't have room to actually draw
 
     if (null != thisArmy)
-      CommanderOverlayArtist.drawCommanderOverlay(myG, thisArmy, false);
+    {
+      MapPerspective drawableMap = getDrawableMap(myControl.getGame());
+      boolean fogOn = myControl.getGame().isFogEnabled();
+      boolean hideFunds = fogOn && thisArmy.isEnemy(drawableMap.viewer);
+      CommanderOverlayArtist.drawCommanderOverlay(myG, hideFunds, thisArmy, false);
```
ThatIsAUsername, line 99: Intent here is to show the enemy Commander's unit count/income/etc based on what the active CO can see, right? That feels potentially misleading, since there is no disclaimer. Is this what the carts do?
Also it looks like the vehicle count is counting troops instead of vehicles (see COStateInfo line 46).

src/Engine/GameScenario.java
ThislsAUsername: I really can't justify that being intuitive from any direction, hah.
I think I did leave open a loophole in that local fog generation should be possible, but I'm not convinced I'll ever use it.

src/Terrain/MapPerspective.java
```diff
@@ -194,24 +217,18 @@ public void resetFog()
...
         }
         for( XYCoord xyc : co.ownedProperties )
         {
           revealFog(xyc, true); // Properties can see themselves and anything on them
-          MapLocation loc = master.getLocation(xyc);
-          for( XYCoord coord : Utils.findVisibleLocations(this, loc.getCoordinates(), Environment.PROPERTY_VISION_RANGE) )
-          {
-            revealFog(coord, false);
-          }
+          if( isFogDoR() ) // Trilogy fog does not enable cities to have area vision
```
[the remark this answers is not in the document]
ThislsAUsername: It may make sense to move to a "What vision range do you have, here's the fog type" function call on the `TerrainType`, since Radars are a thing we could add at some point.
But I didn't feel like finagling that when I really just need 2 states and customizing terrains is annoying.

src/Units/UnitContext.java
ThislsAUsername: That's in the next PR. DoR rain doesn't penalize movement, so it's a different weather type.

FWIW, I decided that DS and DoR rain are the same weather just acting differently in each fog mode (since both cause fog and penalize vision with no other effects).

src/UI/Art/SpriteArtist/InfoView.java
ThislsAUsername: Yes, it's been that way since we've had fog.
It's a little weird, sure, but it's not hard to figure out how it works... and it saves you calculation without giving you free info.
The carts (weirdly) give you a complete unit count (and units lost), but nothing on funds. I guess that makes sense when rout is a major objective and the game's focused on single player. They also give you a complete property count if capture limit is on.
> Also it looks like the vehicle count is counting troops instead of vehicles (see COStateInfo line 46).

You appear to be correct.

src/Units/UnitContext.java
ThatIsAUsername: Okay, maybe I'll go complain about it in that PR then.

PR #414, opened 2024-01-05
Fancy Weather

ThislsAUsername:
This adds support for all the weathers in all 4 games.
It's worth noting that Drake has never actually been immune to the vision penalty rain brings.

I decided to merge the Rains from DS and DoR since they both do the same vague thing (force fog, reduce vision)
I called this `SMOKE`, since "rain" is taken.

I moved all the logic for handling weather (and immunity to weather) into the relevant central logic (mostly `UnitContext`, but `StrikeParams` and `GamePath` each got one).
That gets rid of the hacky modifier I set up at the top level, which is nice.

As of this PR, my to-do list before the next release is the following:
Add Comm Towers (Should DoR COs treat towers differently, or should DoR towers be a different terrain type? They give +5/5 stats and income in DoR)
Some tweaks/bugfixes to my COs.
Make JakeMan and maybe Muriel stop yeeting infs so much (I think I messed something up with the HP changes).
Finish this round of Wally changes.
Maybe some other feature(s) if I get distracted again.
Add COs. All the COs. aaaaaaaaaaaaaaaaaaa


src/Terrain/Environment.java
```diff
@@ -15,7 +15,21 @@ public class Environment implements Serializable
 
   public enum Weathers
   {
-    CLEAR, SNOW, RAIN, SANDSTORM
+    //      AW2 snow           AW2
+    CLEAR, SNOW(true, false), RAIN,
+    //  AWDS    AWDS snow
+    SANDSTORM, CHILL(true, false),
+    //   DoR snow        AW3/4 rain          DoR sandstorm
+    SLEET(true, false), SMOKE(false, true), SIROCCO;
```
ThatIsAUsername, line 23: Is there any good way to tell what weather is in effect while in-game? I'm certainly not gonna memorize 8 different subtly-different overlay colors. Or maybe I haven't gotten to that yet.

ThatIsAUsername: Also is RANDOM weather just going to cycle between everything? Or have game-themed subsets of weather? do you just not plan to support that?

ThislsAUsername: Uh, no, not really.
It's at least always tied to a global option or a CO activating it, at this point, which really limits the options.
Clearly, I need to add a little weather dial like Wargroove has, but it changes based on the tile you're mousing over.
(that is so not in scope for now, though)

Random is... yeah, I dunno. It's not really something I care about, and I really doubt the players will ask for it.
I would probably do separate randoms tied to each game, yeah.
It'd be fun to do a randomized local weather system, but that + overlays (like DoR COs) could make for some very unreadable gamestates (even if we made custom assets for every weather, which is _so_ not happening).

src/Units/UnitContext.java
```diff
@@ -246,8 +250,16 @@ public int calculateVision()
...
         visionRange += env.terrainType.getVisionBoost();
-      if( env.weatherType == Weathers.RAIN )
+      if( env.weatherType == Weathers.RAIN && !CO.immuneToClouds )
         --visionRange;
+      if( env.weatherType == Weathers.SMOKE && !CO.immuneToClouds )
+      {
+        // SMOKE sets vision to 1 (before buffs but after mountains) in DoR, but Drake's version in DS only subtracts 1 from vision.
+        if( null != unit && unit.CO.gameRules.fogMode.dorMode )
+          visionRange = 1;
+        else // Default to Trilogy logic, since "has most of normal vision" will stick out more vs "vision = 1" than vice versa.
```
ThatIsAUsername, line 260: I can't parse this as a coherent English sentence. You are saying the trilogy logic will stick out more, which typically connotes badness, but you also have it as the default, and then you have "a" vs "b" vs vice versa ("b" vs "a"?)

src/Engine/GameEvents/GlobalWeatherEvent.java
```diff
@@ -41,14 +42,24 @@ public GameEventQueue sendToListener(GameEventListener listener)
...
+        loc.setForecast(weather, (map.game.armies.length * duration) - 1);
+      }
+    }
+    if( weather.startsFog )
+    {
+      map.game.setFog(duration);
+      for( Army a : map.game.armies )
+      {
+        if( !a.isEnemy(map.game.activeArmy) )
+          continue; // This is probably technically laxer than it should be (allows allies to watch your turn with full vision in DoR fog), but like... it seems more fun to see things than not.
```
ThatIsAUsername, line 61: This only impacts the weather COs at this point, right? Or doe allies also get vision if fog is on by default (and is there a way to consolidate fog-vision handling to one location?)

src/Units/UnitContext.java
ThislsAUsername: "If badness occurs, the badness will be more obvious".
I like bugs being obvious-er

src/Engine/GameEvents/GlobalWeatherEvent.java
ThislsAUsername: So, this only matters if fog is normally off.
If DoR fog turns on *during your turn*, you keep full map vision. That's cart-accurate.
This check also lets your allies keep full map vision during your turn.

src/Units/UnitContext.java
ThatIsAUsername: Okay. I like bugs being obvious. Fail fast and all that. And I can infer from the code that you want to default to Trilogy behavior because you think it will be more noticeable when erroneously applied. I just don't get that from the comment.
"A sticks out more than B than vice versa" just reads like nonsense.

ThislsAUsername: "A will stick out more vs B than vice versa"
"A on the backdrop of B is more visible than the reverse"

ThatIsAUsername: That makes sense. The structure of the comment sentence is just too tortured and terse to be sure that's what it means.

ThislsAUsername: I do enjoy terseness.

ThatIsAUsername: And I commend clarity.

ThatIsAUsername:
Think I'm done with this one.

PR #415, opened 2024-01-14
Get Muriel and JakeMan to stop throwing inf at tanks

ThislsAUsername:
It was bugging me.

### DIFF
--- a/src/AI/JakeMan.java
+++ b/src/AI/JakeMan.java
@@ -52,6 +52,7 @@ public class JakeMan extends ModularAI
   private static final double FIRSTSTRIKE_ON_THREAT_WEIGHT = 2.0;
   private static final int    STAY_UNHURT_BIAS = 1000;
   private static final int    STAY_ALIVE_BIAS = 2000;
+  private static final int    BIG_THREAT_THRESHOLD = 80; // Enemy health at which I double the expected value of dealing damage
   // Fraction of the unit to remove from the counter-threat power of my unit type if I'm not attacking
   private static final double PEACEFUL_SELF_THREAT_RATIO = 1;
   private static final int    UNIT_HEAL_THRESHOLD = 60; // Health at which units heal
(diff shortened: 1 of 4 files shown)


src/AI/JakeMan.java
```diff
@@ -52,6 +52,7 @@ public AIMaker getAIInfo()
   private static final double FIRSTSTRIKE_ON_THREAT_WEIGHT = 2.0;
   private static final int    STAY_UNHURT_BIAS = 1000;
   private static final int    STAY_ALIVE_BIAS = 2000;
+  private static final int    BIG_THREAT_THRESHOLD = 80; // Enemy health at which I double the expected value of dealing damage
```
ThatIsAUsername, line 55: This is to make attacking stronger/high-HP defenders more likely? As in, it's twice as valuable to chip down threats to below 9 HP as it is to do damage below that? Sure, I guess.

ThislsAUsername: It's basically an incentive to firststrike the most threatening thing, since 2 5 HP tanks is usually less valuable immediately than 1 full HP tanks (because of how counters/etc work)

src/AI/Muriel.java
```diff
@@ -457,15 +457,15 @@ private boolean queueUnitAction(GameMap gameMap, Unit unit, boolean allowDeferri
...
         // Sift through all attack actions we can perform.
-        double damageValue = AICombatUtils.scoreAttackAction(unit, action, gameMap,
+        int damageValue = AICombatUtils.scoreAttackAction(unit, action, gameMap,
             (results) -> {
-              double healthDamage = Math.min(results.defender.getPreciseHealthDamage(), results.defender.unit.health);
+              int healthDamage = Math.min(results.defender.getPreciseHealthDamage(), results.defender.unit.health);
 
               if( shouldAttack(unit, results.defender.unit, gameMap) )
-                return (results.defender.unit.getCost() / 10) * healthDamage;
+                return (results.defender.unit.getCost() / UnitModel.MAXIMUM_HEALTH) * healthDamage;
```
ThatIsAUsername, line 465: Shoulda known better than to use a magic number, even for something as "unchangeable" as unit max HP. Silly me.

ThatIsAUsername:
Looks dandy.

PR #416, opened 2024-01-14
Fix Grit buffing everyone's range

ThislsAUsername:
Dunno how I missed this, since I coulda sworn I tested it.

### DIFF
--- a/src/CommandingOfficers/AW1/BM/Grit.java
+++ b/src/CommandingOfficers/AW1/BM/Grit.java
@@ -66,7 +66,7 @@ public class Grit extends AW1Commander
   public void modifyAttackRange(UnitContext uc)
   {
     if( uc.weapon != null && uc.weapon.rangeMax > 1 )
-      uc.weapon.rangeMax += 1;
+      uc.rangeMax += 1;
   }
 
   private static class SnipeAttack extends AW1BasicAbility


ThatIsAUsername:
> Dunno how I missed this, since I coulda sworn I tested it.

I don't see any tests for it...

src/CommandingOfficers/AW1/BM/Grit.java
```diff
@@ -66,7 +66,7 @@ public void modifyUnitAttack(StrikeParams params)
   public void modifyAttackRange(UnitContext uc)
   {
     if( uc.weapon != null && uc.weapon.rangeMax > 1 )
-      uc.weapon.rangeMax += 1;
+      uc.rangeMax += 1;
```
ThatIsAUsername, line 69: I guess that's whatTwo things:
1) `WeaponModel` is mutable? I probably knew this at one point, but I also know I would only have named the class `*Model` if I intended for all of its members to be unchangeable.
2) Again, we probably hashed this out before, but why copy fields from `WeaponModel` into `UnitContext`, when the latter already owns a copy of the former? Just asking for trouble like this (especially when the immutable backstop isn't there).

ThislsAUsername:
> > Dunno how I missed this, since I coulda sworn I tested it.
> 
> I don't see any tests for it...

I playtested it.
Do you really expect me to write a unit test for every CO?

src/CommandingOfficers/AW1/BM/Grit.java
ThislsAUsername: > `WeaponModel` is mutable?

The old design actually did modify the `WeaponModel`s and `MoveType`s directly for powers/D2Ds, which is why they had to be cloned for each CO's `UnitModel`s. I believe they're mostly still cloned because I'm not massively worried about a few bytes of duplicated data here or there.

Pretty much everything is mutable. We just don't change them during gameplay, because that would be weird.
I believe that's what we agreed on.
> I also know I would only have named the class *Model if I intended for all of its members to be unchangeable.

We _could_ make everything in the `*Model` classes immutable, but then we'd need some kind of builder pattern for Tech's mechs. I would be content with doing that via a library, but I really don't feel like writing that code myself.

> Again, we probably hashed this out before, but why copy fields from `WeaponModel` into `UnitContext`, when the latter already owns a copy of the former?

I imagine I did it to make it easy to substitute a `UnitContext` easily for either a `Unit` or a `UnitModel`.
I don't think we've hashed this one out, actually.
> Just asking for trouble like this (especially when the immutable backstop isn't there).

Fair criticism.

ThatIsAUsername:
> > > Dunno how I missed this, since I coulda sworn I tested it.
> > 
> > 
> > I don't see any tests for it...
> 
> I playtested it. Do you really expect me to write a unit test for every CO?

What? No no no, of course not. I'd only expect unit tests for the COs you want to be sure are working correctly.

Anyway, this appears to address the immediate problem, and a larger fix for the structural issues can wait until

Actually, I don't see much point in finishing that sentence.

ThislsAUsername:
> What? No no no, of course not. I'd only expect unit tests for the COs you want to be sure are working correctly.

I wouldn't have thought to check for that in a unit test anyway, so...
> Anyway, this appears to address the immediate problem, and a larger fix for the structural issues can wait until
> 
> Actually, I don't see much point in finishing that sentence.

Heh. Yes.
I mean, I'm down with adding e.g. Lombok if you are: https://projectlombok.org/features/
That would fix the structural issues in a way I don't hate.

ThatIsAUsername:
> Lombok

I know you have mentioned that before, but it's kinda in the "I don't know what this is or what it does, and therefore don't see any reason we would need it" category of things.

Anyway, this should be fine to merge.

ThislsAUsername:
> > Lombok
> 
> I know you have mentioned that before, but it's kinda in the "I don't know what this is or what it does, and therefore don't see any reason we would need it" category of things.

It's boilerplate automation, where boilerplate is code in certain generic patterns that makes your project better/work but is very tedious to write.
Even better than just "automatically writes it", you don't have to look at the code for it at all. It gets compiled into some cache and "just works".
It autogenerates code for getters and setters, `equals()`/`getHashCode()`, and also more advanced stuff like the Builder pattern (based on annotations it adds).

As for "how do I set it up", it's an IDE plugin installer + library in the same file, via hacker magic. You just put that file with your dependencies, run it to install the IDE plugin, and you're off to the races.

ThatIsAUsername:
I don't have strong feelings on Lombok, but if you feel inspired to use it, I'm down for learning something new.

PR #417, opened 2024-01-22
Remove running dir dependence

ThislsAUsername:
I'm tired of people sending me screenshots of black squares.

### DIFF
--- a/src/Engine/Driver.java
+++ b/src/Engine/Driver.java
@@ -8,6 +8,7 @@ import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
+import java.io.File;
 import java.util.Stack;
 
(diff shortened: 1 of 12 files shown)


ThatIsAUsername:
Cool, looks good. Although: People are just running the jar from some other directory like absolute lunatics? Madmen.

Also: I feel like it would be possible to just bundle the resources into the jar as another way to make the problem go away. Maybe an issue for another day (plus it might make modding less convenient, so maybe we don't do that anyway?).

ThislsAUsername:
> Although: People are just running the jar from some other directory like absolute lunatics? Madmen.

That is (for some reason) the default behavior of a bunch of GUI runners where you just double-click the jar.
> Also: I feel like it would be possible to just bundle the resources into the jar as another way to make the problem go away. Maybe an issue for another day (plus it might make modding less convenient, so maybe we don't do that anyway?).

Yes. I don't want bundling.

PR #418, opened 2024-01-25
Persistent Tags

ThislsAUsername:
I forgot I hadn't done this yet.
It's just as silly as I'd hoped.

### DIFF
--- a/src/CommandingOfficers/AW2/AW2Commander.java
+++ b/src/CommandingOfficers/AW2/AW2Commander.java
@@ -29,8 +29,8 @@ public abstract class AW2Commander extends AW2And3CommanderBase
     if( minion == null || enemy == null )
       return 0;
 
-    int guiHPLoss  = minion.getHealthDamage() / 10;
-    int guiHPDealt =  enemy.getHealthDamage() / 10;
+    int guiHPLoss  = minion.getHPDamage();
+    int guiHPDealt =  enemy.getHPDamage();
 
     int power = 0; // value in funds of the charge we're getting
 
--- a/src/CommandingOfficers/AW2/AW2Commander.java
+++ b/src/CommandingOfficers/AW2/AW2Commander.java
@@ -24,8 +24,6 @@ public abstract class AW2Commander extends AW2And3CommanderBase
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
-    if( null != getActiveAbility() )
-      return 0;
     if( minion == null || enemy == null )
       return 0;
 
@@ -34,9 +32,7 @@ public abstract class AW2Commander extends AW2And3CommanderBase
 
     int power = 0; // value in funds of the charge we're getting
 
-    // Add up the funds value of the damage done to both participants.
     power += guiHPLoss * minion.unit.getCost() / 10;
-    // The damage we deal is worth half as much as the damage we take, to help powers be a comeback mechanic.
     power += guiHPDealt * enemy.unit.getCost() / 10 / 2;
 
     return power;
--- a/src/CommandingOfficers/AW2/AW2Commander.java
+++ b/src/CommandingOfficers/AW2/AW2Commander.java
@@ -21,12 +21,10 @@ public abstract class AW2Commander extends AW2And3CommanderBase
     super(info, rules);
   }
 
+  // Charge based on funds damage taken + 1/2 dealt
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
-    if( minion == null || enemy == null )
-      return 0;
-
     int guiHPLoss  = minion.getHPDamage();
     int guiHPDealt =  enemy.getHPDamage();
 
--- a/src/CommandingOfficers/AW4/RuinedCommander.java
+++ b/src/CommandingOfficers/AW4/RuinedCommander.java
@@ -168,7 +168,7 @@ public abstract class RuinedCommander extends DeployableCommander
     if( !isInZone(chargeSource) )
       return 0;
 
-    int guiHPDealt = enemy.getHealthDamage() / 10;
+    int guiHPDealt =  enemy.getHPDamage();
 
     int power = 0; // value in funds of the charge we're getting
 
--- a/src/CommandingOfficers/AW4/RuinedCommander.java
+++ b/src/CommandingOfficers/AW4/RuinedCommander.java
@@ -153,13 +153,13 @@ public abstract class RuinedCommander extends DeployableCommander
         ++zoneRadius;
     }
   }
+
+  // Charge based on HP damage dealt in zone
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
     if( zoneIsGlobal )
       return 0; // No charging while the zone is global, that's a little OP
-    if( minion == null || enemy == null )
-      return 0;
 
     // isCounter tells us who the attacker is, so we can figure out which one we care about being in the zone
     UnitContext chargeSource = minion.after;
--- a/src/CommandingOfficers/AWBW/AWBWCommander.java
+++ b/src/CommandingOfficers/AWBW/AWBWCommander.java
@@ -36,10 +36,18 @@ public abstract class AWBWCommander extends Commander
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
-    if( null != getActiveAbility() )
+    if( minion == null || enemy == null )
       return 0;
-    // TODO: Copy/paste super's code and delete the HP-based bit once the parent PR settles down.
-    return super.calculateCombatCharge(minion, enemy, isCounter);
+
+    int guiHPLoss  = minion.getHPDamage();
+    int guiHPDealt =  enemy.getHPDamage();
+
+    int power = 0; // value in funds of the charge we're getting
+
+    power += guiHPLoss * minion.unit.getCost() / 10;
+    power += guiHPDealt * enemy.unit.getCost() / 10 / 2;
+
+    return power;
   }
   @Override
   public int calculateMassDamageCharge(Unit minion, int lostHP)
--- a/src/CommandingOfficers/AWBW/AWBWCommander.java
+++ b/src/CommandingOfficers/AWBW/AWBWCommander.java
@@ -33,12 +33,10 @@ public abstract class AWBWCommander extends Commander
     roundUpRepairs = false;
   }
 
+  // Charge based on funds damage taken + 1/2 dealt
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
-    if( minion == null || enemy == null )
-      return 0;
-
     int guiHPLoss  = minion.getHPDamage();
     int guiHPDealt =  enemy.getHPDamage();
 
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -253,14 +253,10 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
     return Color.RED;
   }
 
-  protected boolean taggingDeniesCharge()
-  {
-    return gameRules.tagMode.supportsMultiCmdrSelect
-        && getAbilityText().length() > 0;
-  }
-
   public void awbwTagsCharge(Function<Commander, Integer> chargeEvaluator)
   {
+    if( !cos[0].canAcceptCharge() )
+      return;
     final int primaryCharge = chargeEvaluator.apply(cos[0]);
     cos[0].modifyAbilityPower(primaryCharge);
     final int tagDivisor = 2;
@@ -272,6 +268,8 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
   }
   public void persistentTagsCharge(Function<Commander, Integer> chargeEvaluator)
   {
+    if( !cos[0].canAcceptCharge() )
+      return;
     // As persistent tags is meant to be mostly a sidegrade, give half charge to the primary, and split the rest among the rest.
     final int primaryCharge = chargeEvaluator.apply(cos[0]);
     if( cos.length == 1 )
@@ -294,8 +292,6 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
   @Override
   public GameEventQueue receiveBattleEvent(final BattleSummary summary)
   {
-    if( taggingDeniesCharge() )
-      return null;
     // The lambdas demand the args be effectively final, so we're duplicating checks today
     boolean amAttacking = this == summary.attacker.CO.army;
     if( !amAttacking && this != summary.defender.CO.army )
@@ -327,10 +323,11 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
         break;
       case Team_Merge:
       case OFF:
-      {
-        final int ownerCharge = chargeEvaluator.apply(minion.CO);
-        minion.CO.modifyAbilityPower(ownerCharge);
-      }
+        if( minion.CO.canAcceptCharge() )
+        {
+          final int ownerCharge = chargeEvaluator.apply(minion.CO);
+          minion.CO.modifyAbilityPower(ownerCharge);
+        }
         break;
     }
     return null;
@@ -344,8 +341,6 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
   {
     if( attacker != null && this == attacker.army )
       return null; // Punching yourself shouldn't make you angry
-    if( taggingDeniesCharge() )
-      return null;
 
     for( Entry<Unit, Integer> damageEntry : lostHealth.entrySet() )
     {
@@ -363,10 +358,11 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
           break;
         case Team_Merge:
         case OFF:
-        {
-          final int ownerCharge = chargeEvaluator.apply(minion.CO);
-          minion.CO.modifyAbilityPower(ownerCharge);
-        }
+          if( minion.CO.canAcceptCharge() )
+          {
+            final int ownerCharge = chargeEvaluator.apply(minion.CO);
+            minion.CO.modifyAbilityPower(ownerCharge);
+          }
           break;
       }
     }
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -17,7 +17,6 @@ import AI.AILibrary;
 import AI.AIMaker;
 import CommandingOfficers.Commander;
 import CommandingOfficers.CommanderAbility;
-import Engine.GameScenario.TagMode;
 import Engine.Combat.BattleSummary;
 import Engine.GameEvents.GameEventListener;
 import Engine.GameEvents.GameEventQueue;
@@ -253,9 +252,9 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
     return Color.RED;
   }
 
-  protected boolean awbwDeniesCharge()
+  protected boolean taggingDeniesCharge()
   {
-    return TagMode.AWBW == gameRules.tagMode
+    return gameRules.tagMode.supportsMultiCmdrSelect
         && getAbilityText().length() > 0;
   }
 
@@ -265,7 +264,7 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
   @Override
   public GameEventQueue receiveBattleEvent(final BattleSummary summary)
   {
-    if( awbwDeniesCharge() )
+    if( taggingDeniesCharge() )
       return null;
 
     // We only care who the units belong to, not who picked the fight.
@@ -301,6 +300,24 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
           }
         }
           break;
+        case Persistent:
+        {
+          // As persistent tags is meant to be mostly a sidegrade, give half charge to the primary, and split the rest among the rest.
+          final int primaryCharge = cos[0].calculateCombatCharge(minion, enemy, isCounter);
+          if( cos.length == 1 )
+            cos[0].modifyAbilityPower(primaryCharge);
+          else
+          {
+            cos[0].modifyAbilityPower(primaryCharge / 2);
+            final double tagMultiplier = 0.5 / (cos.length - 1);
+            for( int i = 1; i < cos.length; ++i )
+            {
+              final int tagCharge = cos[i].calculateCombatCharge(minion, enemy, isCounter);
+              cos[i].modifyAbilityPower((int) (tagCharge * tagMultiplier));
+            }
+          }
+        }
+          break;
         case Team_Merge:
         case OFF:
         {
@@ -321,7 +338,7 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
   {
     if( attacker != null && this == attacker.army )
       return null; // Punching yourself shouldn't make you angry
-    if( awbwDeniesCharge() )
+    if( taggingDeniesCharge() )
       return null;
 
     for( Entry<Unit, Integer> damageEntry : lostHealth.entrySet() )
@@ -340,6 +357,24 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
             final int tagCharge = cos[i].calculateMassDamageCharge(minion, damageEntry.getValue());
             final int tagDivisor = 2;
             cos[i].modifyAbilityPower(tagCharge / tagDivisor);
+          }
+        }
+          break;
+        case Persistent:
+        {
+          // As persistent tags is meant to be mostly a sidegrade, give half charge to the primary, and split the rest among the rest.
+          final int primaryCharge = cos[0].calculateMassDamageCharge(minion, damageEntry.getValue());
+          if( cos.length == 1 )
+            cos[0].modifyAbilityPower(primaryCharge);
+          else
+          {
+            cos[0].modifyAbilityPower(primaryCharge / 2);
+            final double tagMultiplier = 0.5 / (cos.length - 1);
+            for( int i = 1; i < cos.length; ++i )
+            {
+              final int tagCharge = cos[i].calculateMassDamageCharge(minion, damageEntry.getValue());
+              cos[i].modifyAbilityPower((int) (tagCharge / tagMultiplier));
+            }
           }
         }
           break;
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -11,6 +11,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.UUID;
+import java.util.function.Function;
 
 import AI.AIController;
 import AI.AILibrary;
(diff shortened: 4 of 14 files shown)


src/Engine/Army.java
```diff
@@ -253,9 +252,9 @@ public Color getMarkingColor(Unit unit)
     return Color.RED;
   }
 
-  protected boolean awbwDeniesCharge()
+  protected boolean taggingDeniesCharge()
   {
-    return TagMode.AWBW == gameRules.tagMode
+    return gameRules.tagMode.supportsMultiCmdrSelect
         && getAbilityText().length() > 0;
   }
```
ThislsAUsername, line 259: I doublechecked AWDS, and realized I was mistaken about the secondary CO getting charge being gained during the primary CO's COP.
This does have the weird effect that COs who can normally charge during a power can't if tagging is on, but I dunno how I want it to work.

ThatIsAUsername: I assume this discrepancy (COs who can/can't charge during a power) only exists across games, and not within them?

ThislsAUsername: There are literally no COs in a cart game that can charge during a power.
It's a common suggested D2D for Adder/custom CO mechanic, but hasn't been in an official game.

ThatIsAUsername:
I tried running this with AI to remind myself of what it is (We should really have a tooltip to explain the tag modes), but it keeps getting a NPE when trying to build things. Muriel and Wally both hit the problem (Wally hit as soon as I tried to launch the game), but JakeMan seems immune so far. Also the AIs don't seem to swap command, if that's supposed to be a thing.
Wally Stack (though I guess you can probably produce it just as easily):
```java
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at AI.CommanderProductionInfo.getNumFacilitiesFor(CommanderProductionInfo.java:155)
	at AI.WallyAI.queueUnitProductionActions(WallyAI.java:1080)
	at AI.WallyAI.access$9(WallyAI.java:969)
	at AI.WallyAI$BuildStuff.getNextAction(WallyAI.java:545)
	at AI.ModularAI.getNextAction(ModularAI.java:99)
	at Engine.Army.getNextAIAction(Army.java:416)
	at Engine.MapController.animationEnded(MapController.java:596)
	at Engine.MapController.startNextTurn(MapController.java:672)
	at Engine.MapController.<init>(MapController.java:68)
	at UI.PlayerSetupController.handlePlayerSetupInput(PlayerSetupController.java:149)
	at UI.PlayerSetupController.handleInput(PlayerSetupController.java:95)
	at UI.GameOptionSetupController.handleInput(GameOptionSetupController.java:91)
	at Engine.Driver.keyPressed(Driver.java:125)
	at java.awt.Component.processKeyEvent(Component.java:6497)
	at java.awt.Component.processEvent(Component.java:6316)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Window.processEvent(Window.java:2025)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.KeyboardFocusManager.redispatchEvent(KeyboardFocusManager.java:1954)
	at java.awt.DefaultKeyboardFocusManager.dispatchKeyEvent(DefaultKeyboardFocusManager.java:834)
	at java.awt.DefaultKeyboardFocusManager.preDispatchKeyEvent(DefaultKeyboardFocusManager.java:1102)
	at java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(DefaultKeyboardFocusManager.java:973)
	at java.awt.DefaultKeyboardFocusManager.dispatchEvent(DefaultKeyboardFocusManager.java:799)
	at java.awt.Component.dispatchEventImpl(Component.java:4760)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:760)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:84)
	at java.awt.EventQueue$4.run(EventQueue.java:733)
	at java.awt.EventQueue$4.run(EventQueue.java:731)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:730)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

ThislsAUsername:
> (We should really have a tooltip to explain the tag modes)

Yeah, or at least that readme I've mentioned.
GUI design is hard.
> but it keeps getting a NPE when trying to build things. Muriel and Wally both hit the problem (Wally hit as soon as I tried to launch the game), but JakeMan seems immune so far.

Huh. Lovely. I'm not super worried about that re: Wally since I have a bazillion pending changes on him, but crashing is bad.
> Also the AIs don't seem to swap command, if that's supposed to be a thing.

They don't have swapping logic yet, no. I have notional plans for AWBW style, but making something at all intelligent for persistent tags in the general case kind of seems like a losing battle.

ThislsAUsername:
Also, I notice a `NOTE: This assumes there is no overlap in what different property types can produce` on `removeBuildLocation()`, which is actually not true already because Strong exists.

src/Engine/Army.java
ThatIsAUsername: Wait, then why mention "COs who can normally charge during a power" when there is no such thing?

src/Engine/Army.java
```diff
@@ -301,6 +300,24 @@ public GameEventQueue receiveBattleEvent(final BattleSummary summary)
           }
         }
           break;
+        case Persistent:
+        {
+          // As persistent tags is meant to be mostly a sidegrade, give half charge to the primary, and split the rest among the rest.
```
ThatIsAUsername, line 305: Is this what the games do? It seems like this would basically be a downgrade, since having two (or more) COs just means you get to use powers less frequently. A single CO would be able to use his much more often.

src/Engine/Army.java
```diff
@@ -343,6 +360,24 @@ public GameEventQueue receiveMassDamageEvent(Commander attacker, Map<Unit, Integ
...
+          if( cos.length == 1 )
+            cos[0].modifyAbilityPower(primaryCharge);
+          else
+          {
+            cos[0].modifyAbilityPower(primaryCharge / 2);
+            final double tagMultiplier = 0.5 / (cos.length - 1);
+            for( int i = 1; i < cos.length; ++i )
+            {
+              final int tagCharge = cos[i].calculateMassDamageCharge(minion, damageEntry.getValue());
+              cos[i].modifyAbilityPower((int) (tagCharge / tagMultiplier));
```
ThatIsAUsername, line 376: Should this be `tagCharge * tagMultiplier` to match line 316? Also, can we put the logic in one place so we don't have this problem?

src/Engine/GameScenario.java
```diff
@@ -37,7 +37,7 @@ public GameEventQueue initTurn(GameMap map)
 
   public enum TagMode
   {
-    OFF(false), AWBW(true), Team_Merge(false); // Cartridge, Persistent,
+    OFF(false), AWBW(true), Persistent(true), Team_Merge(false); // Cartridge
```
ThatIsAUsername, line 40: I know we don't have a UI explanation of what the tag modes are, but while we fix that, can we at least get a high-level explanation here in a comment?

src/Units/Unit.java
```diff
@@ -100,7 +100,11 @@ public boolean capture(MapMaster map)
     int captureThreshold = map.getEnvironment(x, y).terrainType.getCaptureThreshold();
     if( captureProgress >= captureThreshold )
     {
-      map.setOwner(CO, x, y);
+      Commander newOwner = CO;
+      // Capture territory for your primary CO in persistent tags
+      if( CO.gameRules.tagMode.supportsMultiCmdrSelect )
+        newOwner = CO.army.cos[0];
+      map.setOwner(newOwner, x, y);
```
ThatIsAUsername, line 107: Why is unit setting the property owner? Isn't that handled by `CaptureEvent`?

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
```diff
@@ -116,8 +117,12 @@ public GameEventQueue getEvents(MapMaster map)
             {
               // Someone is losing their big, comfy chair.
               ArmyDefeatEvent defeat = new ArmyDefeatEvent(captureLocation.getOwner().army);
+              Commander newOwner = actor.CO;
+              // Capture territory for your primary CO in persistent tags
+              if( newOwner.gameRules.tagMode.supportsMultiCmdrSelect )
+                newOwner = newOwner.army.cos[0];
```
ThatIsAUsername, line 123: Is the first CO in the array always the currently-active one (as in, we are shuffling them around in the array when tags happen)?
Even if we are doing that, it would be cleaner to have a `getActiveCO` or something rather than relying on outside classes directly accessing implicitly-meaningful array indices.

src/Engine/Army.java
ThislsAUsername: It's currently a thing (for our custom COs), though that could be changed.
But also I don't want to rule it out as a mechanic.

src/Engine/Army.java
ThislsAUsername: No games have persistent tags in them.
Cart/AWBW behavior is that the second CO gets half the primary's charge.

It *is* a downgrade for power timing... which is the intent. The intent is to balance persistent tags vs a single CO, though obviously the error bars here are pretty big.
In trade, you get a lot:
Flexibility to mix and match different D2Ds as you see fit.
The ability to stack different COs' D2Ds and powers.
The ability to hold more charge than a single CO can.
More powers to choose from.

Flexibility is power.

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThislsAUsername: Yes, for simplicity/consistency.
I suppose that's a fair point, though.

src/Engine/Army.java
ThislsAUsername: Good catch.

src/Units/Unit.java
ThislsAUsername: I'm pretty sure it just calls `unit.capture()`

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThislsAUsername: When I went to actually write that code, I recalled why I didn't do it.
`getActiveCO() -> cos[0]` is useful for AWBW/persistent tags and harmless for single-CO... but team-merge doesn't have a "primary CO" concept.

Also, `getActiveCO()` is a lot more characters than `cos[0]`, which bothers me more than it should.

src/Engine/Army.java
ThatIsAUsername: That's ... pretty niche, but _okay_

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThatIsAUsername: > team-merge doesn't have a "primary CO" concept.

But then you'd still be using `cos[0]` anyway, yeah? I don't see how this change would have a negative impact (or change anything) even in that case.

> which bothers me more than it should

As a fan of verbose naming and clear intents, I agree that this should not bother you.
We all know that the real problem is not about variable length, but whether adjacent variables are justified properly. Fortunately, [there is help](https://github.com/timvieira/justified-variables).

src/Units/Unit.java
ThatIsAUsername: So it does. Though `CaptureEvent` also has this interesting snippet in its constructor:
```java
captureAmount = unit.getHP(); // TODO: Apply CO buffs.
```

src/Engine/Army.java
```diff
@@ -265,50 +292,43 @@ protected boolean awbwDeniesCharge()
   @Override
   public GameEventQueue receiveBattleEvent(final BattleSummary summary)
   {
-    if( awbwDeniesCharge() )
-      return null;
+    // The lambdas demand the args be effectively final, so we're duplicating checks today
```
ThatIsAUsername, line 295: And why, oh why was your solution to use a lambda in the first place instead of simply allowing the `*Charge` function to accept `(Commander co, UnitDelta minion, UnitDelta enemy, boolean isCounter)` as parameters and calling `co.calculateCombatCharge(minion, enemy, isCounter)` inside the functions?

It adds indirection and complexity, all for the sake of calling a function that is readily available already without wrapping it and passing it around.

ThatIsAUsername: Ah, okay. Should have read farther before complaining. Makes more sense since the underlying calculation can change between battles and mass damage.
So while I still don't love lambdas as much as you, I get why you are using one here.

src/Engine/Army.java
```diff
@@ -265,50 +292,43 @@ protected boolean awbwDeniesCharge()
...
-        case OFF:
+      case AWBW:
+        awbwTagsCharge(chargeEvaluator);
+        break;
+      case Persistent:
+        persistentTagsCharge(chargeEvaluator);
+        break;
+      case Team_Merge:
+      case OFF:
+        if( minion.CO.canAcceptCharge() )
```
ThatIsAUsername, line 326: Should we just do the `canAcceptCharge()` check inside `modifyAbilityPower()` to avoid needing to remember to check it elsewhere?

src/CommandingOfficers/AW4/RuinedCommander.java
```diff
@@ -168,7 +168,7 @@ public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCo
     if( !isInZone(chargeSource) )
       return 0;
 
-    int guiHPDealt = enemy.getHealthDamage() / 10;
+    int guiHPDealt =  enemy.getHPDamage();
```
ThatIsAUsername, line 171: Someone doesn't have auto-formatting enabled. Me. I don't. But also you don't.

src/Engine/StateTrackers/DamageDealtToIncomeConverter.java
```diff
@@ -57,10 +56,10 @@ private void profitize(UnitDelta attacker, UnitDelta defender)
 
   private static int calculateProfit(UnitDelta delta, double myIncomeRatio)
   {
-    int healthLoss = delta.getHealthDamage();
+    int healthLoss = delta.getHPDamage();
     int unitCost = delta.unit.getCost();
     // Do the necessary math, then round to the nearest int.
-    int income = (int) (healthLoss * (unitCost / UnitModel.MAXIMUM_HEALTH) * myIncomeRatio + 0.5);
+    int income = (int) (healthLoss * (unitCost / 10) * myIncomeRatio + 0.5);
```
ThatIsAUsername, line 62: Why change this back to `10` instead of `UnitModel.MAXIMUM_HEALTH`?.

ThatIsAUsername: ... or an otherwise more correct non-magic named quantity?

src/CommandingOfficers/AW2/AW2Commander.java
```diff
@@ -24,19 +24,15 @@ public AW2Commander(CommanderInfo info, GameScenario.GameRules rules)
...
-    int guiHPDealt =  enemy.getHealthDamage() / 10;
+    int guiHPLoss  = minion.getHPDamage();
+    int guiHPDealt =  enemy.getHPDamage();
 
     int power = 0; // value in funds of the charge we're getting
 
-    // Add up the funds value of the damage done to both participants.
     power += guiHPLoss * minion.unit.getCost() / 10;
-    // The damage we deal is worth half as much as the damage we take, to help powers be a comeback mechanic.
     power += guiHPDealt * enemy.unit.getCost() / 10 / 2;
```
ThatIsAUsername, line 36: Here and elsewhere you took away the comments but left the code. Why?

src/CommandingOfficers/AWBW/AWBWCommander.java
```diff
@@ -36,10 +36,18 @@ public AWBWCommander(CommanderInfo info, GameScenario.GameRules rules)
   @Override
   public int calculateCombatCharge(UnitDelta minion, UnitDelta enemy, boolean isCounter)
   {
-    if( null != getActiveAbility() )
+    if( minion == null || enemy == null )
```
ThatIsAUsername, line 39: Do we often try to calculate combat charge for combat between nonexistent combatants?

src/Engine/GameScenario.java
```diff
@@ -37,12 +37,18 @@ public GameEventQueue initTurn(GameMap map)
 
   public enum TagMode
   {
-    OFF(false), AWBW(true), Team_Merge(false); // Cartridge, Persistent,
+    OFF       (false, "Dunno what you expected"),
+    AWBW      (true , "Multi-CO; only one active at a time"),
+    Persistent(true , "Multi-CO; only one builds or COPs at a time"),
+    Team_Merge(false, "All teams become a single player");
+    // Cartridge
```
ThatIsAUsername, line 44: Cartridge?

src/Engine/GameScenario.java
```diff
@@ -37,12 +37,18 @@ public GameEventQueue initTurn(GameMap map)
 
   public enum TagMode
   {
-    OFF(false), AWBW(true), Team_Merge(false); // Cartridge, Persistent,
+    OFF       (false, "Dunno what you expected"),
+    AWBW      (true , "Multi-CO; only one active at a time"),
+    Persistent(true , "Multi-CO; only one builds or COPs at a time"),
+    Team_Merge(false, "All teams become a single player");
```
ThatIsAUsername, line 43: All teams become a single player? You mean all COs in the same army, right? Does that mean they all get an equal share of power, and all powers are available for use, and all passives are simultaneously active? That could be a bit cray cray.

Also an important element of Persistent (at least if my reading of the code is correct) is that the units keep the attributes of the CO that built them  when control shifts; the CO in command just "owns" whatever is built during his watch (but does this mean other units don't benefit from the active CO's passive?).

src/Units/Unit.java
ThislsAUsername: To be fair, Unit probably shouldn't own this game logic.
That said, `CaptureEvent()` looks like this on my end:
```java
  public CaptureEvent( Unit u, MapLocation loc )
  {
    unit = u;
    location = loc;
    XYCoord unitXY = new XYCoord(u.x, u.y);
    if( null != location && location.isCaptureable() && unit.CO.isEnemy(location.getOwner()) )
    {
      priorCaptureAmount = (unitXY.equals(location.getCoordinates()) ? unit.getCaptureProgress() : 0);
      UnitContext uc = new UnitContext(unit);
      captureAmount = uc.calculateCapturePower();
    }
    else
    {
      System.out.println("WARNING! Attempting to capture an invalid location!");
      priorCaptureAmount = 0;
      captureAmount = 0;
    }
  }
```
You may be looking at old code.

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThislsAUsername: > But then you'd still be using `cos[0]` anyway, yeah?

Sure.
> I don't see how this change would have a negative impact (or change anything) even in that case.

Names have power, and "active/primary CO" gives a false sense of that concept always being concretely true.
> As a fan of verbose naming and clear intents, I agree that this should not bother you.

"Active", "primary", and "first" are all synonyms here anyway (for the cases you care about), so why have a problem with `cos[0]`?
> We all know that the real problem is not about variable length, but whether adjacent variables are justified properly. Fortunately, [there is help](https://github.com/timvieira/justified-variables).

Amazing. I'm not sure I wanted to know this exists.

src/Engine/Army.java
ThislsAUsername: I'm gonna say "no", since I think non-standard providers of charge should still work while a power is active.
And in any case, it's not called `maybeModifyAbilityPower()`

src/CommandingOfficers/AW4/RuinedCommander.java
ThislsAUsername: It's formatted to align with all the other statements with the same meaning across other CO classes.

...and yeah, I've given up on compulsive auto-format because I'm a vertical alignment gremlin.

src/Engine/StateTrackers/DamageDealtToIncomeConverter.java
ThislsAUsername: `MAXIMUM_HEALTH` is 100.
I guess we could *also* have `MAXIMUM_HP`, but that doesn't exactly align with all the usecases we have for "10".

I am roughly equally annoyed by long variable names and magic numbers, and naming all the different reasons we'd want to use "10" is hard, so I default to lazy.

src/CommandingOfficers/AW2/AW2Commander.java
ThislsAUsername: The design intent isn't ours.

src/CommandingOfficers/AWBW/AWBWCommander.java
ThislsAUsername: I don't think so. That null check has been there a long time, I think.

src/Engine/GameScenario.java
ThislsAUsername: In a pear tree.

src/Engine/GameScenario.java
ThislsAUsername: > All teams become a single player? You mean all COs in the same army, right?

No. You'll note that you can't pick multiple COs in Team Merge.
If I'm on Final Front with Grit, Andy, and Kanbei, and I pick Team Merge, I have *one* turn for all 3 COs.
It also does share money and the "powers benefit everyone" thing, but Kanbei has yellow buildings and builds Kanbei dudes and there's no swapping.
> Does that mean they all get an equal share of power

No. Same charging rules as solo.
> and all powers are available for use

Yes. Simultaneously, if you want.
> and all passives are simultaneously active?

On their individual dudes, yes.
As a side note, "stack D2Ds" is how Commander Wars does tags, and it almost makes less sense than cart tags IMO.
> That could be a bit cray cray.

That's usually the plan.
> Also an important element of Persistent (at least if my reading of the code is correct) is that the units keep the attributes of the CO that built them when control shifts

Yes. That is the primary reason Persistent exists.
I suppose I could have clarified better, probably by specifying more on the AWBW entry.
> the CO in command just "owns" whatever is built during his watch (but does this mean other units don't benefit from the active CO's passive?).

Yes. Colin can save money so Kanbei can buy units, but he can't buy units and make them Kanbei's.

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThatIsAUsername: I might debate with you about those all being first (even "primary" and "first" are commonly used in different contexts), but it's more just that other code generally shouldn't need to know internal data structures of `Commander` to interact with it. It also just makes the intent more clear.

src/Units/Unit.java
ThatIsAUsername: So I am. Not sure how that happened.

src/Engine/Army.java
ThatIsAUsername: Ugh. Gross.

src/CommandingOfficers/AW2/AW2Commander.java
ThatIsAUsername: Might have to unpack that for me. The comments explain what the code is doing, right? Why remove that?

src/Engine/GameScenario.java
ThatIsAUsername: These could still be clearer I think, but given that this isn't user facing at this point, I won't sweat it too much.

src/Engine/UnitActionLifecycles/CaptureLifecycle.java
ThislsAUsername: Most of the codebase is interested in the internal data structures of the stuff that makes up our game state.
`GameInstance` to a lesser degree, but `Army`/map/terrain/units are all very fair game.
(And it's `Army.cos[0]`, `Commander` doesn't have a list of COs.)
> It also just makes the intent more clear.

I disagree, as noted above.

I think encapsulation for `Commander` makes a fair amount more sense (unless it's e.g. an AI that cares to know stuff like "am I a DoR CO?"), and to a lesser extent `UnitModel`.

src/CommandingOfficers/AW2/AW2Commander.java
ThislsAUsername: The first comment isn't any clearer than the code itself. It's clutter.

The second comment's first clause mentions "half", which is cool I guess.
The second comment's conclusion is about design intent, and the design intent isn't really relevant since we're just matching cart behavior there.

I guess I could put in "Charge based on funds damage taken + half dealt (to match the cart)", but:
That's still just kind of duplicating the code.
The whole point of the class is to match cart logic, so mentioning that is redundant.

ThatIsAUsername: I'd probably agree that it's irrelevant to mention that we are mimicking cart behavior (though not because that's the purpose of the class, since the class doesn't state that elsewhere either). Whether the design intent is ours is also irrelevant; comments should communicate the intent regardless of whose intent it initially was.

I do agree that pairing lines of code with equivalent comments is a bit silly, but I like having a comment to explain a block of code in plain English since 1) is easier to parse, 2) knowing intent helps to parse the code too, and 3) it also provides confirmation of correctness if they match, and an immediate red flag if they don't. Without the intent you just have to trust the code is right.

But this code also isn't that complex; it's pretty clear as is, aside from the magic numbers (though the part about comeback mechanics in the game's design doesn't come through, but hey). I'd still leave a comment (and purge the magic, ofc), but up to you I s'pose.

ThislsAUsername: > I'd probably agree that it's irrelevant to mention that we are mimicking cart behavior (though not because that's the purpose of the class, since the class doesn't state that elsewhere either).

It's in the name: `AW2Commander`
It'd feel pretty redundant to slap on a "To implement AW2 COs" docstring on there.
> Whether the design intent is ours is also irrelevant; comments should communicate the intent regardless of whose intent it initially was.

Suuuure, but we don't actually _know_ what the original design intent was.
We can make assumptions, but that's what they'd be.
The intent of *this* code is to... do whatever the cart code does.
> I do agree that pairing lines of code with equivalent comments is a bit silly, but I like having a comment to explain a block of code in plain English since 1) is easier to parse, 2) knowing intent helps to parse the code too, and 3) it also provides confirmation of correctness if they match, and an immediate red flag if they don't. Without the intent you just have to trust the code is right.
> 
> But this code also isn't that complex; it's pretty clear as is, aside from the magic numbers (though the part about comeback mechanics in the game's design doesn't come through, but hey). I'd still leave a comment (and purge the magic, ofc), but up to you I s'pose.

Fair.
I'm not sure killing the magic in this instance really helps much, but "document the intended behavior" seems pretty defensible.

ThatIsAUsername:
Looks groovy.

PR #420, opened 2024-01-25
Fog Fixes

ThislsAUsername:
I fixed the vision-reset thing on the playtest branch and made a note to fix it here a while back.
I found my note, and also found some other fog badness in testing it.

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -132,6 +132,8 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
     {
       events.add(new CommanderAbilityRevertEvent(myActiveAbility));
       events.addAll(myActiveAbility.getRevertEvents(map));
+      // Deactivate now because we need to get rid of vision buffs before fog is reset
+      myActiveAbility.deactivate(map);
       myActiveAbility = null;
     }
 
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -59,7 +59,7 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
   public boolean immuneToSand   = false; // Sandstorm/Sirocco
 
   public ArrayList<CommanderAbility> myAbilities = null;
-  protected CommanderAbility myActiveAbility = null;
+  public CommanderAbility myActiveAbility = null;
 
   public Commander(CommanderInfo info, GameScenario.GameRules rules)
   {
@@ -103,28 +103,7 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
     return ca;
   }
 
-  public void endTurn()
-  {
-  }
-
-  /**
-   * Deactivate any active ability, and init all units
-   */
-  public GameEventQueue initTurn(MapMaster map)
-  {
-    GameEventQueue events = new GameEventQueue();
-
-    events.addAll(revertActiveAbility(map));
-
-    for( Unit u : units )
-    {
-      events.addAll(u.initTurn(map));
-    }
-
-    return events;
-  }
-
-  public GameEventQueue revertActiveAbility(MapMaster map)
+  public final GameEventQueue getAbilityRevertEvents(MapMaster map)
   {
     GameEventQueue events = new GameEventQueue();
 
@@ -132,11 +111,26 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
     {
       events.add(new CommanderAbilityRevertEvent(myActiveAbility));
       events.addAll(myActiveAbility.getRevertEvents(map));
-      myActiveAbility = null;
     }
 
     return events;
   }
+
+  @Override
+  public final GameEventQueue receiveTurnInitEvent(MapMaster map, Army army, int turn)
+  {
+    if( army != this.army)
+      return null;
+
+    GameEventQueue events = new GameEventQueue();
+
+    onTurnInit(map, events);
+    for( Unit u : units )
+      events.addAll(u.model.getTurnInitEvents(u, map));
+
+    return events;
+  }
+  protected void onTurnInit(MapMaster map, GameEventQueue events) {}
 
   /**
    * @return whether these COs would like to kill each other
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -132,8 +132,6 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
     {
       events.add(new CommanderAbilityRevertEvent(myActiveAbility));
       events.addAll(myActiveAbility.getRevertEvents(map));
-      // Deactivate now because we need to get rid of vision buffs before fog is reset
-      myActiveAbility.deactivate(map);
       myActiveAbility = null;
     }
 
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -89,6 +89,7 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
     {
       events.addAll(co.initTurn(map));
     }
+    myView.resetFog();
 
     if( null != aiController )
     {
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -89,7 +89,6 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
     {
       events.addAll(co.initTurn(map));
     }
-    myView.resetFog();
 
     if( null != aiController )
     {
--- a/src/Engine/Army.java
+++ b/src/Engine/Army.java
@@ -76,35 +76,32 @@ public class Army implements GameEventListener, Serializable, UnitModList, UnitM
       co.deInitForGame(game);
   }
 
-  /**
-   * Collect income and handle any COModifiers.
-   * @param map
-   */
-  public GameEventQueue initTurn(MapMaster map)
+  public GameEventQueue getAbilityRevertEvents(MapMaster map)
   {
     GameEventQueue events = new GameEventQueue();
 
-    money += getIncomePerTurn();
     for( Commander co : cos )
     {
-      events.addAll(co.initTurn(map));
-    }
-
-    if( null != aiController )
-    {
-      aiController.initTurn(myView);
+      events.addAll(co.getAbilityRevertEvents(map));
     }
 
     return events;
   }
 
   @Override
+  public GameEventQueue receiveTurnInitEvent(MapMaster map, Army army, int turn)
+  {
+    if( null != aiController )
+      aiController.initTurn(myView);
+
+    return null;
+  }
+
+  @Override
   public GameEventQueue receiveTurnEndEvent(Army army, int turn)
   {
-    if( aiController != null ) aiController.endTurn();
-    myView.resetFog();
-    for( Commander co : cos )
-      co.endTurn();
+    if( aiController != null )
+      aiController.endTurn();
 
     return null;
   }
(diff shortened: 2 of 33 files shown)


src/Engine/OverlayCache.java
```diff
@@ -59,6 +59,9 @@ public ArrayList<GameOverlay> getNormalOverlays(final OverlayMode mode, final Ma
       ArrayList<Unit> threats = new ArrayList<>();
 
       final Army viewer = drawableMap.viewer;
+      if( null == viewer )
+        return new ArrayList<>();
```
ThatIsAUsername, line 63: I guess I should ask the usual question about asking for a perspective for the null viewer. Do we do that?

src/CommandingOfficers/Commander.java
```diff
@@ -132,6 +132,8 @@ public GameEventQueue revertActiveAbility(MapMaster map)
     {
       events.add(new CommanderAbilityRevertEvent(myActiveAbility));
       events.addAll(myActiveAbility.getRevertEvents(map));
```
ThatIsAUsername, line 134: Do we not activate these events before the user gets control? Or is it just that the turn technically starts before the events fire, and by then the player has already gained the vision?

src/Engine/OverlayCache.java
ThislsAUsername: Yes. It's in the commit message: multi-human+AI fog games.
The map is fully hidden during AI turns there, I believe... though maybe it should account for "all the humans are allied".

src/CommandingOfficers/Commander.java
ThislsAUsername: The debuff needs to reset before the player's fog vision is reset for the turn (especially for the DoR case), which happens in `Army.initTurn()`.
I guess you could argue that resetting the vision/income/etc should also go in an event that ends up on the end of the "start turn" queue.

ThatIsAUsername: I could argue that. It might even be cleaner really (not a super fan of moving the contents of `CommanderAbilityRevertEvent` somewhere else), though I don't think of a vision reset as a "game event" per se.

Another possible approach: have `TurnInitEvent.performEvent()` reset vision for the active army and [move its creation to the end of the event queue in `GameInstance.turn()`](https://github.com/ThislsAUsername/DefendPeace/blob/master/src/Engine/GameInstance.java#L276)? I.e, from line 276 down to 291. That seems to fix it on master without modifying Commander, Army or the ability revert event.

ThislsAUsername: Hmmmhmmhmm.

I think what I might like to do is:
Get rid of Army/CO `initTurn()`
`Unit.initTurn()` loses its event stuff.
Add a new Army/CO hook for generating new-turn events that doesn't modify state.
If COs need to meddle with their internal state on turn init, `receiveTurnInitEvent()` exists.
`TurnInitEvent` calls `Unit.initTurn()` and does fog reset stuff interactions.

So then `GameInstance.turn()` looks like so:
Queue `ModifyFundsEvent` for income (just so it's first)
Do weather stuff
Queue CO/unit init events
Queue `TurnInitEvent`

This also neatly solves the fog problem by making sure the CO's power revert events go through before the `TurnInitEvent`.

There is one fly in this ointment, and her name is Ave.
Her `initTurn()` both modifies internal state (the snow map) and generates events (applying the snow map).
I guess I could just leave all that in the event-generation hook since it'll only get called once per turn anyway, but it makes me sad.
...or I can just have all her logic go into `receiveTurnInitEvent()`, since listeners are allowed to inject events. Huh.

Thoughts?

ThislsAUsername: I ended up doing something a bit different. Plans, enemy, all that.
I had to tweak the tests a bit because now units depend on the `GameEventListener`-returns-events behavior.
But like, it's bad that we weren't accounting for that before anyway, so.

I had to add a separate `ResetVisionEvent` because otherwise you'd see the buffed vision *during the `TurnInitEvent` animation*, since `perform()` happens at the end of that.
But I think I like this design a bit better than I had before.

src/Engine/Army.java
```diff
@@ -76,35 +76,32 @@ public void deInitForGame(GameInstance game)
...
+  public GameEventQueue getAbilityRevertEvents(MapMaster map)
   {
     GameEventQueue events = new GameEventQueue();
 
-    money += getIncomePerTurn();
     for( Commander co : cos )
     {
-      events.addAll(co.initTurn(map));
+      events.addAll(co.getAbilityRevertEvents(map));
     }
```
ThatIsAUsername, line 86: That's a bold move. I guess only one CO per army can have an ability active at once?

ThislsAUsername: No, but they should all deactivate at the same time.

ThatIsAUsername: Or does persistent tags allow multiple actives at once?

ThislsAUsername: Team Merge does.
Persistent does not.

src/Test/TestAIBehavior.java
```diff
@@ -293,13 +294,14 @@ private boolean testTankWadeThroughInfs(AIMaker ai)
...
     Unit inf = addUnit(testMap, testCo1, "Md Tank", new XYCoord(1, 0)); // Create tank last just to make the AI compensate.
+    turn(testGame);
 
     // The inf will want to cap the enemy HQ. Make sure the tanks get out of the way.
     boolean testPassed = true;
     int turnLimit = 4; // This should be enough time to cap the enemy HQ.
     for( int tt = 0; tt < turnLimit; ++tt )
     {
-      testCo1.army.initTurn(testMap);
+      turn(testGame, 2);
```
ThatIsAUsername, line 304: `testGame.armies.length`? I guess that might not track correctly if COs are defeated; I don't remember if we remove them from the list.

ThislsAUsername: We don't.
I dislike magic even less in test code, though.
Tests are supposed to be at least a little brittle.

src/Test/TestCombatMods.java
```diff
@@ -162,7 +162,7 @@ private boolean testRetribution()
     // Clean up
     testMap.removeUnit(infA);
     testMap.removeUnit(infB);
-    performEvents(testGame, venge.army.initTurn(testMap));
+    turn(testGame, 2);
```
ThatIsAUsername, line 165: Huh. I wonder why we even do another turn at the end of this test. Also why it would require two turn changes instead of just one.

ThislsAUsername: The power won't end if the turn doesn't end.
2 is to get back to the turn we started on.

src/Test/TestGameEvent.java
```diff
@@ -296,9 +305,12 @@ private boolean testUnitDieEvent()
...
     GameEventQueue geq = new GameEventQueue();
     Utils.enqueueDeathEvent(boat, geq);
-    testPassed &= validate(geq.size() == 3, "    Utils.enqueueDeathEvent built the wrong number of events!");
+    // One event for each of the 3 units, and also one for losing
+    testPassed &= validate(geq.size() == 4, "    Utils.enqueueDeathEvent built the wrong number of events!");
 
-    // No cleanup required.
+    new UnitDieEvent(man).performEvent(testMap);
+    new UnitDieEvent(car).performEvent(testMap);
+    new UnitDieEvent(boat).performEvent(testMap);
```
ThatIsAUsername, line 313: Why create new events instead of just executing the ones in the queue? Not that it matters, I guess.

ThislsAUsername: Didn't feel like having a CO defeat event in there.

src/Test/TestHealing.java
```diff
@@ -65,7 +65,7 @@ private boolean testHealing()
       prevMoney = testCo1.army.money; // Track money.
       prevHP = victim.getHealth(); // Track HP.
 
-      GameEventQueue events = victim.initTurn(testMap); // Make the unit try to heal itself.
+      GameEventQueue events = victim.model.getTurnInitEvents(victim, testMap); // Make the unit try to heal itself.
```
ThatIsAUsername, line 68: Why not use `turn` for this one?

ThislsAUsername: That would grant income.

src/Engine/GameEvents/TurnInitEvent.java
```diff
@@ -47,12 +50,14 @@ public GameAnimation getEventAnimation(MapView mapView)
...
-    return listener.receiveTurnInitEvent(army, turn);
+    return listener.receiveTurnInitEvent(map, army, turn);
   }
 
   @Override
-  public void performEvent(MapMaster gameMap)
+  public void performEvent(MapMaster map)
   {
+    for( Unit u : army.getUnits() )
+      u.initTurn(map);
```
ThatIsAUsername, line 60: This appears to fire for every unit, every turn. We previously had a caveat for off-map units (though we did init cargo units - I don't recall why, or how else a unit could be at an invalid location).

src/CommandingOfficers/Commander.java
ThatIsAUsername: Agree, this does clean things up a fair bit.

src/Engine/GameEvents/TurnInitEvent.java
ThislsAUsername: We didn't at time of writing.
In fact, cargo was init'd *twice* - once by `Commander`, once by its holder.

As for "Why init cargo?", `LAUNCH` depends on whether that unit has acted this turn or not.

src/Test/TestGameEvent.java
ThatIsAUsername: mumble mumble, comments for non-obvious intent, mumble.

ThislsAUsername: I put a comment in there talking about how the defeat happens...

src/Engine/GameEvents/TurnInitEvent.java
ThatIsAUsername: > LAUNCH

Right, fair.

I'm still seeing this get hit for each army, each turn, which is strange.

ThislsAUsername: I would expect only one `TurnInit` to go through each turn, so yeah, that's pretty weird.

ThatIsAUsername: Looks like the `TurnInitEvent` is hitting every unit, every turn, but `Commander` only hits `UnitModel.turnInitEvent` on his own turn. Meaning that repairs happen correctly, but fuel burn is effectively multiplied by the number of players.

ThatIsAUsername: Looks like `TurnEndEvent` inherits from `TurnInitEvent` but doesn't override everything.

ThislsAUsername: Whoops. Thanks.

src/Engine/GameEvents/TurnEndEvent.java
```diff
@@ -27,6 +27,11 @@ public GameEventQueue sendToListener(GameEventListener listener)
     return listener.receiveTurnEndEvent(army, turn);
   }
 
+  @Override
+  public void performEvent(MapMaster map)
```
ThatIsAUsername, line 31: Is there some reason `TurnEndEvent` needs to override `TurnInitEvent` at all, if we are just gonna override behavior to stop it? It sounds kinda sketchy from the word go.

ThislsAUsername: I know I had a reason, but I can't recall it anymore.
...I think I just didn't feel like copy/pasting the "do nothing" methods.
That seems like a "me" thing to do.

ThatIsAUsername: Haha, well I hope we learned a valuable lesson today.

ThatIsAUsername:
Looks good

PR #424, opened 2024-03-16
Fix Max having amazing D2D inf

ThislsAUsername:
I didn't do an early out since "distance units" doesn't explicitly call out non-inf?
But also neither does his buff description?

I dunno.

### DIFF
--- a/src/CommandingOfficers/AW1/OS/Max.java
+++ b/src/CommandingOfficers/AW1/OS/Max.java
@@ -62,7 +62,7 @@ public class Max extends AW1Commander
   @Override
   public void modifyUnitAttack(StrikeParams params)
   {
-    if( params.battleRange < 2 )
+    if( params.battleRange < 2 && params.attacker.model.isNone(UnitModel.TROOP) )
       params.attackPower += 50;
     if( params.attacker.model.isAny(UnitModel.INDIRECT) )
     {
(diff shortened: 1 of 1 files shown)


ThatIsAUsername:
> doesn't explicitly call out non-inf

Sort of. It says "Direct combat units have high firepower", but also clarifies "+50/0 vehicle direct combat".

> early out

You mean using `&&` instead of `||`? That would have left infs with a close-range buff.

ThislsAUsername:
> Sort of. It says "Direct combat units have high firepower", but also clarifies "+50/0 vehicle direct combat".

The parenthetical is my more-precise addition. It is not the "canonical text".
> > early out
> 
> You mean using `&&` instead of `||`? That would have left infs with a close-range buff.

No, I mean checking for `TROOP` at the start, which would exclude them from consideration for the indirect debuff.

ThatIsAUsername:
Gotcha, I think I'm tracking now. Yeah, neither explicitly mentions "non-infantry" units getting the buff, but in practice only non-infantry get the buff, and I guess there's room for interpretation on whether hypothetical future ranged infantry should get the debuff. I could go either way.

Looks good.

ThatIsAUsername:
Simple 'nuff. Ship it.

PR #425, opened 2024-03-24
Make the Models more model-y

ThislsAUsername:
It turns out I messed up Max's range dealie too and didn't notice that either.

I set up the Builder pattern with `UnitModel` ~with lots of regex~ so we can *clone* them however we want, but you (mostly - `role` is an exception) can't mess with the actual values... without changing the normal usage interface.
However, it's worth noting that the collections don't have an easy way to add immutability, AFAIK. You can prevent someone from changing *which* collection you have, but not prevent changing what's *in* the collection unless you protect yourself with copies or something.

`WeaponModel` prevents write access to anybody who's not its child. We don't go particularly deep with weapon modification at this point, so I figure Builder would be overkill there (and an API change isn't so huge there).

You will probably want to go into the `src` directory and run the jar to install the Eclipse plugin if you want to try this out.


ThatIsAUsername:
> You will probably want to go into the src directory and run the jar to install the Eclipse plugin if you want to try this out.

Is that... optional? Do we not just need Lombok in order to run at all with this change?

.classpath
```diff
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
 	<classpathentry kind="src" path="src"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
```
ThatIsAUsername, line 8: I see adding lombok below. Did this entry also need to change? Or is that just Eclipse doing what Eclipse wants with no forthcoming explanation?

src/AI/Muriel.java
```diff
@@ -1096,6 +1097,7 @@ public int compareTo(ModelValuePair other)
     }
   }
 
+  @Data
   private static class UnitModelPair
```
ThatIsAUsername, line 1101: Apologies in advance for what will probably be many questions - I'm inferring that the `@Data` attribute just handles the `hashCode` and `equals` implementations behind the scenes? Does that mean we just missed something (implemented those wrong, or failed to implement something additional)? I'm pretty sure this did work correctly at some point, though I did notice Muriel recreating matchups more recently.

ThislsAUsername:
> > You will probably want to go into the src directory and run the jar to install the Eclipse plugin if you want to try this out.
> 
> Is that... optional? Do we not just need Lombok in order to run at all with this change?

Good question. I've never tried using it without the IDE plugin.

.classpath
ThislsAUsername: I dunno, I just did the thing in the GUI and that's what Eclipse did.

src/AI/Muriel.java
ThislsAUsername: https://projectlombok.org/features/Data
It does more than just `hashCode` and `equals`, but yes.

And yeah, I'm not entirely sure why this worked and it didn't work before either.

ThislsAUsername: So, they have a tool for figuring out what's going on with the generated code:
https://projectlombok.org/features/delombok

I'm noticing they check `instanceof` instead of whether the classes are equal. I think that might be a difference, since `UnitModel`s ~until this PR~ are usually different classes?
The only other difference I'm seeing is null checks we don't otherwise have.

`ModelForCO`:
```java
  @java.lang.Override
  @java.lang.SuppressWarnings("all")
  public boolean equals(final java.lang.Object o) {
    if (o == this) return true;
    if (!(o instanceof ModelForCO)) return false;
    final ModelForCO other = (ModelForCO) o;
    if (!other.canEqual((java.lang.Object) this)) return false;
    final java.lang.Object this$co = this.getCo();
    final java.lang.Object other$co = other.getCo();
    if (this$co == null ? other$co != null : !this$co.equals(other$co)) return false;
    final java.lang.Object this$um = this.getUm();
    final java.lang.Object other$um = other.getUm();
    if (this$um == null ? other$um != null : !this$um.equals(other$um)) return false;
    return true;
  }

  @java.lang.SuppressWarnings("all")
  protected boolean canEqual(final java.lang.Object other) {
    return other instanceof ModelForCO;
  }

  @java.lang.Override
  @java.lang.SuppressWarnings("all")
  public int hashCode() {
    final int PRIME = 59;
    int result = 1;
    final java.lang.Object $co = this.getCo();
    result = result * PRIME + ($co == null ? 43 : $co.hashCode());
    final java.lang.Object $um = this.getUm();
    result = result * PRIME + ($um == null ? 43 : $um.hashCode());
    return result;
  }
```

`UnitModelPair`:
```java
    @java.lang.Override
    @java.lang.SuppressWarnings("all")
    public boolean equals(final java.lang.Object o) {
      if (o == this) return true;
      if (!(o instanceof Muriel.UnitModelPair)) return false;
      final Muriel.UnitModelPair other = (Muriel.UnitModelPair) o;
      if (!other.canEqual((java.lang.Object) this)) return false;
      final java.lang.Object this$first = this.getFirst();
      final java.lang.Object other$first = other.getFirst();
      if (this$first == null ? other$first != null : !this$first.equals(other$first)) return false;
      final java.lang.Object this$second = this.getSecond();
      final java.lang.Object other$second = other.getSecond();
      if (this$second == null ? other$second != null : !this$second.equals(other$second)) return false;
      return true;
    }

    @java.lang.SuppressWarnings("all")
    protected boolean canEqual(final java.lang.Object other) {
      return other instanceof Muriel.UnitModelPair;
    }

    @java.lang.Override
    @java.lang.SuppressWarnings("all")
    public int hashCode() {
      final int PRIME = 59;
      int result = 1;
      final java.lang.Object $first = this.getFirst();
      result = result * PRIME + ($first == null ? 43 : $first.hashCode());
      final java.lang.Object $second = this.getSecond();
      result = result * PRIME + ($second == null ? 43 : $second.hashCode());
      return result;
    }

    @java.lang.Override
    @java.lang.SuppressWarnings("all")
    public java.lang.String toString() {
      return "Muriel.UnitModelPair(first=" + this.getFirst() + ", second=" + this.getSecond() + ")";
    }
```

ThislsAUsername: Nah, I'm a doofus.
This is the problem in `UnitModelPair`:
```java
      if( first != other.first )
        return false;
      if( second != other.second )
        return false;
```
The `UnitModelForCO` isn't going to ever be the same instance, so you'd need to call `.equals()` to get positive results.

src/CommandingOfficers/DefendPeace/misc/Venge.java
```diff
@@ -88,16 +88,10 @@ public static CommanderInfo getInfo()
...
   {
-    GameEventQueue events = super.initTurn(map);
-    return events;
-  }
-  @Override
-  public void endTurn()
-  {
-    super.endTurn();
     aggressors.clear();
+    return null;
```
ThatIsAUsername, line 94: `null` instead of an empty `GameEventQueue`? I guess that saves the overhead of creating an empty one, but then the receiver needs to handle null checks.

ThatIsAUsername: Though I assume we must already be handling those, or this wouldn't be this way.

ThislsAUsername: Yup, and also those changes were on the fog branch.

src/CommandingOfficers/DefendPeace/RoseThorn/Tech.java
```diff
@@ -570,23 +570,27 @@ private static UnitModel createBattleMechModel(GameReadyModels grms)
...
-    BattleMech.healableHabs = new HashSet<TerrainType>(); // BattleMechs have specialized parts, not easy to repair.
+    var BattleMech = mdTank.toBuilder();
+    BattleMech.name(BATTLEMECH_NAME);
+    BattleMech.costBase(mdTank.costBase*2 + (antiAir.costBase/2));
+    BattleMech.abilityPowerValue(20);
+    BattleMech.maxFuel(30);
+    BattleMech.maxAmmo(10);
+    BattleMech.visionRange(2);
+    BattleMech.baseMovePower(4);
+    BattleMech.baseMoveType(new FootMech());
```
ThatIsAUsername, line 581: This doesn't set `role` or `healableHabs`. I don't think `setCalculatedProps` handles the `SURFACE_TO_AIR` property, either.
Guess I need to read up on lombok builders, too.

src/CommandingOfficers/DefendPeace/RoseThorn/Tech.java
```diff
@@ -570,23 +570,27 @@ private static UnitModel createBattleMechModel(GameReadyModels grms)
...
 
-    WeaponModel ratatat = antiAir.weapons.get(0).clone();
     ArrayList<WeaponModel> weapons = new ArrayList<WeaponModel>();
-    weapons.add(BattleMech.weapons.get(0));
+    weapons.add(mdTank.weapons.get(0));
+    WeaponModel ratatat = antiAir.weapons.get(0).clone();
```
ThatIsAUsername, line 585: Why do we clone the AA, but not the Md Tank weapon? I know it was like that before, but still. Past me should have left a comment, probs.

ThislsAUsername: I think we were cloning as a matter of general principles before, but I don't think there's much point now.

src/Units/WeaponModel.java
```diff
@@ -6,15 +6,22 @@
 import Units.AWBWUnits.AWBWUnitModel;
 import Units.DoRUnits.DoRUnitModel;
 import Units.KaijuWarsUnits.KaijuWarsUnitModel;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.experimental.Accessors;
 
+@Accessors(fluent = true)
+@Getter(AccessLevel.PUBLIC)
```
ThatIsAUsername, line 15: Isn't PUBLIC the default?

ThatIsAUsername: [This](https://projectlombok.org/features/GetterSetter) seems to confirm that, but extra clarity also doesn't hurt.

ThatIsAUsername:
Okay, I think I get what's going on here. I'm admittedly dubious of the value it provides, especially at the cost of requiring a new dependency. There is a reduction in LoC I guess, but it's slight. What is the selling feature here?

Also lombok didn't detect eclipse when I tried to install it, and I couldn't use the "Specify location" option because it doesn't show hidden folders in its file picker. I finally got it to install using the command-line version, but now eclipse crashes when I try to run it from the quick-launch, annoyingly (but it does start if I run it from a terminal, weirdly).

ThislsAUsername:
The reduction in LoC is nice, but the real selling feature is that the "missing lines of code" are correct.
This branch fixes a problem in current `master`.

There's also the fact that the models are mostly unmodifiable after creation, enforced by the compiler (instead of enforced by the developer not being an idiot).

> I couldn't use the "Specify location" option because it doesn't show hidden folders in its file picker

ctrl-H commonly works for showing hidden files in GUIs.

> I finally got it to install using the command-line version, but now eclipse crashes when I try to run it from the quick-launch, annoyingly (but it does start if I run it from a terminal, weirdly).

That's very weird. Maybe you're running two different binaries/configurations?

ThatIsAUsername:
Yeah, "correct invisible code" is nice, as long as it's also well-understood. And reducing builder boilerplate is nice, but I don't see much reduction. I dunno, I guess I'm not opposed, I just wouldn't have bothered with it. How'd you get into lombok, anyway?

> ctrl-H

Nope

> two binaries

Pretty sure not. I have eclipse installed via flatpak, so I'm just putting in the full filepath to the eclipse dir I put into the lombok install command. If I manually unmodify eclipse.ini, it starts again. I also tried modifying the .desktop file to call eclipse directly instead of doing a `flatpak run`, but it still isn't happy that way.

ThislsAUsername:
> And reducing builder boilerplate is nice, but I don't see much reduction.

Right, because we didn't have a builder before because it was too much boilerplate.
So instead of LoC reduction, you see the "new feature" of being unmodifiable, with a different API for construction.

"The models should be unmodifiable" is how I interpreted your point about "structural problems" in the other PR, anyhow.

I imagine "We don't have enough tests" is another "structural problem", but a test that tests something that isn't and won't be broken feels like boilerplate to me.

> I dunno, I guess I'm not opposed, I just wouldn't have bothered with it.

I didn't have a strong opinion on it either, but you kept prodding about making the models unmodifiable and this is IMO "the way" to do it.
> How'd you get into lombok, anyway?

I was really annoyed about there being no out-of-the-box way to interact directly with fields via reflection.
I forget the details, but having a "bean" (has a `getX()` and `setX()` for every X) added another out-of-the-box reflection tool that was suitable for what I was doing.
And Lombok can generate all those getters and setters.
(The implementation was still like 50% longer than my C# version, but "also runs on Linux" is technically an advantage.)

> Nope

It looks like you can put hidden folders into this box?
image
Dunno if that solves your general problem, though.

> I have eclipse installed via flatpak, so I'm just putting in the full filepath to the eclipse dir I put into the lombok install command.

Oh boy, is this another working directory problem? I love working directory problems.
If the lib doesn't work for you, I can fix the few bugs this fixes other ways and we can go on without it... but I wouldn't go on to make the models unmodifiable, so those issues could technically recur.

I guess a "third option" is to `delombok` the code here and check it in. I'm not entirely sure that's a win, though.

ThatIsAUsername:
> My C# version

You wrote a C# program to write boilerplate Java? I don't know whether to be impressed or horrified.

> It looks like you can put hidden folders into this box?

Tried it, actually. I just treats that as "the folder" instead of expanding it in the picker above. Maybe if I put in the complete/absolute path that would work.

> Oh boy, is this another working directory problem?

I don't think so? It doesn't matter where I am when I launch eclipse from the command line, anyway. It's also not a huge deal to launch eclipse from a terminal if needed, and I probably wouldn't suggest delomboking everything either.

ThislsAUsername:
> You wrote a C# program to write boilerplate Java? I don't know whether to be impressed or horrified.

No, I implemented the same set of functions in C# and Java.
The Java implementation was longer, even with lombok turning lots (hundreds?) of getter/setter definitions into a single `@Data` per class.

> Tried it, actually. I just treats that as "the folder" instead of expanding it in the picker above. Maybe if I put in the complete/absolute path that would work.

That's what I was thinking.

> I probably wouldn't suggest delomboking everything either.

Okay.
With that and your "not opposed" comment earlier, I think I'm leaning towards keeping this since it does fix a class of errors (a class of errors I expect to be relevant with all the CO conversions).
Thoughts?

ThatIsAUsername:
> With that and your "not opposed" comment earlier, I think I'm leaning towards keeping this since it does fix a class of errors (a class of errors I expect to be relevant with a notable fraction of the COs I'm pulling from the playtest version).
Thoughts?

I'm admittedly generally skeptical of these magic add-on tools, but I'm also not gonna say we shouldn't use something useful just because I'm a skeptic.

I'm also just annoyed it doesn't seem to work right. I figured out how to get eclipse to run from the quick-launcher (apparently the `-javaagent` tag lombok adds to eclipse.ini *must* come before the `-vmargs` tag), but then all the lombok stuff is flagged as errors once it starts. If I move `-javaagent` back to the end of eclipse.ini, then it won't run from the quick-launcher, but if I run from the command line it recognizes lombok. Something is wonky.

ThislsAUsername:
[This SO answer](https://stackoverflow.com/questions/3418865/cannot-make-project-lombok-work-on-eclipse/3425327#3425327) seems to indicate the proper order has `-vmargs` first.
It goes on to mention that you might need to specify the full path to `lombok.jar` to make it work, which sounds like it might be the problem since "running in a different context" (the quick launcher) fails while terminal invocation works.

ThatIsAUsername:
Ah, I was missing the `-Xbootclasspath/a` setting. Not sure why it worked when I ran it from the terminal though.

Though now it's decided to launch a terminal when I start eclipse anyway, so that's cool (I _did_ set `Terminal=true` in the .desktop file to help debug, but I changed it back).

ThislsAUsername:
> Though now it's decided to launch a terminal when I start eclipse anyway, so that's cool

Yeah, I got nothin'.
I take it this kills your immediate concerns?

ThatIsAUsername:
> I take it this kills your immediate concerns?

Yeah. Nothing blocking me from being able to use it

PR #427, opened 2024-03-31
Com Towers

ThislsAUsername:
I added both flavors, since I absolutely cannot do without being difficult.


ThislsAUsername:
*Mea culpa*, I forgot to make them show up in the preview count thingie
image

src/Engine/Combat/CombatContext.java
```diff
@@ -118,6 +124,32 @@ else if( calcType == CalcType.OPTIMISTIC )
...
+        MapLocation loc = map.getLocation(x, y);
+        Commander owner = loc.getOwner();
+        if( null == owner || uc.CO.army != owner.army )
+          continue;
+        if( loc.getEnvironment().terrainType.equals(TerrainType.DOR_TOWER) )
+          uc.towerCountDoR += 1;
+        if( loc.getEnvironment().terrainType.equals(TerrainType.DS_TOWER) )
+          uc.towerCountDS += 1;
+      }
+    }
```
ThatIsAUsername, line 150: Seems like tower counts would be a useful property to just have on the Commander. This would also let us recount much less frequently.

res/map/Deep_Forest.map
```diff
@@ -0,0 +1,13 @@
...
+  FR  FR  FR  FR  RD  FR  FR  FR  FR  FR  FR  RD  FR  FR  FR  FR
+  FR  FR  FR  T4  RD  FR  FR  FR  FR  FR  FR  RD  CT  FR  FR  FR
+  FR  FR  MT  FR  RD  FR  FR  SR  MT  FR  FR  RD  FR  MT  FR  FR
+  FR  FR  MT  FR  RD  FR  FR  MT  SR  FR  FR  RD  FR  MT  FR  FR
+  FR  FR  FR  CT  RD  FR  FR  FR  FR  FR  FR  RD  T4  FR  FR  FR
+  FR  FR  FR  FR  RD  FR  FR  FR  FR  FR  FR  RD  FR  FR  FR  FR
+ 3FC  FR  FR  FR  RD  RD  RD  RD  RD  RD  RD  RD  FR  FR  FR 1FC
+ 3CT 3FC  FR  CT  FR  FR  T4  FR  FR  CT  FR  FR  CT  FR 1FC 1CT
+ 3HQ 3CT 3FC  FR  FR  FR  FR  MT  MT  FR  FR  FR  FR 1FC 1CT 1HQ
+team, unit type, x, y (author: IntSys, AW4)
```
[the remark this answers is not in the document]
ThislsAUsername: I think supporting subfolders is the first big step there, to make it easy.

src/Engine/Combat/CombatContext.java
ThislsAUsername: On the one hand, yes.
On the other hand, caching that value lead to weirdness in the past.
On the third hand, this collects the comm count *according to a specific observer*, so the comm count isn't leaked by combat previews.

src/Test/TestJavier.java
```diff
@@ -0,0 +1,202 @@
...
+
+  @Override
+  public boolean runTest()
+  {
+
+    boolean testPassed = true;
+
+    testPassed &= validate(testNoTowerStats() , "  Javier 0T test failed!");
+    testPassed &= validate(testDoRTowerStats(), "  Javier DoR tower test failed!");
+    testPassed &= validate(testDSTowerStats() , "  Javier DS tower test failed!");
```
ThatIsAUsername, line 52: I'm not sure how you get to the point of writing these lines without questioning your life choices.

src/UI/Art/SpriteArtist/MapSelectMenuArtist.java
```diff
@@ -174,23 +181,20 @@ private static void drawMapSelectMenu(Graphics g, MapSelectController gameSetup)
...
     {
       for(int x = 0; x < mapInfo.getWidth(); ++x)
       {
-        if(mapInfo.terrain[x][y] == tileType)
+        TerrainType tt = mapInfo.terrain[x][y];
+        if( tt.isCapturable() )
         {
-          ++count;
+          int oldCount = propCounts.getOrDefault(tt, 0);
+          propCounts.put(tt, ++oldCount);
```
ThatIsAUsername, line 194: We didn't formerly show HQ/Lab counts, since that's determined by player count. I guess it isn't a _problem_ to show them, but it seems redundant.

src/Test/TestJavier.java
ThislsAUsername: Javier specifically is less effort to write a math test than to do a playtest.
...and yeah, I'd question my life choices if I actually respected the game design behind Javier, but I don't.

src/UI/Art/SpriteArtist/MapSelectMenuArtist.java
ThislsAUsername: Lab counts != player count.
Also, you can technically just slap down neutral HQs or extra HQs...

ThatIsAUsername: I mean. I guess. Neutral HQs are just cities though. The "objective" structures are also just much less indicative of what kind of map it is than the other things. Not gonna get too worked up about it either way though.

ThislsAUsername: Neutral HQs stay HQs when captured, which presents a liability paired with an extra terrain star.
Is that... good game design? Probably not. But it's not exactly equivalent to a city.

You're right that it's less likely to give a strong signal re: how the map plays.
I guess I could filter them.

PR #428, opened 2024-04-05
Other Stuff from Wally development

ThislsAUsername:
Fixes some bugs I found while testing new Wally code, and adds a feature I found useful.


src/AI/JakeMan.java
```diff
@@ -595,13 +595,19 @@ private GameAction findTravelAction(
         break;
       }
     }
+    // If we have to move and have no destinations, make the start tile the goal
+    if( mustMove && null == goal )
+    {
+      goal = new XYCoord(unit);
+      path = GamePath.stayPut(unit);
```
ThatIsAUsername, line 602: How does this let JakeMan move landers out of the way?

src/Engine/Combat/StrikeParams.java
```diff
@@ -55,7 +55,7 @@ public static StrikeParams buildStrikeParams(
         ( null == attacker.weapon ) ? 0 : attacker.weapon.getDamage(defender),
         isCounter);
 
-    if( gameMap.getEnvironment(attacker.coord).weatherType == Weathers.SIROCCO && !attacker.CO.immuneToSand )
+    if( null != attacker.env && attacker.env.weatherType == Weathers.SIROCCO && !attacker.CO.immuneToSand )
```
ThatIsAUsername, line 58: You've changed this is a few places. If the attacker/defender likely to be on an invalid tile or something?
Note I don't object; its cleaner to have all the info together, but the messages seem to indicate a potential bug.

src/Engine/Combat/CombatContext.java
```diff
@@ -37,6 +37,12 @@ public boolean isSim()
   public CombatContext(GameInstance gi, GameMap map,
                         UnitContext pAttacker, UnitContext pDefender,
                         CalcType pCalcType)
+  {
+    this(gi, map, pAttacker, pDefender, calcBattleRange(pAttacker, pDefender), pCalcType);
+  }
+  public CombatContext(GameInstance gi, GameMap map,
+                        UnitContext pAttacker, UnitContext pDefender, int battleRange,
+                        CalcType pCalcType)
```
ThatIsAUsername, line 45: I'm not sure what problem this addresses, I'm just thrilled that I'm that much closer to being able to implement teleporting ninjas. You can attack them from far away, and they will _still_ stab you in the face.

src/AI/JakeMan.java
ThislsAUsername: Line 605 right below just returns `null` if we don't have any goals.
So if we don't have a goal, we set a goal of staying put (and fail because there's more important stuff going on).
Intuitive algorithm design, I know.

src/Engine/Combat/StrikeParams.java
ThislsAUsername: The new Wally code considers hypotheticals without necessarily providing full gamestate info to the combat calc.
It wasn't a problem before (and I didn't feel like trying to make sure every calc got the map populated in every unit context, blegh)

src/Engine/Combat/CombatContext.java
ThislsAUsername: I'm sure I could dig up the usecase, but I haven't stared at Wally code in a few weeks now.

src/Engine/Combat/StrikeParams.java
ThatIsAUsername: Yeah, UnitContext has always given me the heebie jeebies for this reason. It's too... manual to make sure it has everything it needs.

src/AI/JakeMan.java
ThatIsAUsername: Oh, but of course. How silly of me not to see that before.

src/Engine/Combat/CombatContext.java
ThatIsAUsername: No idea. Haven't tried. Probably so.

src/Engine/Combat/StrikeParams.java
ThislsAUsername: The intent of allowing you to provide subsets is to... allow you to provide subsets.
In my mind, it makes sense to allow e.g. combat calculation without having a whole concrete map to look at. It won't take complex stuff like CO zones into account, but that's not unfixable either.

ThatIsAUsername:
I think this one and comm towers are both good to go. I'm not sure I really understand the pathing one yet.

PR #429, opened 2024-04-05
Optionally find all paths when doing pathfinding

ThislsAUsername:
Sometimes useful, maybe.

### DIFF
--- a/src/AI/CapPhaseAnalyzer.java
+++ b/src/AI/CapPhaseAnalyzer.java
@@ -188,7 +188,7 @@ public class CapPhaseAnalyzer implements Serializable
         if( null == infPath || infPath.getPathLength() < 1 )
           continue; // Can't reach
 
-        int distance = infPath.getFuelCost(inf, map);
+        int distance = infPath.getMoveCost(inf, map);
         if( distance < newOwnerDistance )
         {
           newOwnerDistance = distance;
@@ -363,7 +363,7 @@ public class CapPhaseAnalyzer implements Serializable
           }
           madeProgress = true; // We have somewhere we can still get to
 
-          final int distance = infPath.getFuelCost(inf, map);
+          final int distance = infPath.getMoveCost(inf, map);
           final int currentTotalMove = (last.extraTurns + 1) * infMove;
 
           if( distance <= currentTotalMove )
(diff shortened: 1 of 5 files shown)


ThatIsAUsername:
> Sometimes useful, maybe

Like when?
And while we're at it, how does this work, high-level? There's generally one shortest path to get somewhere, are you just also generating every possible circuitous (overlapping? I guess by definition since the destination will be the same?) route for funzies?

ThislsAUsername:
> Like when?

Like if you want to check if you've fully blocked off an enemy from reaching a tile, and you don't want to mock up a new map state to run a full path calculation on.
> And while we're at it, how does this work, high-level?

It makes the "parent" relationship optionally one-to-many. That optional case has all tiles that you *could* path into this tile from, if you started from the start tile.
> There's generally one shortest path to get somewhere, are you just also generating every possible circuitous (overlapping? I guess by definition since the destination will be the same?) route for funzies?

Nah, it's basically the same code. I was already _doing_ the checks to have the "all possible paths" information, but now I also store that information.
I don't think I have anything that actually builds alternate `GamePath`s - I just want to check hypotheticals.
And yeah, this is another Neo-Wally feature.

ThislsAUsername:
Let's say I have a 3x3 map, with a tank at the top-middle.
I want to prevent the tank from reaching the bottom-middle.
```
| t |
|   |
| g |
```
So I pathfind/floodfill where the tank can go, and keep track of *all* potential parents.
```
| t |
| v |
|>g<|
```
The top one (`v`) is the "main" parent, and would be in your shortest path.
But the tank can still reach `g` if I block that tile, so I'd have to block all of them.
I can also iterate back from those parents to *their* parents, like so:
```
| t |
|vov|
|>g<|
```
So with this info, I can verify that e.g. blocking all 3 middle tiles will block the tank from reaching `g`.

src/AI/CapPhaseAnalyzer.java
```diff
@@ -363,7 +363,7 @@ public void buildBaseCapChains(GameMap map, Army viewer, ArrayList<XYCoord> righ
           }
           madeProgress = true; // We have somewhere we can still get to
 
-          final int distance = infPath.getFuelCost(inf, map);
+          final int distance = infPath.getMoveCost(inf, map);
```
ThatIsAUsername, line 366: What determines whether you use Fuel or Move cost? Other places (e.g. line 280) still use Fuel when it looks like Move might be more apropos.

ThislsAUsername: How tired I was, probably.

ThatIsAUsername:
Sweet. Looks cool. I'd normally be against adding code that doesn't seem to be used anywhere, but I'm guessing this comes into play in the next PR.

PR #430, opened 2024-05-12
Wally version whatever-it-is-now

ThislsAUsername:
The big innovation here is that Wally plans his entire turn in advance, enabling him to "cancel" actions he's already planned if it would give him a better overall result.
I, of course, couldn't resist the urge to throw a bigger factor in my time complexity, but he does seem fairly functional on land-focused cart-style maps.

### DIFF
--- a/src/AI/WallyAI.java
+++ b/src/AI/WallyAI.java
@@ -104,7 +104,7 @@ public class WallyAI extends ModularAI
   private static enum TravelPurpose
   {
     // BUSINESS, PLEASURE,
-    WANDER(0), SUPPLIES(7), KILL(13), CONQUER(42), NA(99);
+    WANDER(0), SUPPLIES(7), KILL(13), CONQUER(42), BESIEGE(64), NA(99);
     public final int priority;
     TravelPurpose(int p)
     {
@@ -1043,7 +1043,8 @@ public class WallyAI extends ModularAI
       ArrayList<ActionPlan> travelPlans = ai.planTravelActions(
                                           this, gameMap, ai.threatMap,
                                           unit, avoidProduction, shouldWander,
-                                          canEvictSiege, evictionValue, EVICTION_DEPTH);
+                                          canEvictSiege, evictionValue, EVICTION_DEPTH,
+                                          null);
 
       if( null == travelPlans )
         return null;
@@ -1085,7 +1086,8 @@ public class WallyAI extends ModularAI
       ArrayList<ActionPlan> travelPlans = ai.planTravelActions(
                                           this, gameMap, ai.threatMap,
                                           unit, avoidProduction, shouldWander,
-                                          canEvictSiege, evictionValue, EVICTION_DEPTH);
+                                          canEvictSiege, evictionValue, EVICTION_DEPTH,
+                                          null);
 
       if( null == travelPlans )
         return null;
@@ -1131,7 +1133,8 @@ public class WallyAI extends ModularAI
       ArrayList<ActionPlan> travelPlans = ai.planTravelActions(
                                           this, gameMap, ai.threatMap,
                                           unit, avoidProduction, shouldWander,
-                                          canEvictSiege, evictionValue, EVICTION_DEPTH);
+                                          canEvictSiege, evictionValue, EVICTION_DEPTH,
+                                          null);
 
       if( null == travelPlans )
         return null;
@@ -1247,7 +1250,10 @@ public class WallyAI extends ModularAI
     {
       goals.addAll(findCombatUnitDestinations(gameMap, allThreats, unit));
       if( !goals.isEmpty() )
-        travelPurpose = TravelPurpose.KILL;
+        if( unit.model.hasImmobileWeapon() )
+          travelPurpose = TravelPurpose.BESIEGE; // Siege units have the highest travel priority since they're most likely to have to waste turns.
+        else
+          travelPurpose = TravelPurpose.KILL;
     }
 
     if( goals.isEmpty() ) // If there's really nothing to do, go to MY HQ
@@ -1334,7 +1340,8 @@ public class WallyAI extends ModularAI
                         AIModule whodunit, GameMap gameMap,
                         ArrayList<TileThreat>[][] threatMap, Unit unit,
                         boolean avoidProduction, boolean shouldWander,
-                        boolean canEvictSiege, int evictionValue, int recurseDepth)
+                        boolean canEvictSiege, int evictionValue, int recurseDepth,
+                        ArrayList<XYCoord> bannedTiles)
   {
     if( evictionStack.contains(unit) )
       return null;
@@ -1357,10 +1364,15 @@ public class WallyAI extends ModularAI
     if( !mustMove && !shouldWander && travelPurpose == TravelPurpose.WANDER )
       return null; // Don't clutter the queue with pointless movements
 
+    if( null != bannedTiles )
+    {
+      destinations.removeAll(bannedTiles);
+      validTargets.removeAll(bannedTiles);
+    }
     if( mustMove ) // If we *must* travel, make sure we do actually move.
     {
-      destinations.remove(new XYCoord(unit.x, unit.y));
-      validTargets.remove(new XYCoord(unit.x, unit.y));
+      destinations.remove(new XYCoord(unit));
+      validTargets.remove(new XYCoord(unit));
     }
     boolean attackEviction = mustMove && null != mapPlan[unit.x][unit.y].toAchieve && mapPlan[unit.x][unit.y].toAchieve.action.getType() == UnitActionFactory.ATTACK;
     if( attackEviction ) // Don't assume we can hop into our evicter's target's space, since that won't work
@@ -1413,7 +1425,7 @@ public class WallyAI extends ModularAI
       // Figure out how to get here.
       boolean spaceFree = null == plannedResident;
       if( !spaceFree &&
-          ( null == ap || ap.fromEviction || ap.purpose.priority > travelPurpose.priority) )
+          ( null == ap || ap.purpose.priority > travelPurpose.priority) )
         continue; // Bail if:
       // There's no other action to evaluate against ours
       // The other action is from an eviction
@@ -1552,7 +1564,7 @@ public class WallyAI extends ModularAI
     } // ~for destinations
 
     ArrayList<ActionPlan> bestPlans = null;
-    // Now that we have an ordered list of our travel locations, figure out the best one we can accomplish (potentially requiring eviction)
+    // Now that we have an ordered list of our travel locations, figure out the best one we can accomplish (potentially requiring eviction of both current and planned residents)
     while (!rankedTravelPlans.isEmpty())
     {
       final Entry<ActionPlan, Integer> entry = rankedTravelPlans.poll();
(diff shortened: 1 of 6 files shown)


src/AI/WallyAI.java
```diff
@@ -6,21 +6,46 @@
...
+import Engine.UnitActionLifecycles.CaptureLifecycle.CaptureAction;
 import Engine.UnitActionLifecycles.WaitLifecycle;
 import Terrain.*;
 import Units.*;
 import Units.MoveTypes.MoveType;
 
-/**
- *  Wally values units based on firepower and the area they can threaten.
- *  He tries to keep units safe by keeping them out of range, but will also meatshield to protect more valuable units.
- */
```
ThatIsAUsername, line 21: Oh, good. I was worried about having any kind of description here would make the file too long.

src/AI/ModularAI.java
```diff
@@ -161,13 +161,13 @@ public GameAction getUnitAction(Unit unit, GameMap map)
     }
   }
 
-  public static abstract class UnitActionFinder implements AIModule
+  public static abstract class UnitActionFinder<T> implements AIModule
```
ThatIsAUsername, line 164: > Throw the CRTP on UnitActionFinder

That doesn't look like CRTP - `UnitActionFinder` inherits from `AIModule`, but is templated with `ModularAI`. In CRTP, child classes template their base classes with themselves. Or am I misreading something?
I'm also not sure what we gain from this template in the first place, since it's just storing the `ModularAI` in a variable.

src/AI/WallyAI.java
```diff
@@ -6,21 +6,46 @@
...
+            new PowerActivator(army, CommanderAbility.PHASE_BUY),
+            new BuildStuff(army, this),
+            new FreeRealEstate(army, this),
+            new SiegeTravel(army, this),
+            new Travel(army, this),
+            new Eviction(army, this), // Getting dudes out of the way
+
+            new FillActionQueue(army, this),
+            new PowerActivator(army, CommanderAbility.PHASE_TURN_END)
+            ));
```
ThatIsAUsername, line 46: Why is `FillActionQueue` at the end, and `DrainActionQueue` is at the beginning? That sounds backwards, which means I don't understand the design intent here.

src/AI/WallyAI.java
ThislsAUsername: It was duplicated with the user-facing one.

src/AI/ModularAI.java
ThislsAUsername: Okay, fair. It's not exactly CRTP - it's the parent class's owned/static class that is templated with the child class in this case. It's CRTP-adjacent if you consider the parent's owned class(es) to be entangled with the parent class in a relevant way.

I gain... like 3 less lines of code per `ActionFinder` definition. Which probably isn't worth it, but I hate boilerplate.

src/AI/WallyAI.java
ThislsAUsername: Quoth Muriel:
```java
  @Override
  public GameAction getNextAction(GameMap gameMap)
  {
    // If we have actions ready to go, don't bother calculating anything.
    if( !queuedActions.isEmpty() )
    {
      GameAction action = queuedActions.poll();
      log(String.format("  Action: %s", action));
      return action;
    }
```
That's the action queue.
The rest of the `ActionFinders` push actions into an unordered planning map (`mapPlan` IIRC), and `FillActionQueue` is responsible for figuring out the order to *execute* those actions.
So:
I do any queued actions
I build a plan
I turn the plan into an action queue (and return the first action)

src/AI/WallyAI.java
```diff
@@ -166,29 +224,456 @@ public void endTurn()
...
+  private void updatePlan(ActionPlan plan)
+  {
+    XYCoord dest = plan.action.getMoveLocation();
+    final UnitPrediction destPredictTile = mapPlan[dest.x][dest.y];
+    ActionPlan evicted = destPredictTile.toAchieve;
+    ActionPlan actorPrevPlan = plansByUnit.get(plan.actor.unit);
+    cancelPlan(evicted);
+    cancelPlan(actorPrevPlan);
+
+    // Assuming our path has been planned well re:terrain, running into enemies is the only obstacle
```
ThatIsAUsername, line 258: This comment references enemies, but the code below does not. It just checks that the generated path isn't too short to be valid (also not sure how you are generating paths such that they might be too short to reach the destination. The comment is also repeated further down, so I suspect pasta.

ThislsAUsername: Seems both lines were added at the same time and it was inaccurate at that time (bd0caac), so agreed. Good catch.

src/AI/WallyAI.java
ThatIsAUsername: Oh man. Choosing actions and then choosing the order to execute them? Sounds dicey (though maybe that explains the logic to catch unexpected divergences and clear the rest of the plan).

ThislsAUsername: Yessir, dicey it is.
I have discovered a lovely little case where Wally will plan a perfectly viable series of attacks, then reassign one of the guys who was supposed to walk away to create space to dive in and attack, borking everything.
Haven't fixed it yet, though I have made progress.

src/AI/WallyAI.java
```diff
@@ -166,29 +225,460 @@ public void endTurn()
...
+        // We iterate from 1 because the first waypoint is the unit's initial position.
+        for( int i = 1; i < waypoints.size(); i++)
+        {
+          XYCoord from = waypoints.get(i-1).GetCoordinates();
+          XYCoord to   = waypoints.get( i ).GetCoordinates();
+          int cost = fff.getTransitionCost(map, from, to, actor.CO.army, false);
+          if( cost > actor.movePower )
+            if( !vacatedTiles.contains(to) )
+              return null;
+        }
```
ThatIsAUsername, line 561: This iterates over all the waypoints, but line 558 only compares `cost` against `actor.movePower` for the current segment's cost. Should `cost` accumulate across loop iterations? Or does `movePower` not account for distance moved so far? I guess this is just a check to see if the unit can't cross the boundary at all, which should only happen if there is an enemy in the way. If that's the case though, it seems like there should be a more straightforward way to check that.

ThislsAUsername: > Should `cost` accumulate across loop iterations?

Arguably, I guess.
> Or does `movePower` not account for distance moved so far? I guess this is just a check to see if the unit can't cross the boundary at all, which should only happen if there is an enemy in the way.

Pretty much.
> If that's the case though, it seems like there should be a more straightforward way to check that.

...Yeah, probably.

src/AI/WallyAI.java
ThatIsAUsername: He creates the plan, then instead of executing the plan, he just... changes the plan? That's a strange way to use a plan. Good luck though, I guess. These changes are getting pretty in the weeds.

ThislsAUsername: He considers a re-plan whenever he's planned to kill off a new unit, because that opens up new tiles for new attacks.
But I wasn't properly accounting for the order the actions would have to be in, which would break things.
With [769b3ee](https://github.com/ThislsAUsername/DefendPeace/pull/430/commits/769b3ee3a9d4b74e78640b4a7fdfc3a132a0b2f2), I managed to fix that such that (if you disabled the travel calc steps), he could plan and execute moves like this:
(Wally is blue, yellow is an ally)
image
image

But as I mentioned, having the other compute steps on messed it up.
I tried to address that, then got distracted with other stuff.

I've been tinkering with this on and off for ages, so sorry it's kind of all over the place.

ThislsAUsername: The tests I've added should give some insight into the sorts of things I'm trying to do, as well.
Was the "good luck" an indication that I can merge this whenever I decide it's ready? hah.

src/Test/TestAIBehavior.java
```diff
@@ -62,17 +69,19 @@ public boolean runTest()
   {
     boolean testPassed = true;
 
+    testPassed &= validate(testWalkInLine(Muriel.info), "  "+Muriel.info.getName()+" failed line walking test.");
```
ThatIsAUsername, line 72: Does Wally not pass this one? Or does it just assign the closer infantry and the other one is jobless? Hm; should maybe modify the test to put two cappable properties on the right.

ThislsAUsername: Yeah, he started failing it at some point in the process and I decided I didn't care :P
That's not a bad idea, though.

src/AI/WallyAI.java
```diff
@@ -579,314 +1066,737 @@ public GameAction getNextAction(PriorityQueue<Unit> unitQueue, GameMap gameMap)
...
+      return maxEvictionDepth - evictionStack.size();
+    }
+    /** Find the tiles a unit on startTile shouldn't go, based on any planned actions that need that tile to be open. */
+    public ArrayList<XYCoord> calcPostReqBans()
+    {
+      var bannedTiles = new ArrayList<XYCoord>();
+
+      for( ActionPlan plan : postrequisites )
+      {
+        var tile = new XYCoord(plan.actor.unit);
```
ThatIsAUsername, line 1287: This used to be `var tile = plan.actor.coord`. How is this different? Is the unit at a different coordinate than itself?

ThislsAUsername: There are (at least) 3 places the "unit's coordinate" is stored.
- The unit's actual coordinate. Wally cares about this some of the time, but not all of the time.
- The coordinate in `mapPlan` that the `ActionPlan` is stored (i.e. where the unit will move to complete its planned action this turn). Wally almost always cares about this one.
- The coordinate in the `ActionPlan`'s actor `UnitContext`. This should match the above.

It's pretty convoluted, but I don't think there's a cleaner way to enable pre-planning like this - It's kind of baked into the notion of considering "where I am" and "where I want to be" for your entire army at once.

src/AI/WallyAI.java
```diff
@@ -981,212 +1917,533 @@ private Map<XYCoord, UnitModel> queueUnitProductionActions(GameMap gameMap)
...
-        Collections.sort(availableUnitModels, new UnitPowerComparator(enemyToCounter, this));
-
-        // Grab the best counter.
-        UnitModel idealCounter = availableUnitModels.get(0);
-        availableUnitModels.remove(idealCounter); // Make sure we don't try to build two rounds of the same thing in one turn.
         // I only want combat units, since I don't understand transports
-        if( !idealCounter.weapons.isEmpty() )
+        if( currentCounter.um.weapons.isEmpty() )
+          continue;
+        // Filter stuff I can't afford; this breaks if we have localized discounts, but those don't exist (yet)
```
ThatIsAUsername, line 2016: We _do_ have localized price increases, if that makes a difference.

ThatIsAUsername: Though doesn't line 2022 get the location-specific cost anyway?

ThislsAUsername: It doesn't, in this case.
If I can't buy a tank at the normal price, it doesn't matter that I can also buy it at 8k.

I also have some thoughts about how I should (re)design his buying code, but distractions are very cruel.
You can imagine why I hadn't put this up for review for so long - there are always more changes I can/should make.

ThislsAUsername: > Though doesn't line 2022 get the location-specific cost anyway?

Technically yes, but that's a single location-specific cost (from an effectively arbitrary production location, since another unit type may be chosen to build there first)

src/AI/WallyAI.java
ThatIsAUsername: > I can merge this whenever I decide it's ready?

More or less, yeah. Wally is your thing, and I'm not gonna get too involved with changes to his code, plus there is a _lot_ going on here, with a good mix of "yeah, makes sense" and "how does this even run?". I've at least looked at all the changes at this point, but I'm trying to do more of a skim than a dive.

src/AI/WallyAI.java
ThatIsAUsername: > The coordinate in the ActionPlan's actor UnitContext. This should match the above.

Match _which_ of the above? The first two don't sound like they should match.

> considering "where I am" and "where I want to be" for your entire army at once

This problem statement would seem to imply you need two stored values, not three.

src/AI/WallyAI.java
ThislsAUsername: Understandable.
I am fairly interested in examples of the latter.

src/AI/WallyAI.java
ThislsAUsername: The coordinate where the `ActionPlan` is stored.
The `UnitContext` gets its coordinate set (to the end of the path) when I feed it the path it will use for its action.
And yes, the first two usually don't match (but units can act in place, so they *can* match).
> This problem statement would seem to imply you need two stored values, not three.

You're quite correct, but I also want to be able to get at the coordinate without having to iterate through the entire `mapPlan` looking for an identity match.

src/AI/WallyAI.java
ThatIsAUsername: I mean it's more "this doesn't make immediate sense, and I don't feel like spending half an hour figuring out how it does make sense, and if it didn't work you would probably have fixed it so it did".

Like, `canReachHitFromZone` - name doesn't really tell me what it's doing, and the first thing it does is grab the starting/root node and check if it has a parent; if not, it returns immediately. Root nodes are typically orphans (cry), so at first blush it seems like the rest of the function is unreachable.

I'm doubtful it would be worth complaining every time I didn't immediately understand what was going on, especially in Wally-specific code.

ThislsAUsername: Ah, yeah.

"HitFromZone" is supposed to indicate "the places this particular threat can hit this particular tile from".
Each one of them is a reachable tile for said threat, assuming there are no units in the way. (and the parent of each node is the tile that you would path through to get to it)
The function tells you if all of the HitFromZone is blocked (by either having units on or blocking access to those tiles).

Probably fair.
It's also probably fair to say I should add more documentation, but the function name itself is usually enough for me.

ThatIsAUsername: > the function name itself is usually enough for me.

And that's the difference between coding for yourself, and coding for the homicidal maniac who knows where you live and will be maintaining your code after you.

Or in my case, "future me who forgot everything I've written more than a year ago". Maybe you don't have that problem, but I sometimes do, and especially when I'm reading code written by not-me.

ThislsAUsername: Said homicidal maniac should know that this house of cards is necessarily interdependent and stateful, so making almost any change is at least a little fraught.
Like, there's a flag that controls whether Wally's units should fear death or not. Turning it off generally leads to better swarming/inf wall tactics, but also means his artillery think kissing enemy tanks is a good hobby.
I've tried a few strategies for toggling it, but each has had substantial failings.
We have decent test coverage for "Can you plan your way out of this paper bag?", but building good unit tests for complex tactical scenarios sounds hard.

For a slightly better example, I think I've gotten the "plan a string of actions that will be executable in a clogged map state" thing set up to the point where most of the state it needs is "on stack", and it can generate a sequence of actions that makes sense.
But trying to make it generic would add complexity I don't need, and would probably rule out some of the "other state" interactions it has to deconflict with already-planned action sequences.
Maybe that's a good argument for more documentation, not less.

I do forget sometimes as well, but my style of naming things is a pretty high-density language in my head, heh.

PR #431, opened 2024-05-18
Discriminators

ThislsAUsername:
I didn't feel like manually editing a billion extra CO portrait PNGs, so I made code to do it.


ThatIsAUsername:
Cool, looks good. Though I just noticed, why is Sturm VS's face kinda messed up?

ThislsAUsername:
He has the smoke trail thing that was his excitement portrait for winning unit combats.
It'd work better if it were animated, probably.
Or I could just let him have a discriminator and call it a day. I dunno.

ThatIsAUsername:
Looks good

PR #432, opened 2024-06-12
Fix towers not applying to acts of vandalism

ThislsAUsername:
Whoops


ThatIsAUsername:
This is why I don't like the manual nature of the "context" way of doing things. Not that I have a better suggestion at the moment, except maybe add testing?
Anyway, good catch.

ThislsAUsername:
The context being manually set up is kind of the whole point.
If we can't manually vary parameters, we can't ask theoretical questions very well.

I guess we could default to using `UnitContext` constructors that take lots of inputs, but having all that info populated is usually not necessary.

PR #433, opened 2024-06-12
Fix CO Zone edge cases

ThislsAUsername:
Dunno why I didn't think of what would happen if the COU was loaded/invisible.

### DIFF
--- a/src/CommandingOfficers/AW4/RuinedCommander.java
+++ b/src/CommandingOfficers/AW4/RuinedCommander.java
@@ -30,6 +30,7 @@ import Engine.StateTrackers.StateTracker;
 import Engine.UnitMods.UnitModifierWithDefaults;
 import Terrain.GameMap;
 import Terrain.MapMaster;
+import Terrain.MapPerspective;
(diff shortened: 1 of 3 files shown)


src/CommandingOfficers/AW4/RuinedCommander.java
```diff
@@ -208,14 +209,16 @@ public boolean isInZone(UnitContext uc)
 
     for( Unit cou : COUs )
     {
+      if( !army.myView.isLocationValid(cou.x, cou.x) )
+        continue; // Loaded/dead COUs shouldn't boost the top left, derp
```
ThatIsAUsername, line 213: Hah. Maybe our "not present" coordinates should have been farther away.

ThatIsAUsername:
Looks fine from here

src/CommandingOfficers/AW4/RuinedCommander.java
ThislsAUsername: Arguably, but it's also arguably a logic error to not account for "off map" at all, so...

PR #434, opened 2024-06-17
Markings markup

ThislsAUsername:
Seems we had symbols for markings the whole time, and I was just too dumb to add them to the map.

I also figure we probably only need one "extra" spot for unit status; it's not like every match is gonna have statuses, and we don't gain _that_ much from splitting them up when everything but HP can cycle.


res/ui/main/letters_lowercase.png
ThatIsAUsername, line 1: Unused until and unless we need to regenerate the font file, yeah.
Which, hopefully we won't need to do, and if so, then git history will still have it.
Plus removing the files eliminates them as a future red herring, so probably worthwhile.

ThatIsAUsername: Actually we've still got the dpfont_glyphs files with the letter/image mappings anyway, don't we? In that case, it's doubly good to remove the unused things.

ThislsAUsername: We sure do

src/UI/Art/SpriteArtist/UnitSpriteSet.java
```diff
@@ -234,26 +234,26 @@ public void drawUnit(Graphics g, AnimState state, int imageIndex, int drawX, int
   /**
    * Draw icons detailing any relevant unit information.
    * Icons are arranged as follows:
-   * Upper Left: Status icons, e.g. when stunned or low on fuel or ammo.
-   * Upper Right: Special markings, usually applied by Commander abilities.
-   * Lower Right: Activity icons, e.g. transporting units or capturing property.
-   * Lower Left: HP when not at full health (this can extend to the lower right if the unit has >10 HP).
+   * Lower Left: HP when not at 10 (this can extend to the lower right if the unit has >10 HP).
+   * Lower Right: Status/activity icons (ammo/fuel/stun/transport/capture), Commander markings, the ones' digit of HP>10.
+   * NOTE: Upper right is reserved for terrain marks, to disambiguate them from unit marks.
```
ThatIsAUsername, line 239: What was the driver for moving everything around? I thought we had hashed this out before.

ThislsAUsername: We did, you're right.
Somebody pointed out the capture mark was on the wrong side, which got me thinking about the issue again.
I *was* going to move it to the left side, but then I remembered about the multi-digit HP thing.
I previously wanted to have CO stuff on the upper right because then it wouldn't overlap with activity/HP overflow stuff, but I think overlap is rare enough that the visibility/readability gains of having the top of the unit unobscured most of the time are a bigger deal.

Also, we made that decision before COs could put marks on terrain. I think the gain of splitting unit and terrain marks is also pretty clear.

ThatIsAUsername: > capture mark was on the wrong side

Yeah, has been forever.

> the multi-digit HP thing

We _could_ move it to the bottom right and then overflow leftward (the way normal numeral places work), but that increases complexity, and I was never too worried about matching exactly.

> having the top of the unit unobscured

Yeah, that's a good point. There are some things (in particular, status conditions like stun that prevent you from using the unit) where that might even be desirable, but there's a good argument to keep the top clear in general.

> before COs could put marks on terrain

They could do that before this change too, or am I hallucinating?

ThislsAUsername: > We _could_ move it to the bottom right and then overflow leftward (the way normal numeral places work), but that increases complexity

I think it would reduce complexity. However, it means that the "more significant" HP digit is the one that cycles when there are other icons active.
If we didn't support HP > 10, I would swap it because matching is nice.

> and I was never too worried about matching exactly.

Very fair. I think UX quality > matching UX.

> Yeah, that's a good point. There are some things (in particular, status conditions like stun that prevent you from using the unit) where that might even be desirable, but there's a good argument to keep the top clear in general.

That's a good point about stun. I think trying to mix the icons for that would be a pain, though (since the two icon draw functions would have to coordinate).

> They could do that before this change too, or am I hallucinating?

You aren't.
We made the original icon allocation choices.
I later added terrain marks on top of unit marks for COs. (Cinder is to blame, heh)
I've now re-evaluated in light of terrain marks existing.

ThatIsAUsername:
Looks ship-shape

PR #435, opened 2024-06-22
AWBW ports with COU mechanics

ThislsAUsername:
Ported the AWBW ports from the playtest branch.
Not gonna lie, I think the originals are a lot stronger unless the map is high income.

### DIFF
--- a/src/CommandingOfficers/AWBW/COUableCommander.java
+++ b/src/CommandingOfficers/AWBW/COUableCommander.java
@@ -1,0 +1,125 @@
+package CommandingOfficers.AWBW;
+
+import Engine.GameScenario;
+
+import java.util.ArrayList;
+
+import CommandingOfficers.CommanderInfo;
+import CommandingOfficers.DeployableCommander;
+import CommandingOfficers.AWBW.AWBWCommander.AWBWAbility;
+import Engine.Combat.StrikeParams;
+import Engine.Combat.StrikeParams.BattleParams;
+import Engine.UnitMods.UnitDamageModifier;
+import Engine.UnitMods.UnitDefenseModifier;
+import Engine.UnitMods.UnitModifier;
+import Engine.UnitMods.UnitTypeFilter;
+import Terrain.MapMaster;
+import Units.Unit;
+import Units.UnitDelta;
+
+/**
+ * An AWBW-style CO that also has the COU mechanic, but no zone.
+ */
+public abstract class COUableCommander extends DeployableCommander
+{
+  private static final long serialVersionUID = 1L;
+
+  public final int couPowBase; // static values that define the power the COU should stay at
+  public final int couDefBase;
+  private int couPow; // floating values that dip on powers to match the above
+  private int couDef;
+
+  @Override
+  public int getCOUCount() {return 1;}
+
+  public COUableCommander(int atk, int def, CommanderInfo info, GameScenario.GameRules rules)
+  {
+    super(info, rules);
+    couPowBase = atk;
+    couDefBase = def;
+    couPow = couPowBase;
+    couDef = couDefBase;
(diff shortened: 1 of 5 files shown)


src/CommandingOfficers/AWBW/COUableCommander.java
```diff
@@ -0,0 +1,125 @@
...
+
+  protected static class NonStackingBoost extends AWBWAbility
+  {
+    private static final long serialVersionUID = 1L;
+    COUableCommander COcast;
+    private int atk, def;
+    UnitTypeFilter powMod;
+    UnitTypeFilter defMod;
+
+    protected NonStackingBoost(COUableCommander commander, String name, int cost, CostBasis basis, int pAtk, int pDef)
```
ThatIsAUsername, line 91: Some commentary on intent here would be helpful. If I apply two of these, one is ignored? If I have another boost and then this one is applied, does it only apply the difference? What is expected behavior?

src/CommandingOfficers/AWBW/COUableCommander.java
```diff
@@ -0,0 +1,125 @@
...
+      defMod = new UnitTypeFilter(new UnitDefenseModifier(def));
+    }
+
+    @Override
+    protected void perform(MapMaster gameMap)
+    {
+      COcast.couPow -= atk;
+      COcast.couDef -= def;
+      powMod.allOf = COcast.canDeployMask;
+      defMod.allOf = COcast.canDeployMask;
```
ThatIsAUsername, line 107: `perform`ing this boost subtracts from the COU's attack and defense?
Ohhh, so is the idea that it buffs all the units except the COU itself?
Or rather, all units _that could have been made into a COU__, but not others or the COU itself.

src/CommandingOfficers/AWBW/IDS/TabithaBW.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Tabitha", UIUtils.SourceGames.AWBW, UIUtils.IDS);
+      infoPages.add(new InfoPage(
+          "Can grant a single Mega Boost of +"+POWER+"/"+DEFENSE+" stats.\n"
+        + "This powerup lasts until the unit is destroyed.\n"));
+      infoPages.add(new InfoPage(
+          "Firestorm (6):\n"
+        + "4 HP strike that targets HP with gunboats counting half.\n"
```
ThatIsAUsername, line 40: HP strike targets HP, huh? Seems odd to call out (twice) that strikes can cause HP damage.

Also, "gunboats counting half" - as in they take half damage, or something else?

src/CommandingOfficers/AWBW/COUableCommander.java
ThislsAUsername: Yes. Agreed that I should have commented.

src/CommandingOfficers/AWBW/IDS/TabithaBW.java
ThislsAUsername: It _targets_ HP instead of funds value.
Gunboats again, count as half (HP, IIRC) value for targeting.

ThatIsAUsername: Or, reparsing this, I'm guessing it is an AOE attack that tries to inflict the maximum amount of overall HP damage (or maybe it tries to target units with the most HP? not necessarily the same), but counts gunboats as less important. If so, verbiage could be clearer, and also list the strike's AOE range.

ThatIsAUsername:
Cool, looks good. Though I was at first confused when the changes here had nothing to do with terrain or sea unit production. Other kind of ports, hah.

PR #436, opened 2024-06-22
Kaiju Wars unit tweaks

ThislsAUsername:
The main relevant thing here is the final commit: it fixes coordinate-based move buffs not showing up in hypothetical threat radius calcs.

To be a bit more concrete:
This PR adds a move boost to KW land units on roads/bridges.
I noticed that the threat radius calculated while hovering a move onto a road tile was wrong (i.e. it didn't take the road-move-boost into account for the threat calc).
I fixed it.


ThatIsAUsername:
Coool.

PR #438, opened 2024-06-29
TIL I can't read

ThislsAUsername:
The CO zone is probably not supposed to disappear when you're on certain tiles.


ThatIsAUsername:
oof

PR #443, opened 2024-07-15
More Jake

ThislsAUsername:
Counterbuilds:
Handle planes at all
Build MDs to stall assaults on your factories

### DIFF
--- a/src/AI/FightClub.java
+++ b/src/AI/FightClub.java
@@ -11,7 +11,7 @@ import java.util.List;
 
 import CommandingOfficers.Commander;
 import CommandingOfficers.CommanderInfo;
-import CommandingOfficers.DefendPeace.CyanOcean.Patch;
+import CommandingOfficers.CommanderLibrary.NotACO;
 import Engine.Army;
 import Engine.GameAction;
 import Engine.GameAction.EndTurnAction;
@@ -45,7 +45,7 @@ public class FightClub
     // How many bouts per map?
     int numGamesPerSet = 3;
     // Select CO(s).
-    List<CommanderInfo> COs = Arrays.asList(Patch.getInfo(), Patch.getInfo());
+    List<CommanderInfo> COs = Arrays.asList(NotACO.getInfo(), NotACO.getInfo());
     // Select AI(s).
     List<AIMaker> AIs = Arrays.asList(JakeMan.oldSchoolCool, JakeMan.info);
 
--- a/src/AI/FightClub.java
+++ b/src/AI/FightClub.java
@@ -47,7 +47,7 @@ public class FightClub
     // Select CO(s).
     List<CommanderInfo> COs = Arrays.asList(Patch.getInfo(), Patch.getInfo());
     // Select AI(s).
-    List<AIMaker> AIs = Arrays.asList(Muriel.info, JakeMan.info);
+    List<AIMaker> AIs = Arrays.asList(JakeMan.oldSchoolCool, JakeMan.info);
 
     // Run a set of games on each map.
     for( int setNum = 0; setNum < maps.size(); ++setNum )
--- a/src/AI/FightClub.java
+++ b/src/AI/FightClub.java
@@ -338,7 +338,7 @@ public class FightClub
         stopwatches.put(game.activeArmy, stopwatches.get(game.activeArmy) + thinkTimeNanos);
 
         // Map should-ish be covered in units by turncount == map area
-        if(game.getCurrentTurn() > game.gameMap.mapWidth * game.gameMap.mapHeight)
+        if(game.getCurrentTurn() > Math.max(game.gameMap.mapWidth, game.gameMap.mapHeight))
         {
           if(game.armies[0].getUnits().size()/2 > game.armies[1].getUnits().size() )
           {
--- a/src/AI/JakeMan.java
+++ b/src/AI/JakeMan.java
@@ -58,10 +58,9 @@ public class JakeMan extends ModularAI
 
   // What % base damage I'll ignore when checking safety
   private static final int    INDIRECT_THREAT_THRESHOLD = 7;
-  private static final int    DIRECT_THREAT_THRESHOLD = 13;
+  private static final int    DIRECT_THREAT_THRESHOLD   = 30;
   // Value to scale the funds damage I deal to something that threatens me
   private static final double FIRSTSTRIKE_ON_THREAT_WEIGHT = 2.0;
-  private static final int    STAY_UNHURT_BIAS = 1000;
   private static final int    STAY_ALIVE_BIAS = 2000;
   private static final int    BIG_THREAT_THRESHOLD = 80; // Enemy health at which I double the expected value of dealing damage
   // Fraction of the unit to remove from the counter-threat power of my unit type if I'm not attacking
@@ -693,8 +692,6 @@ public class JakeMan extends ModularAI
 
             // Convert to abstract value
             int extraLoss = 0;
-            if( loss >= 10 && unit.getHealth() == UnitModel.MAXIMUM_HEALTH )
-              extraLoss += STAY_UNHURT_BIAS;
             if( loss >= unit.getHealth() )
               extraLoss += STAY_ALIVE_BIAS;
             loss *= unit.getCost();
@@ -781,17 +778,19 @@ public class JakeMan extends ModularAI
     // If there are no threats we can't handle, dude is free.
     if( threatCounts.size() < 1 )
       return true;
-    double totalThreat = 0;
-    for( UnitModel threat : threatCounts.keySet() )
-    {
-      totalThreat += threatCounts.get(threat);
-      // Be extra scared of stuff we can't hit back against
-      if( isWeakTo(unit.model, threat) )
-        totalThreat += threatCounts.get(threat) * 2;
-    }
-    // If we have threats, but we have good terrain, that's good enough
+
+    // Count dude as free if:
+    //   we get 3+ terrain stars
+    //   the threat surplus is small
+    //   that threat is of the same unit type
     final int defLevel = gameMap.getEnvironment(xyc).terrainType.getDefLevel();
-    if( defLevel > totalThreat )
+    if( defLevel < 3 || unit.model.isAirUnit() )
+      return false;
+    if( threatCounts.size() > 1 )
+      return false;
+    if( !threatCounts.containsKey(unit.model) )
+      return false; // Our one threat is not same-type
+    if( threatCounts.get(unit.model) < 1.3 )
       return true;
     return false;
   }
--- a/src/AI/JakeMan.java
+++ b/src/AI/JakeMan.java
@@ -59,6 +59,7 @@ public class JakeMan extends ModularAI
   // What % base damage I'll ignore when checking safety
   private static final int    INDIRECT_THREAT_THRESHOLD = 7;
   private static final int    DIRECT_THREAT_THRESHOLD   = 30;
+  private static final int    MASSIVE_THREAT_THRESHOLD  = 70;
   // Value to scale the funds damage I deal to something that threatens me
   private static final double FIRSTSTRIKE_ON_THREAT_WEIGHT = 2.0;
   private static final int    STAY_ALIVE_BIAS = 2000;
@@ -90,6 +91,11 @@ public class JakeMan extends ModularAI
     {
       this.counter = counter;
       this.power = power;
+    }
+    @Override
+    public String toString()
+    {
+      return counter.name + " countering for " + power;
     }
   }
   // For each enemy unit type, my unit types X/Y/Z counter it at this effectiveness percent
@@ -905,6 +911,27 @@ public class JakeMan extends ModularAI
 
           if( marginalCost <= budget )
           {
+            boolean safeToBuild = true; // Don't build counter units in range of units they're squishy to.
+            for( UnitModel threat : unitMapEnemy.keySet() )
+            {
+              if( unitMapEnemy.get(threat).containsKey(coord) )
+              {
+                double power = unitMapEnemy.get(threat).get(coord);
+                UnitContext tc = new UnitContext(myArmy.cos[0], threat);
+                tc.chooseWeapon(ratio.counter);
+                if( null == tc.weapon )
+                  continue;
+                int damage = (int) (tc.weapon.getDamage(ratio.counter) * power);
+                if( damage > MASSIVE_THREAT_THRESHOLD )
+                {
+                  safeToBuild = false;
+                  break;
+                }
+              }
+            }
+            if( !safeToBuild )
+              continue;
+
             builds.put(coord, ratio.counter);
             budget -= marginalCost;
             shouldSave = false; // If we've already built some counter units, don't worry about saving for more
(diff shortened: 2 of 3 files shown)


src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+      return;
+
+    // tanks: no need to calc them beyond the Md clause
+    // copters: 1.5 copters or 1AA per
+    if( null != copter )
+    {
+      counterOrder.add(copter);
+      unitTypeToCounterMap.put(copter, new ArrayList<>());
+      var copterCounters = unitTypeToCounterMap.get(copter);
+      CounterRatio copterCounterCopter = new CounterRatio(copter, 200/3);
```
ThatIsAUsername, line 1150: Whence 200/3?

src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+    var mdCounters = unitTypeToCounterMap.get(mdTank);
+    var bomber = myArmy.cos[0].getUnitModel(UnitModel.AIR_TO_SURFACE | UnitModel.JET, false);
+    if( null != bomber )
+      mdCounters.add(new CounterRatio(bomber, 250));
+    var neoTank = allTanks.get(2);
+    if( null != neoTank )
+      mdCounters.add(new CounterRatio(neoTank, 150));
+
+    // 1 fighter per 1.7 (yes) bombers
+    //   (trunctate after the multiplicarion, so you get one in response to one bomber and a second in response to the third),
```
ThatIsAUsername, line 1168: multiplicarion

src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+        counterTypeNegationMap.put(bomber, new HashMap<>());
+        var bomberCounterNegators = counterTypeNegationMap.get(bomber);
+        bomberCounterNegators.put(antiAir, new ArrayList<>());
+        var aaCounterCounters = bomberCounterNegators.get(antiAir);
+        aaCounterCounters.add(new CounterRatio(copter, 100));
+      }
+    }
+
+    // stealth: have 1 healthy fighter on the board if a stealth is present
+    var stealths = myArmy.cos[0].getAllModels(UnitModel.AIR_TO_SURFACE | UnitModel.AIR_TO_AIR | UnitModel.JET, false);
```
ThatIsAUsername, line 1191: This is still a super awkward way to identify a specific unit type.

src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+      {
+        for( var actionType : s.baseActions )
+          if( actionType instanceof TransformLifecycle.TransformFactory )
+          {
+            stealthIsStealth = true;
+            break;
+          }
+        if( stealthIsStealth )
+          break;
+      }
```
ThatIsAUsername, line 1204: This is just double-checking that the model is hidden and has the ability to change? Defensive programming against other hypothetical air-to-air, air-to_ground, jet units?

src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+    if( stealthIsStealth && null != fighter )
+      for( var s : stealths )
+      {
+        counterOrder.add(s);
+        unitTypeToCounterMap.put(s, new ArrayList<>());
+        var stealthCounters = unitTypeToCounterMap.get(s);
+        CounterRatio fighterCounterStealth = new CounterRatio(fighter, 100);
+        fighterCounterStealth.roundTargetPercentUp = true; // Fighter needs to be healthy to deal
+        stealthCounters.add(fighterCounterStealth);
```
ThatIsAUsername, line 1213: I'm not sure I understand what's happening here either. It looks like you are mapping the fighter unit model as a counter to the stealth unit model, but both models are drawn from `myArmy.cos[0]` above - are we countering our own unit types?

src/AI/FightClub.java
```diff
@@ -338,7 +338,7 @@ public GameResults runGame(GameInstance game, PrintStream defaultOut)
         stopwatches.put(game.activeArmy, stopwatches.get(game.activeArmy) + thinkTimeNanos);
 
         // Map should-ish be covered in units by turncount == map area
-        if(game.getCurrentTurn() > game.gameMap.mapWidth * game.gameMap.mapHeight)
+        if(game.getCurrentTurn() > Math.max(game.gameMap.mapWidth, game.gameMap.mapHeight))
```
ThatIsAUsername, line 341: So like, Spann Island matches would have to be over in 15 turns? Seems short, especially for large, squarish maps.

ThatIsAUsername: Just ran a Jakeman/Strong v Muriel/Patch on Spann that took 70 turns. Jakeman looked like he should have had it in the bag around 15, but then it swung back the other way (Muriel upgraded to Mechs for standard infantry and built Md/Neo tanks, but Jake stuck with inf/tanks).
And sure, Spann isn't exactly peak design for competitive maps, but that time limit seems overly restrictive if the goal is to see how things end.

src/AI/JakeMan.java
```diff
@@ -785,16 +867,188 @@ private Map<XYCoord, UnitModel> queueUnitProductionActions(GameMap gameMap)
...
     log("Evaluating Production needs");
-    ArrayList<UnitModel> wantedTypes = new ArrayList<>();
 
+    ArrayList<FactoryThreatState> potentialMdBuilds = new ArrayList<>();
+    ArrayList<MapLocation> mdBases = CPI.getAllFacilitiesFor(mdTank);
+    UnitContext theVeh = new UnitContext(myArmy.cos[0], tank); // for getting the zone of concern
+    for( MapLocation loc : mdBases )
+      potentialMdBuilds.add(new FactoryThreatState(this, loc.getCoordinates(), gameMap, theVeh));
+
+    // "Net health" of the units I might want to build counters for, or use as counters.
```
ThatIsAUsername, line 898: Net health? Assets minus liabilities? Current health minus intended damage? Or do you mean total cumulative health?
... I'm just being pedantic.

src/AI/JakeMan.java
```diff
@@ -785,16 +867,188 @@ private Map<XYCoord, UnitModel> queueUnitProductionActions(GameMap gameMap)
...
+    HashMap<UnitModel, Integer> niceHealth = new HashMap<>();
+    HashMap<UnitModel, Integer> meanHealth = new HashMap<>();
```
ThatIsAUsername, line 900: [nice/mean](https://github.com/timvieira/justified-variables)?

src/AI/JakeMan.java
```diff
@@ -785,16 +867,188 @@ private Map<XYCoord, UnitModel> queueUnitProductionActions(GameMap gameMap)
...
+        remainingHealth  -= counterPower;
+        if( ratio.roundTargetPercentUp )
+        {
+          int roundable = remainingHealth % UnitModel.MAXIMUM_HEALTH;
+          if( roundable >= CounterRatio.ROUND_UP_HEALTH )
+            remainingHealth += UnitModel.MAXIMUM_HEALTH - roundable;
+        }
+      }
+      if( remainingHealth < UnitModel.MAXIMUM_HEALTH )
+        continue; // We consider it fully-countered
```
ThatIsAUsername, line 984: If a unit type has less than one unit's worth of cumulative HP left, Jakeman ignores it?

src/AI/JakeMan.java
```diff
@@ -846,29 +1107,122 @@ else if( um == copter ) // Consider downgrading a tank to an inf to get a copter
...
+
+    counterBuildSetup();
+
+    allCounterContestants = new HashSet<>();
+    for( UnitModel counterable : unitTypeToCounterMap.keySet() )
+    {
+      allCounterContestants.add(counterable);
+      for( CounterRatio ratio : unitTypeToCounterMap.get(counterable) )
+        allCounterContestants.add(ratio.counter);
+    }
```
ThatIsAUsername, line 1127: You are adding both the type we want to counter, and the type that counters it? And then the only place this is used is to ignore cumulative health for units that aren't in this map? So it's more like a cache of units that Jake is willing to think about?
Also it seems like this might have overlap with `unitTypeToCounterMap`, which is also used in `queueUnitProductionActions`

src/AI/JakeMan.java
ThislsAUsername: And *truncate.

Best I can do is `[sic]`, since this is my spec :wink:

src/AI/JakeMan.java
ThislsAUsername: Would you prefer a search by name? lol
The intent is to default to searching by *relevant properties*, so you aren't stuck with e.g. not affecting Neotanks or whatever.

src/AI/JakeMan.java
ThislsAUsername: Yes... and Kaiju Wars fighters realize this hypothetical already.

src/AI/JakeMan.java
ThislsAUsername: We are mapping from and to *universal* `UnitModel`s.
Doesn't matter if we're Kanbei or Colin, or who we're fighting - our behavior is the same, so we don't care about the specifics.

src/AI/FightClub.java
ThislsAUsername: FWIW, it checks if there is an overwhelming value advantage instead of having a *strict* turn limit.
I guess we could raise it back up, but I'm not sure how much I care about the amount of noise there is since there's so much variance in the rest of the system.

src/AI/JakeMan.java
ThislsAUsername: Enemy assets vs enemy liabilities you've already built, based on the defined ratios for each relationship, cumulatively.
Or I could just use scare quotes.

src/AI/JakeMan.java
ThislsAUsername: Yessir.

ThislsAUsername: I guess I could go evil/good, but I feel like that's a little strong.

src/AI/JakeMan.java
ThislsAUsername: Yes, because it enables him to use different rounding modes for different units.
IIRC fighter and bopter (as counter units) are the relevant examples.

src/AI/JakeMan.java
ThislsAUsername: > You are adding both the type we want to counter, and the type that counters it?

Yes. This is a cache for the types of units we count for counterbuild logic purposes (vs the ones we ignore, like tanks and infantry).
> And then the only place this is used is to ignore cumulative health for units that aren't in this map?

No, see previous line.
> So it's more like a cache of units that Jake is willing to think about?

Yes, actually. It's the set of units that factor into his special/extra planning.
> Also it seems like this might have overlap with `unitTypeToCounterMap`, which is also used in `queueUnitProductionActions`

It intentionally overlaps, but I have to count both things that I might *want* to counter and might *build as* a counter.
Without both categories, JakeMan will either not counterbuild at all, or he'll just keep spamming AA as long as you have a jet or two. ...I say this from experience.

src/AI/JakeMan.java
ThislsAUsername: 2/3, but in percent because we int mathing.
I have to have 1.5 copters per enemy copter to consider all the copters copter-countered, so my copters count as 1/1.5 of an enemy copter. Multiply both by 2 for int-ness and it's 200/3.

ThatIsAUsername: Okay, so this (and the other magic force-proportion numbers) are just SME-provided rules of thumb. Cool I guess, as long as they don't occur in more than one place.

src/AI/JakeMan.java
ThatIsAUsername: Sure, I get wanting to be able to get a unit or list of units based on shared properties. On the other hand, it's also super convenient to be able to just grab a specific unit type, which is what you are trying to do here.
Besides which, the existence of a way to grab a specific unit type in no way prevents using attribute-based filtering. The real argument for this being the "preferred" mechanism is that it allows us to make the code apply across all unit sets, but then you still have code like `mdTank = allTanks.get(1);`, which betrays that pretense.

src/AI/JakeMan.java
ThatIsAUsername: Sigh.

src/AI/FightClub.java
ThatIsAUsername: Ah, so this is just allowing an earlier end to the fight, gotcha. It could still skew results (and I think there's a good chance it would have in the fight I just ran), but that might still be more or less reasonable.

src/AI/JakeMan.java
ThislsAUsername: That's the plan. And naming these at all sanely would be... difficult.

src/AI/JakeMan.java
ThislsAUsername: > On the other hand, it's also super convenient to be able to just grab a specific unit type, which is what you are trying to do here.

I suppose. I guess it'd be possible to make a universal unit enum that doesn't interact with the damage tables (since that would be a lot of bloat).
But it feels weird to set up something like that with what's ultimately pretty niche utility? And you'd still have to check for absence of that unit type, which IMO the current API makes a little more obvious.
> then you still have code like `mdTank = allTanks.get(1);`, which betrays that pretense.

I dunno. It does a semi-sane thing in KaijuWars instead of exploding.
It's obviously not _ideal_, but this AI isn't designed with even DoR in mind, let alone that.

ThatIsAUsername: > I guess it'd be possible to make a universal unit enum

One with values like TANK_SMALL, TANK_MD, TANK_BIG or whatever, that may or may not exist for a given unit set? I think I might lean toward (for cases like Jake where you have UnitSet-specific logic) to have a function that just explicitly sets his internal tank/mdTank/etc variables to the right UnitModel based on which UnitSet is active (and maybe fall back to the current logic if the UnitSet isn't the expected one?). It would consolidate the mess, though it's debatable whether it's a real improvement.

> It does a semi-sane thing in KaijuWars instead of exploding.

Sure, but it's far from a general solution, though a general solution is admittedly a hard ask here.

ThatIsAUsername:
The unit set convo is sort of adjacent to the point of this MR, so I think you can merge when ready.

src/AI/JakeMan.java
ThislsAUsername: > I think I might lean toward (for cases like Jake where you have UnitSet-specific logic) to have a function that just explicitly sets his internal tank/mdTank/etc variables to the right UnitModel based on which UnitSet is active (and maybe fall back to the current logic if the UnitSet isn't the expected one?).

So what, it would downcast the unit types and check against the damage chart enum values? Or maybe "downcast" like the visiting that happens for weapons?
I assume you're not proposing the the unit set expose a function specifically for picking units for JakeMan.
> Sure, but it's far from a general solution, though a general solution is admittedly a hard ask here.

`ASSAULT` in flag terms is basically "essence of tank", and we've made the assumption that unit sets should order their units in rough cost/practicality order.
I think it's about as general as it's getting (without digging into the damage table algorithmically).

PR #444, opened 2024-07-21
Add the missing BM COs, and also fix a bunch of stupid mistakes

ThislsAUsername:
I went on an adventure to discover every Power of Money implementation is slightly different.

I added the 0-funds option 'case the AWBW Sasha description mentioned it.
This gave me an idea to test/play a stupid thing, which lead to me finding a bug with our repair logic.

Also, I guess I should clean up my repo folder 'cause there's a bunch of untracked files and it's a real drag when I forget to include a critical file like an idiot.


src/UI/GameOptionSetupController.java
```diff
@@ -19,7 +19,7 @@ public class GameOptionSetupController implements IController
 
   private GameOption<FogMode> fowOption = new GameOption<FogMode>("Fog of War", FogMode.values(), 0);
   private GameOption<Integer> startingFundsOption = new GameOptionInt("Starting Funds", 0, 50000, 1000, 0);
-  private GameOption<Integer> incomeOption = new GameOptionInt("Income", 250, 20000, 250, 1000);
+  private GameOption<Integer> incomeOption = new GameOptionInt("Income", 0, 20000, 250, 1000);
```
ThatIsAUsername, line 22: RIP Colin, I guess

ThatIsAUsername: Or maybe he'd do better since could make more money with his power? Hm. Maybe RIP Kanbei.

ThislsAUsername: No, Colin can't create money if he doesn't have money.
Kanbei, OTOH, just wins with the predeployed units.

src/Test/TestDumbMath.java
```diff
@@ -0,0 +1,98 @@
...
+import Terrain.MapInfo;
+import Terrain.MapLibrary;
+import Terrain.MapMaster;
+import Units.UnitContext;
+import Units.UnitModel;
+
+/**
+ * Tests Power of Money variants because I make poor life choices
+ */
+public class TestDumbMath extends TestCase
```
ThatIsAUsername, line 19: Setting aside the issue of the quality of your life choices, I think I'd prefer a different name for this class.

ThatIsAUsername: On the other hand, I guess it will technically only ever matter if a test fails, and then context will be there to save the day, but I mean, still.

src/Test/TestDumbMath.java
```diff
@@ -0,0 +1,98 @@
...
+    hit2 = shot2.calculateDamage();
+    testPassed &= validate(hit2 == 228,     "    Colin2 PoM is wrong");
+    shot2.attackerHealth = 40;
+    hit2 = shot2.calculateDamage();
+    testPassed &= validate(hit2 ==  91,     "    Colin2 PoM is wrong");
+    shot2.attackerDamageMultiplier = CommandingOfficers.AW2.BM.Colin.calcSuperMult(640660);
+    shot2.attackerHealth = 60;
+    hit2 = shot2.calculateDamage();
+    testPassed &= validate(hit2 == 657,     "    Colin2 PoM is wrong");
+
```
ThatIsAUsername, line 94: How'd you pick the funds amounts for these tests? And why use different ones for each test? How am I supposed to know which Colin is best Colin when numbers not same?

res/co/Grit_2.png
ThatIsAUsername, line 1: But Grid didn't have the brim in AW2? I am confuse.

src/CommandingOfficers/AW2/OS/Max.java
```diff
@@ -61,7 +61,7 @@ public Max(GameScenario.GameRules rules)
   {
     super(coInfo, rules);
 
-    CommanderAbility.CostBasis cb = new CommanderAbility.CostBasis(CHARGERATIO_FUNDS);
+    CommanderAbility.CostBasis cb = getGameBasis();
```
ThatIsAUsername, line 64: Since every CO needs to call this, can we just move this now-universal call into the parent constructor?

ThatIsAUsername: You could also make `getGameBasis()` abstract at the base `Commander` level, and only have one place where this is called.

src/Test/TestDumbMath.java
ThislsAUsername: I was thinking I might eventually put more things in it, but I guess Javier already exists and has his own file.
Thoughts on a "cart math" test file for these purposes?
...though that would technically only cover *one* Javier...

Naming is hard.

src/Test/TestDumbMath.java
ThislsAUsername: I played some matches (badly) on cart, and tested numbers with savestates.
They aren't the same numbers because math is hard.

Best Colin is unambiguously DS Colin since he charges Gold Rush faster, and it doesn't really matter whether you have +1000 attack or +1100 attack during PoM... but his PoM is also strongest:
AW2 PoM is weakest.
AWBW and DS Colin PoM are close, but \*3/1000 < \*1/300, respectively.

res/co/Grit_2.png
ThislsAUsername: He did in Japan!
I thought about bringing back white-bearded Olaf too, for kicks, but it turns out I'm lazy.

src/CommandingOfficers/AW2/OS/Max.java
ThislsAUsername: Every CO needs _at least one_ basis. Ave's an example that has multiple.

My initial thought on this was "but only the child has access to the powers at construction time", but we are technically already constructing many powers with a null basis for description purposes.
I see two "easy options" for a single callsite:
1) Parent defines a basis instance variable and fills it out. Child uses instance var.
2) Child COs construct their powers with no basis, and the parent fills them out.

1 is simple and breaks nothing, but it's misleading (The Commander->basis relationship is `1:n`, not `1:1`... even if `n` is usually 1).
2 can not break Ave by doing null checks on the basis first, I guess. But it's multi-stage initialization for something really simple and otherwise-intuitive.

src/Test/TestDumbMath.java
ThatIsAUsername: One of the two hard problems, yeah.

This file is specific to Collin and his money-based buffs, so I'd prefer to name it for that, and have a separate file for verifying Javier functionality (which you have). A global "cart math" file is probably too general.

src/Test/TestDumbMath.java
ThatIsAUsername: Ah, makes sense. As long as we answer the important question here.

src/CommandingOfficers/AW2/OS/Max.java
ThatIsAUsername: I was thinking along the lines of 1), but you're right, Ave is the fly in the soup here. All right, I'm sorry I ruin things, and please carry on.

ThatIsAUsername:
Rename the Collin money-power test file if you feel so inspired, and then this looks good to go.

src/CommandingOfficers/AW2/OS/Max.java
ThislsAUsername: I ruin everything often enough. You need your turn in the spotlight.

PR #447, opened 2024-08-09
Add a Unit Cap

ThislsAUsername:
This has to go on top of the YC changes since they both tweak `CreateUnitEvent` - YC to expose the new unit's health for meddling (for Sensei), this to enable preventing it from spawning.

Pursuant to your complaint, I also added a GUI thing so options can be flagged as invalid:
image
This, of course, led to me throwing the option of invalid options onto arty firing and production actions, for UX-y reasons.

I'm not at all certain this is a good design, but it does seem to function, which is a nice feature.

I am also aware this kinda-sorta breaks our AIs, so I've added an AI constraint test on another branch: `aiunitcap`
I'm not gonna make the test pass just yet, since there are two AI PRs in the air at the moment.

### DIFF
--- a/src/Engine/GameInput/SelectUnitProduction.java
+++ b/src/Engine/GameInput/SelectUnitProduction.java
@@ -91,12 +91,7 @@ class SelectUnitProduction extends GameInputState<Engine.GameInput.SelectUnitPro
     ArrayList<UnitBuildOption> menuOptions = new ArrayList<>();
     int maxNameLength = 0;
     int maxPriceLength = 0;
-    int armyCount = 0;
-    boolean disableAll = false;
-    for( Commander coi : co.army.cos )
-      armyCount += coi.units.size();
-    if( co.gameRules.unitCap <= armyCount )
-      disableAll = true;
+    boolean disableAll = !co.army.canBuildUnits();
 
     // Start by getting just the unit names.
     for(UnitModel model : models)
--- a/src/Engine/GameInput/SelectUnitProduction.java
+++ b/src/Engine/GameInput/SelectUnitProduction.java
@@ -6,25 +6,24 @@ import CommandingOfficers.Commander;
 import Engine.GameAction;
 import Engine.GameActionSet;
 import Engine.XYCoord;
+import UI.InGameMenu;
+import UI.InGameMenu.MenuOption;
 import Units.UnitModel;
 
 /************************************************************
  * Presents options for building a unit.                    *
  ************************************************************/
-class SelectUnitProduction extends GameInputState<String>
+class SelectUnitProduction extends GameInputState<Engine.GameInput.SelectUnitProduction.UnitBuildOption>
 {
   private Commander builder;
-  private ArrayList<String> myStrings;
   private ArrayList<UnitModel> myUnitModels = null;
   private XYCoord myProductionLocation = null;
 
-  @SuppressWarnings("unchecked")
   public SelectUnitProduction(StateData data, Commander builder, ArrayList<UnitModel> buildables, XYCoord buildLocation)
   {
     super(data);
     this.builder = builder;
     myUnitModels = buildables;
-    myStrings = (ArrayList<String>) data.menuOptions;
     myProductionLocation = buildLocation;
   }
 
@@ -33,25 +32,22 @@ class SelectUnitProduction extends GameInputState<String>
   {
     OptionSet options = null;
     if( null != myStateData.menuOptions )
-    {
       options = new OptionSet(myStateData.menuOptions.toArray());
-    }
     return options;
   }
 
   @Override
-  public GameInputState<?> select(String option)
+  public GameInputState<?> select(UnitBuildOption option)
   {
     GameInputState<?> next = this;
 
-    if( null != option && null != myUnitModels )
+    if( null != option && option.enabled && null != myUnitModels )
     {
-      for( String buyable : myStrings )
+      for( UnitModel buyable : myUnitModels )
       {
-        if( option == buyable )
-          {
-          UnitModel model = myUnitModels.get(myStrings.indexOf(buyable));
-          myStateData.actionSet = new GameActionSet(new GameAction.UnitProductionAction(builder, model, myProductionLocation), false);
+        if( option.item == buyable )
+        {
+          myStateData.actionSet = new GameActionSet(new GameAction.UnitProductionAction(builder, buyable, myProductionLocation), false);
           next = new ActionReady(myStateData);
         }
       }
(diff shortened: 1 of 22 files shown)


ThatIsAUsername:
AAAAAH
image

ThislsAUsername:
Yeah, your maps with previous saved options will _probably_ default to 0 unit cap.
I guess I could add logic to make it use the defaults if it doesn't have values for everything?

ThatIsAUsername:
Ah, okay. And yeah, that seems sensible.
Semi-related; I also noticed that if you have a map with saved tag-team settings, you can:
 - go into CO select, verify the teams' COs were remembered
 - back out to settings, switch to no tags
 - forward again to CO select, and it'll still show multiple COs per team.

src/UI/Art/SpriteArtist/SpriteUIUtils.java
```diff
@@ -24,21 +29,19 @@ public static BufferedImage makeTextFrame(String item, int hBuffer, int vBuffer)
   }
   public static BufferedImage makeTextFrame(Color bg, Color frame, int hBuffer, int vBuffer)
   {
-    return makeTextMenu(bg, frame, bg, new ArrayList<String>(), 0, hBuffer, vBuffer);
+    return makeTextMenu(bg, frame, bg, new ArrayList<>(), 0, hBuffer, vBuffer);
```
ThatIsAUsername, line 32: Are generic type arguments discouraged now?

ThislsAUsername: I was messing with the function signature while I was trying to get it to work, and not specifying is easier to "update".

...and it doesn't make any difference to the generated code. `ArrayList` is always of `Object`; the compiler just inserts downcasts for you.

src/Engine/GameInput/SelectUnitProduction.java
```diff
@@ -33,87 +32,84 @@ protected OptionSet initOptions()
...
+    }
+    @Override
+    public String toString()
+    {
+      return label;
+    }
+    public void bakeLabel(int nameLen, int priceLen)
+    {
+      String fmt = "%-"+nameLen+"s %"+priceLen+"d";
+      label = String.format(fmt, item.name, price);
```
ThatIsAUsername, line 82: Hah. Clever.

ThislsAUsername: Formatting format strings is my passion

src/Engine/UnitActionLifecycles/BattleLifecycle.java
```diff
@@ -37,6 +37,33 @@ public static class BattleFactory extends UnitActionFactory
   {
     private static final long serialVersionUID = 1L;
 
+    // Provide an invalid FIRE option on selecting an action when:
+    // Weapon could attack, but nothing's in range
+    // Weapon is out of ammo
+    @Override
```
ThatIsAUsername, line 43: Why bother showing a grayed-out menu item for units that can't fire? I'm no great fan of magically-disappearing menu items in general, but I've never felt that was a problem in the Advance Wars context; there really aren't alternate menus the mission option _might_ be in, and it's generally very clear why you can/can't do something.

ThislsAUsername: On current `master`, if I move onto an arty that's vaguely near the front and hit enter twice, one of two things happens:
I get the targeting reticle and can pick who I want to shoot.
I wait the arty in place.

With this change, the second possibility becomes "I am still on the action menu".
It's misclick prevention (and, for what it's worth, accurate to AW2)

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
```diff
@@ -28,12 +28,32 @@ public UnitProduceFactory(UnitModel type)
...
+      GameActionSet realActions = getPossibleActions(map, movePath, actor, ignoreResident);
+      if( null == realActions )
+      {
+        XYCoord moveLocation = movePath.getEndCoord();
+        if( moveLocation.equals(actor.x, actor.y)
+            && actor.hasMaterials() )
+        {
+          GameActionSet gas = new GameActionSet(new UnitProduceAction(this, actor), false);
+          gas.isInvalidChoice = true;
+          return gas;
```
ThatIsAUsername, line 44: Remind me what `hasMaterials` is checking - it's `return materials > 0 || !model.needsMaterials;`, but what is `needsMaterials`? It's defaulted to true, which seems wrong if it's a marker for which units can use materials, it isn't referenced anywhere else, and there isn't a comment explaining anything.
Also I tried to see if I could make this invalid option appear, but I'm still (intermittently?) running into the "you cannot build anything" bug.

ThatIsAUsername: Further inspection implies that `needsMaterials` is a flag for whether the unit should cost a material, if produced by another unit. This is kinda the opposite of what the variable name implies.

src/UI/GameOptionSetupController.java
```diff
@@ -61,14 +63,15 @@ public GameOptionSetupController(GameBuilder aGameBuilder)
...
     if( initialPicksMap.containsKey(GAME_OPTIONS_CONFIG_KEY) )
     {
+      GameOption<?>[] chronologicalOptions = new GameOption<?>[] {fowOption, startingFundsOption, incomeOption, weatherOption, unitSchemeOption, tagsOption, securityOption, unitCapOption};
       String fullConfig = initialPicksMap.get(GAME_OPTIONS_CONFIG_KEY);
-      String[] s = fullConfig.split("\\s+"); // Split (string integers)
-      int si = 0; // Split Index
-      while (s[si].isEmpty()) ++si; // Skip any sneaky empty splits
+      String[] configInts = fullConfig.split("\\s+"); // Split (string integers)
+      int configIndex = 0; // Split Index
+      while (configInts[configIndex].isEmpty()) ++configIndex; // Skip empty splits at the start
```
ThatIsAUsername, line 70: Do empty splits correspond to default values? Why would there be throwaway data in there?

src/Engine/UnitActionLifecycles/BattleLifecycle.java
ThatIsAUsername: Okay, cool. makes sense.

ThatIsAUsername:
Mostly looks good. I didn't attempt to characterize the overall design, so I also can't tell you if there's a better way, but "working" is probs good enough. Though not being able to build any units is a bit of a downer, so you should probs fix that, and then this should be good to go.

Also, a side note: tersely-worded commit messages are great, so long as they convey the necessary meaning; ideally the intent of the change and maybe how it was accomplished. However, too much brevity can be a problem. For example, I'm still not sure what "Add metadata disabled drawing to menu items" is supposed to mean. It appears to be composed of English words in precisely the same way that [this song](https://www.youtube.com/watch?v=77Ms1oCiDH4) is composed of Spanish ones. Should it be "metadata-disabled drawing"? Does it even make more sense with the hyphen? I'm not even sure it does.
Something like "Introduced new MenuOption class to allow disabling menu items" might have been more clear, assuming that adequately captures the change. Anyway, I nitpick, but clear messages really do help to orient the reviewer.

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
ThislsAUsername: It's probably worth noting that while reviewing code, you can actually go to the definition by clicking on the function name:
image
> It's defaulted to true, which seems wrong if it's a marker for which units can use materials, it isn't referenced anywhere else

It's a marker for which units can run out of materials, i.e. have finite materials. 0 is a finite number of materials.
I put that in for KaijuWars because the sky carrier needs to be able to produce planes forever (and evidently forgot to make DoR APCs building things use `hasMaterials()`, derp).

Your point of "not referenced anywhere" is kind of a flaw in Lombok - the generated builder/setter for that value are how it gets set, and you can't see the definitions (or find all references unless you've found at least one).

I suppose an equivalent option would be to provide versions of those actions that don't cost materials, but this seemed simpler.
> there isn't a comment explaining anything.

Touche. IIRC we discussed the name at length, but a name isn't always enough.

> Also I tried to see if I could make this invalid option appear, but I'm still (intermittently?) running into the "you cannot build anything" bug.

You need a DoR Carrier, materials, and no money/unit count for a Seaplane.
As for the "bug", I still think that's saved options biting you (because now the order has changed, so any options you saved in the interim would have an unrelated value in the final slot). Try deleting `res/map_options`.
> Further inspection implies that `needsMaterials` is a flag for whether the unit should cost a material, if produced by another unit. This is kinda the opposite of what the variable name implies.

I have no idea how you came up with that.
`needsMaterials` is only used in `hasMaterials()`, which is called on the actor (the unit that exists that wants to build another unit) to find out whether it has materials.
A DoR Carrier normally has 4 materials. If we changed it to have `needsMaterials = false`, then it would have infinite materials.

src/UI/GameOptionSetupController.java
ThislsAUsername: I think the first version of the code started the line with a space, so I ended up putting guard in there.
Arguably, this should be more robust since it's reading from the filesystem, but /shrug.

ThislsAUsername:
> Mostly looks good. I didn't attempt to characterize the overall design, so I also can't tell you if there's a better way, but "working" is probs good enough.

I'm not gonna complain too hard, given your time constraints.
> Though not being able to build any units is a bit of a downer, so you should probs fix that, and then this should be good to go.

I think you have stored `Protect Turns = OFF` values in the final slot of your map options, from the version before I added the newest commit.
> Also, a side note: tersely-worded commit messages are great, so long as they convey the necessary meaning; ideally the intent of the change and maybe how it was accomplished.

Oh, good. You do read them.
> [this song](https://www.youtube.com/watch?v=77Ms1oCiDH4)

...nice.
> Should it be "metadata-disabled drawing"? Does it even make more sense with the hyphen? I'm not even sure it does.

Okay, so this is an unfortunate consequence of me doing the thing where I think about [alternate interpretations](https://xkcd.com/37/).
"Metadata-disabled drawing" and "metadata disabled-drawing" are pretty much equivalent (in this context, which would push me toward the second hyphenation if I was writing the message over right now).
It means I added drawing logic for items disabled by metadata.
Arguably, the correct answer is to hyphenate the whole phrase, but I guess I either didn't think of it or didn't feel like it.
> Something like "Introduced new MenuOption class to allow disabling menu items" might have been more clear, assuming that adequately captures the change.

A fair chunk of the change was the actual drawing logic, which is why I didn't write something like that.
> Anyway, I nitpick, but clear messages really do help to orient the reviewer.

Agreed.

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
ThatIsAUsername: > You need a DoR Carrier, materials, and no money/unit count for a Seaplane.

So after running out of materials entirely, even the invalid menu option disappears. Also, do we have a visual display for materials? I don't see one (or remember adding one). Be a good issue to create if not (as if we ever revisit those).

> As for the "bug", I still think that's saved options biting you

That fixed it. BUT, is this also going to bite every person online who slaps in the new jar? Is there some way we can detect the old version and update (or just ignore) it?

> I have no idea how you came up with that.

On further further inspection, I think I misread `performEvent`, which checks `builder.model.needsMaterials`. I think I saw `model`, and assumed `builder` was the Lombok thing, instead of, y'know, the unit doing the building. Sigh.

> needsMaterials is only used in hasMaterials()

And `UnitProduceLifecycle.performEvent()`. Heh.

ThatIsAUsername:
Ideally I'd like to protect users from the config file change problem, but then I think this is good to go.

src/Engine/UnitActionLifecycles/UnitProduceLifecycle.java
ThislsAUsername: > So after running out of materials entirely, even the invalid menu option disappears.

Right. The invalid option says "if the state of the world changes a bit, you could be building a seaplane right now".
Greyfield is the only way to restore materials, though.
...on that note, should Jess restore materials? I feel like I should make her restore materials.
> Also, do we have a visual display for materials? I don't see one (or remember adding one). Be a good issue to create if not (as if we ever revisit those).

Nnnnnope. I think I considered it and didn't feel like making an icon.
> That fixed it. BUT, is this also going to bite every person online who slaps in the new jar?

It would have, I think.
> Is there some way we can detect the old version and update (or just ignore) it?

Yes, that's the new feature that gave you such problems just now.
It now orders the *saved* items in the order they were added to the code, instead of menu order.
This just happened to cause problems for people unlucky enough to run the code between adding the new option and changing that order.

> On further further inspection, I think I misread `performEvent`, which checks `builder.model.needsMaterials`. I think I saw `model`, and assumed `builder` was the Lombok thing, instead of, y'know, the unit doing the building. Sigh.

Heh, fair enough.

PR #448, opened 2024-08-10
Weathernanigans

ThislsAUsername:
This PR fixes some interactions with DS/DoR-style rain and powers that can cancel the fog it brings (since fog state and weather state aren't linked in the same way here).

I'm not giving this effect to AW1/2 Olaf/Drake, for a few reasons:
They never interact with fog-bringing weather on cart.
They produce only a single day of weather, which would still (in the current implementation) cancel the entire N-day sequence.
Their weather effects are substantive on their own.

This does mean DS Olaf gets to cancel one of Penny's fog-rain days for free, but I'm fine with that:
It'd be kind of annoying to make the implementation more complex to handle that case properly (and then I'd have to reexamine the cases above...).
~It's not actually a clear-cut advantage for Penny to reinstate fog for a single turn when Olaf's rain runs out - she would end up having a fog-turn right after Olaf's non-fog turn.~ Nevermind, I did my math wrong: The sequence is Penny/Rain, Olaf/Snow, P/S, O/S, P/S, O/R, which totals to 3 full days. The point that it's a niche inter-game interaction that'd take a lot of effort stays, though.

I guess one could argue that Penny should still get the primary use of her rain. If people thought that was important, I'd probably just add a clause of "activates fog for a turn if the COU is in `SMOKE` at turn start".
That would technically be very weird jank, but would solve this case in the most advantageous way for Penny - she would get an un-fogged turn (in DoR-style fog but not Trilogy, lol), while Olaf would have to cope with fog as normal, much like the COP activation turn.
This would also have the effect that any localized source of permanent `SMOKE` would let Penny turn fog on globally forever, which is funny.

### DIFF
--- a/src/CommandingOfficers/AW3/BM/Olaf.java
+++ b/src/CommandingOfficers/AW3/BM/Olaf.java
@@ -38,10 +38,12 @@ public class Olaf extends AW3Commander
           + "Winter weather poses no problem for Olaf or his troops. Snow causes his firepower to rise, and his troops can move with no penalty.\n"
           + "(Normal movement and +20 attack in cold weather.)"));
       infoPages.add(new InfoPage(new Blizzard(null, null),
-            "Causes chill (doubled fuel consumption) for two days.\n"
+            "Causes cold weather (doubled fuel consumption) for two days.\n"
+          + "Disables temporary fog, if it's active.\n"
           + "+10 attack and defense.\n"));
       infoPages.add(new InfoPage(new WinterFury(null, null),
-            "A mighty blizzard causes two HP of damage to all enemy troops. The chill also persists for two days.\n"
+            "A mighty blizzard causes two HP of damage to all enemy troops. The cold weather also persists for two days.\n"
+          + "Disables temporary fog, if it's active.\n"
           + "+10 attack and defense.\n"));
       infoPages.add(new InfoPage(
             "Hit: Warm Boots\n"
@@ -95,10 +97,11 @@ public class Olaf extends AW3Commander
     @Override
     public GameEventQueue getEvents(MapMaster map)
     {
-      GameEvent event = new GlobalWeatherEvent(Weathers.CHILL, 2);
+      GlobalWeatherEvent weather = new GlobalWeatherEvent(Weathers.CHILL, 2);
+      weather.canCancelFog = true;
 
       GameEventQueue events = new GameEventQueue();
-      events.add(event);
+      events.add(weather);
 
       return events;
     }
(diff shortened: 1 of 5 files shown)


src/CommandingOfficers/AW3/BM/Olaf.java
```diff
@@ -38,10 +38,12 @@ public instantiator()
...
           + "(Normal movement and +20 attack in cold weather.)"));
       infoPages.add(new InfoPage(new Blizzard(null, null),
-            "Causes chill (doubled fuel consumption) for two days.\n"
+            "Causes cold weather (doubled fuel consumption) for two days.\n"
+          + "Disables temporary fog, if it's active.\n"
           + "+10 attack and defense.\n"));
       infoPages.add(new InfoPage(new WinterFury(null, null),
-            "A mighty blizzard causes two HP of damage to all enemy troops. The chill also persists for two days.\n"
+            "A mighty blizzard causes two HP of damage to all enemy troops. The cold weather also persists for two days.\n"
+          + "Disables temporary fog, if it's active.\n"
```
ThatIsAUsername, line 46: Maybe "weather-induced" fog, instead of "temporary"? Are there any other sources of temporary fog? If someone sets the game weather to persistent rain (or smoke I guess), would this disable the fog of war then?

src/CommandingOfficers/AW4/IDS/Penny.java
```diff
@@ -83,11 +87,17 @@ private static class Stormfront extends RuinedAbility
...
     public GameEventQueue getEvents(MapMaster map)
     {
-      Weathers[] candidates = { Weathers.SLEET, Weathers.SMOKE, Weathers.SIROCCO };
-      int rand = map.game.getRN(candidates.length);
+      ArrayList<Weathers> candidates = new ArrayList<>();
+      candidates.add(Weathers.SLEET);
+      candidates.add(Weathers.SMOKE);
+      candidates.add(Weathers.SIROCCO);
+      for( Unit u : COcast.COUs ) // Don't roll the same weather that's currently active
+        candidates.remove(map.getEnvironment(u.x, u.y).weatherType);
```
ThatIsAUsername, line 95: And what if Penny has a COU in each of the three weather types? Wait, that can't happen? We should give Penny multiple COUs, then.

src/CommandingOfficers/AW3/BM/Olaf.java
ThislsAUsername: "Weather-induced" would imply that the fog is tied to the weather state, which it isn't.
> Are there any other sources of temporary fog?

Not yet. Maybe eventually.
> If someone sets the game weather to persistent rain (or smoke I guess), would this disable the fog of war then?

Those weathers do not inherently make the game fog in this engine, and I don't really want to change that (since weather is local).
I guess it'd be doable to have fog rules apply anywhere a fog weather is in effect (but not otherwise), but I at least don't want to add that scope right now.

But to answer the question, "no".
If fog is on, it would counteract the vision penalty effects of those weathers.

src/CommandingOfficers/AW4/IDS/Penny.java
ThislsAUsername: If you wanna be really pedantic about it, I guess I could check for the weather the *unit who activated the power* is in.

ThatIsAUsername:
Looks good. I'm fine with new weather powers interrupting prior ones; it would almost be more confusing to do it differently (unless you add a "forecast" display...)

ThatIsAUsername:
Is this still waiting on anything?

ThislsAUsername:
I was thinking you might have something else to say on the one thread before I hit the button.
I guess not.

PR #449, opened 2024-08-10
Fix resupply moving your cursor

ThislsAUsername:
It bothered me


ThatIsAUsername:
Does what it says on the tin. The real question is: should APC resupply actions result in a cursor move to the target just like attack/other actions, and just suppress it for turn-init/structure-based resupplies? (answer: No, that is not really the question)

PR #450, opened 2024-08-10
Fix AI pathing

ThislsAUsername:
Muriel and ISAI can actually participate on 2020 now.


ThatIsAUsername:
Doesn't look like you changed ISAI, and he's still sorting by distance rather than travel time. Otherwise looks good (though I haven't tried to break it or anything).

Could have sworn I fixed Muriel at some point, but maybe I just did it locally or something. Good to get this fixed.

PR #451, opened 2024-09-03
Make Wally spam arty

ThislsAUsername:
He really needs some kind of aggression, and to handle fuel better, and to... actually wall. Those are hard problems, though.
But he finally actually feels like a distinct playstyle instead of just "bad, but occasionally builds arty".

### DIFF
--- a/src/AI/WallyAI.java
+++ b/src/AI/WallyAI.java
@@ -1586,6 +1586,12 @@ public class WallyAI extends ModularAI
       {
         evictionFailure |= ( ec.evictionStack.contains(ev) );
         evictionFailure |= ( ev.isTurnOver && ev.CO.army == myArmy );
+        if( !evictionFailure )
+        {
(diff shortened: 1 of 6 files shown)


src/AI/WallyAI.java
```diff
@@ -649,7 +652,8 @@ public GameAction getUnitAction(Unit unit, GameMap gameMap)
     {
       if( ai.plansByUnit.containsKey(unit) )
         return null;
-      boolean isSiege = unit.model.hasImmobileWeapon();
+      boolean isSiege = unit.model.isAny(UnitModel.SIEGE);
+      isSiege |= unit.model.hasImmobileWeapon();
```
ThatIsAUsername, line 656: ಠ_ಠ

ThatIsAUsername:
More arty, better unit flags, better balance of unit types in army composition. Got it. I'm not gonna try to scrutinize the AI logic too closely unless it's doing weird things to core game stuff.

PR #452, opened 2024-09-06
Black Hole COs

ThislsAUsername:
I noticed we were missing a mass heal event and a listen hook, so I made something up.
I think having no separate mass heal hook is probably fine?


src/CommandingOfficers/AWBW/OS/Andy.java
```diff
@@ -69,13 +71,15 @@ private static class HyperRepair extends AWBWAbility
...
-      super.perform(gameMap);
+      var heal = new MassHealEvent(null, myCommander.army.getUnits(), HEAL*10);
+      heal.roundUp = myCommander.roundUpRepairs;
+
+      GameEventQueue events = new GameEventQueue();
+      events.add(heal);
+
+      return events;
     }
   }
```
ThatIsAUsername, line 84: Are we not using `perform` anymore? We may want to update some docstrings if we are changing how abilities are normally implemented.

src/Engine/GameEvents/MassHealEvent.java
```diff
@@ -0,0 +1,80 @@
...
+import Units.Unit;
+import lombok.var;
+
+/**
+ * Heals an arbitrary number of units, potentially at cost<p>
+ * Note: Healing dealt will not be tracked correctly if the event is executed twice
+ */
+public class MassHealEvent implements GameEvent
+{
+  // Records how much health each victim lost; doubles as our victim storage area
```
ThatIsAUsername, line 21: "victim"? "lost"?

src/CommandingOfficers/AWBW/OS/Andy.java
ThislsAUsername: My thought was that events are the "preferred" mechanism, since they can be animated and listened to.
`perform` is still useful for stuff that doesn't have an associated event, though.

src/CommandingOfficers/AWBW/BH/Kindle.java
```diff
@@ -0,0 +1,152 @@
...
+      super("Kindle", UIUtils.SourceGames.AWBW, UIUtils.BH);
+      infoPages.add(new InfoPage(
+            "Kindle (AWBW)\n"
+          + "Units (even air units) gain +40% attack while on urban terrain. HQs, bases, airports, ports, cities, labs, and comtowers count as urban terrain.\n"));
+      infoPages.add(new InfoPage(new UrbanBlight(null, null),
+            "All enemy units lose -3 HP on urban terrain. Urban bonus is increased to +80%.\n"
+          + "+10 attack and defense.\n"));
+      infoPages.add(new InfoPage(new HighSociety(null, null),
+            "Urban bonus is increased to +130%, and attack for all units is increased by +3% for each of your owned urban terrain.\n"
+          + "+10 attack and defense.\n"));
```
ThatIsAUsername, line 41: Your git commit says Kindle's AWBW SCOP has 10 too much urban boost.
AW3 Kindle gets +80, AWBW gets +130. 130-80 != 10. Please explain.

src/CommandingOfficers/AWBW/OS/Andy.java
ThatIsAUsername: Concur that `getEvent` should be preferred. Maybe `perform` will go away altogether eventually.

src/CommandingOfficers/AWBW/BH/Kindle.java
ThislsAUsername: This description is a copy/paste of the AWBW description, for obvious reasons.
AWBW's descriptions (when modifying a stat in a power) say "increased to", i.e. giving the new total for *that* boost.
So AWBW Kindle goes from 40 to 130 urban boost on SCOP,
If she has 3 cities, that's an additional +9 attack anywhere.
Add the generic +10/10, and she has 249 on urban and 119 attack elsewhere.

I otherwise prefer to simply give the relative boosts, with totals (which will include generic stats/etc) being optional.
DS Kindle goes from 40 to 120 urban boost on SCOP (a difference of 80).

130 -120 = 10

ThatIsAUsername: Okay, so just a result of the generic boost difference between the two games. Or no, it looks like you add that separately. Did AWBW just give her an extra 10 power for no reason? I mean, I trust you did your homework, but that seems weird.

ThislsAUsername: They totally did. It really doesn't matter that much, TBH, given the enormous total in either case.

...and I made the mistake of telling them, so they might fix it. Oh, well.

ThatIsAUsername: Oh. Well then. I guess we can go ahead and merge this then.

PR #453, opened 2024-09-06
Add GE COs

ThislsAUsername:
I'm tired of `AIR_LOW | AIR_HIGH`, so I got rid of that.


src/CommandingOfficers/AWBW/GE/Drake.java
```diff
@@ -0,0 +1,164 @@
...
+  private static class Tsunami extends AWBWAbility
+  {
+    private static final long serialVersionUID = 1L;
+    private static final String NAME = "Tsunami";
+    private static final int COST = 4;
+
+    Tsunami(Drake commander, CostBasis cb)
+    {
+      super(commander, NAME, COST, cb);
+      AIFlags = 0; // Less efficient damage than SCOP
```
ThatIsAUsername, line 98: Could be situational. But yeah and AI smart enough to figure that out wouldn't need to rely on the flags.

src/CommandingOfficers/AWBW/GE/Drake.java
```diff
@@ -0,0 +1,164 @@
...
+      return events;
+    }
+    @Override
+    protected void perform(MapMaster map)
+    {
+      super.perform(map);
+      ArrayList<Unit> victims = COutils.findMassDamageTargets(map, myCommander);
+      for( Unit vic : victims )
+      {
+        vic.fuel /= 2; // Does this rate an event?
```
ThatIsAUsername, line 121: Uhh... anti-resupply? Or a generic debuff notification of some kind? Probably not, at least for now.

src/CommandingOfficers/AWBW/GE/Drake.java
ThislsAUsername: 4 for 1 vs 7 for 2.
If you want to use Drake COP, it's probably for fuel or timing reasons.

src/CommandingOfficers/AWBW/GE/Drake.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+  private static class instantiator extends CommanderInfo
+  {
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Drake", UIUtils.SourceGames.AWBW, UIUtils.GE);
+      infoPages.add(new InfoPage(
+            "Drake (AWBW)\n"
+          + "aval units gain +1 movement and +10 defense. Air units lose -30% attack. Unaffected by rain (except vision), and has a higher chance of Rain in random weather.\n"));
```
ThatIsAUsername, line 37: 'N' <-- Here, you dropped this.

ThatIsAUsername: Also your capitalization of "rain" is inconsistent.

src/CommandingOfficers/AW2/GE/Jess.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private static class JessAbility extends AW2Ability
+  {
+    private static final long serialVersionUID = 1L;
+    UnitTypeFilter moveMod, attMod;
+
+    JessAbility(Commander commander, CostBasis cb, String name, int cost, int tankMove, int tankAttack)
+    {
+      super(commander, name, cost, cb);
+      AIFlags = PHASE_TURN_START;
```
ThatIsAUsername, line 90: So if all her abilities inherit from this, she'll just always use the cheaper one? Not meta, I think.

src/CommandingOfficers/AWBW/GE/Drake.java
ThislsAUsername: Ain't my capitalization https://awbw.amarriner.com/co.php

ThatIsAUsername: Well, FINE then.

ThatIsAUsername:
Sweet, looks good

src/CommandingOfficers/AW2/GE/Jess.java
ThislsAUsername: If she has both charged, the AI should SCOP?
Quoth `AIUtils`:
```java
  public static CommanderAbility queueCromulentAbility(Queue<GameAction> q, Army co, int flags)
  {
    CommanderAbility retVal = null;
    ArrayList<CommanderAbility> abilities = co.getReadyAbilities();
    for( CommanderAbility ab : abilities )
    {
      if( flags == (ab.AIFlags & flags) )
      {
        retVal = ab;
      }
    }
    if (null != q && null != retVal)
      q.offer(new GameAction.AbilityAction(retVal));
    return retVal;
  }
```

ThatIsAUsername: Yeah, if they are in order, which they are. That means she lost quite a bit of materiel in one turn though.

I mean, I'm not sure what the better way is to handle that except maybe to disable the COP? Eh, your call.

ThatIsAUsername: If any of the DoR COs have a resupply ability, and it does/doesn't work on material, that would give us precedent.
If not... I mean, none of the other resupply abilities currently do, right? The best argument might be that you said her powers are bad, so maybe she needs the help, but I have trouble breaking consistency for that.

ThislsAUsername: Greyfield resupplies materials. He's the only other CO in the entire series that does resupply.
But also, Jess powers have buffs to move and attack. Greyfield's *entire* effect (outside global zone) is resupply.
But also², Greyfield has 3 zone, so he can charge really quickly (and 120/150 for copters isn't _nothing_).
> I mean, I'm not sure what the better way is to handle that except maybe to disable the COP? Eh, your call.

The "better way" is to have an AI that knows what a move/attack boost on a limited set of units is good for, and whether hitting the button this turn would be efficient/effective.
But that's wildly not in scope.

The cart AI had some conditions for each CO to decide whether it hit buttons, but that is also not in scope and it's not like I actually know what those are.

ThatIsAUsername: There are only two resupply abilities in the whole series? I thought there were more, like Andy or someone, I dunno. If there's only one other, and it works for materials, then sure, why not?
Re: Jess having more stuff for her powers than Greyfield - they are from different games with different designs, and I'm not worried about interpretively balancing them.

ThislsAUsername: Andy just gives HP. This makes sense, since he'd probably give his dudes fuel if he knew what an airport is.

Fair enough, lol
Thanks.

PR #455, opened 2024-09-07
CO Picker UX

ThislsAUsername:
I added the thing we talked about, where swapping COs in the tag picker also moves the main selection.
I also rediscovered why I didn't do it before. That code is definitely spaghetti.

And since I was in there, I also fixed the visual bug you mentioned with the tag COs.
If you go *into* a match, you wouldn't get to have extra COs anyway, but not looking buggy is also good.

### DIFF
--- a/src/UI/UIUtils.java
+++ b/src/UI/UIUtils.java
@@ -219,6 +219,8 @@ public class UIUtils
 
   public static String getCanonicalFactionName(COSpriteSpec spec)
   {
+    if( Color.LIGHT_GRAY.equals(spec.color) )
+      return "MISC";
     return getCanonicalFactionName(getPaletteName(spec.color), spec.faction.name);
   }
   public static String getCanonicalFactionName(String palette, String faction)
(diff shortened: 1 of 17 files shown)


src/UI/UIUtils.java
```diff
@@ -281,7 +284,7 @@ public String toString()
     {
       if( paletteNames.containsKey(color) )
         return getCanonicalFactionName(this);
-      return color.toString() + " " + faction.name;
+      return faction.name;
```
ThatIsAUsername, line 287: Why cut out the color here?

ThatIsAUsername:
Well, I can't say the implementation sparks joy, but I don't think you expected it to, and it works.

And continuing my tradition of complaining about tangentially-related things in PRs:
Weirdly, if I sort by game-then-faction, and move the cursor up from Blue Moon to Orange Star (BW or AW3), the cursor jumps to the end of the line (Hachi) no matter where it was before. It doesn't do that when going down though, and it doesn't do that on any other transitions.

Anyway, looks good.

src/UI/UIUtils.java
ThislsAUsername: To avoid "MISC MISC"

PR #456, opened 2024-09-09
Materialism

ThislsAUsername:
What's an issue if not an inferior PR?


res/unit/icon/material.png
ThislsAUsername, line 1: I think the dynamic/teamcolor icon would work better as a status icon, like how DoR proper uses it to indicate that the unit's "low on material"...
But I don't think an indicator like that makes a lot of sense - it's more clutter on a unit type that already has cargo capacity.
And there's a lot more freedom to make a recognizable icon when you don't give up like 20% of your pixels to a border.

src/Units/UnitModel.java
```diff
@@ -83,18 +83,19 @@ public static String standardizeID(String input)
...
   @Builder.Default public final int fuelBurnPerTile = 1;
   @Builder.Default public final int maxMaterials = 0;
-  @Builder.Default public final boolean needsMaterials = true;
+  @Builder.Default public final boolean needsMaterials = true; // Unit has a finite materials count, which is likely 0.
   public final int visionRange;
-  @Builder.Default public final boolean visionPierces = false;
-  @Builder.Default public final boolean hidden = false;
-  public final Set<TerrainType> healableHabs = new HashSet<TerrainType>();
+  @Builder.Default public final boolean visionPierces = false; // Can see into cover terrain in fog of war.
+  @Builder.Default public final boolean hidden = false; // Invisible even outside of fog, in any terrain; usually on on a variant of a unit type that transforms.
```
ThatIsAUsername, line 89: "on on"

res/unit/icon/material.png
ThatIsAUsername: Yeah, I'm good with having it in the detail pane rather than as a status icon. I like the colored icon version too.

ThatIsAUsername:
> What's an issue if not an inferior PR?

A call to action; a description of a problem; a task assignment. Probably other things too, none of them being a PR, because these are completely separate things.
Does this PR resolve an outstanding issue? Should probably link it if so to close it when this merges.
Anyway, looks good.

src/Units/UnitModel.java
ThislsAUsername: This isn't a typo.
"(Usually on) on a variant of a unit type that transforms."
I could use a less ambiguous word, but then I wouldn't have "on on".

ThislsAUsername:
You mentioned in another PR that we didn't have a materials count in the GUI.
So I made a PR instead of an issue.

src/Units/UnitModel.java
ThatIsAUsername: I... uh. I mean, I'd suggest at least adding a comma after the first "on", but I suspect you'd object because then it wouldn't have "on on". Whatever, it's a comment, and at least intelligible enough as is.

ThatIsAUsername:
> You mentioned in another PR that we didn't have a materials count in the GUI.
So I made a PR instead of an issue.

Touché; I see the logic now.

src/Units/UnitModel.java
ThislsAUsername: I wanted to say it wouldn't grammatically fit, but it actually would.
Fair enough.

PR #457, opened 2024-09-20
Organize ye olde map folder

ThislsAUsername:
image

Guru meditation:
Maybe instead of showing the first map of the folder, it should show the list of maps in that folder? But I am lazy.
Should pushing left and right now go up and down folder levels?
Two maps with the same name in different folders would collide their saved game options. I don't know if that's a feature or a bug.


ThatIsAUsername:
A few observations, in no particular order:
0) This is super cool. Also some of these map designs are crazy. That is not how rivers are supposed to work. I definitely did not envision some of these terrain combinations when building the tileset. Also I don't think landers belong in trees. Or on HQs? Anyway.

1) My map appears to be one pixel off from yours (overlapping the left border by 1 px)? Odd.
image

2) Some of the categories aren't clear (New?). Plus the folder names aren't visually differentiated from map names, which is odd. I don't (currently) object to just displaying the first map in the folder though.

3) We now have multiple copies of some maps (e.g. Blood on my Hands is at the top level and also in "league"

4) Maybe we should capitalize folder names.

5) Spann is in `built-in`, but should probably by in `cart`? Also the built-in maps are all terrible; I'd almost vote to just nuke 'em all, except Test Range is still useful for...testing.

ThislsAUsername:
> This is super cool.

Thanks.
> Also some of these map designs are crazy.

I have a few up my sleeves to bully AIs with at some point later. Do you also want those?
> That is not how rivers are supposed to work. I definitely did not envision some of these terrain combinations when building the tileset.

To your credit, I feel like they still work pretty well.
> Also I don't think landers belong in trees. Or on HQs?

I'm sure Sami would love to hear that, lol

>     1. My map appears to be one pixel off from yours (overlapping the left border by 1 px)? Odd.

It appears to be a thing w.r.t. the specific size of the map vs size of the window.
I believe it's now fixed.
>     2. Some of the categories aren't clear (New?).

It's an AWBW category, one that Jay had in the old map folder.
I just work here, man.
> Plus the folder names aren't visually differentiated from map names, which is odd.

We don't really have anything like bold font...
Drawing with the small-caps font or something could work, but it would look weird.
>     3. We now have multiple copies of some maps (e.g. Blood on my Hands is at the top level and also in "league"

I think that's the only example, but good catch.
>     4. Maybe we should capitalize folder names.

Done.
>     5. Spann is in `built-in`, but should probably by in `cart`? Also the built-in maps are all terrible; I'd almost vote to just nuke 'em all, except Test Range is still useful for...testing.

`built-in` is for the maps that are literally built into the jar and not in the map folder.
I had assumed they were all cart maps, but Firing Range and Cage Match don't appear to be.
I tend to use Firing Range as a short/simple test map, since it has a bit more realistic factory/city ratio than Spann. Maybe a better map would be better, but it probably wouldn't be so short.
If we replaced Firing Range with something that fits my requirements, we could get rid of the built-in category if we move Spann `cart`.
Fair enough.

src/Terrain/MapLibrary.java
```diff
@@ -1,36 +1,58 @@
...
   private static void loadMapInfos()
   {
     availableMaps = new ArrayList<MapInfo>();
     availableMaps.add(TestRange.getMapInfo());
-    availableMaps.add(FiringRange.getMapInfo());
-    availableMaps.add(SpannIsland.getMapInfo());
-    availableMaps.add(CageMatch.getMapInfo());
-    availableMaps.addAll(MapReader.readMapData());
+
+    rootMap = MapReader.readMapData();
```
ThatIsAUsername, line 38: Instead of changing the `MapReader` API and introducing a bunch of loop logic here, why not keep the API the same; let `MapReader` do the directory traversal internally (maybe split into a utility function) and just return the list of maps as before? That way `MapLibrary` is back to just worrying about maps, not where they come from.

ThislsAUsername: Because the `rootMap` is the API used by the GUI, returning two separate results from a function isn't convenient, and I don't really want `MapReader` to have static state.

res/map/AWBWMapConverter.py
```diff
@@ -141,6 +141,8 @@ def countryCodeToPlayerID(x):
 		'pl': 13,
 		'ar': 14,
 		'wn': 15,
+		'aa': 16,
+		'ne': 17,
```
ThatIsAUsername, line 145: No SC?

ThislsAUsername: I think it's newer than my changes here, lol

src/UI/GameOptionSetupController.java
```diff
@@ -75,11 +75,13 @@ public GameOptionSetupController(GameBuilder aGameBuilder)
...
-      final int incomeIndexDefault = 3; // 250 = 0, increment 250
+      final int incomeIndexDefault = 4; // 0 = 0, increment 250
       final int incomeIndexHF = incomeIndexDefault + 4;
-      if( gameBuilder.mapInfo.mapName.startsWith("HF") )
+      if( gameBuilder.mapInfo.dirPath.contains("HF") ||
+          gameBuilder.mapInfo.mapName.startsWith("HF") )
         incomeOption.setSelectedOption(incomeIndexHF);
-      if( gameBuilder.mapInfo.mapName.startsWith("FoW") )
+      if( gameBuilder.mapInfo.dirPath.contains("FoW") ||
+          gameBuilder.mapInfo.mapName.startsWith("FoW") )
```
ThatIsAUsername, line 84: Clever

ThatIsAUsername:
> It's an AWBW category, one that Jay had in the old map folder.
I just work here, man.

Are we beholden to the AWBW categories? I guess it's convenient to people coming from there, which is... most of them, probably.

> Drawing with the small-caps font or something could work, but it would look weird.

Maybe draw the folders with a leading (or trailing) "> " to show that it's an openable category?
We could even get real fancy and put the subfolder map names in the map pane instead of drawing the first map.

> Firing Range

We can keep it around if you want.


Also, do 2020 and Aria of War not belong in any of the existing folders?

res/map/AWBWMapConverter.py
ThatIsAUsername: I didn't realize they were still adding new ones, though I guess AA and NE are also relatively new. SC does look pretty cool, though the tanks should probably be like stegosaurs or something (blah readability blah OK fine).

ThislsAUsername: That reminds me. I should steal the official sprites at some point, too.
I'm thinking I wait until they get the full intended 4.

ThislsAUsername:
> Maybe draw the folders with a leading (or trailing) "> " to show that it's an openable category? We could even get real fancy and put the subfolder map names in the map pane instead of drawing the first map.

I thought about something like this, but `>` isn't a character we can draw and most characters we _can_ draw are valid characters for a map/folder name.
But then I realized there's a stupid simple solution.

> > Firing Range
> 
> We can keep it around if you want.

It doesn't really fit in any category, and having it at the top level feels like a waste of space?

> Also, do 2020 and Aria of War not belong in any of the existing folders?

I don't know.

ThatIsAUsername:
> But then I realized there's a stupid simple solution.

That'll work

> I don't know.

Wait. Where did they come from? How did the other maps get located?

ThatIsAUsername:
Also, we might eventually need to acquire that cool Japan map we saw a while back, and start an Earth Geography folder.
Also Also, I'm pretty sure Spann Teamland belongs under "League".

ThislsAUsername:
> Wait. Where did they come from?

I'm pretty sure 2020 and Aria of War are AWBW maps, but I'm pretty sure you added Aria of War.
I think 2020 *was* league at some point, but I dunno if it still is.
There's like, a map rotation, I think? I don't play AWBW, lol

> How did the other maps get located?

Jay converted and named them with categories (so they'd be alphabetized into categories)

> Also, we might eventually need to acquire that cool Japan map we saw a while back, and start an Earth Geography folder.

Converting maps is very easy. Just pass the map ID in on the command line...
> Also Also, I'm pretty sure Spann Teamland belongs under "League".

Truth be told, I have a local folder called "teletile" that it would fit nicely in...

res/map/AWBWMapConverter.py
ThatIsAUsername: There's one more new one planned after AA, NE, and SC? Cool.

ThatIsAUsername:
I probably did add Aria of War; not sure about 2020. I don't remember where they came from, but it seems like they'd be at home in `League/`, barring any more coherent organizing principle. The current categories seem sensible enough, except maybe for `New`, which is guaranteed to become outdated and confusing.
Also, did you purposely put Test Range above the folders and the other maps below?

Anyway, other than some vague misgivings about how the maps are organized, This all looks good to go.

ThislsAUsername:
> The current categories seem sensible enough, except maybe for `New`, which is guaranteed to become outdated and confusing.

It's already confusing. I blame Jay.
> Also, did you purposely put Test Range above the folders and the other maps below?

Let's say "yes".
> Anyway, other than some vague misgivings about how the maps are organized, This all looks good to go.

kk

ThatIsAUsername:
> It's already confusing. I blame Jay.

Nuh-uh. You are perpetuating the chaos. It is your chaos now.

ThislsAUsername:
Touche.

PR #461, opened 2024-11-16
Fix my dudes

ThislsAUsername:
The Kanbei and Colin here ended up entangled in the rest of the changes because I was working unorganizedly.
This Kanbei gets *fewer* units instead of lower-tech ones, which I feel like fits his lore identity a bit better.
This Colin gets more units instead of cheaper ones, because normal Kanbei and Colin are also opposites. I was also gonna say something about him getting more units making lore sense, but I just realized that doesn't actually fit.

Yeah, I recognize you want COs to have art. I may or may not actually fix that.

Venge is not working as intended on `master` (his marks expire at the wrong time), and also this design seems more amusing to me.

I also have design tweaks for Meridian and Bear&Bull, but I haven't gotten to that (and probably won't for the next version, since it's just design meddling rather than fixing problems).

### DIFF
--- a/src/CommandingOfficers/Commander.java
+++ b/src/CommandingOfficers/Commander.java
@@ -196,7 +196,7 @@ public class Commander implements GameEventListener, Serializable, UnitModifierW
         : new ArrayList<UnitModel>();
   }
 
-  private UnitContext getCostContext(UnitModel um, XYCoord coord)
+  protected UnitContext getCostContext(UnitModel um, XYCoord coord)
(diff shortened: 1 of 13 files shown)


ThatIsAUsername:
If Colin builds a double unit, and it gets destroyed before it moves, it looks like he doesn't get the duplicate.

ThatIsAUsername:
Cool, looks good pending resolution on the Colin thing.

PR #462, opened 2024-11-16
Cursed CO Variants

ThislsAUsername:
Building the Kanbei and Colin variants awoke something terrible in me.

The base concept here is to make "one change" to the base CO's design that radically changes how the CO plays, without compromising the CO's identity.
And yes, I now realize the Colin one (in the other PR) doesn't entirely match. Whoops.
I've been standardizing on AWBW rules for these out of habit.

The commit `VM Sami->AW1 Sami, redesign her penalty` was me being confused.
The original cursed Sami I added was "Mechs cost double and get 2 actions" which is uh... something.
I reworked "lightning Sami" to "VM Sami" and didn't commit it, then I changed her stats to AW1-like since "permanent Victory March Sami" probably doesn't care very much for using Victory March.

And, of course, playtesting and rereading stuff helped me find bugs.


ThislsAUsername:
I kind of want 2 CO variants for each nation, but having a bunch of pending changes and being lazy allied against it for now.

Feel free to ignore the rest of this comment, and I'll just do whatever I was gonna do before.

The count of cursed variants as of this PR is 1 OS, 2 BM, 2 GE, 1 YC, no BH
The YC slot is "Sensei with foot Lightning Strike instead of dude spawns", which I didn't make since YC wasn't merged yet.
The OS slot could either be a second Sami, or Rachel. I don't have any sufficiently-silly-but-simple ideas for the rest.
I have mixed feelings on BH. Excluding them would be fine, and getting "good" cursed designs for them feels hard. The guy who described JakeMan did give me some acceptable ones (and had good suggestions on Rachel), though.

Here are the designs I've been considering:

Cursed Sensei (`ZAP`)
As AW2 (+50 attack foot/copters, -10 attack non-air, +1 move transports)
xxXXXX
Copter Command: Copters +1 move and +20 attack (180 total)
Airborne Assault: COP, and footsoldiers reactivate with 0.5x capture speed.

Cursed Rachel (`MSL`)
On building a missiles unit, queue a COST silo missile to fire at the start of your next turn
xxxXXX
COP: Heal 10 HP to your units on urban terrain.
SCOP: 0-79% luck damage.

Leveling Sami (`LVL`)
Normal Sami, but foot units start bland.
When your foot units kill or finish a capture, they get a stacking boost: 1.2x/0.9x damage dealt/taken, +0.5x capture power
Double Time (2): Footsoldiers get +1 perfect move and 1.2x/0.9x damage dealt/taken

I think the *gameplay experience* of this Sami variant would be a lot more in-line with how "Sami should play" than the VM version — you nurture your elite footsoldier dudes, and they Solve Problems™ for you.
I think VM Sami is more *mechanically* in line than the leveling variant.
Another idea for the level scaling: it goes only to 3, but with +30/10 per level and +1 move at vet.
I would definitely have to playtest to feel out how it should work, but I don't really want to put in the effort if we want Rachel instead.

Rize's BH suggestions:
Adder: +1 move, -10% *infantry* capture power (as in, mechs have 3 move and capture normally). No powers
VB: 125/125 baseline, units level up to 3 with kills, -15/-15 per level. No powers

src/CommandingOfficers/DefendPeace/OS/Sami.java
```diff
@@ -0,0 +1,162 @@
...
+  {
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Sami", UIUtils.SourceGames.DEFEND_PEACE, UIUtils.OS, "VM");
+      infoPages.add(new InfoPage(
+            "Sami (victory)\n"
+          + "AW1 Sami, but with instant captures and AWBW rules.\n"
+          + "Footsoldiers +20/10 stats.\n"
+          + "20x capture rate, lose 5 HP on capture.\n"
```
ThatIsAUsername, line 47: ridonculous

src/CommandingOfficers/DefendPeace/OS/Sami.java
```diff
@@ -0,0 +1,162 @@
...
+          + "Cities do not give income if there is any unit (friendly or not) on them.\n"
```
ThatIsAUsername, line 48: How is this Sami themed?

src/CommandingOfficers/DefendPeace/OS/Sami.java
```diff
@@ -0,0 +1,162 @@
...
+          + "Unarmed transports +1 move. -10/0 direct vehicle combat.\n"));
```
ThatIsAUsername, line 49: This feels like a... _strong_ passive

src/CommandingOfficers/DefendPeace/OS/Sami.java
ThislsAUsername: It is "make this design not T -1" themed.
It also makes her D2D somewhat symmetric, since she loses the city income as soon as you start capturing.

But yeah, it probably could be better.

src/CommandingOfficers/DefendPeace/OS/Sami.java
ThislsAUsername: It ain't +1 capacity...

ThatIsAUsername:
Looks fine.

PR #463, opened 2024-11-16
Fix a null dereference in the color/faction GUI on hovering MISC (21 line diff)

ThislsAUsername:


ThatIsAUsername:
Looks like this fixes #468. Cool.

PR #465, opened 2024-11-23
Fix the missing units in HF maps (the unit section header was missing)

ThislsAUsername:
The FTA counter is a lot stronger when you start with *two* more inf than your opponent...


ThatIsAUsername:
Life is fair and balanced once again.

PR #467, opened 2024-11-25
Make the AIs not try to break the unit cap (constraint test+minimal fixes)

ThislsAUsername:
I think I probably will want to add deeper logic to Wally at some point, but it ain't gonna be soon.

### DIFF
--- a/src/Test/TestAIConstraints.java
+++ b/src/Test/TestAIConstraints.java
@@ -89,10 +89,11 @@ public class TestAIConstraints extends TestCase
 
   private boolean testBuildTooManyMans(AIMaker ai)
   {
-    setupTest(TestRange.getMapInfo(), ai, new GameScenario(new AWBWUnits(), 1000, 1000, 1, FogMode.OFF_DOR, TagMode.OFF));
+    final int unitCap = 1;
+    setupTest(TestRange.getMapInfo(), ai, new GameScenario(new AWBWUnits(), 1000, 1000, unitCap, FogMode.OFF_DOR, TagMode.OFF));
     final Army armyOne = testGame.armies[0];
     armyOne.team = 9;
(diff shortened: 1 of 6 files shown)


src/Test/TestAIConstraints.java
```diff
@@ -82,4 +86,30 @@ private boolean testUnmovedFriend(AIMaker ai)
 
     return testPassed;
   }
+
+  private boolean testBuildTooManyMans(AIMaker ai)
+  {
+    setupTest(TestRange.getMapInfo(), ai, new GameScenario(new AWBWUnits(), 1000, 1000, 1, FogMode.OFF_DOR, TagMode.OFF));
```
ThatIsAUsername, line 92: Probably make a temp variable for the altered unit cap, since that's the point of the test.

src/Test/TestAIConstraints.java
```diff
@@ -82,4 +86,30 @@ private boolean testUnmovedFriend(AIMaker ai)
...
+    final Army armyOne = testGame.armies[0];
+    armyOne.team = 9;
+    testGame.armies[1].team = armyOne.team;
```
ThatIsAUsername, line 95: Why do they have to be friends?

src/Test/TestAIConstraints.java
```diff
@@ -82,4 +86,30 @@ private boolean testUnmovedFriend(AIMaker ai)
...
+    // Run through a round of init-turns so that our allies are ready-to-act
+    day(testGame);
```
ThatIsAUsername, line 97: Is this test acting on allies at all? It looks like it just lets `armyOne` act until it's done with its first turn.

ThatIsAUsername:
Looks good, pending comments.

PR #472, opened 2025-08-13
New version, new features

ThislsAUsername:


src/CommandingOfficers/AWBW/OS/NellWarCrimes.java
```diff
@@ -0,0 +1,85 @@
...
+  private static class instantiator extends CommanderInfo
+  {
+    private static final long serialVersionUID = 1L;
+    public instantiator()
+    {
+      super("Nell", UIUtils.SourceGames.AWBW, UIUtils.OS, "crime");
+      infoPages.add(new InfoPage(
+            "Nell (warcrimes)\n"
+          + "Nell, but she always rolls max luck.\n"
+          + "Luck on attacks is +19%.\n"));
```
ThatIsAUsername, line 33: What kind of broken garbage is this? Dibs.

src/Engine/UnitMods/LuckBaseModifier.java
```diff
@@ -0,0 +1,20 @@
+package Engine.UnitMods;
+
+import Engine.Combat.StrikeParams;
+
+public class LuckBaseModifier implements UnitModifierWithDefaults
```
ThatIsAUsername, line 5: Could use a docstring to explain intent. If I'm understanding correctly, it coerces the luck value in all combat to the provided luck value. If so, calling it a "Base" modifier seems misleading, as that might just imply changing the luck floor.

res/map/AWBWMapConverter.py
```diff
@@ -143,6 +147,8 @@ def countryCodeToPlayerID(x):
 		'wn': 15,
 		'aa': 16,
 		'ne': 17,
+		'sc': 18,
+		'uw': 19,
```
ThatIsAUsername, line 151: All right, so now we've got Alien (but dinosaurs, and from underground) and Planet of the Apes-slash-W40K orks. Should we also add their sprites if we are supporting them in map data?
Also if AWBW only goes up to 16 players, why do they have 20 different placeable civs?

src/CommandingOfficers/AWBW/OS/NellWarCrimes.java
ThislsAUsername: It was requested for testing against such paragons of balance as Javier on a 2T map or Hachi in high funds.

src/Engine/UnitMods/LuckBaseModifier.java
ThislsAUsername: Nope. The "base" luck is the "zero point" that the good luck and bad luck vary from.

e.g. AW1 Sonja has single-roll -15 to +9, so she has 25 good luck and -15 base luck.

res/map/AWBWMapConverter.py
ThislsAUsername: They've also changed the soldiers of PC and I think some of the other countries.

I'm not too worried about the sprites since there's nothing in the game that respects which country goes in which slot, and I'm a very lazy boi.

As for faction count, I guess they just wanted more silly options.

src/Engine/UnitMods/LuckBaseModifier.java
ThatIsAUsername: I recall you saying something about two rolls for luck - was that in other games?

Makes sense then. Raise the floor, set variation to zero == always get max luck. Still think a docstring would be helpful.

Dumb question just because I don't remember, but do we need to verify which unit is attacking in `modifyUnitAttack` to make sure we don't also buff enemies? I remember needing to check sides for at least some of the combat modifiers.

ThatIsAUsername:
Fire when ready.

src/Engine/UnitMods/LuckBaseModifier.java
ThislsAUsername: Yup. AW1 is the only game that has a single RN for bad and good luck.

> Still think a docstring would be helpful.

I am mildly resistant to the idea because it's a class that **just** modifies a variable (which is in the name of the class), and that variable is documented... so it ends up being kind of duplicate info.

> do we need to verify which unit is attacking in `modifyUnitAttack` to make sure we don't also buff enemies

Nope. That's only for the context hook, since that lets you swap the combatants around. Otherwise, you know which combatant is "yours" based on the hook being called.

PR #473, opened 2026-04-03
Add some deranged cart accuracy

ThislsAUsername:
I wish I lived in a world where I didn't know this.
AW2
image image

AW1
image image

AWDS
image image

### DIFF
--- a/src/Test/TestColinMath.java
+++ b/src/Test/TestColinMath.java
@@ -25,11 +25,43 @@ public class TestColinMath extends TestCase
   @Override
   public boolean runTest()
   {
-
     boolean testPassed = true;
 
+    testPassed &= validate(testCombatOOO(), "  Order of Operations test failed!");
     testPassed &= validate(testPowerOfMoney() , "  Power of Money test failed!");
 
+    return testPassed;
+  }
+
+  private boolean testCombatOOO() // Order of Operations
+  {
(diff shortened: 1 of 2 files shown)


src/Test/TestColinMath.java
```diff
@@ -25,14 +25,46 @@ public class TestColinMath extends TestCase
...
+    UnitContext arty = new UnitContext(colin1, colin1.getUnitModel(UnitModel.INDIRECT, false));
+    arty.terrainStars = 1;
+
+    BattleParams shot = new CombatContext(game, map, tank, arty, 1, CalcType.NO_LUCK).applyModifiers().getAttack();
+
+    int hit = shot.calculateDamage();
+    testPassed &= validate(hit == 44,     "    Colin 8 HP tank damage vs plains arty is wrong");
+
+    game.endGame();
+    return testPassed;
```
ThatIsAUsername, line 65: So this checks that the AW2 logic results in a damage of 44%; do we also need to check that other game calcs still result in a damage of 45%?
The answer, of course, is do what you want.

ThatIsAUsername:
Cursed indeed. Why did you have to learn this? Change looks reasonable, but I did not attempt to test it, so go ahead and merge if you're happy with it. Actually it's pretty clear you aren't happy, so I guess just merge if you think the change is good to go.

PR #474, opened 2026-04-27
Audio

ThislsAUsername:
re: library choice\
The choices available were basically jOGG, jOrbis, or some wrapper around one of 'em.\
I went for a wrapper that allows me to use the normal Java APIs instead of having to directly wrangle the file data myself.

I figured I should make the hooks into audio stuff be relatively minimal, since it's an optional subsystem.\
As a result, the thing basically just plays menu music until a game starts, then picks up the current CO from the game state.\
This isn't thread-safe, but I really don't foresee serious problems from race conditions.

I'm not sure it was a great idea to commit all the audio. It's fairly hefty.

In general, I'm happier with this than I expected I'd be.\
I probably still should add the GUI for selecting your CO's themes, or I could punt on that until people bug me to add it.

I'm definitely not adding sound effects yet. That seems gnarly.


src/UI/AudioEngine.java
```diff
@@ -0,0 +1,273 @@
...
+    public void join()
+    {
+      killThread = true;
+      try
+      {
+        myThread.join();
+      }
+      catch(InterruptedException ie)
+      {
+        System.out.println("[InfantrySpritePreloader] Exception! Details:\n  " + ie.toString());
```
ThatIsAUsername: Clearly should be Infantry**Sound**Preloader

src/UI/AudioUtils.java
```diff
@@ -0,0 +1,192 @@
...
+      return allThemes.get(UNKNOWN_CO_FORMAT);
+
+    // We don't have it, so we need to load it.
+    allThemes.put(themeFormat, new LoopMusic(themeFormat));
+    return allThemes.get(themeFormat);
+  }
+
+  ///////////////////////////////////////////////////////////////////
+  //  Code for loading music
+  //  Note: as these paths are stored in the savefile, the paths should be relative to the game directory and not absolute.
```
ThatIsAUsername, line 59: The audio file paths are included in the save file? That seems... unnecessary.

ThatIsAUsername:
> This isn't thread-safe, but I really don't foresee serious problems from race conditions.

IIRC, the only threads we explicitly create are for loading infantry sprites, so I'm not worried about racing.
Ah, I see you added one for audio. I'm inclined to agree that it's fine until proven otherwise.

> I'm not sure it was a great idea to commit all the audio. It's fairly hefty.

Saaaame, but yeah, lossless will do that for you. OTOH, What's half a gig in this day and age?

> I'm definitely not adding sound effects yet. That seems gnarly.

Yet you added the audio files, which might invite questions (If files there, why no hear?)

Also - I'm not hearing audio on Linux. I tried cycling through the available options to no avail.
Not sure if this will be any help:
```
Port NVidia [hw:0] is NOT supported
Port Unknown Name is NOT supported
default [default] is supported
NVidia [plughw:0,3] is supported
NVidia [plughw:0,7] is supported
NVidia [plughw:0,8] is supported
NVidia [plughw:0,9] is supported
sofhdadsp [plughw:2,0] is NOT supported
sofhdadsp [plughw:2,1] is NOT supported
sofhdadsp [plughw:2,3] is supported
sofhdadsp [plughw:2,4] is supported
sofhdadsp [plughw:2,5] is supported
sofhdadsp [plughw:2,6] is NOT supported
sofhdadsp [plughw:2,7] is NOT supported
```
If I get a chance (maybe in a few days) I can try to triage.

src/UI/AudioUtils.java
ThislsAUsername: My idea was to have a single fixed default for any CO that doesn't have a specific theme, but let people pick their CO theme(s) during a game.

This can help the issue of themes being super repetitive over a long game.

I then skipped making that, because UI is hard and I am lazy.

ThislsAUsername:
> Saaaame, but yeah, lossless will do that for you. OTOH, What's half a gig in this day and age?

We do squash, so taking a subset would still save us repo size in the future.

> Yet you added the audio files, which might invite questions (If files there, why no hear?)

Fair. I could purge those and all the RBC music, at least.\
The AW1 music isn't all there either, so I could just say "we're only committing AW2 music" and really shrink things.

> Also - I'm not hearing audio on Linux. I tried cycling through the available options to no avail.

Curious. I actually haven't tested in Windows at all, though I sent a build to someone who has reported the sound working but being "funky".\
Note that you have to *select* a device for the sound thread to be restarted pointing at that device. It takes like half a second to stop and spin back up.

ThatIsAUsername:
I take it back. It turns out that  42 is apparently too low. The volume falloff from 100 is... pretty rapid. Even at 80 it's barely audible unless I crank my volume quite a bit.

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+  public static GameOption<?>[] allOptions = { soundDeviceOption, volumeOption };
+  public static OptionSelector highlightedOption = new OptionSelector(allOptions.length);
+  public static SlidingValue animHighlightedOption = new SlidingValue(0);
+
+  private static SoundThread soundThread;
+  public static void initialize()
+  {
+    soundDeviceOption.optionList.clear();
+    for (var opt : DEFAULT_SOUND_DEVICES)
+      soundDeviceOption.optionList.add(opt);
```
ThatIsAUsername, line 59: Is this needed? I have three working options in my audio list:
- default
- default default
- sofhdadsp [plughw:2,0]  (This one oddly doesn't always show up)
Looks like the first one is a placebo and maybe selecting it only works because "default default" happens to contain "default", so the check on line 244 passes.

ThatIsAUsername: Oh, and empirically the answer is "No", at least on my machine.

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+      String name = m.getName();
+      //System.out.println("Got " + name);
+      if (! mix.isLineSupported(info))
+        continue;
+      soundDeviceOption.optionList.add(name);
+    }
+    soundDeviceOption.reset(soundDeviceOption.optionList.size()); // Allow selecting the new options
+
+    // Load saved settings from disk, if they exist.
+    loadSettingsFromDisk();
```
ThatIsAUsername, line 87: In the wild world of audio mixer availability, is there any chance that the saved mixer will no longer exist?
The other options we have aren't dynamically assembled, so that wasn't really a concern.

ThatIsAUsername: I mostly wonder because of the sometimes-present `sofhdadsp [plughw:2,0]` option (though I haven't gotten anything bad to happen by selecting it and restarting).

Even more weirdly, when it is there, the first time I select it I sometimes see this error (but if I select something else then come back to it, it'll work):
```java
javax.sound.sampled.LineUnavailableException: line with format PCM_SIGNED 44100.0 Hz, 16 bit, stereo, 4 bytes/frame, little-endian not supported.
	at java.desktop/com.sun.media.sound.DirectAudioDevice$DirectDL.implOpen(DirectAudioDevice.java:484)
	at java.desktop/com.sun.media.sound.AbstractDataLine.open(AbstractDataLine.java:115)
	at java.desktop/com.sun.media.sound.AbstractDataLine.open(AbstractDataLine.java:357)
	at UI.AudioEngine$SoundThread.run(AudioEngine.java:259)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+          if( !m.getName().contains(mixerName) )
+            continue;
+          var mixTemp = AudioSystem.getMixer(m);
+          if( !mixTemp.isLineSupported(info) )
+            continue;
+          mix = mixTemp;
+          break;
+        }
+        if( null == mix )
+          return;
```
ThatIsAUsername, line 275: Is this a pun? If we fail to make noise, we fail silently?

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+        SourceDataLine line = (SourceDataLine) mix.getLine(info);
+
+        if( null == line )
+          return;
+        line.open();
+        FloatControl volumeKnob = null;
+        if (line.isControlSupported(FloatControl.Type.VOLUME))
+          volumeKnob = (FloatControl) line.getControl(FloatControl.Type.VOLUME);
+        if (line.isControlSupported(FloatControl.Type.MASTER_GAIN))
+          volumeKnob = (FloatControl) line.getControl(FloatControl.Type.MASTER_GAIN);
```
ThatIsAUsername, line 285: Should this be an else if (and I guess flip the order to still prioritize `MASTER_GAIN`)?

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+
+  //////////////////////////////////////////////////////////////////////
+  //  Actual sound control.
+  //////////////////////////////////////////////////////////////////////
+  public static class SoundThread implements Runnable
+  {
+    private Thread myThread;
+    private String mixerName;
+    private boolean killThread = false;
+    private static enum LoopState { INTRO, PRELOOP, LOOP };
```
ThatIsAUsername, line 193: If it plays intro->preloop->loop[->loop], why are intro and preloop distinct?

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+        FloatControl volumeKnob = null;
+        if (line.isControlSupported(FloatControl.Type.VOLUME))
+          volumeKnob = (FloatControl) line.getControl(FloatControl.Type.VOLUME);
+        if (line.isControlSupported(FloatControl.Type.MASTER_GAIN))
+          volumeKnob = (FloatControl) line.getControl(FloatControl.Type.MASTER_GAIN);
+
+        line.start();
+        @SuppressWarnings("unused")
+        int nBytesRead = 0, nBytesWritten = 0;
+        int failedReadCombo = 0; // The stream reader API arbitrarily returns -1 even when the song isn't over, so I guess we just play the odds here.
```
ThatIsAUsername, line 292: Sure does. Constantly. Why? Is there not a better way to check for the end of the audio file? I'm... somewhat in disbelief that this is represents the state of audio playback in Java.
On the other hand, it's working.

src/UI/AudioEngine.java
```diff
@@ -0,0 +1,338 @@
...
+        while (!killThread)
+        {
+          LoopMusic newLoop = null;
+          if( null == activeGame || null == activeGame.activeArmy )
+            newLoop = AudioUtils.loadMenuTheme();
```
ThatIsAUsername: This line is getting hit CONSTANTLY. And I guess the next one down is once we get into a game. Feels kinda wonky to be constantly re-creating `newLoop` when nothing has changed.

ThatIsAUsername: On closer inspection, it's not actually reloading anything after the first time. I might make them getters instead of loaders for clarity? Maybe. Otherwise this seems fine.

res/audio/t-co-select.ogg
ThatIsAUsername, line 1: Why do so many (but not all) tracks have a "t-" prefix? Is it just a t for track?

src/UI/AudioUtils.java
```diff
@@ -0,0 +1,192 @@
...
+  private static HashMap<String, LoopMusic> allThemes = new HashMap<>(); // Map from music format string to the audio streams
+  private static HashMap<UIUtils.SourceGames, String> sourceGameToCategory;
+  static
+  {
+    sourceGameToCategory = new HashMap<>();
+    sourceGameToCategory.put(UIUtils.SourceGames.AW1, "aw1/");
+    sourceGameToCategory.put(UIUtils.SourceGames.AW2, "aw2/");
+    sourceGameToCategory.put(UIUtils.SourceGames.AW3, "aw-ds/");
+    sourceGameToCategory.put(UIUtils.SourceGames.AW4, "aw-dor/");
+    sourceGameToCategory.put(UIUtils.SourceGames.AWBW, "aw-bw/");
```
ThatIsAUsername, line 30: These two folders don't exist, but "aw-rbc" does.

src/UI/AudioUtils.java
```diff
@@ -0,0 +1,192 @@
...
+      if( canRead(ntt) )
+      {
+        passiveTheme = ntt;
+        break;
+      }
+    co.themePathFormats.put(null, passiveTheme);
+
+    String allyOrBH = "ally";
+    if( whichCO.baseFaction.faction.name.equalsIgnoreCase(UIUtils.BH.faction.name) )
+      allyOrBH = "bh";
```
ThatIsAUsername: Why special ally-vs-bh-handling? I know I say this  lot, but comments can be really helpful for explaining the intent behind special handling or other non-obvious logic.

src/UI/AudioEngine.java
ThislsAUsername: I suppose that depends on how one defines "needed".

"default" as the first option guarantees that if you don't touch the setting, it will always(?) use whatever sound device is the selected default.\
I have no idea what sort of guarantees (if any) exist about the order of the values in `AudioSystem.getMixerInfo()`, so I wanted to hedge my bets.

"None" is a hardcoded option that entirely disables the sound thread, and that also should be at a consistent index.

I guess this belongs in a comment.

src/UI/AudioEngine.java
ThislsAUsername: > In the wild world of audio mixer availability, is there any chance that the saved mixer will no longer exist? The other options we have aren't dynamically assembled, so that wasn't really a concern.

The chance seems pretty high. Mixers appear to sometimes(?) map to hardware devices (for example, one of my mixers plays sound from my second monitor).\
I thought about trying to save the name directly, but I dunno what to do in the case where the selected option is gone.\
Saving the int index is not an ideal solution, but it required no thought and arguably even less code, so it was really easy to default to.
> though I haven't gotten anything bad to happen by selecting it and restarting

I wouldn't expect selecting a missing option to make anything weird happen, since the index would loop back around to "default" if the list was "too short".
> I sometimes see this error (but if I select something else then come back to it, it'll work)

I'm gonna guess it's a timing/init issue?\
I'll say I've resisted adding audio partly because interacting with hardware is a great way to start writing once and debugging everywhere :P

src/UI/AudioEngine.java
ThislsAUsername: Would you rather it crash?

I guess printing something would be reasonable.

src/UI/AudioEngine.java
ThislsAUsername: I dunno, man. I've never seen one that supports both.

src/UI/AudioEngine.java
ThislsAUsername: The AWBW music player (which I stole the audio from) has an option to skip intros.

Preloop is just "the first loop"

src/UI/AudioEngine.java
ThislsAUsername: I dunno either.

I guess it's possible there's some jank with the library that makes this happen?

res/audio/t-co-select.ogg
ThislsAUsername: I assumed it meant "theme", as in "theme (for/of) Andy"

The real answer is I didn't feel like changing nomenclature in case that project gets more stuff set up that I can steal.

src/UI/AudioUtils.java
ThislsAUsername: I figured I should have some path defined in case people wanted to get themes set up for those :P

src/UI/AudioEngine.java
ThatIsAUsername: No, I mean: "default" as the default option only works if there is a mixer with "default" in the name (due to the check on line 244 for `if( !m.getName().contains(mixerName) )`).
Is it guaranteed that we will have a "*default*" mixer? If it is, the more correct thing to do is probably to find that mixer and select it if one isn't already selected. As it is, a system with five mixers will have six selectable options here; the five real ones plus "default".

src/UI/AudioEngine.java
ThatIsAUsername: Then an else if would still be more efficient, no? Gotta save those compute cycles in our highly performant game code.

src/UI/AudioEngine.java
ThatIsAUsername: I mean that doesn't answer my question. If preloop and loop are identical, why have both?
I'm not complaining, I'm just confused.

src/UI/AudioEngine.java
ThislsAUsername: I had assumed that the system default would always have "default" in the name. I should get around to verifying that on Windows...
> If it is, the more correct thing to do is probably to find that mixer and select it if one isn't already selected.

I feel like that's not obvious? I thought the point of having a device selection is to let you make a persistent choice on which device to select.\
I guess it'd be reasonable to just use the default and not have an option at all. In my head, that translates to a greater expectation that the game will detect when you change the OS default, and I have no idea how to do that other than polling.

If we're to have a selection beyond "on/off", though, I definitely think a "just use the current default" option is sensible because it will follow the OS selection (on boot) without needing to change anything.

If I change my in-game volume, close the game, then switch my default device in the OS, I feel like I would expect the game to use the new default.\
With your suggestion, it would use the old default unless we implement storing the string name (and even then, it would only mismatch and use the new default because "default" is now missing from the old name... unless of course the "real" name of it includes "default")

If I changed my in-game device to a different one (even explicitly selecting the current default), I would be surprised and likely annoyed if the game switched to using the (new, current) default after I restart it.

src/UI/AudioEngine.java
ThislsAUsername: They're not identical. RBC Jess has both: the intro is the bit that doesn't sound like the loop, and the preloop is the part that sounds similar to but not exactly the same as the loop.

The distinction is pretty semantic if we don't have a "skip intros" option, but I guess I could add one.

src/UI/AudioEngine.java
ThatIsAUsername: > If we're to have a selection beyond "on/off", though, I definitely think a "just use the current default" option is sensible because it will follow the OS selection (on boot) without needing to change anything.

Ah, I see. I might then make that option more obviously "special" - it doesn't map to a specific mixer, so I'd maybe display it as "Auto (<name of the discovered default mixer>)". Then whatever is in parens might change if the system mixer changes, but the user can still just go pick that mixer or a different one directly.
This also allows for "Auto (not found)" in case that's something we need to worry about.

src/UI/AudioUtils.java
```diff
@@ -0,0 +1,192 @@
...
+    public final String path;
+    public final BufferedInputStream intro; // non-loop intro
+    public final BufferedInputStream preloop; // first loop
+    public final BufferedInputStream loop;
+
+    public LoopMusic(String nameFormat)
+    {
+      path    = Engine.Driver.JAR_DIR + String.format(nameFormat, "");
+      var tmp = loadAudioStream(path);
+      af      = tmp.getFormat();
```
ThatIsAUsername: Given that `loadAudioStream()` can explicitly return `null`, should we check for that here?

src/UI/AudioEngine.java
ThatIsAUsername: All right, well now I have both `Auto (default)` and `default default` mixer options, and I wasn't able to get the Auto option to change by messing with system settings, but it seems to work anyway.

ThatIsAUsername:
All right, I think I'm out of criticisms, so if you're happy with things, feel free to merge.

PR #475, opened 2026-06-07
Drop static test members since they aren't needed

ThislsAUsername:
I like it when the garbage collector is allowed to collect garbage.


ThatIsAUsername:
Huh. Not sure why those were static to begin with. Are you running memory profilers now?
Anyway, looks good to me.

ThislsAUsername:
Nah, I just noticed it while I was thinking about memory profile for reachability caching.

PR #476, opened 2026-06-16
Turn-iterated pathing

ThislsAUsername:
While poking at other things, I ended up writing this as part of figuring out what I actually wanted to do.

There's a performance penalty and extra complexity, and the new information isn't *that* useful, so I'm kinda iffy on the change.

The one objective advantage is that now multi-turn pathing should properly account for "lost" movepoints (e.g. if my inf is 3 tiles away from a mountain, he can't get halfway up it. He's stuck using only 2/3 movepoints).

The first 4 commits are gonna happen either way, but I figured I'd isolate this bit from my bigger feature.


src/Test/TestPathfindingPerf.java
```diff
@@ -0,0 +1,83 @@
...
+
+    turn(testGame);
+  }
+
+  @Override
+  public boolean runTest()
+  {
+    boolean testPassed = true;
+    if( testPassed )
+      return testPassed;
```
ThatIsAUsername, line 37: Hm. Good test. Run only as needed, eh?

src/Engine/PathCalcParams.java
```diff
@@ -170,17 +198,29 @@ private void expandSearchNode(SearchNode currentNode, Queue<SearchNode> searchQu
...
       // If we can move more cheaply than previously discovered,
       // then update the power grid and re-queue the next node.
-      int oldPower = powerGrid[currentNode.x][currentNode.y];
-      int oldNextPower = powerGrid[next.x][next.y];
+      if( decodeTurns(powerGrid[next.x][next.y]) < decodeTurns(powerGrid[currentNode.x][currentNode.y]) )
+        continue; // Is from a previous turn
+
       final int transitionCost = mt.getTransitionCost(map, currentNode, next, team, canTravelThroughEnemies);
+      if( transitionCost >= MoveType.IMPASSABLE || transitionCost >= initialMovePower )
+        continue; // We cannot enter this tile even in principle.
```
ThatIsAUsername: Should the second part just be `transitionCost > initialMovePower`? I would think if they are equal, then it is in principle possible to move there.

src/Engine/PathCalcParams.java
```diff
@@ -40,16 +40,44 @@ public PathCalcParams(UnitContext uc, GameMap gameMap)
...
+  private static int decodeMovePower(int overallPower)
+  {
+    if( (0x8000 & overallPower) != 0 ) // The new sign bit matches
+      return -1;
+    return 0xFFFF & overallPower;
+  }
+  private int decodeTurns(int overallPower)
+  {
+    return -1 * ((overallPower >> 16) - maxTurns);
+  }
```
ThatIsAUsername: I see your vendetta against commented code continues.
Clever, yes. Obvious, less so. Necessary? Debatable.
I guess it keeps the `SearchNodeComparator.compare` implementation streamlined.

src/Engine/PathCalcParams.java
```diff
@@ -170,17 +198,29 @@ private void expandSearchNode(SearchNode currentNode, Queue<SearchNode> searchQu
     {
       // If we can move more cheaply than previously discovered,
       // then update the power grid and re-queue the next node.
-      int oldPower = powerGrid[currentNode.x][currentNode.y];
-      int oldNextPower = powerGrid[next.x][next.y];
+      if( decodeTurns(powerGrid[next.x][next.y]) < decodeTurns(powerGrid[currentNode.x][currentNode.y]) )
+        continue; // Is from a previous turn
```
ThatIsAUsername: Should this skip in the place of equality too? If the best I can do from here is no better than the previously-discovered route, why bother?

src/Engine/PathCalcParams.java
```diff
@@ -194,12 +234,24 @@ private void expandSearchNodeWithParents(SearchNode currentNode, Queue<SearchNod
...
+
+      int oldPower     = decodeMovePower(powerGrid[currentNode.x][currentNode.y]);
+      int oldNextPower = decodeMovePower(powerGrid[next.x][next.y]);
       int newNextPower = oldPower - transitionCost;
+      int newTurns     = currentNode.turn;
+      if( newNextPower < 0 && currentNode.turn < maxTurns )
+      {
+        newNextPower = initialMovePower - transitionCost;
+        newTurns    += 1;
+      }
```
ThatIsAUsername, line 236: There's quite a bit of duplicated code between `expandSearchNode` and `expandSearchNodeWithParents`. Though I guess that's pre-existing, and fixing that is out of scope here.

src/Engine/PathCalcParams.java
ThatIsAUsername: Okay, now I see the commit message about performance. That's more significant than I would have expected, but I guess that's why we profile. Good stuff.

ThatIsAUsername:
I can go either way. On the one hand, the "lost" movement points are unlikely to be super important, and nobody likes losing performance.
On the other hand, it's nominal, and individual pathing operations are still going to be quick. Plus the cases where the "lost" movements matter (e.g., cap phase), could have an outsized importance on some maps (or so I imagine).

I don't have any practical examples to consider to help decide, and I think I naturally lean toward more correctness.
Your call.

src/Engine/PathCalcParams.java
ThislsAUsername: Goooood catch.

src/Engine/PathCalcParams.java
ThislsAUsername: Hey now, there are clearly comments in that code.
I sort of assumed "pack two values into one int" would be self-explanatory, though maybe I assume too much.

Either way, I didn't wanna spend a bunch of time commenting it if I was gonna throw it away.

src/Engine/PathCalcParams.java
ThislsAUsername: If it's equal, then we are comparing the values within the same turn.
Let's say I have costs like:
```
1 1 1
1 7 1
1 x 1
```
I will reach the top middle tile from the middle tile first, but I definitely want to update the total movecost down to 4 later.

src/Engine/PathCalcParams.java
ThislsAUsername: It pains me too, but I dunno how to fix it.
If I had gotten inspiration, it may have magically become in-scope.

src/Engine/PathCalcParams.java
```diff
@@ -170,17 +182,29 @@ private void expandSearchNode(SearchNode currentNode, Queue<SearchNode> searchQu
     {
       // If we can move more cheaply than previously discovered,
       // then update the power grid and re-queue the next node.
-      int oldPower = powerGrid[currentNode.x][currentNode.y];
-      int oldNextPower = powerGrid[next.x][next.y];
+      if( powerGrid[next.x][next.y] >= powerGrid[currentNode.x][currentNode.y] )
+        continue; // Movecosts cannot be negative, so our next power cannot be higher than the current one if it is equal or greater already.
+
       final int transitionCost = mt.getTransitionCost(map, currentNode, next, team, canTravelThroughEnemies);
+      if( transitionCost >= MoveType.IMPASSABLE )
```
ThatIsAUsername, line 189: The `|| transitionCost > initialMovePower` check "annoys your new code"? What does that even mean?

ThislsAUsername: I put in a movepower of 1 because I'm an idiot, and then attributed my brainfart to the code not working together as expected (which it wasn't, because my expectation was wrong).

It doesn't really affect runtime or correctness, so it's whatever.

ThatIsAUsername:
Looks good from here.

PR #478, opened 2026-07-13
Make ISAI consistently prefer attacking to capturing

ThislsAUsername:
Deciding arbitrarily is silly


ThatIsAUsername:
Fine with me.

PR #479, opened 2026-07-19
AI Shipping

ThislsAUsername:
Herein lies the scaffolding for basic transport logic.
The utils' API was hard to figure out. Dunno how I feel about it.
In particular, passing `SearchNode`s around feels weird, but I don't wanna lose the info they have. On the other hand, the info could be misleading if you don't use the right assumptions? I dunno.
The function divisions seemed reasonable enough at the time, but they feel kinda tied at the hip and that's not encouraging. Maybe caching unit goals would help, but we've been avoiding AIs storing stuff between turns.

Despite being divided up, they'd be a little annoying to test. That, combined with my lack of certainty on where divisions should go, kinda killed any desire I might have had to write tests.
Also, I just kinda wanted to see dudes move and then make a PR to limit scope.

I added a new thing to `PathCalcParams`: you can start from a cloud of tiles instead of just 1.
I considered making the goals a list, but I think this is as intuitive and should be more performant (it changes setup instead of adding an O(n) to check each potential end goal.
This is another thing I should probably test more, especially since I'm thinking it may expand to take over the iterative search for the load intercept.

`ReachabilityCache` seems like a thing that could have made some AI things easier, but there are some weird implementation details that make using it kind of annoying (as you'll see).
I sorta want to make it live in a `StateTracker` so it can recalculate, but only AIs use it and AIs don't currently add `StateTrackers` (and arguably should not, since those do count as game state?)
It ended up being much cheaper to build than I expected, so maybe the moral of the story here is that the AI can just recalculate it every turn like I have in here? I dunno.

You will note this was added to ISAI.
I wanted something really dumb and minimal because the utils had to be so large to enable it.
I have some comments about the sort of stuff that might need to go in scaffolding.

Is commented a bit more than usual since it's util code and such?

### DIFF
--- a/src/AI/AITransportUtils.java
+++ b/src/AI/AITransportUtils.java
@@ -31,7 +31,7 @@ public class AITransportUtils
     {
       if( modelT.baseCargoCapacity < 1 )
         continue;
-      if( !modelT.isCargoRole(cargo.model.role) )
+      if( !modelT.canTransport(cargo.model.role) )
         continue;
       transportTypes.add(modelT);
     }
--- a/src/AI/AITransportUtils.java
+++ b/src/AI/AITransportUtils.java
@@ -3,6 +3,9 @@ package AI;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+
+import Engine.PathCalcParams;
+import Engine.PathCalcParams.SearchNode;
 import Engine.Utils;
 import Engine.XYCoord;
 import Terrain.GameMap;
@@ -14,7 +17,8 @@ public class AITransportUtils
 {
   /**
    * Build a map of potential transport types, and where they could unload the unit to reach its destination.<p>
-   * Note that this only does a 1-ply search from the <b>destination</b>, so a recursive doughnut island would need extra scaffolding.
+   * Note that this only does a 1-ply search from the <b>destination</b>, so a recursive doughnut island would need extra scaffolding.<p>
+   * Note also that this will not calc transports with the exact same movetype as the thing they're transporting, since they use the same island.
    */
   public static HashMap<UnitModel, HashSet<XYCoord>> findUnloadTiles(GameMap map, ReachabilityCache rc, UnitContext cargo, XYCoord dest)
   {
@@ -58,9 +62,86 @@ public class AITransportUtils
 
     return transportToBeachMap;
   }
-//    PathCalcParams pcp = new PathCalcParams(uc, gameMap);
-//    pcp.start = origin;
-//    pcp.includeOccupiedSpaces = true; // We assume the enemy knows how to manage positioning within his turn
-//    ArrayList<SearchNode> destinations = pcp.findAllPaths();
+
+  public static class InterceptPaths
+  {
+    SearchNode transport, cargo;
+  }
+
+  /**
+   * Determines the places that cargo can load into transport with optimal turn usage (accounting for both units' turn status).
+   * <p>This assumes cargo should load into transport while transport is still ready to act, if possible.
+   * <p>If a UC's unit pointer is not populated, this function will assume the unit is to be built (on the UC coordinate).
+   */
+  // TODO: Consider enabling PCP to take in a list of start tiles, so less needs to be recalculated on each iteration?
+  public static HashMap<XYCoord, InterceptPaths> findOptimalLoadIntercepts(GameMap map, ReachabilityCache rc, UnitContext transport, UnitContext cargo)
+  {
+    HashMap<XYCoord, InterceptPaths> result = new HashMap<>();
+    var cargoIslands = rc.getAdjacentIslands(cargo, map);
+    if( cargoIslands.isEmpty() )
+      return result; // Cargo can't move, so GGs
+
+    boolean canLoad = false;
+    if( cargoIslands.contains(rc.getIsland(cargo.model.baseMoveType, transport.coord)) )
+      canLoad = true; // Cargo can reach the transport, even if the transport can't move.
+    if( !canLoad )
+    {
+      var transportIslands = rc.getAdjacentIslands(transport, map);
+      for( var cis : cargoIslands )
+        for( var tis : transportIslands )
+          if( cis.overlapIslands.contains(tis) )
+          {
+            canLoad = true; // Cargo can reach a place the transport can
+            break;
+          }
+    }
+    if( !canLoad )
+      return result;
+
+    int turnsCargo     = 0;
+    int turnsTransport = 0;
+    // We want the cargo to load while the transport has its turn.
+    // This means cargo gets +1 turn advantage if equal initiative, +2 if it has TA, or +0 if it's waited and transport is not
+    if( !cargo.isTurnOver )
+      turnsCargo += 1;
+    if( transport.isTurnOver )
+      turnsCargo += 1;
+
+    PathCalcParams pcpCargo = new PathCalcParams(cargo, map);
+    PathCalcParams pcpTransport = new PathCalcParams(transport, map);
+    // I think accounting for current enemy blocking doesn't make a lot of sense for multi-turn calcs.
+    // If I can already load this turn, I shouldn't be calling this function.
+    // The only other scenario this calc would be important for is SFW-style transports, but I don't plan to add that because they're just bad in a not-fun way.
+    pcpCargo.    canTravelThroughEnemies = true;
+    pcpTransport.canTravelThroughEnemies = true;
+    while( turnsCargo < 10 ) // just so it doesn't go infinite
+    {
+      pcpCargo.    maxTurns = turnsCargo;
+      pcpTransport.maxTurns = turnsTransport;
+      ArrayList<SearchNode> destinationsCargo     = pcpCargo.findAllPaths();
+      ArrayList<SearchNode> destinationsTransport = pcpTransport.findAllPaths();
+      for( var snC : destinationsCargo )
+      {
+        for( var snT : destinationsTransport )
+        {
+          if( !snC.equals(snT) )
+            continue; // Doing a nested for loop instead of .contains() for convenience.
+          var env = map.getEnvironment(snC);
+          if( transport.model.unloadExclusionTerrain.contains(env.terrainType) )
+            continue;
+          InterceptPaths intercept = new InterceptPaths();
+          intercept.cargo     = snC;
+          intercept.transport = snT;
+          result.put(snC, intercept);
+        }
+      }
+      if( !result.isEmpty() )
+        break;
+      ++turnsCargo;
+      ++turnsTransport;
+    }
+
+    return result;
+  }
 
 }
--- a/src/AI/AITransportUtils.java
+++ b/src/AI/AITransportUtils.java
@@ -111,9 +111,9 @@ public class AITransportUtils
     int turnsTransport = 0;
     // We want the cargo to load while the transport has its turn.
     // This means cargo gets +1 turn advantage if equal initiative, +2 if it has TA, or +0 if it's waited and transport is not
-    if( !cargo.isTurnOver )
+    if( !cargo.isTurnOver && null != cargo.unit )
       turnsCargo += 1;
-    if( transport.isTurnOver )
+    if( transport.isTurnOver || null == cargo.unit )
       turnsCargo += 1;
 
     PathCalcParams pcpCargo = new PathCalcParams(cargo, map);
--- a/src/AI/AITransportUtils.java
+++ b/src/AI/AITransportUtils.java
@@ -1,0 +1,66 @@
+package AI;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import Engine.Utils;
+import Engine.XYCoord;
+import Terrain.GameMap;
+import Units.UnitContext;
+import Units.UnitModel;
+import lombok.var;
+
+public class AITransportUtils
+{
+  /**
+   * Build a map of potential transport types, and where they could unload the unit to reach its destination.<p>
+   * Note that this only does a 1-ply search from the <b>destination</b>, so a recursive doughnut island would need extra scaffolding.
+   */
+  public static HashMap<UnitModel, HashSet<XYCoord>> findUnloadTiles(GameMap map, ReachabilityCache rc, UnitContext cargo, XYCoord dest)
+  {
+    var transportToBeachMap = new HashMap<UnitModel, HashSet<XYCoord>>();
+    var grms = map.game.rules.unitModelScheme.getGameReadyModels();
+    var transportTypes = new ArrayList<UnitModel>();
+
+    for( var modelT : grms.unitModels )
+    {
+      if( modelT.baseCargoCapacity < 1 )
+        continue;
+      if( !modelT.isCargoRole(cargo.model.role) )
+        continue;
+      transportTypes.add(modelT);
+    }
+
+    for( var modelT : transportTypes )
+    {
+      var destIsland = rc.getIsland(cargo.model.baseMoveType, dest);
+      var beaches = new HashSet<XYCoord>();
+      for( var connectedIsland : destIsland.overlapIslands )
+      {
+        if( connectedIsland.moveType != modelT.baseMoveType )
+          continue;
+        // Note that there could be multiple islands this transport could approach on.
+        for( var xyc : connectedIsland.coords )
+        {
+          if( !destIsland.coords.contains(xyc) )
+            continue; // Cargo mobility on the transport's terrain (i.e. island overlap) is required to unload
+          var env = map.getEnvironment(xyc);
+          if( modelT.unloadExclusionTerrain.contains(env.terrainType) )
+            continue;
+          ArrayList<XYCoord> dropoffLocations = Utils.findUnloadLocations(map, null, xyc, cargo.model.baseMoveType);
+          if( !dropoffLocations.isEmpty() )
+            beaches.add(xyc);
+        }
+      }
+
+      transportToBeachMap.put(modelT, beaches);
+    }
+
+    return transportToBeachMap;
+  }
+//    PathCalcParams pcp = new PathCalcParams(uc, gameMap);
+//    pcp.start = origin;
+//    pcp.includeOccupiedSpaces = true; // We assume the enemy knows how to manage positioning within his turn
+//    ArrayList<SearchNode> destinations = pcp.findAllPaths();
+
+}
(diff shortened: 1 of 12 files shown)


src/AI/ReachabilityCache.java
```diff
@@ -0,0 +1,162 @@
...
+        Map<MoveType, Island> coordIslandMap = coordToIslandByMoveType[x][y];
+        XYCoord xyc     = new XYCoord(x, y);
+        Environment env = map.getEnvironment(x, y);
+        for( var mt : islandSetsByMoveType.keySet() ) // all unique movetypes
+        {
+          if( !mt.canStandOn(env) )
+            continue; // This might not look like it accounts for teletiles, but teletiles aren't actually part of any island: they just connect islands.
+          if( coordIslandMap.containsKey(mt) )
+            continue; // Already part of an island
+          PathCalcParams pcp = new PathCalcParams(mt, 3, xyc, map); // Chosen by a fair die roll. Guaranteed to be random.
```
ThatIsAUsername, line 137: I notice you bumped this from 1 to 3 - Seems like that will make it assess tiles more times? Your commit comments indicate this change fixed some problem, but didn't make it clear to me what the problem was (and turning on the debug viz code and toggling the value did not make it clear to me).

src/Units/UnitModel.java
```diff
@@ -274,6 +274,11 @@ public boolean isAll(long input)
   {
     return (role & input) == input;
   }
+  public boolean isCargoRole(long type)
```
ThatIsAUsername: `isCargoRole(type)` sounds like it's asking a question about the model itself. e.g., "is this model cargo for this type", vs "can this model type hold that type".
Maybe `canTransport(type)` would be clearer? Or maybe I'm just silly and pedantic.

src/Engine/Utils.java
```diff
@@ -88,16 +88,21 @@ else if (includeTerrain && resident == null && // Peeps ain't there.
...
+  }
+  public static ArrayList<XYCoord> findUnloadLocations(GameMap map, Unit transport, XYCoord moveLoc, final MoveType cargoMoveType, boolean includeOccupied)
   {
     ArrayList<XYCoord> locations = findLocationsInRange(map, moveLoc, 1);
     ArrayList<XYCoord> dropoffLocations = new ArrayList<XYCoord>();
-    final MoveType cargoMoveType = new UnitContext(cargo).calculateMoveType();
     if( cargoMoveType.canStandOn(map.getEnvironment(moveLoc)) )
       for( XYCoord loc : locations )
       {
         // Add any location that is empty and supports movement of the cargo unit.
```
ThatIsAUsername: This comment insists that the drop zone must be empty, but now there is a flag to toggle that check.

src/AI/AITransportUtils.java
```diff
@@ -0,0 +1,193 @@
...
+
+    // Grab relevant transport types for this cargo.
+    for( var modelT : grms.unitModels )
+    {
+      if( modelT.baseCargoCapacity < 1 )
+        continue;
+      if( !modelT.isCargoRole(cargo.model.role) )
+        continue;
+      transportTypes.add(modelT);
+    }
```
ThatIsAUsername: You've got this transport-type aggregation in multiple places - should it move to `unitModelScheme.getTransportModels()`?

src/AI/AITransportUtils.java
```diff
@@ -0,0 +1,193 @@
...
+
+    // Figure out the tiles this transport could drop off dudes from.
+    for( var modelT : transportTypes )
+    {
+      var destIsland = rc.getIsland(cargo.model.baseMoveType, dest);
```
ThatIsAUsername: This declaration doesn't depend on the loop variable, so it can be moved outside the loop.

src/AI/AITransportUtils.java
```diff
@@ -0,0 +1,193 @@
...
+          continue;
+        // Note that there could be multiple islands this transport could approach on.
+        for( var xyc : connectedIsland.coords )
+        {
+          if( !destIsland.coords.contains(xyc) )
+            continue; // Cargo mobility on the transport's terrain (i.e. island overlap) is required to unload
+          var env = map.getEnvironment(xyc);
+          if( modelT.unloadExclusionTerrain.contains(env.terrainType) )
+            continue;
+          ArrayList<XYCoord> dropoffLocations = Utils.findUnloadLocations(map, null, xyc, cargo.model.baseMoveType);
```
ThatIsAUsername: Doesn't `findUnloadLocations` find the places that can unload *to* `xyc`?
But here don't we want to find tiles `xyc` that can unload to `dest` instead? Maybe I'm reading something backward.

src/AI/AITransportUtils.java
```diff
@@ -0,0 +1,193 @@
...
+  public static HashSet<SearchNode> findShortestUnloadTrips(GameMap map, UnitContext transport, HashMap<XYCoord, InterceptPaths> pickups, HashSet<XYCoord> drops)
+  {
+    var result = new HashSet<SearchNode>();
+
+    // We use extraStarts here because we don't know which load tile will give us the shortest trip, and this allows calculating that in one shot.
+    PathCalcParams pcp = new PathCalcParams(transport, map);
+    pcp.canTravelThroughEnemies = true;
+    for( var p : pickups.keySet() )
+    {
+      pcp.start = p; // Arbitrary; just make sure we don't path from the transport's current position.
```
ThatIsAUsername, line 163: Why not? Are we sure the transport's current position won't be a valid meetup?

src/AI/AIUtils.java
```diff
@@ -55,7 +55,7 @@ public class AIUtils
     Map<XYCoord, ArrayList<GameActionSet> > actions = new HashMap<XYCoord, ArrayList<GameActionSet> >();
 
     PathCalcParams pcp = new PathCalcParams(unit, gameMap);
-    pcp.includeOccupiedSpaces = includeOccupiedDestinations;
+    pcp.includeOccupiedSpaces = true; // Must be true to allow Load
```
ThatIsAUsername, line 58: Are other callers still depending on this flag? Or are you now just depending on `unit.getPossibleActions()` below to filter out actions on the occupied spaces?

src/AI/ReachabilityCache.java
ThislsAUsername: I thought it was kinda obvious, which is why I didn't say anything specific.

It's the move power of the "unit doing pathfinding", which means anything over that value is treated as impassable (even in a situation where we have "infinite movepower" from multi-turn).
It shouldn't affect runtime much, though I guess I haven't tested that.

The only movecost above 3 is 4 for inf on snowy mountains, which is actually impassable without a move boost.
There's probably a better solution than picking an arbitrary number, but I don't know what it is and don't want to make the setup more complicated :P

EDIT:
Specifically, different units can have the same movetype but different actual reachability graphs if movecost > move power for only some unit types.
Probably the simplest solution in this case would be at the unit set level: just have the same data in a different instance of the MoveType class, so the cache considers them separately.
I have no good solution for what to do if e.g. permanent snow is on but your CO is Adder or Drake, so you can _technically_ move inf over mountains but it's a slog that requires a power activation (and each solves the problem in a different way).

src/AI/AITransportUtils.java
ThislsAUsername: Seems reasonable.

Aside:
I feel like I didn't really intend for the UMS to be a thing that we query a lot, but I guess the design has changed.
I was thinking CO-based calculation would be the standard since that'd let COs insert stuff like "this unit can transport [a new unit type] now", but a util function prolly doesn't make that much sense to live on `Commander` either... and `Army` complicates the picture to the point where it probably couldn't make sense.
Also, I don't really see that as an extension point we'd be liable to actually use, especially if we want AI design to be tractable (lol)

src/AI/AITransportUtils.java
ThislsAUsername: Nope, it
> Returns a list of locations at distance 1 from transportLoc that cargo can move on.

Outdated param names aside (whoops), we're not looking for any *direct* relationship between `dest` and our candidates, we're just checking if each candidate transport move target has a valid tile to dump the cargo off on (because otherwise... we can't unload).

In retrospect, I think the main thing being accomplished here is that it excludes 1-tile islands from having an unload point, since the "island" for the cargo is definitionally places it can stand.
...OTOH, "discontinuous islands" exist due to teletiles, so nevermind.
Maybe there should be a comment.

EDIT: Would it make sense to check `unloadExclusionTerrain` inside this function body? It'd feel a little odd to return an empty list without actually checking adjacent tiles, but it's prolly cleaner this way?

src/AI/AITransportUtils.java
ThislsAUsername: The transport's current position *can be* a valid meetup point, but it isn't necessarily (and assumed not since the meetup point is implicitly something the transport is pathing to)

If I had a Lander sitting on another island's port (where the inf wanted to go), the Lander would say "gee, this meetup point results in a really short path" and not move.

EDIT: Note that the meetup points are not checked for viability because they're assumed to be all equally good. This is obviously not true in the general case (since we're just considering turncount), but baaaaaby steps.

src/AI/AIUtils.java
ThislsAUsername: Yes, I'm depending on `unit.getPossibleActions()` to filter actions because the player's code path obviously needs that to work and won't be going through `AIUtils`.

However, it's also a thing that the only call path is currently via `getAvailableUnitActionsByType()` from ISAI and Muriel, and Muriel doesn't ever consider LOAD.
So it's safe from both directions.

src/AI/ReachabilityCache.java
ThatIsAUsername: Oh, I see. Yeah, seems like there should definitely be a better/more efficient way to do this. Seems like you need the initial checks (canStandOn, alreadyInIsland), and then just grab the four adjacent, push them into a queue, and move on. This feels like it's abusing the wrong tool (path planning) and repeating work when a simple flood fill would be more efficient (just check adjacency and standability instead of worrying about move power and all that).

Then again, I haven't done the full complexity/efficiency analysis (though triply nested loops is rarely ideal), and this already works, so hey.

src/AI/AITransportUtils.java
ThatIsAUsername: Yeah, comments are usually helpful.

If you need to return an empty list (because you can't actually unload from where you are), then why bother checking the adjacent tiles. Makes sense to me. Also isn't it already doing that check on line 54-55? Or do you mean inside `findUnloadLocations`? Sure, it might make sense to move the check into there.

Also it's strange to call `findUnloadLocations` from inside `findUnloadTiles` - it seems like they advertise doing the same thing.

src/AI/AITransportUtils.java
ThatIsAUsername: > If I had a Lander sitting on another island's port (where the inf wanted to go), the Lander would say "gee, this meetup point results in a really short path" and not move.

By that logic it would just move one tile to the adjacent/next-shortest path - if it's at a port on the target island, that tile shouldn't be in consideration anyway (since the inf can't get to it). Or is that what you mean that they aren't checked for viability?

src/AI/InfantrySpamAI.java
```diff
@@ -150,8 +164,41 @@ public GameAction getNextAction(GameMap gameMap)
...
+        foundAction = true;
+      }
+      if(foundAction)break; // Only one action per getNextAction() call, to avoid overlap.
+
+      if( unit.hasCargoSpace(UnitModel.TROOP) && unit.heldUnits.isEmpty() )
+        continue; // Don't bother with moving infantry transports around until they're called.
+
       // If no attack/capture actions are available now, just move towards a non-allied building.
-      Utils.sortLocationsByTravelTime(unit, unownedProperties, gameMap);
+      Utils.sortLocationsByDistance(new XYCoord(unit), unownedProperties); // Landers want this ordering, and support it for infantry.
```
ThatIsAUsername: Why sort by distance instead of travel time (or at least path length)? Seems like you'd need to take into account sailing around intermediate islands or whatever.

src/AI/InfantrySpamAI.java
```diff
@@ -240,4 +308,135 @@ public GameAction getNextAction(GameMap gameMap)
...
+    loadPoints.put(new XYCoord(unit), ip);
+    HashMap<UnitModel, HashSet<XYCoord>> modelTToUnloadPoints = AITransportUtils.findUnloadTiles(gameMap, rc, new UnitContext(cargo), goal);
+    HashSet<SearchNode> unloadPaths = AITransportUtils.findShortestUnloadTrips(gameMap, new UnitContext(unit), loadPoints, modelTToUnloadPoints.get(unit.model));
+
+    ArrayList<GameAction> unloadActions = unitActionsByType.get(UnitActionFactory.UNLOAD);
+    if( null != unloadActions && !unloadActions.isEmpty() )
+      for( GameAction action : unloadActions )
+      {
+        // Rather than dig into the unload action's bits, just assume we're dropping off the first cargo unit.
+        if( unloadPaths.contains(action.getMoveLocation()) )
```
ThatIsAUsername, line 346: Just unload at the first opportunity? Seems like that would disallow APC games for optimizing early captures.
On the other hand, this is in ISAI (so probs no need for capture optimization), and baby steps are okay.

src/AI/ReachabilityCache.java
ThislsAUsername: `pcp.findAllPaths()` is a flood fill. It just happens to keep the parent metadata because that's usually useful (just not in this case). It also natively handles all the weird edge cases with one pre-existing code path.

If I'm just grabbing the "four adjacent", then I'm not sure whether those "really count" as being in an island yet? It also really doesn't account for teletiles, while fully calculating the island immediately does.

I dunno where the "repeating work" comes in, either.
For each tile in the map, for each movetype, check if I have an island yet.
If I have no island, calculate that island.


On that note, this is what was inspiring the earlier pathfinding changes and the perf test.
I tried using the reachability cache as the perf test, but it was a bit too fast for me to feel confident in the test setup (seconds for my whole map library vs minutes with the current setup).
This operation ended up far cheaper than I expected it to be, so asking for more feels like it'd just make the code less intuitive for not much benefit. I guess I could be wrong.

src/AI/AITransportUtils.java
ThislsAUsername: Yeah, I was talking about moving the check into `findUnloadLocations()`.
EDIT: The unload check depends on the `Unit` arg being identity, but this function doesn't have a unit in context. I could make up a unit or split out a `UnitModel` arg, but I don't like either choice.

Your point on nomenclature is good. This function is kind of hard to name (maybe because it's doing more than one job?).
It produces a map of transport types you could use to get `cargo` onto the same island as `dest`, and the tiles they could do that final unload on.
I guess it could be `calcUnloadTilesOntoIsland()` or `calcUnloadTilesToReach[Dest]()`?

src/AI/AITransportUtils.java
ThislsAUsername: > if it's at a port on the target island, that tile shouldn't be in consideration anyway (since the inf can't get to it). Or is that what you mean that they aren't checked for viability?

Yessir, that's what I meant.

`HashMap<XYCoord, InterceptPaths> pickups` is the list of "tiles where I can pick up `cargo` that optimize turn count".
I want to pathfind from only those tiles, so that I produce a route from a valid pickup point to a valid drop point.
`PCP` was not built to assume the unit is pathing from a location it is not (since we haven't picked up `cargo` yet...), so we have to prevent it from making that assumption.

It's sort of an abuse of the API, but I didn't wanna unify on a broader standard most of the codebase doesn't need. (i.e. `extraStarts` -> `starts` and get rid of `start` so most calcs just have `starts` contain one element: the old `start` value)

src/AI/InfantrySpamAI.java
ThislsAUsername: That is indeed a pessimization for the case where we're having our infantry just hoof it.
I had been testing on a small map with a bunch of simple islands, so I guess I forgot about that.

`unit` is the infantry, so sorting by travel time doesn't give us anything on the properties the inf can't reach on its own.
Sailing around islands would be useful to consider, but we don't have the transport in context to do that check yet.

I guess the solution would be to make `sortLocationsByTravelTime()` also sort unreachable tiles by Manhattan distance. I think that cleanly solves all the problems we can solve on this line.

src/AI/InfantrySpamAI.java
ThislsAUsername: Yeah, this is ISAI code. ISAI should never build APCs, because APCs don't help infantry get to new places (which was useful for highlighting some logic bugs, heh).

It also results in unloading pretty stupidly (sometimes you unload only one of your cargo instead of both, and the `unloadPaths.contains()` check is required to make it unload on the right island at all), but yeah... baby steps.
This kinda seems like its own sort of mildly hard problem I'm not entirely sure how to solve.

ThislsAUsername: ...on that note, I feel like ISAI's description should get an update if we're keeping these changes.
~'twould be a pity to ruin perfection, though~

src/AI/InfantrySpamAI.java
ThislsAUsername: Having now tried it, there is a silly implication:
Landers become mind controlled by the presence of an unowned port and beeline for it to the exclusion of everything else (because the port is always reachable, and nothing else is).

I actually dunno how to fix this in a way that respects real lander travel time and also targets land tiles "in range".
...I guess it could... target the closest unload spot that's closest to a property? But I dunno how to... wrangle that. It definitely don't feel in scope for this PR.

src/AI/InfantrySpamAI.java
ThatIsAUsername: I was kinda assuming you wouldn't have done all this work if you weren't planning on keeping the changes, and I'm really not here to gatekeep. Up to you ~~if you want to ruin perfection~~.

ThislsAUsername: Oh, I definitely want the scaffolding. I'm less sure about the ISAI-specific changes, but I think it's fine since it's scoped to maps where it'd be helpless otherwise.
But you complain if I add stuff that isn't used, so I added a usecase that wouldn't blow up the scope too much.

src/AI/InfantrySpamAI.java
ThatIsAUsername: To the exclusion of everything except ports? Why not shoals? Oh, because you are specifically searching among `unownedProperties`, right. I mean, as long as the lander has cargo, that's not necessarily a bad thing? But definitely not optimal, agreed.

Yeah, getting the behavior "right" is certainly a bigger chore, and it's already at least functional.

ThatIsAUsername:
I'm not sure I have much more to say here. I'm sure there are efficiency improvements to be made, but there's a lot to be said for "functional", and I'm not gonna try to audit the complexity or whatever. It worked fine on a 4-for-all on a midsized map, so hey.

PR #481, opened 2026-08-03
Some small DoR bugs

ThislsAUsername:


ThislsAUsername:
I doublechecked, and it looks like the only "parent" CO subclass that uses the turn init hook is Deployable.

I guess it could be argued that Deployable should itself override the function that calls the turn init hook, so its childrens' turn inits don't have to call back up the chain.

